PR #186, opened 2024-02-22
feat: Add UI for custom host worker capability requirements

YutongLi291:
The worker requirements tab has a section on Custom Worker requirements that hasn't been implemented yet.

* Add UI that's similar to the one in the farm creation workflow on the Monitor for adding new attribute requirements and new amount requirements
* Submit/Export will populate the defined custom requirements
* Many misc. changes and fixes from the 1st draft https://github.com/casillas2/deadline-cloud/pull/81
* Field validation
* Add warning popup when custom requirements aren't valid, e.g. when not all fields are filled out and when min > max for the amount requirements.
* Also added the tab to the CLI job submitter for dev

User with access to host requirements tab can input custom requirements.  For submitter without access to host requirements tab, there are no changes.

Tested by temporarily enabling host requirements tab on Dev GUI Submitter 

To test, set show_host_requirements_tab=True in job_bundle_submitter.py L166, and add the below in `def on_create_job_bundle_callback`.
```
 # If "HostRequirements" is provided, inject it into each of the "Step"
        if host_requirements:
            print(f"{host_requirements}")
            # for each step in the template, append the same host requirements.
            for step in template["steps"]:
                step["hostRequirements"] = copy.deepcopy(host_requirements)
```

Everything works as expected on submitters from my testing

#### Deadline cloud submitter

image

#### Maya Submitter
image

#### Nuke Submitter

image

#### Example output
```

specificationVersion: jobtemplate-2023-09
name: Timeout Example
description: |
  This job runs tasks that run indefinitely with a random probability, and
  requires timeouts to function.
parameterDefinitions:
- name: HangProbability
  type: FLOAT
  description: Probability that a task will hang.
  default: 0.2
  minValue: 0
  maxValue: 1
- name: TaskCount
  type: INT
  description: How many tasks to run.
  default: 50
  minValue: 1
steps:
- name: CliScript
  parameterSpace:
    taskParameterDefinitions:
    - name: TaskIndex
      type: INT
      range: 1-{{Param.TaskCount}}
  script:
    embeddedFiles:
    - name: runScript
      type: TEXT
      data: |
        import random
        import time

        if random.random() < {{Param.HangProbability}}:
            while True:
                time.sleep(60)
    actions:
      onRun:
        command: python
        args:
        - '{{Task.File.runScript}}'
        timeout: 10
  hostRequirements:
    amounts:
    - name: amount.worker.Capacity
      min: 1
      max: 12
    attributes:
    - name: attr.worker.Fruit
      anyOf:
      - Banana
      - Grape
```

No

No


jericht (commented):
Looks good! I just have a few questions since I'm not too familiar with how Qt works

justinsaws (approved):

jericht (approved):

PR #284, opened 2024-04-03
fix: set QT_API to pyside6 or pyside2 for deadline-cli

epmog:

If a user has PyQt5/6 installed in their environment, launching the gui from the deadline client cli will throw an error saying `TypeError: 'f' is an unknown keyword argument`. This is because PyQt5/6 does not define f as a keyword arg in the QDialog constructor, even though Qt, PySide2, and PySide6 do.

This error will occur even if PySide6 or PySide2 are in the environment.

For now, set the QT_API environment variable to use/prioritize our supported Qt bindings.

ref: https://github.com/spyder-ide/qtpy/tree/master?tab=readme-ov-file#requirements

Can always add PyQt5/6 support in the future if needed.

Users with additional dependencies in their environment will continue to work!

Before the fix:
```
hatch shell
pip install PyQt5 PySide6
deadline config gui
```
received this error, since `f` is not a keyword for a QDialog in PyQt5/6

image

After fix without changing environment:
```
deadline config gui
```

And the gui launched properly

N/A

Nope

----

### DIFF
--- a/src/deadline/client/ui/__init__.py
+++ b/src/deadline/client/ui/__init__.py
@@ -60,6 +60,7 @@ def gui_context_for_cli():
         app.exec()
     """
     import importlib
+    import os
     from os.path import basename, dirname, join, normpath
     import shlex
     import shutil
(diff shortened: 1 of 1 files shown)


src/deadline/client/ui/__init__.py
```diff
@@ -69,8 +70,9 @@ def gui_context_for_cli():
 
     import click
 
-    has_pyside = importlib.util.find_spec("PySide6") or importlib.util.find_spec("PySide2")
-    if not has_pyside:
+    has_pyside6 = importlib.util.find_spec("PySide6")
```
viknith, line 73: Is this logic change just for code quality purpose or does this actually fix something?

viknith (approved):

src/deadline/client/ui/__init__.py
epmog: Makes it easier to check which pyside to prioritize loading further on in the code.

marofke (approved):

PR #285, opened 2024-04-03
fix(job_attachments): pass original exception to AssetSyncError

gahyusuh:
In certain cases, when raising `AssetSyncError` from another exception using 
```
except Exception as e:
    raise AssetSyncError from e
```
the error message was not being displayed correctly. This makes it hard to debug and identify the root cause of the issue.

Change `raise AssetSyncError from e` to `raise AssetSyncError(e) from e`, passing the original exception instance `e` to the `AssetSyncError`, so that we can ensure that the error message is preserved and displayed correctly.

Improves the overall debugging experience and make it easier to understand the root cause of issues.

- Made sure all unit tests passed by running `hatch run test`
- Made sure all integration tests passed by running`hatch run integ:test`
- Manually tested a scenario where the `[WinError 183] Cannot create a file when that file already exists` error occurred when running the output download command, `deadline job download-output < resource ids >  --output json --conflict-resolution CREATE_COPY`, on windows. Before this change, the download would fail silently without any error message. After the change, the error message was correctly displayed.

No.

No.

----

### DIFF
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -122,7 +122,7 @@ def get_manifest_from_s3(
             error_details=str(bce),
         ) from bce
     except Exception as e:
-        raise AssetSyncError from e
+        raise AssetSyncError(e) from e
 
 
 def _get_output_manifest_prefix(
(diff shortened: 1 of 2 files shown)


marofke (approved):
Good find! This should help surface errors to users much better!

:ship: :shark:

lucaseck (approved):

PR #287, opened 2024-04-06
fix(job_attachments): handle case-insensitive path extraction on Windows

gahyusuh:
On Windows, asset paths are not treated as case-insensitive, leading to incorrect grouping of assets with paths that differ only in case. For example, if two paths "C:\Username\user\test1.txt" and "c:\uSerNaMe\user\test2.txt" exist, their shared root path should be extracted as "C:\Username\user". However, currently, these paths are treated as having different root paths due to the case-sensitivity issue.

The `_get_asset_groups` function has been modified to ensure that the root path extraction is case-insensitive on Windows. 

Resolves the issue of incorrect asset grouping on Windows due to case-sensitivity

- A new unit test case has been added to validate the case-insensitive root path extraction.
- Existing unit tests and integration tests pass successfully.
- Manual end-to-end testing: submitting a job with a following parameter:
  ```
  name: AssetsDir1
  type: PATH
  objectType: DIRECTORY
  dataFlow: IN
  default: ./INPUTS_1
  ```
Where the actual input directory's name on the file system is `inputs_1`. The job submission and job execution on a worker was working, and generated output files as expected.

No.

No.

----

### DIFF
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -832,7 +832,8 @@ class S3AssetManager:
                 abs_path=abs_path,
                 local_type_locations=local_type_locations,
             )
-            groupings[matched_root].inputs.add(abs_path)
+            matched_group = self._get_matched_group(matched_root, groupings)
+            matched_group.inputs.add(abs_path)
 
         for _path in output_paths:
             abs_path = Path(os.path.normpath(Path(_path).absolute()))
@@ -849,7 +850,8 @@ class S3AssetManager:
                 abs_path=abs_path,
                 local_type_locations=local_type_locations,
             )
-            groupings[matched_root].outputs.add(abs_path)
+            matched_group = self._get_matched_group(matched_root, groupings)
+            matched_group.outputs.add(abs_path)
 
         for _path in referenced_paths:
             abs_path = Path(os.path.normpath(Path(_path).absolute()))
@@ -865,7 +867,8 @@ class S3AssetManager:
                 abs_path=abs_path,
                 local_type_locations=local_type_locations,
             )
-            groupings[matched_root].references.add(abs_path)
+            matched_group = self._get_matched_group(matched_root, groupings)
+            matched_group.references.add(abs_path)
 
         # Finally, build the list of asset root groups
         for asset_group in groupings.values():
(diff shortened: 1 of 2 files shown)


test/unit/deadline_job_attachments/test_upload.py
```diff
@@ -2122,6 +2122,24 @@ def test_get_asset_groups(
                 {},  # File System Location (SHARED type)
                 [],
             ),
+            (
+                {"c:\\USERNAME\\DOCS\\inputs\\input1.txt"},  # input paths
+                {"C:\\username\\docs\\outputs"},  # output paths
```
loachri: question: Do we need/want test cases where the output path uses a lowercase `c`?

gahyusuh: Sure, I can add it

src/deadline/job_attachments/upload.py
```diff
@@ -906,10 +906,11 @@ def _find_matched_root_from_local_type_locations(
                 )
             return matched_root
         else:
+            keys_upper = [key.upper() for key in groupings.keys()]
```
mwiebe: Could we use `os.path.normcase` (see https://docs.python.org/3/library/os.path.html#os.path.normcase) instead of upper for this purpose? With that, the same code can have correct OS-specific behavior.

test/unit/deadline_job_attachments/test_upload.py
```diff
@@ -2122,6 +2122,24 @@ def test_get_asset_groups(
...
+                set(),  # referenced paths
+                {},  # File System Location (LOCAL type)
+                {},  # File System Location (SHARED type)
+                [
+                    AssetRootGroup(
+                        root_path="C:\\username\\docs",
+                        inputs={
+                            Path("C:\\username\\docs\\inputs\\input1.txt"),
```
ttblanchard: Can you share what kind of E2E testing we did on this change? It seems this test case is implying that it's okay to modify the provided input/output path (UPPER/lower). My understanding of the preferred solution here would to only perform a case-insensitce check, but store the paths in the asset manifest as provided. eg. removing the `.upper()` calls on line 912 and 913.

ttblanchard: Also would this not break on non-Windows systems as they would have case sensitive paths?

gahyusuh: For the E2E testing, I have submitted a (non-rendering) job that has a following parameter in its template:
```
name: AssetsDir1
type: PATH
objectType: DIRECTORY
dataFlow: IN
default: ./INPUTS_1
```
Where the actual input directory's name on the file system is `inputs_1`. The job submission and job execution on a worker was working, and generated output files as expected.
And then checked that the input manifest has the dir name `INPUTS_1` in its paths:
```
{
    "hash":"002537c669b48332b051bae047beed61",
    "mtime":1706715912000000,
    "path":"INPUTS_1/medium_test_1_0.txt",
    "size":52428800
}
```
Would there be anything I have missed?

ttblanchard: hmm, I still think we'd want to leave the provided paths unmodified.

src/deadline/job_attachments/upload.py
gahyusuh: I fixed it to use `os.path.normcase` instead of `upper`. I also added a test case for non-Windows.

loachri (dismissed):
LGTM! Ship it :shipit:

ttblanchard (dismissed):

loachri (dismissed):

mwiebe (approved):

loachri (approved):

ttblanchard (approved):

PR #290, opened 2024-04-08
fix: Throw error in sync-inputs if total input size is too large

marofke:
When using the [COPIED](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/models.py#L156) Job Attachments file system type, a worker could end up getting "disk full" errors because there isn't enough space on the machine to download all of the inputs required for the job.

Since we count the total size of all inputs in the manifest file, we'll just check if the free disk space is enough to fit all of the files we're planning to download, and otherwise throw an error so users can see they need to increase their storage, or reduce the size of their job.

Jobs should fail fast if there are too many inputs for workers.

- Added new unit test
- spun up a CMF with my changes and did the following tests:
  - confirmed a normal job downloaded inputs successfully
  - modified the input manifest for a job, setting the file sizes really high, and confirmed it failed in the Sync Attachments session action. Here's the log:
```
No older logs at this moment.
2024/04/09 11:32:28-05:00  
2024/04/09 11:32:28-05:00 ==============================================
2024/04/09 11:32:28-05:00 --------- Job Attachments Download for Job
2024/04/09 11:32:28-05:00 ==============================================
2024/04/09 11:32:28-05:00 Syncing inputs using Job Attachments
2024/04/09 11:32:28-05:00 Error occurred while attempting to sync input files: Total file size required for download (100.0 TB) is larger than available disk space (255.15 GB)
  File "/opt/deadline/worker/lib/python3.9/site-packages/deadline_worker_agent/sessions/actions/sync_input_job_attachments.py", line 135, in _on_done
    future.result()
  File "/home/marofke/.local/share/rtx/installs/python/3.9.16/lib/python3.9/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/home/marofke/.local/share/rtx/installs/python/3.9.16/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/home/marofke/.local/share/rtx/installs/python/3.9.16/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/opt/deadline/worker/lib/python3.9/site-packages/deadline_worker_agent/sessions/session.py", line 934, in sync_asset_inputs
    (download_summary_statistics, path_mapping_rules) = self._asset_sync.sync_inputs(
  File "/opt/deadline/worker/lib/python3.9/site-packages/deadline/job_attachments/asset_sync.py", line 502, in sync_inputs
    self._ensure_disk_capacity(session_dir, total_input_size)
  File "/opt/deadline/worker/lib/python3.9/site-packages/deadline/job_attachments/asset_sync.py", line 349, in _ensure_disk_capacity
    raise AssetSyncError(
deadline.job_attachments.exceptions.AssetSyncError: Error occurred while attempting to sync input files: Total file size required for download (100.0 TB) is larger than available disk space (255.15 GB)
No newer logs at this moment.
```

N/A

No

----

### DIFF
--- a/src/deadline/job_attachments/asset_sync.py
+++ b/src/deadline/job_attachments/asset_sync.py
@@ -3,6 +3,7 @@
 """ Module for File Attachment synching """
 from __future__ import annotations
 import os
+import shutil
 import sys
 import time
 from io import BytesIO
@@ -337,6 +338,19 @@ class AssetSync:
                 abs_path = str(Path(local_root) / manifest_path.path)
                 self.synced_assets_mtime[abs_path] = Path(abs_path).stat().st_mtime_ns
 
+    def _ensure_disk_capacity(self, session_dir: Path, total_input_bytes: int) -> None:
+        """
+        Raises an AssetSyncError if the given input bytes is larger than the available disk space.
+        """
+        disk_free: int = shutil.disk_usage(session_dir).free
+        if total_input_bytes > disk_free:
+            input_size_readable = _human_readable_file_size(total_input_bytes)
+            disk_free_readable = _human_readable_file_size(disk_free)
+            raise AssetSyncError(
+                "Error occurred while attempting to sync input files: "
+                f"Total file size required for download ({input_size_readable}) is larger than available disk space ({disk_free_readable})"
+            )
+
     def sync_inputs(
         self,
         s3_settings: Optional[JobAttachmentS3Settings],
@@ -447,14 +461,16 @@ class AssetSync:
 
         # Merge the manifests in each root into a single manifest
         merged_manifests_by_root: dict[str, BaseAssetManifest] = dict()
+        total_input_size: int = 0
         for root, manifests in grouped_manifests_by_root.items():
             merged_manifest = merge_asset_manifests(manifests)
 
             if merged_manifest:
                 merged_manifests_by_root[root] = merged_manifest
+                total_input_size += merged_manifest.totalSize  # type: ignore[attr-defined]
 
         # Download
-
+        # Virtual Download Flow
         if (
             attachments.fileSystem == JobAttachmentsFileSystem.VIRTUAL.value
             and sys.platform != "win32"
@@ -482,6 +498,8 @@ class AssetSync:
                     f"Virtual File System not found, falling back to {JobAttachmentsFileSystem.COPIED} for JobAttachmentsFileSystem."
                 )
 
+        # Copied Download flow
+        self._ensure_disk_capacity(session_dir, total_input_size)
         try:
             download_summary_statistics = download_files_from_manifests(
                 s3_bucket=s3_settings.s3BucketName,
--- a/test/unit/deadline_job_attachments/test_asset_sync.py
+++ b/test/unit/deadline_job_attachments/test_asset_sync.py
@@ -21,6 +21,7 @@ from deadline.job_attachments.asset_sync import AssetSync
 from deadline.job_attachments.os_file_permission import PosixFileSystemPermissionSettings
 
 from deadline.job_attachments.exceptions import (
+    AssetSyncError,
     VFSExecutableMissingError,
     JobAttachmentsS3ClientError,
     VFSOSUserNotSetError,
@@ -276,6 +277,7 @@ class TestAssetSync:
         default_queue: Queue,
         job_fixture_name: str,
         s3_settings_fixture_name: str,
+        test_manifest_one: dict,
         request: pytest.FixtureRequest,
     ):
         """Asserts that a specific error message is raised when getting 404 errors synching inputs"""
@@ -288,6 +290,7 @@ class TestAssetSync:
             message="File not found",
         )
         job: Job = request.getfixturevalue(job_fixture_name)
+        test_manifest = decode_manifest(json.dumps(test_manifest_one))
         s3_settings: JobAttachmentS3Settings = request.getfixturevalue(s3_settings_fixture_name)
         default_queue.jobAttachmentSettings = s3_settings
         dest_dir = "assetroot-27bggh78dd2b568ab123"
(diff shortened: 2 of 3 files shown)


src/deadline/job_attachments/asset_sync.py
```diff
@@ -482,6 +498,8 @@ def sync_inputs(
                     f"Virtual File System not found, falling back to {JobAttachmentsFileSystem.COPIED} for JobAttachmentsFileSystem."
                 )
 
+        # Copied Download flow
+        self._ensure_disk_capacity(session_dir, total_input_size)
```
jericht, line 502: Just at a glance, it looks like we'll always be ensuring disk capacity, even if using VFS. Does the above code make the session dir have infinite space when inspected by `shutil.disk_usage`?

test/unit/deadline_job_attachments/test_asset_sync.py
```diff
@@ -465,6 +486,78 @@ def test_sync_inputs_with_step_dependencies_same_root_vfs_on_posix(
...
+            ("default_job"),
+        ],
+    )
+    @pytest.mark.parametrize(
+        ("s3_settings_fixture_name"),
+        [
+            ("default_job_attachment_s3_settings"),
+        ],
+    )
+    def test_sync_inputs_no_space_left(
```
jericht, line 501: Should we add a test case that this doesn't raise an error if VFS is being used?

leongdl (approved):
Looks good to me!

src/deadline/job_attachments/asset_sync.py
leongdl: I think VFS returns at line479/ 495, starting at 501 is regular JA downloads?

mwiebe: That's a good point, we should allow the VFS case to fail dynamically, because if the job only accesses a small fraction of a really large dataset, it will use just a small amount of EBS space.

src/deadline/job_attachments/asset_sync.py
```diff
@@ -337,6 +338,19 @@ def _record_attachment_mtimes(
...
+        """
+        Raises an AssetSyncError if the given input bytes is larger than the available disk space.
+        """
+        disk_free: int = shutil.disk_usage(session_dir).free
+        if total_input_bytes > disk_free:
+            input_size_readable = _human_readable_file_size(total_input_bytes)
+            disk_free_readable = _human_readable_file_size(disk_free)
+            raise AssetSyncError(
+                "Error occurred while attempting to sync input files: "
+                f"Total file size required for download ({input_size_readable}) is larger than available disk space ({disk_free_readable})"
```
epmog, line 351: What're your thoughts on adding more guidance for the user? From what I understand they have 2 options available to them:
1. Increase storage capacity on the workers
2. Switch to VFS if they believe each task COULD fit within the storage capacity

Can you think of a succinct way to make that message user-friendly? Right now, I'm not certain how users find out that VFS is an option

src/deadline/job_attachments/asset_sync.py
marofke: As David mentioned, we return early if using VFS. We fallback to COPIED if some settings aren't set correctly for VFS to run.

src/deadline/job_attachments/asset_sync.py
marofke: We do have a whole docs page for VFS, so customers should have ways to find out about it: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/storage-virtual.html, I also put a link to this same page in the JA readme

I could link that in the code, but not sure if we've done that anywhere / if that's a thing we want to do?

test/unit/deadline_job_attachments/test_asset_sync.py
marofke: Sure yea I can do that

marofke: Oh wait, there's no need, I forgot I'm doing an `assert_not_called` in this VFS test, which is doing the test you suggested: https://github.com/aws-deadline/deadline-cloud/blob/ed9539a02474183b17bb214afb013ccaaf7d8153/test/unit/deadline_job_attachments/test_asset_sync.py#L474

src/deadline/job_attachments/asset_sync.py
marofke: I'm struggling to come up with a succinct way to word this as well. As I don't believe customers can increase the storage of SMF workers, so point 1 doesn't help in all cases. So really we have:

1. If SMF, try VFS if a total step's input size < the total storage capacity
2. Otherwise if SMF, just make smaller jobs, or try your own CMF with larger storage
3. If CMF, increase your storage

I don't think the worker knows if it's SMF or CMF in this context, so I'd have to do a rather large message, how does this sound?

```
Error occurred while attempting to sync input files:
Total file size required for download (12PB) is larger than available disk space (155GB).
Try using the 'VIRTUAL' job attachments filesystem option if the total input size for each step (including step dependencies) is less than the available space on the worker.
Otherwise if using a CMF worker, try increasing the storage capacity.
```

marofke: > Switch to VFS if they believe each task COULD fit within the storage capacity

We can't say 'task' since a session spans a step, so if a session spans all tasks for a step we need capacity for it. Also need to account for step dependency inputs, so the message could get pretty bloated

epmog: Chatted offline, I think we can do better to guide customers, but for now we'll leave this as-is. The info probably belongs in a troubleshooting section due to the decision tree that presents itself

epmog (approved):
I do notice that there's still a few scenarios where we should theoretically know not to download anything. Consider the scenario:

1. job downloads fit on disk
2. step downloads do NOT fit on disk

Since they happen as separate actions, the first action will succeed and the second would fail. So there's still some wasted downloading.

I believe any fix we want there would require us to change how we're handling the actions within the worker though, so that's for another time.

Ship it!

PR #291, opened 2024-04-08
fix: Write manifests locally when submitting

marofke:
When making job submissions with Job Attachments, it isn't easy to see what actual manifests were submitted with your job. You'd have to call `deadline job get` to see your manifest paths, and then call `aws s3 cp` to download that file locally.

Write the manifest files locally in the job bundle history directory, in a `manifests` sub-directory. Note that currently only GUI submissions create job history directories, so we aren't saving manifests for CLI submissions.

It'll be a lot easier to debug Job Attachments submission issues now that customers can just look at what assets were collected for the job submission by looking through their manifest files.

- Updated unit test
- Did multiple job submissions from GUI and confirmed the manifests directory was created in each job bundle history submission.
- Did a CLI submission and confirmed it worked as expected, and did not write a manifest file locally

Updated the readme

No

----

### DIFF
--- a/src/deadline/job_attachments/README.md
+++ b/src/deadline/job_attachments/README.md
@@ -41,6 +41,8 @@ These snapshots are encapsulated in one or more [`asset_manifests`](asset_manife
 
 When starting work, the worker downloads the manifest associated with your job, and recreates the file structure of your submission locally, either downloading all files at once, or as needed if using the [virtual][vfs] job attachments filesystem type. When a task completes, the worker creates a new manifest for any outputs that were specified in the job submission, and uploads the manifest and the outputs back to your S3 bucket.
 
+Manifest files are written to a `manifests` directory within each job bundle that is added to the job history if submitted through the GUI (default: `~/.deadline/job_history`). The file path inside the `manifests` directory corresponds to the S3 manifest path in the submitted job's job attachments metadata.
+
 [vfs]: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/storage-virtual.html
 
 ## Local Cache Files
--- a/src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
+++ b/src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
@@ -339,6 +339,7 @@ class SubmitJobProgressDialog(QDialog):
                 manifests=manifests,
                 on_uploading_assets=_update_upload_progress,
                 s3_check_cache_dir=config_file.get_cache_directory(),
+                manifest_write_dir=self._job_bundle_dir,
             )
 
             logger.info("Finished uploading job attachments files.")
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -144,6 +144,7 @@ class S3AssetUploader:
         file_system_location_name: Optional[str] = None,
         progress_tracker: Optional[ProgressTracker] = None,
         s3_check_cache_dir: Optional[str] = None,
+        manifest_write_dir: Optional[str] = None,
     ) -> tuple[str, str]:
         """
         Uploads assets based off of an asset manifest, uploads the asset manifest.
@@ -160,15 +161,6 @@ class S3AssetUploader:
         Returns:
             A tuple of (the partial key for the manifest on S3, the hash of input manifest).
         """
-        # Upload assets
-        self.upload_input_files(
-            manifest,
-            job_attachment_settings.s3BucketName,
-            source_root,
-            job_attachment_settings.full_cas_prefix(),
-            progress_tracker,
-            s3_check_cache_dir,
-        )
 
         # Upload asset manifest
         hash_alg = manifest.get_default_hash_alg()
@@ -177,6 +169,15 @@ class S3AssetUploader:
             f"{file_system_location_name or ''}{str(source_root)}".encode(), hash_alg
         )
         manifest_name = f"{manifest_name_prefix}_input"
+
+        if manifest_write_dir:
+            local_manifest_file = os.path.join(
+                manifest_write_dir, "manifests", partial_manifest_prefix, manifest_name
+            )
+            logger.info(f"Creating local manifest file: {local_manifest_file}\n")
+            os.makedirs(os.path.dirname(local_manifest_file), exist_ok=True)
+            with open(local_manifest_file, "w") as file:
+                file.write(manifest.encode())
 
         if partial_manifest_prefix:
             partial_manifest_key = _join_s3_paths(partial_manifest_prefix, manifest_name)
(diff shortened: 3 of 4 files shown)


src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -418,10 +421,15 @@ def _default_update_upload_progress(upload_metadata: Dict[str, str]) -> bool:
     if not upload_progress_callback:
         upload_progress_callback = _default_update_upload_progress
 
+    # Create job history dir for the manifest file(s)
```
mwiebe: This creates a directory in the history dir only for holding the manifests. I don't think that's quite what we want, it seems better to use a temp directory in this code path.

src/deadline/job_attachments/README.md
```diff
@@ -41,6 +41,8 @@ These snapshots are encapsulated in one or more [`asset_manifests`](asset_manife
 
 When starting work, the worker downloads the manifest associated with your job, and recreates the file structure of your submission locally, either downloading all files at once, or as needed if using the [virtual][vfs] job attachments filesystem type. When a task completes, the worker creates a new manifest for any outputs that were specified in the job submission, and uploads the manifest and the outputs back to your S3 bucket.
 
+Manifest files are written locally in the job history directory for debugging purposes (default: `~/.dealine/job_history`).
```
npmacl: dealine

mwiebe: There's a bit of impedance mismatch between this comment, that they're "for debugging purposes", and putting them in a directory called `manifests`. I like the idea of placing these manifest files in the job bundles, and think it would be better to make it about that instead of about the job history directory. The job history directory is about storing the history of job bundle submissions. Each job bundle submission includes the manifests that were submitted.

So for this comment, what about changing the description to "Manifest files are written to a `manifests` directory within each job bundle that is added to the job history."? I don't think there's a reason to say it's for debugging purposes, I can think of a lot of other reasons to use these files too.

Another comment to add here is that the file path inside this `manifests` directory corresponds to the S3 manifest path in the submitted job's job attachments metadata.

mwiebe (changes requested):

src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -418,10 +421,15 @@ def _default_update_upload_progress(upload_metadata: Dict[str, str]) -> bool:
...
+    manifest_write_dir = os.path.join(
```
npmacl: Is there a way to disable this feature with deadline config? If not I think it would make sense to add that ability.

Also I think it would make sense to have a print giving the local path the manifest was saved during job submission. I think it would be more convenient to the user than having to remember/check the path to the manifest directory, and then know to call `job get` to see what the name of the manifest would be.

src/deadline/job_attachments/README.md
marofke: At least it wasn't `dealdine`, but pretty close haha. Fixed!

marofke: Right good call, I've updated this and mentioned the new flag for CLI that Nathan suggested

src/deadline/client/api/_submit_job_bundle.py
marofke: Yea great idea, I've added the CLI flag `--save-manifests` which will write to a temporary directory, and I added an info log statement to print where manifests are written if this is specified

src/deadline/client/api/_submit_job_bundle.py
marofke: I thought the job history dir would be a good common area for users to look in, but with Nathan's suggestion for a specific flag and printing the path for the CLI flow I think a temp dir works better.

src/deadline/job_attachments/README.md
```diff
@@ -41,6 +41,8 @@ These snapshots are encapsulated in one or more [`asset_manifests`](asset_manife
 
 When starting work, the worker downloads the manifest associated with your job, and recreates the file structure of your submission locally, either downloading all files at once, or as needed if using the [virtual][vfs] job attachments filesystem type. When a task completes, the worker creates a new manifest for any outputs that were specified in the job submission, and uploads the manifest and the outputs back to your S3 bucket.
 
+Manifest files are written to a `manifests` directory within each job bundle that is added to the job history (default: `~/.deadline/job_history`) if submitted through the GUI, or written to a temporary directory if submitted through the CLI with the `--save-manifests` flag. The file path inside the `manifests` directory corresponds to the S3 manifest path in the submitted job's job attachments metadata.
```
npmacl: This wouldn't be accurate any more since it's now either the temp dir or job bundle, right? 
`default: ~/.deadline/job_history`

Also whats the reasoning for why we would want to save in a temp directory if submitted with the CLI, vs the job bundle if submitted through GUI? I would think we would want these actions to be equivalent, just with a different interface.

marofke: When you submit a job through the GUI, we copy the job bundle to the job history directory, which is in `~/.deadline/job_history` by default but can be changed. We don't do this for CLI submissions currently. So, for GUI submissions, since we're already always making a job bundle history directory, we write the manifests to the job bundle history directory. As [Mark suggested above](https://github.com/aws-deadline/deadline-cloud/pull/291#discussion_r1556535820), for CLI submissions we just write the manifests to a temporary directory since it doesn't make sense to have a job bundle history directory that just contains manifests (refactoring CLI submissions to also write the job bundle to the job history dir is out of scope of this PR).

npmacl: I see, thanks for the explanation

npmacl (dismissed):

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -87,6 +87,11 @@ def validate_parameters(ctx, param, value):
     is_flag=True,
     help="Skip any confirmation prompts",
 )
+@click.option(
+    "--save-manifests",
```
mwiebe: If we add a `--save-to-history` option here, the combination of that and `--save-manifests` seems redundant and potentially confusing. I think you mentioned that you looked at adding a flag like that, would it be very much extra scope to do that instead of `--save-manifests` as a new option?

mwiebe (changes requested):

src/deadline/client/cli/_groups/bundle_group.py
mwiebe: I think if we want to keep something like `--save-to-history` out of scope, I would remove `--save-manifests` and see about adding `--save-to-history` as a follow-up PR sooner rather than later.

marofke: We chatted offline and decided to remove saving manifests for the CLI flow for now so we don't bloat the CLI interface

mwiebe (approved):

npmacl (approved):

PR #295, opened 2024-04-11
fix(job_attachments): fix output syncing when using identically named local File System Locations across different OS

gahyusuh:
We have identified a bug related to output syncing when using identically named local File System Locations across Windows submitter and Linux workers.

The issue occurs in the following scenario:
1. A queue has storage profiles attached with identically named File System Locations (e.g., FSL1) of the LOCAL type.
2. A job bundle is submitted on Windows, and the bundle is located under the FSL1 path.
3. The job runs on a Linux worker. In this case, the input assets are downloaded to the Linux worker's FSL1 as the root path, and the output files are generated under the same root path.

However, our current implementation (see [this PR](https://github.com/aws-deadline/deadline-cloud/pull/225)) skips files outside the session directory during output syncing. As a result, the output files generated on the Linux worker are skipped because they are outside the session directory.

we updated the output syncing logic, adding the line `session_dir = local_root` when a storage profile path is available. We use this `session_dir` to filter out any files resolved to a location outside of that directory. If storage profile path is available, we can consider the `session_dir` to be the storage profile path.

Ensures that output files generated under the identically named FSL path are properly synced back to S3.

Manual E2E testing on both Linux and Windows, and made sure generated output are synced correctly.

No.

No.

----

### DIFF
--- a/src/deadline/job_attachments/asset_sync.py
+++ b/src/deadline/job_attachments/asset_sync.py
@@ -562,6 +562,7 @@ class AssetSync:
         storage_profiles_source_paths = list(storage_profiles_path_mapping_rules.keys())
 
         for manifest_properties in attachments.manifests:
+            session_root = session_dir
             local_root: Path = Path()
             if (
                 len(storage_profiles_path_mapping_rules) > 0
@@ -571,6 +572,10 @@ class AssetSync:
                     local_root = Path(
                         storage_profiles_path_mapping_rules[manifest_properties.rootPath]
                     )
+                    # We use session_root to filter out any files resolved to a location outside
+                    # of that directory. If storage profile's path mapping rules are available,
+                    # we can consider the session_root to be the mapped-storage profile path.
+                    session_root = local_root
                 else:
                     raise AssetSyncError(
                         "Error occurred while attempting to sync output files: "
@@ -584,7 +589,7 @@ class AssetSync:
                 manifest_properties,
                 s3_settings,
                 local_root,
-                session_dir,
+                session_root,
             )
             if output_files:
                 output_manifest = self._generate_output_manifest(output_files)


marofke (dismissed):
Thanks for fixing this so quickly!

src/deadline/job_attachments/asset_sync.py
```diff
@@ -571,6 +571,10 @@ def sync_outputs(
                     local_root = Path(
                         storage_profiles_path_mapping_rules[manifest_properties.rootPath]
                     )
+                    # We use this session_dir to filter out any files resolved to a location
+                    # outside of that directory. If storage profile path is available, we can
+                    # consider the session directory to be the storage profile path.
+                    session_dir = local_root
```
mwiebe: Does this cover cases where the job simultaneously has the following?

1. A manifest that is for file system location A in the storage profile.
2. Another manifest that is for file system location B in the storage profile.
3. A third manifest that is not for a file system location, so appears in the session directory.

marofke: We iterate over each manifest here: https://github.com/aws-deadline/deadline-cloud/blob/f67d0a1458879dac8f253b5445f599cbb0912fed/src/deadline/job_attachments/asset_sync.py#L564

Each manifest object here has it's own associated root path that may or may not map to a storage profile location: https://github.com/aws-deadline/deadline-cloud/blob/f67d0a1458879dac8f253b5445f599cbb0912fed/src/deadline/job_attachments/models.py#L162

So I believe this covers the cases you're talking about, as each manifest would be grouped under a different root path

marofke: Oh I see what you mean though, since we're overwriting the session_dir, we would have it overwritten for the next iteration of the loop. Yea we should just make a local variable for the session_dir

gahyusuh: Thank you. I added a variable `session_root` local to that loop instead of overwriting the `session_dir`:
```
for manifest in manifests:
    session_root = session_dir
    ...
    if local storage profile is available:
      session_root = local_root
    ...
    _get_output_files(..., local_root, session_root)
```

marofke (approved):

mwiebe (approved):

PR #296, opened 2024-04-12
fix(job_attachments)!: use correct profile for GetStorageProfileForQueue API

gahyusuh:
When submitting a job with a Storage Profile ID, the `GetStorageProfileForQueue` API was using the default AWS profile instead of the Deadline Cloud submitter's configured profile. If the two profiles pointed to different regions, the API couldn't find the storage profile resource, resulting in a job submission failure

Modified the code to use the correct boto3 session (using the Deadline Cloud submitter's profile) when calling the `GetStorageProfileForQueue` API. The API is no longer called from `src/deadline/job_attachments/upload.py`. Instead, the API is called from the below places:
1. src/deadline/client/api/_submit_job_bundle.py
2. src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
and the `prepare_paths_for_upload()` function now receives `storage_profile` object itself instead of `storage_profile_id`.

Fixes the job submission issue when involving a Storage Profile in a different region than the default AWS profile. 

1. I created AWS CLI's `[default]` profile and Deadline Cloud default profile, where each profile is pointing to different regions.
2. Before the change, job submission to a queue with a Storage Profile failed with a "storage profile does not exist" error.
2. After the change, job submission to the same queue succeeded, and the job could be created.
3. A new job was submitted to a Queue with a Storage Profile, and I executed a worker (CMF) to verify that asset sync and output downloads worked as expected.

No.

Yes, there are changes in the public interface (`S3AssetManager`'s `prepare_paths_for_upload()`).

----

### DIFF
--- a/src/deadline/client/api/_submit_job_bundle.py
+++ b/src/deadline/client/api/_submit_job_bundle.py
@@ -71,7 +71,7 @@ def create_job_from_job_bundle(
 
     /template.json|yaml (required): An Open Job Description job template that specifies the work to be done. Job parameters
             are embedded here.
-    /parameter_values.yson|yaml (optional): If provided, these are parameter values for the job template and for
+    /parameter_values.json|yaml (optional): If provided, these are parameter values for the job template and for
             the render farm. AWS Deadline Cloud-specific parameters are like "deadline:priority".
             Looks like:
             {
@@ -126,7 +126,7 @@ def create_job_from_job_bundle(
         hashing_progress_callback / upload_progress_callback / create_job_result_callback (Callable -> bool):
                 Callbacks periodically called while hashing / uploading / waiting for job creation. If returns false,
                 the operation will be cancelled. If return true, the operation continues. Default behavior for each
-                is to not cancel the operation. hashing_progress_callback and upload_progress_callback both recieve
+                is to not cancel the operation. hashing_progress_callback and upload_progress_callback both receive
                 ProgressReport as a parameter, which can be used for projecting remaining time, as in done in the CLI.
     """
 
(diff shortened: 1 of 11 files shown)


src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -171,8 +175,25 @@ def create_job_from_job_bundle(
     }
 
     storage_profile_id = get_setting("settings.storage_profile_id", config=config)
+    storage_profile = None
     if storage_profile_id:
         create_job_args["storageProfileId"] = storage_profile_id
+        storage_profile_response = deadline.get_storage_profile_for_queue(
```
marofke: Nit: since we're doing this in two places, we might want to make a function wrapper like what we've done in other places, like queue params https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/api/_queue_parameters.py

Not blocking though, since we didn't hit the rule of 3 ;)

marofke (dismissed):

test/integ/deadline_job_attachments/test_job_attachments.py
```diff
@@ -1333,8 +1333,9 @@ def test_upload_bucket_wrong_account(external_bucket: str, job_attachment_test:
 
     # WHEN
     with pytest.raises(
-        AssetSyncError, match=f"Error checking if object exists in bucket '{external_bucket}'"
+        AssetSyncError, match=f"Error uploading binary file in bucket '{external_bucket}'"
```
chocecil: I'm curious about the exception. When I ran this test, I received an "access denied" error. Was the related logic updated in this PR?

chocecil (dismissed):

test/integ/deadline_job_attachments/test_job_attachments.py
gahyusuh: Yes, I also saw the "access denied" error message. There were no updates related to this in this PR. I found it a bit strange that this integ test was failing... The error message that's currently showing up seems correct though - it's expected that PutObject would be denied if we try to upload a manifest file to a bucket that we don't have access to.

src/deadline/client/api/_submit_job_bundle.py
gahyusuh: Thank you for the advise. I have dded a function wrapper `get_storage_profile_for_queue` in client/api/_get_storage_profile_for_queue.py.

marofke (approved):

chocecil (approved):

PR #319, opened 2024-04-26
feat: use AWS CRT for faster transfers

crowecawcaw:
Transfers to and from S3 can be slow and do not use all available bandwidth.

Use AWS CRT ([more info](https://aws.amazon.com/blogs/storage/improving-amazon-s3-throughput-for-the-aws-cli-and-boto3-with-the-aws-common-runtime/)) which uses optimized low level libraries under the hood to gain dramatic boosts in performance. Boto3 abstracts away the differences invisibly.

If CRT is installed, the S3 transfer manager will automatically use the CRT flavor of manager when running on some EC2 instances, but its performance boosts are useful in non-EC2 contexts too so this PR opts into the CRT transfer manager unless it's not compatible.

I submitted jobs with 10GB attachments and compared transfer speeds. Before this change, transfers went at ~260MB/s. After this change, they went at ~960MB/s. My machine is an EC2 instance in a different region from the S3 bucket. The connection to S3 was over the internet.

Submitting jobs and observing the job attachment speed. I ran `hatch run test` and `hatch run test_docker`.

No change to documentation.

No. CRT for boto is compatible will all Unix/Linux, Mac, and Windows.

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,9 @@ classifiers = [
 # Applications that consume this library should be the ones that are more strictly
 # limiting dependencies if they want/need to.
 dependencies = [
-    "boto3[crt] >= 1.34.75",
+    # See CR platform and Python version support here: https://pypi.org/project/awscrt/#files
+    "boto3[crt] >= 1.34.75; python_version >= '3.7' and python_version <= '3.11'",
+    "boto3 >= 1.34.75; python_version < '3.7' or python_version > '3.11'",
     "click >= 8.1.7",
     "pyyaml >= 6.0",
     # Job Attachments
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ classifiers = [
 # Applications that consume this library should be the ones that are more strictly
 # limiting dependencies if they want/need to.
 dependencies = [
-    "boto3 >= 1.34.75",
+    "boto3[crt] >= 1.34.75",
     "click >= 8.1.7",
     "pyyaml >= 6.0",
     # Job Attachments
--- a/src/deadline/job_attachments/_aws/aws_clients.py
+++ b/src/deadline/job_attachments/_aws/aws_clients.py
@@ -7,11 +7,7 @@ from functools import lru_cache
 from typing import Optional
 
 import boto3
-<<<<<<< HEAD
-from boto3.s3.transfer import create_crt_transfer_manager
-=======
 from boto3.s3.transfer import create_crt_transfer_manager, create_transfer_manager
->>>>>>> 203eff5 (feat: support AWS CRT for faster transfers)
 
 import botocore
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -35,7 +35,7 @@ classifiers = [
 # Applications that consume this library should be the ones that are more strictly
 # limiting dependencies if they want/need to.
 dependencies = [
-    "boto3 >= 1.34.75",
+    "boto3[crt] >= 1.34.75",
```
ddneilson: On what platforms is CRT not available? This change to dependencies makes this package non-installable on those platforms, so understanding what we're cutting off is important.

crowecawcaw: Adding a link in a comment here to the [CRT distributable files](https://pypi.org/project/awscrt/#files) which show the supported platforms and Python versions. Also adding a fallback to vanilla boto3 for Python versions that don't have a precompiled version of CRT.

ddneilson: Thanks!

ddneilson (approved):

gahyusuh (commented):
When previously trying to enable CRT (https://github.com/aws-deadline/deadline-cloud/pull/191), Preston raised some concerns about its compatibility and fallback behaviour. He questioned whether CRT, being compiled C code, would work seamlessly across different systems and environments, and if it would gracefully fall back to Python in case of any issues. He also inquired about its compatibility with Windows. Could you please share your insights on this?
Also, based on this [doc](https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#:~:text=optimized%20for%20the%20crt%20transfer%20client), for CRT to be effective, the host instance running the AWS CLI should be one of the p4, p5, trn1, or trn1n instance types. I'm curious if you were able to achieve the performance boost by using one of those optimized instance types, or if you found a good approach to optimize the performance?

crowecawcaw:
@gahyusuh Both points are valid. CRT is precompiled for Python 3.7 through 3.11 on a wide range of systems including Windows, Mac, Linux, and Unix. I added a link to the [list of supported CRT versions](https://pypi.org/project/awscrt/#files) as a comment and added conditions to the dependency list so that regular boto3 will be used in if CRT is not compiled for the particular version of Python (especially 3.12).

The normal S3 transfer manager will use CRT automatically only if `preferred_transfer_client` is `auto` and if it's running on a particular list of preconfigured instance types. This CR bypasses the automatic detection and instead uses the CRT transfer manager as long as CRT is installed and if compatible with the boto session (which is when the session and bucket are in the same region). By bypassing the CRT opt-in logic, we can use CRT in more cases. In my testing, it still produced a big speedup on a instance type that was not preconfigured.

gahyusuh (approved):
Thanks for the explanation! It's awesome that we can leverage CRT's performance gains while ensuring compatibility. Appreciate your work!

PR #320, opened 2024-04-29
fix: fix typos in error messages

chocecil:
- Need to improve error messages and fix Minor typos with setup/login/logout
- Deadline Cloud Monitor should be `Deadline Cloud monitor`

image

Opening Deadline Cloud Monitor  Please login before

- improved error messages and fixed Minor typos with setup/login/logout
- replaced `Monitor` with `monitor`

There won't be impact, as it doesn't change any logic

tested in my local

No

----

### DIFF
--- a/src/deadline/client/api/_loginout.py
+++ b/src/deadline/client/api/_loginout.py
@@ -58,7 +58,8 @@ def _login_deadline_cloud_monitor(
             time.sleep(0.5)
     except FileNotFoundError:
         raise DeadlineOperationError(
-            f"Could not find Deadline Cloud Monitor at {deadline_cloud_monitor_path}. Please ensure Deadline Cloud Monitor is installed correctly and setup the {profile_name} profile again."
+            f"Could not find Deadline Cloud Monitor at {deadline_cloud_monitor_path}."
+            f"Please ensure Deadline Cloud Monitor is installed correctly and set up the {profile_name} profile again."
         )
     if on_pending_authorization:
         on_pending_authorization(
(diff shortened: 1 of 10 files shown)


src/deadline/client/api/_loginout.py
```diff
@@ -58,7 +58,8 @@ def _login_deadline_cloud_monitor(
             time.sleep(0.5)
     except FileNotFoundError:
         raise DeadlineOperationError(
-            f"Could not find Deadline Cloud Monitor at {deadline_cloud_monitor_path}. Please ensure Deadline Cloud Monitor is installed correctly and setup the {profile_name} profile again."
+            f"Could not find Deadline Cloud Monitor at {deadline_cloud_monitor_path}."
+            f"Please ensure Deadline Cloud Monitor is installed correctly and set up the {profile_name} profile again."
```
epmog: Based on the branding guidelines, `monitor` should actually be lower-cased here while Deadline Cloud remains capitalized. Might be worth updating more than just here for that.

epmog (changes requested):

src/deadline/client/api/_loginout.py
gahyusuh: Nit: Do you think we should put a space " " between sentences? Same for the other DeadlineOperationError messages.

chocecil: Actually that was one of my questions, because here we were using `M`onitor unlike other places. Will update it.

chocecil: For CLI, it was ok looking at my console message. Im wondering how to put messages with `\n` on the UI, it didn't work although I added an '\n'. Let me double check.

epmog: Gahyun is specifically calling out the no space between the period and the sentence. 

docs: https://docs.python.org/3.12/reference/lexical_analysis.html#string-literal-concatenation

> Multiple adjacent string literals (delimited by whitespace), possibly using different quoting conventions, are allowed, and their meaning is the same as their concatenation. Thus, "hello" 'world' is equivalent to "helloworld".

So in this case the second line is adjacent to the period.

chocecil: Added a space between the period and the next sentence.

gahyusuh: Thank you!

gahyusuh (approved):

epmog (approved):

PR #323, opened 2024-04-30
chore: update labels to sentence casing

epmog:

We were capitalizing every word in every label which doesn't match the style guidelines. We should be using sentence casing where applicable (almost everywhere)

Swapped to sentence casing across the shared UI in the library. There will need to be future changes to individual integrations for their specific widgets.

A more consistent experience across AWS products.

```
hatch run fmt
hatch build
hatch run lint
hatch run test
```

I then also launched the gui-submit in a `hatch shell` to inspect the resulting GUI to ensure I didn't miss any.

N/A

Nope

----

### DIFF
--- a/src/deadline/client/ui/dev_application.py
+++ b/src/deadline/client/ui/dev_application.py
@@ -42,11 +42,11 @@ class DevMainWindow(QMainWindow):
         )
 
     def setup_ui(self):
-        submit = self.menuBar().addMenu("&Submit Job")
-        submit.addAction("Submit Job Bundle...", self.submit_job_bundle)
-        submit.addAction("Submit CLI Job...", self.submit_cli_job)
+        submit = self.menuBar().addMenu("&Submit job")
+        submit.addAction("Submit job bundle...", self.submit_job_bundle)
+        submit.addAction("Submit CLI job...", self.submit_cli_job)
         account = self.menuBar().addMenu("&Account")
-        account.addAction("AWS Deadline Cloud Workstation Configuration...", self.configure)
+        account.addAction("AWS Deadline Cloud workstation configuration...", self.configure)
         account.addAction("Log in to AWS Deadline Cloud...", self.login)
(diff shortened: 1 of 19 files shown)


src/deadline/client/ui/dev_application.py
```diff
@@ -95,7 +95,7 @@ def app() -> None:
     app.setStyle(QStyleFactory.create("fusion"))
 
     # Set the application info.
-    app.setApplicationName("AWS Deadline Cloud Client Test GUI")
+    app.setApplicationName("AWS Deadline Cloud client test gui")
```
mwiebe: GUI could stay capitalized, since it's an acronym.

src/deadline/client/ui/resources/blender_example_bundle/template.yaml
```diff
@@ -24,10 +24,10 @@ parameterDefinitions:
     description: Choose the file format to render as.
     default: PNG
     allowedValues: [TGA, RAWTGA, JPEG, IRIS, IRIZ, PNG, HDR, TIFF, OPEN_EXR, OPEN_EXR_MULTILAYER, CINEON, DPX, DDS, JP2, WEBP]
-  - name: RezPackages
+  - name: CondaPackages
```
mwiebe, line 27: We can include both here, so the job bundle would work with either a Rez queue environment or a Conda queue environment providing the software.

mwiebe (dismissed):

src/deadline/client/ui/dev_application.py
epmog: updated!

src/deadline/client/ui/resources/blender_example_bundle/template.yaml
epmog: Now including both RezPackages and CondaPackages

mwiebe (approved):

jusiskin (approved):

Issue #308, opened 2024-04-19
Feature request: Customizing DCC Submitters (UI, settings, hooks, custom steps)


Ahuge:
### Use Case

This is a discussion around scoping out a feature request that I initially asked in #292 around adding a "pre-submit" and "post-submit" callback.

After discussing with @epmog we have determined that some additional scoping around this is needed and I'd like to solicit opinions on it. 

### Proposed Solution

Initially the PR focused around two hooks that ran after the UI had been created but just before and just after the job was being submitted.

This was designed to solve my own use case of modifying a Nuke scene before a job submission and then reverting the changes for the artist just afterwards.
Our studios had been doing this around the "SubmissionHook.py" in the old Deadline application to strip out ShotGrid Toolkit Write Nodes from Nuke and convert them to generic write nodes so that the farm didn't have to deal with SGTK stuff.

epmog suggested potential features around the ability to create/embed custom gui controls, pre-populating values/overriding certain sticky settings, translate assets into openjd compatible ones, add custom steps/modify existing ones.

I can see these being grouped into "Before UI" and "After UI" options, with the "After UI" one being another chance to modify the scene.

Ahuge:
### Custom GUI Controls
I think the first thing we need to determine here is if it is ok to have all custom controls on a separate tab, or if we need to provide the ability to add custom elements to each existing tab (Shared Job Settings, Job-Specific Settings, Job Attachments, Host Requirements)

#### Shared Job Settings
I don't have any suggestions for what may be added to this section. Please suggest ideas.

#### Job-Specific Settings
This is a DCC submitter provided `QWidget` object however we could allow for a developer provided widget to replace the existing widget.

#### Job Attachments
I don't have any suggestions for what may be added to this section. Please suggest ideas.
I think it is more likely that developers will be adding to the translated OpenJD assets and those will be automatically shown inline here as well.

#### Host Requirements
The main modification that I anticipate seeing here would be setting up default settings within this Host Requirements widget and I see that either via passing this widget and implementing a programmatic api on it that can be set, or allowing these settings to be pre-defined by some sort of developer provided "settings override".



Based on my initial look into the potential UI overrides, I think the most likely place a developer may want to override UI controls is in the Job-Specific Settings. I think we will want to allow them to provide their own UI Settings object as well.

Ahuge:
### Overriding Settings / Default Settings
The `deadline-cloud` already provides settings for everything except the host_requirements, so adding those would be valuable.

[...]

PR #327, opened 2024-05-06
[WIP] Customizing DCC Submitters (UI, settings, hooks, custom steps)

Ahuge:
This PR is my WIP implementation to be discussed in https://github.com/aws-deadline/deadline-cloud/issues/308
This *currently starts* to implement the UI Hook, Post Submission Hook, and the remaining JOB_CALLBACK has not been implemented yet.
This is a major change, however it is intended to be an "opt-in" change where you need to provide callbacks to run, and if none are provided, no change will be noticed.
Not at the testing point yet. Will need to write some tests

----

### DIFF
--- a/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
+++ b/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
@@ -527,7 +527,7 @@ class SubmitJobToDeadlineDialog(QDialog):
             # Execute any PostSubmission function defined.
             if self.on_post_submit_callback:
                 self._call_post_submit_hook(
-                    job_id=self.create_job_response["jobId"],
+                    job_id=self.create_job_response.get("jobId"),
                 )
         except UserInitiatedCancel as uic:
             logger.info("Canceling submission.")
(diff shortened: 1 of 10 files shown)


src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -82,6 +83,8 @@ def __init__(
         auto_detected_attachments: AssetReferences,
         attachments: AssetReferences,
         on_create_job_bundle_callback,
+        on_ui_callback=None,
+        on_post_submit_callback=None,
```
mwiebe, line 87: I know we started down this road of callbacks being passed in here, but I wonder if these two cases are better served via a plugin structure like what @epmog alluded to in https://github.com/aws-deadline/deadline-cloud/issues/308#issuecomment-2103569689 ?

The repo + blogpost https://github.com/wax911/plugin-architecture looks like a decent high level way to think about plugins - with discovery and registration being the key concepts to think about first. With callbacks via the parameter here, it is the code that is using the deadline-cloud library that has to perform any discovery of plugins available.

For `on_post_submit_callback`, does this meet the goals of the callback? If we want to let someone write code that runs on every single job submission, this won't work, because every downstream dependency would need to implement specific code to support that, and other ways of submitting jobs like the CLI `deadline bundle submit` don't use these GUI code flows at all.

If we have a "post submit" action that a plugin can register, and within the deadline-cloud library every path to submitting a job runs the plugins (probably with an explicit opt-out `--no-plugins` option or similar to help with testing whether a plugin is responsible for an issue or not), maybe that's closer to the extensibility we want?

PR #344, opened 2024-06-07
fix: bundles are stored in job_history in their original format

crowecawcaw:
The bundle GUI submitter does not always save asset references or parameter values in the original file format when the bundle is saved to the `job_history` directory.

Job bundles have data files for the template, asset references and parameter values, and these files can be either JSON or YAML. If input job bundle has a JSON template, the saved bundle's template will also by JSON, but the asset references and parameter values will still be saved as YAML.

Update the bundle saving logic so to that asset references and parameter values are saved in the same format as the job template.

This unblocks some other work I'm doing on the Keyshot submitter where Keyshot's Python interpreter can only read and write JSON files. If saved bundle data is all in JSON, the submitter will be able to load it to find previously selected settings.

I created a bundle with a JSON template, opened it with the GUI submitter, changed some parameters and job attachments, then exported the bundle. I verified that the new bundle in `job_history` had all JSON files. I also opened the GUI submitter for the new template and verified it loaded the JSON references and parameters correctly.

No - bug fix.

No

----


src/deadline/client/job_bundle/loader.py
```diff
@@ -116,3 +118,20 @@ def read_yaml_or_json_object(
...
         return None
+
+
+def save_yaml_or_json_to_file(
+    bundle_dir: str,
+    filename: str,
+    file_type: str,
+    data: Any,
+) -> None:
+    with open(
```
epmog: Missing a docstring (and it's public!), we should include something here

src/deadline/client/job_bundle/loader.py
```diff
@@ -116,3 +118,20 @@ def read_yaml_or_json_object(
         return parse_yaml_or_json_content(file_contents, file_type, bundle_dir, filename)
     else:
         return None
+
+
+def save_yaml_or_json_to_file(
```
epmog: Should this be moved to a different module? It's currently under `loader` where everything reads in content. feels like we may need to move it to somewhere like `deadline_yaml_dump` for writing out content. Should its naming be similar, like `yaml_or_json_dump`? Not certain myself. Thoughts?

epmog (changes requested):
Code looks good, just need to ensure it's in the right home! Thanks Stephen

src/deadline/client/job_bundle/loader.py
crowecawcaw: Moved!

lucaseck (dismissed):

epmog (approved):

lucaseck (approved):

PR #354, opened 2024-06-17
feat(asset-cli): added asset subcommand template and placeholder commands

seant-aws:
Additional CLI commands to add exposure to job attachments features

*feat: Create template for the deadline asset subcommands, as well added placeholder commands that print out placeholder text. Added for setting up other commands to implement

New user facing placeholder commands and template of
- snapshot
- upload
- diff
- download

Through user faced testing; checked by calling each command for placeholder output.
Will add further testing throughout development cycle

No, will be further documented after more features.

No, only frontend facing user commands that perform placeholder actions.

### DIFF
--- a/src/deadline/client/cli/_groups/asset_group.py
+++ b/src/deadline/client/cli/_groups/asset_group.py
@@ -1,0 +1,95 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+All the `deadline asset` commands:
+    * snapshot
+    * upload
+    * diff
+    * download
+"""
+
+import click
+
+from .._common import _handle_error
+
+
+@click.group(name="asset")
+@_handle_error
+def cli_asset():
+    """
+    Commands to work with AWS Deadline Cloud Job Attachments.
+    """
+
+
+@cli_asset.command(name="snapshot")
+@click.option("--root-dir", help="The root directory to snapshot. ")
+@click.option("--manifest-out", help="Destination path to directory where manifest is created. ")
+@click.option(
+    "--recursive",
+    "-r",
+    help="Flag to recursively snapshot subdirectories. ",
+    is_flag=True,
+    show_default=True,
+    default=False,
+)
+@_handle_error
+def asset_snapshot(**args):
+    """
+    Creates manifest of files specified root directory.
+    """
+    click.echo("snapshot taken")
+
+
+@cli_asset.command(name="upload")
+@click.option(
+    "--manifest", help="The path to manifest folder of the directory specified for upload. "
+)
+@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
+@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
+@click.option(
+    "--update",
+    help="Flag to update manifest before upload. ",
+    is_flag=True,
+    show_default=True,
+    default=False,
+)
+@_handle_error
+def asset_upload(**args):
+    """
+    Uploads the assets in the provided manifest file to S3.
+    """
+    click.echo("upload done")
+
+
+@cli_asset.command(name="diff")
+@click.option("--root-dir", help="The root directory to compare changes to. ")
+@click.option(
+    "--manifest", help="The path to manifest folder of the directory to show changes of. "
+)
+@click.option(
+    "--format",
+    help="Pretty prints diff information with easy to read formatting. ",
+    is_flag=True,
+    show_default=True,
+    default=False,
+)
+@_handle_error
+def asset_diff(**args):
+    """
+    Check file differences of a directory since last snapshot.
+
+    TODO: show example of diff output
+    """
+    click.echo("diff shown")
+
+
+@cli_asset.command(name="download")
+@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
+@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
+@click.option("--job-id", help="The AWS Deadline Cloud Job to get. ")
+@_handle_error
+def asset_download(**args):
+    """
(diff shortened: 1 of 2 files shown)


src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    click.echo("snapshot taken")
+
+
+@cli_asset.command(name="upload")
+@click.option("--manifest", help="The manifest to ")
+@_handle_error
+def asset_upload(**args):
+    """
+    Preupload files before for a render with specified manifest.
```
jericht: This comment is explaining an example use case of the command rather than what it does. Can we change this to explain what the command does instead? e.g. something like `Uploads the assets in the provided manifest file to S3`)

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,59 @@
...
+@_handle_error
+def asset_snapshot(**args):
+    """
+    Creates manifest of files specified root directory.
+    """
+    click.echo("snapshot taken")
+
+
+@cli_asset.command(name="upload")
+@click.option("--manifest", help="The manifest to ")
```
jericht: `help` text is incomplete

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,59 @@
...
+    Preupload files before for a render with specified manifest.
+    """
+    click.echo("upload done")
+
+
+@cli_asset.command(name="diff")
+@_handle_error
+def asset_diff(**args):
+    """
+    Check file differences of a directory since last snapshot.
```
jericht, line 79: Can we expand on this a bit? Maybe give an example of what the output would be?

If we're still not sure on that, maybe we can have a placeholder with a TODO comment?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    click.echo("diff shown")
+
+
+@cli_asset.command(name="download")
+@_handle_error
+def asset_download(**args):
+    """
+    Downloads input manifest of previously submitted job.
```
jericht, line 93: I think we should remove mentioning "job" here unless we require a `--job-id` parameter, for example.

jericht:
Let's put in all the parameters we know each command will need for now, then we can merge this in.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,63 @@
...
+@_handle_error
+def asset_snapshot(**args):
+    """
+    Creates manifest of files specified root directory.
+    """
+    click.echo("snapshot taken")
+
+
+@cli_asset.command(name="upload")
+@click.option("--manifest", help="The manifest of files to be uploaded. ")
```
godobyte: This is `path to manifest folder` according to design doc? The `help` here is not very descriptive.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,90 @@
...
+@_handle_error
+def cli_asset():
+    """
+    Commands to work with AWS Deadline Cloud Job Attachments.
+    """
+
+
+@cli_asset.command(name="snapshot")
+@click.option("--root-dir", help="The root directory to snapshot. ")
+@click.option("--manifest-out-dir", help="Path to directory for manifest output. ")
```
godobyte: How is this different from other CLIs' manifest option? They all seem to be path to manifest folder according to help. Let's keep them consistent if they're for the same thing.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,90 @@
...
+@cli_asset.command(name="diff")
+@click.option("--root-dir", help="The root directory to compare changes to. ")
+@click.option("--manifest", help="The path to manifest of working directory to show changes of. ")
+@click.option(
+    "--print",
+    help="Pretty prints diff information. ",
+    is_flag=True,
+    show_default=True,
+    default=False,
+)
```
godobyte: Would this be similar to --v (verbosity) use case, or this is about formatting?

seant-aws: This would be a formatting option:
Without --print, it would output changes in a manifest schema/raw json format, while using --print would show in a nicer formatted way, such as with git status

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Made this more clear; it's supposed to be the destination where the created manifest will end up at, in case user wants to specify it.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    Commands to work with AWS Deadline Cloud Job Attachments.
+    """
+
+
+@cli_asset.command(name="snapshot")
+@click.option("--root-dir", help="The root directory to snapshot. ")
+@click.option("--manifest-out-dir", help="Destination path to directory for created manifest. ")
+@click.option(
+    "--r",
```
jericht: nit: For single letter flags, usually we do the single dash, like `-r`. Could we have both `-r` and `--recursive` here? Can also address this later.

src/deadline/client/cli/_groups/asset_group.py
godobyte: In that case, would `--format` makes more sense? According to your reply, this CLI prints output changes regardless of the print flag.

seant-aws: good point tbh, will add

godobyte (approved):

jericht (approved):

PR #357, opened 2024-06-19
feat: support JSON output in bundle GUI submitter

crowecawcaw:
In a DCC integration, I was trying to read out the most recently submitted job settings. The bundle GUI submitter prints a human readable output with the bundle directory path where the settings can be found. but the response is not machine readable. 

Add a JSON output option.

Scripts calling the GUI submitter can easily read back where the job bundle is saved and what the submitted job ID is.

Unit tests and manually running the CLI commands.

Auto-documented in the CLI help.

No - new minor feature.
----

### DIFF
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -325,7 +325,7 @@ def _print_response(
 ):
     if output == "json":
         if submitted:
-            response = {
+            response: dict[str, Any] = {
                 "status": "SUBMITTED",
                 "jobId": job_id,
             }
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -328,11 +328,13 @@ def _print_response(
                 "jobId": job_id,
             }
             if extra_info:
-                response.update({
-                    "jobBundleDirectory": job_bundle_dir,
-                    "parameterValues": parameter_values,
-                    "assetReferences": asset_references,
-                })
+                response.update(
+                    {
+                        "jobBundleDirectory": job_bundle_dir,
+                        "parameterValues": parameter_values,
+                        "assetReferences": asset_references,
+                    }
+                )
             click.echo(json.dumps(response))
         else:
             click.echo(json.dumps({"status": "CANCELED"}))
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 import json
 import logging
 import re
-from typing import Optional
+from typing import Any, Dict, Optional
 
 import click
 from contextlib import ExitStack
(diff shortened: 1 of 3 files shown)


src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -276,10 +290,42 @@ def bundle_gui_submit(job_bundle_dir, browse, **args):
...
+    parameter_values: Optional[list[Dict[str, Any]]],
+    asset_references: Dict[str, Any],
+):
+    if output == "json":
+        if submitted:
+            response = {
+                "status": "SUBMITTED",
+                "jobId": job_id,
+                "jobBundleDirectory": job_bundle_dir,
+                "parameterValues": parameter_values,
```
lucaseck: Is it possible to hide some of these extra fields behind additional flags combined with `--output json` e.g. `--output json --output-parameter-values --output-asset-references` or something to that effect. That way only what is actually required is being included and the default behaviour is to give the exact same information as the human readable format?

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -409,6 +410,19 @@ def on_submit(self):
                 settings.submitter_name, settings.name
             )
 
+            # First filter the queue parameters to exclude any from the job template,
+            # then extend it with the job template parameters.
+            job_parameter_names = {param["name"] for param in settings.parameters}
```
lucaseck, line 415: It's unclear to me from the PR description why this is needed. Is this for the same requirement of printing out additional information or is this a separate issue being fixed?

crowecawcaw: This block is copied from here: https://github.com/aws-deadline/deadline-cloud/blob/b5de504918cc6420d80d5206f9332cb8880471a5/src/deadline/client/ui/job_bundle_submitter.py#L101-L111

I looked at moving this code block into the core submitter class instead of as part of a callback the submitter caller provides, but it appears that the submitter class is a public interface that other things depend on, so I opted not to clean it up. Copying this code block seemed like a practical path forward.

lucaseck: I guess the part I'm curious about is what use case was failing here and is it now confirmed that it's working because of this change? I want to make sure nothing else is missing that would be needed in here like the merging of queue and job parameters etc from that other file.

crowecawcaw: I took a closer look. In the linked code, `parameter_values` contains the merged values and `parameters` contains the merged parameter definitions. I copied over the value merging logic which is sufficient. The linked code has a call to `merge_queue_job_parameters` but (confusingly) it's used for merging definitions for the UI, not the values. Since we're not storing the parameter definitions, we don't need to merge or handle them.

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -276,10 +300,51 @@ def bundle_gui_submit(job_bundle_dir, browse, **args):
...
+    submitted: bool,
+    job_bundle_dir: str,
+    job_id: Optional[str],
+    parameter_values: Optional[list[Dict[str, Any]]],
+    asset_references: Dict[str, Any],
+):
+    if output == "json":
+        if submitted:
+            response: dict[str, Any] = {
+                "status": "SUBMITTED",
```
lucaseck, line 329: I think we should still include the job bundle directory here and not in the extra info like we do for the verbose output. Keeps their defaults the same.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
lucaseck: Okay, so just the values being merged and that's all we need. Thanks for double checking.

src/deadline/client/cli/_groups/bundle_group.py
lucaseck: Looks like this was moved behind `--extra-info` so I'll resolve this thread.

moorec-aws (approved):

lucaseck (approved):

PR #360, opened 2024-06-20
fix: bundle gui-submit fails loading bundles with saved queue parameter values

evanspearman-a:

The bundle submitter GUI will fail to open a job bundle if a parameter has a value in `parameter_values` but the parameter does not exist in the template.

The error was caused when trying to generate a GUI component for the parameter. `parameter_values` does not contain type information for the parameter, only the value. The code that generates a component for a parameter assumes there is type information for all parameters. If type information is not found, we simply skip generating a component for that parameter. Note that we still want to retain these parameter values as they can still be used to populate queue parameter values.

Job bundles as described above can now be opened in the bundle submission GUI. In particular, job bundles generated by the integrated submitters when queue environment parameters were present will now work with the job bundle submission GUI.

I opened a job bundle that was previously not opening.

No

No

----

### DIFF
--- a/src/deadline/client/ui/widgets/openjd_parameters_widget.py
+++ b/src/deadline/client/ui/widgets/openjd_parameters_widget.py
@@ -130,6 +130,13 @@ class OpenJDParametersWidget(QWidget):
             if ":" in parameter["name"]:
                 continue
 
+            # Skip any parameters that do not have a type defined.
+            # This can happen when a queue environment parameter was
+            # saved to the bundle, but the template itself does not contain
+            # that parameter.
+            if "type" not in parameter:
+                continue
+
             control_type_name = get_ui_control_for_parameter_definition(parameter)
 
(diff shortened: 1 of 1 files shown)


jusiskin (changes requested):
Looks good, but can we please rename the PR and commit title as it should appear in the release notes?

jusiskin (approved):

crowecawcaw:
You mentioned that you can now open bundles with queue parameters, but can you also submit them?

crowecawcaw (approved):

crowecawcaw:
This would be better with a unit test.

evanspearman-a:
> You mentioned that you can now open bundles with queue parameters, but can you also submit them?

Yes

> This would be better with a unit test.

This change is in the UI code. This project isn't set up for unit tests involving UI code. I looked into it, but there are some things that would need to be figured out with respect to Qt bindings in order to get that to work. Given that this is a fairly trivial change, I don't think it's worth figuring that out to test this particular change.

PR #364, opened 2024-06-21
ci: Add req to toml file 

AlexTranAmz:

- Integ workflow of this package could not be run successful due to missing dependency

- Error message:

image

- After deep dive into the problem, it turn out that Pytest was missing:

```
Checking dependencies
/bin/sh: pytest: command not found
```

This is due to the pre install file do not contain enough dependencies.

We add the pre install file that needed for unittest to the list of requirement before running Integ test. This will ensure the tests have all dependencies needed to work correctly.
The Integ workflow will now run successfully.
`./pipeline/integ.sh`
NA
No

----

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -47,7 +47,8 @@ make_exe = "python scripts/pyinstaller/make_exe.py --output {env:OUT_FILE}"
 
 [envs.integ]
 pre-install-commands = [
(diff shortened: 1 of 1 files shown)


leongdl:
Can we please put more details into the PR description? It will help all users what we are adding.

jericht (approved):

leongdl (approved):

PR #366, opened 2024-06-24
feat: Squish Automated Tests - Dev Submit GUI Settings flow

rena-cheng:
We currently do not have automated user-facing tests for the Submitters dialogue which would catch regressions/bugs from being released to customers. 

Implement/add tests for the Submitters dialogue using Squish Automated Test Framework. This is a first set of automated tests for the Submitters dialogue, and this is PR focuses on the Dev Submit GUI - Settings flow/config.

Currently none as we don't have a CI/CD pipeline set up (yet). As this repo is open-source, users will have access to these tests (but probably won't be able to run it, yet, as they would need Squish licenses, and probably a ReadMe on how to run these tests).  

I ran my tests locally and ensured that the tests passed with no issues.  

I will have a quip doc/draft of Squish Automated framework, how to get set up with it, debug, etc. 

No.

----

### DIFF
--- a/.github/workflows/mainline_integration_test.yml
+++ b/.github/workflows/mainline_integration_test.yml
@@ -15,3 +15,4 @@ jobs:
       repository: ${{ github.event.repository.name }}
       branch: mainline
       environment: mainline
+      os: linux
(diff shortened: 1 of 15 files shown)


rena-cheng:
Resubmitting new Pull Request.

PR #369, opened 2024-06-24
feat(asset-cli): asset snapshot subcommand to capture manifests

seant-aws:
Allow users to call CLI command to capture directories of assets into manifests.

Added deadline asset snapshot subcommand to allow capture of manifests. Includes:
- capturing manifest file into specified directory of assets
- recursive capture of directories
- can specify destination of where manifest can be created

Exposes functionality of creating manifests

- Added unit tests mocking snapshot functionality at `test/unit/deadline_client/cli/test_cli_asset.py`, passing successfully
- Added integ tests with temporary files in filesystem at `test/integ/cli/test_cli_asset.py`, passing successfully
- Manual end to end tests with sample directories with assets:

````
deadline-cloud-stangch % hatch run deadline asset snapshot --root-dir /Users/stangch/Desktop/test_asset -r
Manifest creation path defaulted to /Users/stangch/Desktop/test_asset 

Hashing Summary:
    Processed 5 files totaling 7.5 KB.
    Skipped re-processing 1 files totaling 37.0 B.
    Total processing time of 0.00932 seconds at 804.07 KB/s.

Manifest created at /Users/stangch/Desktop/test_asset
````

N/A

N/A

----

### CODE REVIEW

File: src/deadline/client/api/_submit_job_bundle.py [hunk 1]
```diff
@@ -396,7 +396,7 @@ def wait_for_create_job_to_complete(
     )
 
 
-def _hash_attachments(
+def hash_attachments(
```
jericht, line 399: Hmm if we are marking this function as "public", I think it should live in a different file. It feels weird to have an asset hashing function live in the `_submit_job_bundle` module. Especially given the description for the `api` submodule from https://github.com/aws-deadline/deadline-cloud/blob/feature_assets_cli/DEVELOPMENT.md#srcclientapi:

> This submodule contains utilities to call boto3 in a standardized way using an aws profile configured for AWS Deadline Cloud, helpers for working with the AWS Deadline Cloud monitor login/logout, and objects representing AWS Deadline Cloud resources.

Hashing assets is not directly related to working with the Deadline Cloud API. How about making this a function inside `job_attachments`?

src/deadline/client/cli/_common.py
```diff
@@ -176,3 +182,42 @@ def _cli_object_repr(obj: Any):
...
+        total_progress = int(upload_metadata.progress)
+        new_progress = total_progress - self._upload_progress.pos
+        if new_progress > 0:
+            self._upload_progress.update(new_progress)
+
+        if total_progress == self._length or not sigint_handler.continue_operation:
+            self._bar_status = self.BAR_CLOSED
+            self._exit_stack.close()
+
+        return sigint_handler.continue_operation
```
jericht, line 224: If the intention is to use this class in other modules, let's add it to the `__all__` list above.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -7,10 +7,19 @@
...
 
-from .._common import _handle_error
+from deadline.client import api
+from deadline.job_attachments.upload import S3AssetManager, S3AssetUploader
+from deadline.job_attachments.models import JobAttachmentS3Settings
+
+from .._common import _handle_error, _ProgressBarCallbackManager
+from ...exceptions import NonValidInputError
+
+IGNORE_FILE: str = "manifests"
```
jericht: Is this variable being used anywhere?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,66 @@ def cli_asset():
     default=False,
 )
 @_handle_error
-def asset_snapshot(**args):
+def asset_snapshot(root_dir, manifest_out, recursive, **args):
```
jericht: Can we add types to these parameters?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,66 @@ def cli_asset():
...
     """
     Creates manifest of files specified root directory.
     """
-    click.echo("snapshot taken")
+    root_dir_basename = os.path.basename(root_dir) + "_"
```
jericht: This seems a bit strange. Why are we appending an `_` to the root dir base name? It looks like this is used further down in the `asset_uploader._write_local_input_manifest` call. Should we instead update that function to do this `_` appending? It would be easier to understand why we are doing it if it lives in the same code that uses it IMO.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,66 @@ def cli_asset():
...
+
+    if not os.path.isdir(root_dir):
+        misconfigured_directories_msg = f"Specified root directory {root_dir} does not exist. "
+        raise NonValidInputError(misconfigured_directories_msg)
+
+    if manifest_out and not os.path.isdir(manifest_out):
+        misconfigured_directories_msg = (
+            f"Specified destination directory {manifest_out} does not exist. "
+        )
+        raise NonValidInputError(misconfigured_directories_msg)
```
jericht: Do we need these `misconfigured_directories_msg` variables? Can we just raise the error with the string directly?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,66 @@ def cli_asset():
...
+    elif manifest_out is None:
+        manifest_out = root_dir
+
+    inputs = []
+    for root, dirs, files in os.walk(root_dir):
+        if os.path.basename(root).endswith("_manifests"):
+            continue
```
jericht: Is this the behavior we want? Are we trying to skip the folders created by Job Attachments that contain the manifests here? This will also skip a folder called `my_super_important_not_job_attachments_manifests`, which is probably not what we want to skip.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,66 @@ def cli_asset():
...
+        for file in files:
+            file_full_path = str(os.path.join(root, file))
+            inputs.append(file_full_path)
```
jericht: nit: this can be simplified to:
```suggestion
        inputs.extend(files)
```

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,66 @@ def cli_asset():
...
+        if not recursive:
+            break
+
+    # Placeholder Asset Manager
+    asset_manager = S3AssetManager(
+        farm_id=" ", queue_id=" ", job_attachment_settings=JobAttachmentS3Settings(" ", " ")
+    )
```
jericht, line 67: If we don't need the `S3AssetManager`'s fully functionality, can we refactor the functionality we do need out of it and call that from here instead? Creating an "empty" instance is probably not how we should go about this.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,66 @@ def cli_asset():
...
+        if asset_root_manifests.asset_manifest is None:
+            continue
+        source_root = Path(asset_root_manifests.root_path)
+        file_system_location_name = asset_root_manifests.file_system_location_name
+        (_, _, manifest_name) = asset_uploader._gather_upload_metadata(
+            asset_root_manifests.asset_manifest, source_root, file_system_location_name
+        )
+        asset_uploader._write_local_input_manifest(
+            manifest_out, manifest_name, asset_root_manifests.asset_manifest, root_dir_basename
+        )
```
jericht: Couple comments:
1. This seems like a good candidate for a common function like `write_manifest`. We are doing a similar operation elsewhere in the code that could be changed to use this common function as well.
2. nit: Using functionality from the `S3AssetUploader` class even though we aren't uploading anything feels weird. Can you consider refactoring the relevant code out of the uploader class into common functions?

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,109 @@
...
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        yield tmpdirname
+
+
+def test_cli_asset_snapshot_valid_root_dir(temp_dir):
+    root_dir = os.path.join(temp_dir, "root_dir")
+    os.makedirs(root_dir)
+    file_path = os.path.join(root_dir, "file.txt")
+    with open(file_path, "w") as f:
+        f.write("test file content")
```
jericht: I think tests that actually create files/folders should be considered as integration tests. Can we move tests that do this into a dedicated test file under the `test/integ` folder?

For the unit test, we should be mocking out any filesystem operations. If there are code paths that have filesystem operations that we run in multiple test cases, consider using fixtures to mock out the relevant function calls to use across all the tests that need them

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,109 @@
...
+    manifest_file_path = os.path.join(manifest_folder_path, manifest_file_name)
+
+    with open(manifest_file_path, "r") as f:
+        manifest_data = json.load(f)
+        print(manifest_data["paths"][0]["hash"])
+        assert manifest_data["paths"][0]["path"] == "file.txt"
+        assert manifest_data["paths"][0]["hash"] == "0741993e50c8bc250cefba3959c81eb8"
+
+
+def test_cli_asset_snapshot_invalid_root_dir(temp_dir):
```
jericht: For tests, it's nice to have an idea of what the test is asserting at a glance. You can do this a couple ways:

1. (My preferred way) having the test function name represent the state and expected outcome. For example, I would rename this test as `test_cli_asset_snapshot_raises_when_given_invalid_root_dir`. This name is quite long, so I normally split up my test files into classes that represent a "category" of tests. e.g.,

```
class TestSnapshot:
    def test_snapshots_directory(): pass  # regular happy path test case
    def test_raises_on_invalid_root_dir(): pass  # this test
    ...etc.
```

2. Add a docstring to the test function explaining what it is testing

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # Check manifest file details to match correct content
+    # since manifest file name is hashed depending on source location, we have to list out manifest
+    manifest_folder_path = os.path.join(root_dir, f"{os.path.basename(root_dir)}_manifests")
+    manifest_file_name = os.listdir(manifest_folder_path)[0]
+    manifest_file_path = os.path.join(manifest_folder_path, manifest_file_name)
+
+    with open(manifest_file_path, "r") as f:
+        manifest_data = json.load(f)
+        print(manifest_data["paths"][0]["hash"])
```
jericht: I don't think we need this print statement here right? Can we remove this?

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,109 @@
...
+        assert manifest_data["paths"][0]["path"] == "file.txt"
+        assert manifest_data["paths"][0]["hash"] == "0741993e50c8bc250cefba3959c81eb8"
```
jericht: nit: Can you add a comment explaining how this hardcoded hash was generated? (which hash algorithm was used). Same for the other hardcoded hashes we have.

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,109 @@
...
+    os.makedirs(os.path.join(root_dir, "subdir1", "subdir2"))
+
+    # Create a file in the root directory
+    root_file_path = os.path.join(root_dir, "root_file.txt")
+    with open(root_file_path, "w") as f:
+        f.write("root file content")
+
+    # Create a file in the subdirectory
+    subdir_file_path = os.path.join(root_dir, "subdir1", "subdir2", "subdir_file.txt")
+    os.makedirs(os.path.dirname(subdir_file_path), exist_ok=True)
```
jericht: This should be unnecessary since we are creating the subdirectory in the above line 54. Let's choose either one. I think this one is cleaner since the subdirectory and file are created one after another

src/deadline/client/cli/_groups/asset_group.py
jericht: Also, if the user specified a `--manifest-out` argument, do we still want to ignore these?

jericht (changes requested):

src/deadline/client/api/_submit_job_bundle.py
seant-aws: That makes sense, but inside the function `hash_attachments` does call `hash_assets_and_create_manifest`, which is the actual job attachments functionality. While it doesn't directly call boto3, it does act like an API which calls the actual hashing functionality. It also works with callback info, which belongs in client ? I think it can stay in client api, but maybe under a new file ?

src/deadline/client/cli/_groups/asset_group.py
seant-aws: was supposed to include this during the parsing instead of just comparing to a string, slipped my mind

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Right, I chose not to do so because `_write_local_input_manifest` is used by other parts of the code, and they already have a standard for passing the parameter name of manifests. I'm adding the '\_' to make a new name for manifests, and that only happens in snapshot. E.g. "stangch_assets" + "\_" + "manifests" for snapshot, while job history only uses it while named "manifests"
It makes sense to move it closer to where it actually gets used though, so I moved it down ?

src/deadline/client/cli/_groups/asset_group.py
seant-aws: I'm following what the `_submit_job_bundle.py` does with inputs, but changing the error `NonValidInputError`, which is a client error. It seems redundant because I only snapshot one file though, so should I follow the style of other functions or make it less redundant ?

src/deadline/client/cli/_groups/asset_group.py
seant-aws: yes, the assumption is that users wont name things ending in `_manifest`, I will make this more robust and check for `root-dir-name` + `manifest`, but I think it's important to stay because subdirectories can also have manifests, and I think its uneccessary to upload it as file data, if it's already being uploaded as a manifest

also `--manifest-out` shouldn't change things, because users could do a first snapshot on the rootdir without specifying, then a second time with `--manifest-out`, and the manifests folder should still be ignored

seant-aws: spoke with jericht, will remove for now, as enforcing files named manifest may be unsafe

src/deadline/client/api/_submit_job_bundle.py
seant-aws: will add to future refactoring

src/deadline/client/cli/_groups/asset_group.py
seant-aws: added to future refactor

src/deadline/client/cli/_groups/asset_group.py
seant-aws: added to future refactor

src/deadline/client/cli/_groups/asset_group.py
jericht: Looks like this is still around. Can we remove it?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +42,61 @@ def cli_asset():
     default=False,
 )
 @_handle_error
-def asset_snapshot(**args):
+def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
```
jericht, line 45: Question for the type annotation: Is `manifest_out` always a `str`? Is it a different type when the user does not provide `--manifest-out`?

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Ah my mistake

src/deadline/client/api/_submit_job_bundle.py
godobyte: Is there any conversation offline? The conversation here doesn't seem to connect. If so, please add summary on what was discussed for visibility.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +40,61 @@ def cli_asset():
...
     Creates manifest of files specified root directory.
     """
-    click.echo("snapshot taken")
+    if not os.path.isdir(root_dir):
+        raise NonValidInputError(f"Specified root directory {root_dir} does not exist. ")
+
+    if manifest_out and not os.path.isdir(manifest_out):
+        raise NonValidInputError(f"Specified destination directory {manifest_out} does not exist. ")
+    elif manifest_out is None:
+        manifest_out = root_dir
```
godobyte, line 55: Could we give caller some visibility for this, maybe via a print message? 

Speaking of which, this asset implementation could use more communication to caller in general.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -33,11 +40,61 @@ def cli_asset():
...
+
+    # Placeholder Asset Manager
+    asset_manager = S3AssetManager(
+        farm_id=" ", queue_id=" ", job_attachment_settings=JobAttachmentS3Settings(" ", " ")
+    )
+    asset_uploader = S3AssetUploader()
+    hash_callback_manager = _ProgressBarCallbackManager(length=100, label="Hashing Attachments")
+
+    upload_group = asset_manager.prepare_paths_for_upload(
+        input_paths=inputs, output_paths=[root_dir], referenced_paths=[]
```
godobyte, line 72: Curious - what is this `referenced_paths`?

src/deadline/job_attachments/upload.py
```diff
@@ -182,7 +179,7 @@ def upload_assets(
 
         if manifest_write_dir:
             self._write_local_manifest(
-                manifest_write_dir, manifest_name, full_manifest_key, manifest
+                manifest_write_dir, manifest_name, full_manifest_key, manifest, None
```
godobyte: What about adding default to the function parameter rather than passing `None` from the caller?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,126 @@
...
+
+from deadline.client.cli import main
+
+
+class TestSnapshot:
+
+    @pytest.fixture
+    def temp_dir(self):
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            yield tmpdirname
```
godobyte: Nit - the naming is a bit confusing, I assume it's tmpdir rather than name?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,126 @@
...
+        # since manifest file name is hashed depending on source location, we have to list out manifest
+        manifest_folder_path = os.path.join(root_dir, f"{os.path.basename(root_dir)}_manifests")
+        manifest_file_name = os.listdir(manifest_folder_path)[0]
+        manifest_file_path = os.path.join(manifest_folder_path, manifest_file_name)
+
+        with open(manifest_file_path, "r") as f:
+            manifest_data = json.load(f)
+            assert manifest_data["paths"][0]["path"] == "file.txt"
+            assert (
+                manifest_data["paths"][0]["hash"] == "0741993e50c8bc250cefba3959c81eb8"
```
godobyte: How did you get this hash, could it be a call with the input? i.e. we don't need to update the hash if there is any change to the input.

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,126 @@
...
+        root_manifest_file_name = [file for file in os.listdir(manifest_folder_path)][0]
+        root_manifest_file_path = os.path.join(manifest_folder_path, root_manifest_file_name)
+        with open(root_manifest_file_path, "r") as f:
+            manifest_data = json.load(f)
+            assert manifest_data["paths"][0]["path"] == "root_file.txt"
+            assert (
+                manifest_data["paths"][0]["hash"] == "a5fc4a07191e2c90364319d2fd503cc1"
+            )  # hashed with xxh128
+            assert manifest_data["paths"][1]["path"] == posixpath.join(
+                "subdir1", "subdir2", "subdir_file.txt"
```
godobyte: In general for tests, the test input can be defined as a variable and pass to the call and assertion. Take a look at how it's done in other test cli implementation.

src/deadline/client/cli/_groups/asset_group.py
seant-aws: There's some comments on this page: https://github.com/ddneilson/deadline-cloud-samples/blob/the_job_bundle/job_bundles/README.md

tldr: it's neither input nor output, but in case job references the name of the folder; in the case of snapshot + upload, the main goal is to upload files we care about, so I don't think referenced_paths are relevant here.

src/deadline/client/cli/_groups/asset_group.py
seant-aws: I think the click library doesn't allow users to specify optional in the function params, rather in the above @click.options, so I added a default=None for --manifest-out

src/deadline/client/cli/_groups/asset_group.py
seant-aws: I've added prints for defaulting the manifest out dir, as well as added a print when function finishes, notifying user where manifest folder is created. 
There are also already hashing prompts notifying user of hashing progress, with callbacks

godobyte:
This CLI implementation seems to capture the basic functionality. Were you able to call it and verify? If so, could you add to testing evidence?

src/deadline/client/api/_submit_job_bundle.py
seant-aws: Refactoring issues were discussed to be put in another commit to be addressed, as there are quite a few things that can be refactored with existing job attachments code, and don't really align with the snapshot functionality.

godobyte (approved):

src/deadline/client/cli/_groups/asset_group.py
jericht: Right, so the typing should say it can either be `None` or `str` right?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+        with open(file_path, "w") as f:
+            f.write(TEST_FILE_CONTENT)
+
+        runner = CliRunner()
+        runner.invoke(main, ["asset", "snapshot", "--root-dir", root_dir])
+
+        # Check manifest file details to match correct content
+        # since manifest file name is hashed depending on source location, we have to list out manifest
+        manifest_folder_path = os.path.join(root_dir, f"{os.path.basename(root_dir)}_manifests")
+        manifest_file_name = os.listdir(manifest_folder_path)[0]
```
jericht: This seems flaky at first glance. This test relies on there being exactly one file in the `manifest_folder_path` right? Can we add an assertion here to check that there is only exactly one file in this directory, so that if this test fails because of this, we know it was due to an incorrect assumption rather than a regression in the code?

We can probably do something like:
```
manifest_files = os.listdir(manifest_folder_path)
assert len(manifest_files) == 1, f"Expected exactly one manifest file, but got {len(manifest_files)}"
manifest_file_name = manifest_files[0]
```

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+        manifest_file_path = os.path.join(manifest_folder_path, manifest_file_name)
+
+        with open(manifest_file_path, "r") as f:
+            manifest_data = json.load(f)
+            assert manifest_data["paths"][0]["path"] == TEST_ROOT_FILE
+            assert manifest_data["paths"][0]["hash"] == hash_file(
+                file_path, HashAlgorithm()
+            )  # hashed with xxh128
```
jericht: nit: do we still need the manifest file opened for the `assert` statements?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def test_raises_on_invalid_root_dir(self, temp_dir):
+        """
+        Snapshot with invalid root directory, raises exception
+        """
+        invalid_root_dir = os.path.join(temp_dir, "invalid_root_dir")
+        runner = CliRunner()
+        result = runner.invoke(main, ["asset", "snapshot", "--root-dir", invalid_root_dir])
+        assert result.exit_code == 1
```
jericht: Couple nits:
1. What makes the root dir we're providing invalid? Can we make that clear in the test description/name? (or maybe even the variable name for the directory?)
2. Can we assert beyond just checking the exit code? How are we sure it did not fail due to some other regression?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+            f.write(TEST_SUB_DIR_FILE_CONTENT)
+
+        runner = CliRunner()
+        runner.invoke(main, ["asset", "snapshot", "--root-dir", root_dir, "--recursive"])
+
+        # Check manifest file details to match correct content
+        # since manifest file name is hashed depending on source location, we have to list out manifest
+        manifest_folder_path = os.path.join(root_dir, f"{os.path.basename(root_dir)}_manifests")
+        root_manifest_file_name = [file for file in os.listdir(manifest_folder_path)][0]
+        root_manifest_file_path = os.path.join(manifest_folder_path, root_manifest_file_name)
```
jericht: Similar to my other comment, I think we can reduce the potential flakiness here. Let's assert that there is exactly one manifest file in the manifest dir before proceeding with the test.

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+            assert manifest_data["paths"][0]["path"] == TEST_ROOT_DIR
+            assert manifest_data["paths"][0]["hash"] == hash_file(
+                root_file_path, HashAlgorithm()
+            )  # hashed with xxh128
+            assert manifest_data["paths"][1]["path"] == posixpath.join(
+                TEST_SUB_DIR_1, TEST_SUB_DIR_2, TEST_SUB_FILE
+            )
+            assert manifest_data["paths"][1]["hash"] == hash_file(
+                subdir_file_path, HashAlgorithm()
+            )  # hashed with xxh128
```
jericht: Same comment here as before, do we need the `assert` statements to be inside this `with` block, keeping the file open?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+            main, ["asset", "snapshot", "--root-dir", root_dir, "--manifest-out", manifest_out_dir]
+        )
+
+        # Check manifest file details to match correct content
+        # since manifest file name is hashed depending on source location, we have to list out manifest
+        manifest_folder_path = os.path.join(
+            manifest_out_dir, f"{os.path.basename(root_dir)}_manifests"
+        )
+        manifest_file_name = os.listdir(manifest_folder_path)[0]
+        manifest_file_path = os.path.join(manifest_folder_path, manifest_file_name)
```
jericht: Same here, let's assert our assumption that there is exactly one manifest file is correct

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,171 @@
...
+        subdir1.mkdir()
+        subdir2.mkdir()
+        (subdir1 / "file1.txt").touch()
+        (subdir2 / "file2.txt").touch()
+
+        with patch.object(
+            S3AssetManager, "prepare_paths_for_upload"
+        ) as prepare_paths_mock, patch.object(
+            api, "hash_attachments", return_value=(Mock(), [])
+        ) as hash_attachments_mock:
```
jericht: I see we've applied both of these mocks three times. Should we consider using [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) for these mocks?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        invalid_root_dir = os.path.join(temp_dir, "invalid_root_dir")
+        runner = CliRunner()
+        result = runner.invoke(main, ["asset", "snapshot", "--root-dir", invalid_root_dir])
+        assert result.exit_code == 1
+
+    def test_pass_valid_root_dir_recursive(self, temp_dir):
+        """
+        Snapshot with valid root directory and recursively
+        """
```
jericht: Should we have a test case for the opposite of this? When we do not provide `--recursive`, the command does not recurse into subdirectories?

test/integ/cli/test_cli_asset.py
```diff
@@ -0,0 +1,139 @@
...
+
+TEST_ROOT_FILE = "root_file.txt"
+TEST_SUB_FILE = "subdir_file.txt"
+
+TEST_ROOT_DIR = "root_dir"
+TEST_SUB_DIR_1 = "subdir1"
+TEST_SUB_DIR_2 = "subdir2"
+
+
+class TestSnapshot:
```
jericht, line 29: I noticed that the unit tests and integration tests have the exact same test cases, except the unit tests use mocks. Could we tighten up the integration test cases a bit more to focus on testing actual integration points (e.g. reading/writing files)?

e.g. we probably do not needed the test cases here for when the command fails due to an invalid root directory, for example, since that is already covered by the unit tests.

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,171 @@
...
+            prepare_paths_mock.return_value = upload_group_mock
+
+            runner = CliRunner()
+            result = runner.invoke(main, ["asset", "snapshot", "--root-dir", root_dir])
+
+            assert result.exit_code == 0
+            prepare_paths_mock.assert_called_once_with(
+                input_paths=[str(temp_file)], output_paths=[root_dir], referenced_paths=[]
+            )
+            hash_attachments_mock.assert_called_once()
```
jericht: Do we still need the `patch.object` with blocks active while we are asserting the test results?

test/integ/cli/test_cli_asset.py
jericht: Please see https://github.com/aws-deadline/deadline-cloud/pull/369#discussion_r1659496238 before addressing this in the integration tests (just address it in the unit tests)

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,157 @@
...
+    def test_snapshot_root_directory_only(
+        self, tmp_path, mock_prepare_paths_for_upload, mock_hash_attachments, upload_group_mock
+    ):
+        """
+        Tests if CLI snapshot command calls correctly with an exiting directory path at --root-dir
+        """
+        root_dir = str(tmp_path)
+
+        temp_file = tmp_path / "temp_file.txt"
+        temp_file.touch()
```
jericht, line 56: nit: Ideally we shouldn't be creating files/dirs in our unit tests, but we can address this later.

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,157 @@
...
+                "--manifest-out",
+                str(manifest_out_dir),
+            ],
+        )
+
+        assert result.exit_code == 0
+        mock_prepare_paths_for_upload.assert_called_once_with(
+            input_paths=[str(temp_file)], output_paths=[root_dir], referenced_paths=[]
+        )
+        mock_hash_attachments.assert_called_once()
```
jericht, line 113: Looks like we're missing an assertion here to verify the command would have actually written the manifest file to `manifest_out_dir`

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -0,0 +1,157 @@
...
+    def test_invalid_root_directory(self, tmp_path):
+        """
+        Tests if CLI snapshot raises error when called with an invalid --root-dir with non-existing directory path
+        """
+        invalid_root_dir = str(tmp_path / "invalid_dir")
+
+        runner = CliRunner()
+        result = runner.invoke(main, ["asset", "snapshot", "--root-dir", invalid_root_dir])
+
+        assert result.exit_code != 0
```
jericht, line 78: nit: if we know what the expected exit code should be, we should assert equality on that expected value rather than checking for non-zero exit code.

test/integ/cli/test_cli_asset.py
jericht: It looks like this is still unaddressed. We can fix this in a future PR but it should be quick to add in here. Consider how this test will behave if there is not exactly one file in the `manifest_folder_path` (will we be able to quickly identify the issue is due to a test assumption being false rather than a regression in the code?)

jericht (approved):
There's still some comments I have on the tests that I think we should address, but we can do that in a future PR. Approved.

jericht (approved):

PR #371, opened 2024-06-25
fix: Implement s3 cache integrity checking on submission time.

benl-2023:
* When assets in S3 get deleted, it can result in the local S3 cache becoming invalid. That is, assets may fail to be uploaded if the client believes an asset is already uploaded, when in reality it is no longer present in S3.
* This results in failed jobs, as workers will not be able to find these assets in S3.

* The solution is to implement sampled cached asset inspections to verify that the contents in the scope of a submission that are cached, are also present in S3.
  * When assets are cached but not present in S3, the client resets its cache. 
* When the cache is reset, all assets will be re-checked for upload.

* Reduces the likelihood of jobs failing due to missing assets in S3.

**The following setup was done:**
* The local submission host was cleared of its caches.
* The S3 job attachments bucket was cleared of all objects.
* A job with 5 attachments was submitted to a test queue. 
  * One of the 5 attachments was deleted from the Job Attachments bucket
  * The job was re-submitted, and it was validated that the deleted attachment again re-appeared in the Job Attachments bucket.

No. This is a bug fix / improvement.

No.

----

### DIFF
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -17,7 +17,7 @@ from datetime import datetime
 from io import BufferedReader, BytesIO
 from math import trunc
 from pathlib import Path, PurePath
-from typing import Any, Callable, Generator, Optional, Tuple, Type, Union
+from typing import Any, Callable, Generator, List, Optional, Tuple, Type, Union
 
 import boto3
 from boto3.s3.transfer import ProgressCallbackInvoker
@@ -174,6 +174,8 @@ class S3AssetUploader:
             A tuple of (the partial key for the manifest on S3, the hash of input manifest).
         """
 
+        logger.debug("S3AssetUploader.upload_assets")
+
         # Upload asset manifest
         (hash_alg, manifest_bytes, manifest_name) = S3AssetUploader._gather_upload_metadata(
             manifest=manifest,
@@ -207,6 +209,9 @@ class S3AssetUploader:
                 key=full_manifest_key,
                 extra_args=manifest_metadata,
             )
+
+        # Reset if there is mismatch between cache and actual in S3
+        self.reset_s3_check_cache(s3_check_cache_dir)
 
         # Upload assets
         self.upload_input_files(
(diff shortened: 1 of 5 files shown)


mwiebe:
>When the cache is reset, all assets will be uploaded.

I think you mean that all assets will be re-checked with s3:HeadObject, right? Assets that are on S3 will only get this check, not be re-uploaded.

benl-2023:
> I think you mean that all assets will be re-checked with s3:HeadObject, right? Assets that are on S3 will only get this check, not be re-uploaded.

Yes, that is correct.

benl-2023:
Regarding **Quality Gate Failed Quality Gate failed**

I believe the context for the use of random.shuffle() function is not security sensitive and is a good approach for generating a random sample of assets.

src/deadline/job_attachments/upload.py
```diff
@@ -208,6 +209,15 @@ def upload_assets(
...
             )
 
+        # Verify S3 hash cache integrity, and reset cache if cached files are missing
+        if not self.verify_hash_cache_integrity(
+            s3_check_cache_dir,
+            manifest,
+            job_attachment_settings.full_cas_prefix(),
+            job_attachment_settings.s3BucketName,
+        ):
+            self.reset_s3_check_cache(s3_check_cache_dir)
```
jairaws, line 219: Q: What's the UX if an exception is thrown by this function?

benl-2023: Good question, it turns out it logs the error but the rest of the submission is proceeded with. This isn't what we want because it will result in a job that can't be successful since not all the assets were uploaded.

benl-2023: I've updated the PR such that in `src/deadline/job_attachments/caches/cache_db.py` we now raise the exception thats thrown when we fail to remove the cache file. That is the core place in the call chain from `self.reset_s3_check_cache(s3_check_cache_dir)` where an exception can occur.

The UX here is that the job submission is stopped and the submission CLI exits. The error will be printed to the terminal/logs

jairaws (approved):

src/deadline/job_attachments/upload.py
```diff
@@ -382,6 +392,78 @@ def upload_input_files(
...
+            except Exception as e:
+                # If error occurs, log warning and return False to indicate cache reset required
+                logger.warning(f"Error occurred while checking {cache_entry}. Exception: {e}")
+                return False
+
+        # Otherwise all hashes exist in S3
+        return True
+
+    def _sample_cache_entries_with_limit(
+        self, cache_entries: List[S3CheckCacheEntry], limit: int = 30
```
leongdl, line 432: Future ask - should we make this configurable?

src/deadline/job_attachments/upload.py
```diff
@@ -382,6 +392,78 @@ def upload_input_files(
...
+        ]
+        with S3CheckCache(s3_check_cache_dir) as s3_cache:
+            cache_entries = [
+                s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}")
+                for upload_key in s3_upload_keys
+                if s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}") is not None
+            ]
+
+            # verify that a sample of the cached entries still exist in S3
+            sampled_cache_entries = self._sample_cache_entries_with_limit(cache_entries)
```
leongdl, line 464: Can we print or add some warnings to let the user know their cache is inconsistent? Once this happens, users may want to just reset the cache instead of paying for failed jobs..

src/deadline/job_attachments/upload.py
```diff
@@ -382,6 +392,78 @@ def upload_input_files(
...
+
+        # Otherwise all hashes exist in S3
+        return True
+
+    def _sample_cache_entries_with_limit(
+        self, cache_entries: List[S3CheckCacheEntry], limit: int = 30
+    ):
+        sampled_count: int = min(len(cache_entries), limit)
+        random.shuffle(cache_entries)
+        return cache_entries[:sampled_count]
```
jericht, line 436: nit: does this need to be a separate method? maybe we can refactor it into a method when we need to reuse it

jericht (approved):

PR #374, opened 2024-06-26
revert: revert json output and schema version

### LINKED ISSUE #376: Bug: Integrated submitters using deadline 0.48.5 cannot submit jobs
### Expected Behaviour

If you maintain a custom install of integrations, installing deadline 0.48.5 into Maya or Nuke should allow you to successfully submit jobs

### Current Behaviour

Submission fails with a message that looks like 

image

### Reproduction Steps

Nothing special, launch the submitter and attempt to submit

### Code Snippet

Root cause is this commit: [aad9a49c9085e67031abc3ce342ca8068a6508d1](https://github.com/aws-deadline/deadline-cloud/commit/aad9a49c9085e67031abc3ce342ca8068a6508d1)

epmog:
Resolves #376 

1. Integrated submitters failed on submit with:

```
'RenderSubmitterUISettings' object has no attribute 'parameters'
```

image

2. jsonschema was updated to a version that required additional code changes

Revert the two commits that introduced issues

integrated submitters work, and we can build submitter installers again

Deployed the new version with Maya and confirmed that we no longer get this error

N/A

Yes - we're removing a CLI option that was added that may have worked for users until we get the proper fix in.

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
     # Pinning due to 3.4 not being available upstream
     "xxhash == 3.4.*",
     # Pinning due to new 4.18 dependencies breaking pyinstaller implementation
-    "jsonschema >= 4.17,< 4.23",
+    "jsonschema == 4.17.*",
     "pywin32 == 306; sys_platform == 'win32'",
     "QtPy == 2.4.*",
 ]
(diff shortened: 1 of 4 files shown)


epmog:
We might be able to make this non-breaking if we only revert the check around job parameters in the submit dialog

lucaseck (approved):

moorec-aws (approved):

PR #375, opened 2024-06-26
[feat][test]: Squish Automated Test Framework - Dev Submit GUI Settings flow

rena-cheng:

We currently do not have automated user-facing tests for the Submitters dialogue which would catch regressions/bugs from being released to customers.

Implement/add tests for the Submitters dialogue using Squish Automated Test Framework. This is a first set of automated tests for the Submitters dialogue, and this is PR focuses on the Dev Submit GUI - Settings flow/config.

Implement/add tests for the Submitters dialogue using Squish Automated Test Framework. This is a first set of automated tests for the Submitters dialogue, and this is PR focuses on the Dev Submit GUI - Settings flow/config.

I ran my tests locally and ensured that the tests passed with no issues.

I will have a quip doc/draft of Squish Automated framework, how to get set up with it, debug, etc.

No.

----

### DIFF
--- a/test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
+++ b/test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
@@ -25,140 +25,141 @@ def launchDeadlineConfigGUI():
     )
 
 
-def setGlobalSettings(profileName: str, jobHistDir: str):
+def setGlobalSettings(profileName: str):
+    # open AWS profile drop down menu
+    squish.mouseClick(
+        squish.waitForObjectExists(gui_locators.global_settings_AWS_profile_QComboBox),
+    )
     test.compare(
-        str(squish.waitForObjectExists(gui_locators.profile_settings_QLineEdit).displayText),
-        jobHistDir,
-        "Expect job history directory default path to be equal.",
+        squish.waitForObjectExists(gui_locators.aWS_profile_deadlinecloud_squish_QModelIndex).text,
+        "deadlinecloud_squish",
+        "Expect AWS profile name to be present in dropdown.",
     )
-    squish.mouseClick(
-        squish.waitForObjectItem(gui_locators.global_settings_AWS_profile_QComboBox, profileName),
-        484,
-        9,
-        squish.Qt.NoModifier,
-        squish.Qt.LeftButton,
-    )
-    squish.mouseClick(
-        squish.waitForObjectItem(gui_locators.global_settings_AWS_profile_QComboBox, profileName),
-        225,
-        6,
-        squish.Qt.NoModifier,
-        squish.Qt.LeftButton,
-    )
+    # select AWS profile: `deadlinecloud_squish`
+    squish.mouseClick(gui_locators.aWS_profile_deadlinecloud_squish_QModelIndex)
     test.log("Set Global settings")
 
 
-def setProfileSettings(farmName: str):
-    squish.mouseClick(
-        squish.waitForObject(gui_locators.profile_settings_QComboBox),
-        407,
-        13,
-        squish.Qt.NoModifier,
-        squish.Qt.LeftButton,
+def setProfileSettings(jobHistDir: str, farmName: str):
+    test.compare(
+        str(squish.waitForObjectExists(gui_locators.profile_settings_QLineEdit).displayText),
+        jobHistDir,
+        "Expect Job history directory default path to be equal.",
     )
-    squish.mouseClick(
-        squish.waitForObjectItem(gui_locators.profile_settings_QComboBox, farmName),
-        189,
-        10,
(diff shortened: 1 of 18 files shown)


test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+        10,
+        squish.Qt.NoModifier,
+        squish.Qt.LeftButton,
+    )
+    test.log("Set Profile settings")
+
+
+def setFarmSettings(queueName: str, storageProfile: str, jobAttachments: str):
+    squish.mouseClick(
+        squish.waitForObject(gui_locators.farm_settings_QComboBox),
```
moorec-aws: Can we add some comments for mouse clicks on what is being tested? Makes it a bit easier to parse. Some of the locator names are not very helpful.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+        squish.waitForObjectItem(gui_locators.global_settings_AWS_profile_QComboBox, profileName),
+        225,
+        6,
+        squish.Qt.NoModifier,
+        squish.Qt.LeftButton,
+    )
+    test.log("Set Global settings")
+
+
+def setProfileSettings(farmName: str):
```
moorec-aws: Is there a test here that we can successfully open the file browser?

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def main():
+    gui_helpers.launchDeadlineConfigGUI()
+    gui_helpers.setGlobalSettings(profileName, jobHistDir)
+    gui_helpers.setProfileSettings(farmName)
+    gui_helpers.setFarmSettings(queueName, storageProfile, jobAttachments)
+    gui_helpers.setGeneralSettings(conflictResOption, loggingLevel)
+    test.log("All settings config have been applied.")
+    gui_helpers.closeDeadlineConfigGUI()
```
erico-aws: I'm a little confused about these functions. It seems like we're using them as tests but they also seem like setup for a test. Like none of them seem to validate that the controls have the values we expect after they're set. Or are these here just to setup before tests are run?

rena-cheng: These tests are run from a customer-facing perspective (so in essence this is a "setup" flow for a customer). The functions will fail if the test cannot find those values (thereby verifying if the Dialogue is able to pull the correct farm info, etc).

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: Will add in the next revision. Good catch. Thanks!

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: Will look into adding comments regarding the mouse-clicks. Thanks!

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+
+def setGlobalSettings(profileName: str, jobHistDir: str):
+    test.compare(
+        str(squish.waitForObjectExists(gui_locators.profile_settings_QLineEdit).displayText),
+        jobHistDir,
+        "Expect job history directory default path to be equal.",
+    )
+    squish.mouseClick(
+        squish.waitForObjectItem(gui_locators.global_settings_AWS_profile_QComboBox, profileName),
+        484,
```
justinsaws: What are these numbers referencing?

justinsaws: If these are pixel placements for the mouse, I think we are going to want to do this another way. Doing it with mouse clicks makes our tests extremely fragile because any changes to the layout of the UI can potentially break our tests. 

Squish has ways of clicking on specific buttons directly which is more accurate and as long as we don't change the name of the underlying UI objects, our tests remain valid.

Check out this: https://doc.qt.io/squish/qt-convenience-api.html#qt-clickbutton-function

justinsaws (changes requested):

rena-cheng:
Will be closing this PR as I submitted a new one here: https://github.com/aws-deadline/deadline-cloud/pull/399

PR #380, opened 2024-06-27
feat: support JSON output from bundle gui-submit

joel-wong-aws:
Copied from https://github.com/aws-deadline/deadline-cloud/pull/357
This is a simplified version of that PR.

In a DCC integration, I was trying to read out the most recently submitted job settings. The bundle GUI submitter prints a human readable output with the bundle directory path where the settings can be found. but the response is not machine readable. 

Add a JSON output option.

Scripts calling the GUI submitter can easily read back where the job bundle is saved and what the submitted job ID is.

Unit tests and manually running the CLI commands.

Auto-documented in the CLI help.

No - new minor feature.
----

### DIFF
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -5,8 +5,10 @@ All the `deadline bundle` commands.
 """
 from __future__ import annotations
 
+import json
 import logging
 import re
+from typing import Any, Optional
 
 import click
 from contextlib import ExitStack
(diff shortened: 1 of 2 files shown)


lucaseck (approved):

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -249,8 +251,20 @@ def _decide_cancel_submission(upload_group: AssetUploadGroup) -> bool:
     is_flag=True,
     help="Allows user to choose Bundle and adds a 'Load a different job bundle' option to the Job-Specific Settings UI",
 )
+@click.option(
+    "--output",
+    type=click.Choice(
+        ["verbose", "json"],
```
epmog, line 257: I'm not sure verbose makes sense here. Any reason to not just do json?

Verbose is usually implied to have a non-verbose mode which is usually the default.

joel-wong-aws: Are you proposing that we remove this option altogether and always provide output in JSON format?

crowecawcaw: I originally had it like you suggested and I think it makes sense for this particular use case. I chatted with @ddneilson about it though and he pointed out that over time, we might add `json` output options to the rest of the CLI commands, and someone might expect the output format to not change the type of data being returned, which seemed reasonable.

crowecawcaw (approved):

src/deadline/client/cli/_groups/bundle_group.py
epmog: Looks like we're following this pattern elsewhere in deadline-cloud currently. Does this mean you can't select verbose AND json? My thoughts led me to `--output-format JSON` or something. But apparently I added one of these options in the past, perhaps not the greatest...

My biggest gripe is that for most CLIs I've every used, verbose is usually a standalone flag rather than a choice on individual commands. 

We're not changing this here though, so the change is fine with me.

epmog (approved):

PR #381, opened 2024-06-27
feat: Run Integration tests on Windows as well

YutongLi291:
We have successfully fixed the integration tests that are being ran for Job Attachments (which are currently the only integ tests in this repo) on Linux. We would like to add this workflow  to also run on Windows.
Add it to the actions
We can verify if the tests run properly on Windows
The linux workflow was tested https://github.com/aws-deadline/deadline-cloud/actions/runs/9704180168 to verify that it works
No
No
----

### DIFF
--- a/.github/workflows/mainline_integration_test.yml
+++ b/.github/workflows/mainline_integration_test.yml
@@ -1,4 +1,4 @@
-name: Mainline Integration Test
+name: Mainline Merge Integration Test
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


moorec-aws (dismissed):

jericht (commented):
Can we add Windows integ tests here too?

YutongLi291:
> Can we add Windows integ tests here too?
Added!

moorec-aws (approved):

jericht (approved):

PR #385, opened 2024-07-02
fix: Fix JA wrong account bucket integration test temp path name

YutongLi291:
The windows integration tests were failing due to `FAILED test/integ/deadline_job_attachments/test_job_attachments.py::test_sync_inputs_bucket_wrong_account[2023-03-03] - ValueError: An error occurred \(403\) is not a normalized and relative path`
Fix the test which is creating a temp path with a weird path to just use a normal path
The tests should pass on windows.

No windows environment, but confirmed that mac tests still work properly by `hatch run integ:test`
No
No
----

### DIFF
--- a/test/integ/deadline_job_attachments/test_job_attachments.py
+++ b/test/integ/deadline_job_attachments/test_job_attachments.py
@@ -1355,7 +1355,7 @@ def test_sync_inputs_bucket_wrong_account(
     )
 
     syncer = asset_sync.AssetSync(job_attachment_test.farm_id)
-    session_dir = tmp_path_factory.mktemp(r"An error occurred \(403\)")
+    session_dir = tmp_path_factory.mktemp("session_dir")
 
     def on_downloading_files(*args, **kwargs):
         return True


marofke (approved):

AlexTranAmz (approved):

test/integ/deadline_job_attachments/test_job_attachments.py
```diff
@@ -1355,7 +1355,7 @@ def test_sync_inputs_bucket_wrong_account(
     )
 
     syncer = asset_sync.AssetSync(job_attachment_test.farm_id)
-    session_dir = tmp_path_factory.mktemp(r"An error occurred \(403\)")
```
leongdl, line 1358: Was this written as a funky folder for a reason before?

YutongLi291: It doesn't make sense to have a funky folder since it doesn't have anything to do with what the test is testing

AlexTranAmz: Even if there is a reason, it's most likely not needed anymore as we already have a regex check to make sure char like "(" and "\" can't be used

YutongLi291:
This fixed the integration tests completely https://github.com/aws-deadline/deadline-cloud/actions/runs/9765921768

test/integ/deadline_job_attachments/test_job_attachments.py
leongdl: Perfect - this would make sense now.

leongdl (commented):
OK with the PR, but we need to find a good way to test our PRs for platforms we are fixing for. Lets find time to go over to IT and get our Windows laptops.

PR #387, opened 2024-07-02
test: Add integration test runs as part of mainline merge and release  bump workflows

YutongLi291:
For better code quality gates, we should run the integration tests that test the Job Attachments functionality on release, as well as on merges to mainline.
On PR merge to mainline, or release of the package, automated integration test workflows will be triggered. If the test workflow fails on release of the package, it will block the release.
Bad code changes that break the entire Job Attachments functionality should be observable after mainline is merged into, 

Furthermore, failing tests will stop the release workflow when release bump is ran.
`hatch build` was still working.

The integration tests do work properly as tested in https://github.com/aws-deadline/deadline-cloud/actions/runs/9765921768
No
No
----

### DIFF
--- a/.github/workflows/release_bump.yml
+++ b/.github/workflows/release_bump.yml
@@ -21,11 +21,36 @@ jobs:
     name: Unit Tests
     uses: ./.github/workflows/code_quality.yml
     with:
+      branch: mainline 
+  LinuxIntegrationTests:
+    name: Linux Integration Tests
+    needs: UnitTests
+    permissions:
+      id-token: write
+      contents: read
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    secrets: inherit
+    with:
+      repository: ${{ github.event.repository.name }}
       branch: mainline
-
+      environment: mainline
+      os: linux
+  WindowsIntegrationTests:
+    name: Windows Integration Tests
+    needs: UnitTests
+    permissions:
+      id-token: write
+      contents: read
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    secrets: inherit
+    with:
+      repository: ${{ github.event.repository.name }}
+      branch: mainline
+      environment: mainline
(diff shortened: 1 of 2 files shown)


leongdl:
Can we improve the "What was the problem/requirement? (What/Why)
"Setup integration tests on PR Merge to increase quality gates" - this repo is public and other users read this!

What was the solution? (How)
"Add these workflows."  - Can we make this more descriptive. EG: "On PR merge to mainline, automated integration test workflows will be triggered"

.github/workflows/release_bump.yml
```diff
@@ -21,8 +21,31 @@ jobs:
     name: Unit Tests
     uses: ./.github/workflows/code_quality.yml
     with:
+      branch: mainline 
+  LinuxIntegrationTests:
+    name: Linux Integration Tests
```
moorec-aws, line 26: We should add `needs: UnitTests` here. No point in running integration tests if Unit Tests are going to fail.

.github/workflows/release_bump.yml
```diff
@@ -21,8 +21,31 @@ jobs:
...
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    secrets: inherit
+    with:
+      repository: ${{ github.event.repository.name }}
       branch: mainline
-
+      environment: mainline
+      os: linux
+  WindowsIntegrationTests:
+    name: Windows Integration Tests
```
moorec-aws, line 39: Same thing here, `needs: UnitTests`.

.github/workflows/release_bump.yml
YutongLi291: Revised and fixed

.github/workflows/release_bump.yml
YutongLi291: Revised and fixed

moorec-aws (approved):

AlexTranAmz (approved):

PR #390, opened 2024-07-03
test: run integration tests every four hours

YutongLi291:
To better the repo's code quality, we should run the integration tests on a scheduled basis, to prevent flakiness, issues with dependencies, etc.
Every 4 hours, the integration tests will be ran as an action workflow in mainline. 
Better repo quality and guarantee that the main functionality is tested.
`hatch build`. Verified that the syntax was correct.
N/A
No
----

### DIFF
--- a/.github/workflows/mainline_integration_test.yml
+++ b/.github/workflows/mainline_integration_test.yml
@@ -1,6 +1,8 @@
 name: Mainline Merge Integration Tests
 
 on:
+  schedule:
+    - cron: '0 */4 * * *'
   push:
     branches:
       - mainline


leongdl (approved):

.github/workflows/mainline_integration_test.yml
```diff
@@ -1,6 +1,8 @@
 name: Mainline Merge Integration Tests
 
 on:
+  schedule:
+    - cron: '0 */4 * * *'
```
leongdl, line 5: Chatted offline, 

Why 4 ? Because worker agent side is running every 4 hours too.

leongdl (approved):

godobyte (approved):

PR #392, opened 2024-07-05
feat(asset-cli): asset upload subcommand to upload assets to S3

seant-aws:
Allow users to call CLI command to upload directories of assets to S3, in order to pre-cache assets before job submission

Added deadline asset upload subcommand to allow for upload of directory content to corresponding S3 bucket, specified with manifest, farmID, and queueID. Along with the option to auto update directory manifests if changed before upload.

Exposes the job attachment upload functionality to user, allows users to choose when and what to put onto their S3 bucket in relevant job attachment format.

- Added unit tests mocking upload functionality at `/Users/stangch/Desktop/deadline-cloud- stangch/test/unit/deadline_client/cli/test_cli_asset.py`, passing successfully
- Manual end to end tests with sample directories with assets:

```
deadline-cloud-stangch % hatch run deadline asset upload --manifest /Users/stangch/Desktop/test_asset/test_asset_manifests --update
Manifest information updated: 1 files updated. 

Uploading Attachments  [####################################]  100%
Upload Summary:
    Processed 1 file totaling 8.2 KB.
    Skipped re-processing 2 files totaling 412.0 B.
    Total processing time of 0.74311 seconds at 11.03 KB/s.

Upload of /Users/stangch/Desktop/test_asset complete. 
```

----

### DIFF
--- a/src/deadline/client/cli/_groups/asset_group.py
+++ b/src/deadline/client/cli/_groups/asset_group.py
@@ -9,15 +9,27 @@ All the `deadline asset` commands:
 """
 import os
 from pathlib import Path
+import concurrent.futures
+from typing import List
+import glob
 
 import click
 
 from deadline.client import api
-from deadline.job_attachments.upload import S3AssetManager, S3AssetUploader
-from deadline.job_attachments.models import JobAttachmentS3Settings
+from deadline.job_attachments.upload import FileStatus, S3AssetManager, S3AssetUploader
+from deadline.job_attachments.models import (
+    JobAttachmentS3Settings,
+    AssetRootManifest,
+)
+from deadline.job_attachments.asset_manifests.decode import decode_manifest
+from deadline.job_attachments.asset_manifests.base_manifest import BaseAssetManifest
+from deadline.job_attachments.caches import HashCache
 
-from .._common import _handle_error, _ProgressBarCallbackManager
-from ...exceptions import NonValidInputError
+from .._common import _apply_cli_options_to_config, _handle_error, _ProgressBarCallbackManager
+from ...exceptions import NonValidInputError, ManifestOutdatedError
+from ...config import get_setting, config_file
+import boto3
+from botocore.client import BaseClient
 
 
 @click.group(name="asset")
@@ -104,7 +116,13 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
 
 @cli_asset.command(name="upload")
 @click.option(
-    "--manifest", help="The path to manifest folder of the directory specified for upload. "
+    "--root-dir",
+    help="The root directory of assets to upload. Defaults to the parent directory of --manifest-dir if not specified. ",
+)
+@click.option(
+    "--manifest-dir",
+    required=True,
+    help="The path to manifest folder of the directory specified for upload. ",
 )
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
@@ -116,11 +134,104 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
     default=False,
 )
 @_handle_error
-def asset_upload(**args):
+def asset_upload(root_dir: str, manifest_dir: str, update: bool, **args):
     """
     Uploads the assets in the provided manifest file to S3.
     """
-    click.echo("upload done")
+
+    if not os.path.isdir(manifest_dir):
+        raise NonValidInputError(f"Specified manifest directory {manifest_dir} does not exist. ")
+
+    if root_dir is None:
+        asset_root_dir = os.path.dirname(manifest_dir)
+    else:
+        if not os.path.isdir(root_dir):
+            raise NonValidInputError(f"Specified root directory {root_dir} does not exist. ")
+        asset_root_dir = root_dir
+
+    config = _apply_cli_options_to_config(required_options={"farm_id", "queue_id"}, **args)
+    upload_callback_manager: _ProgressBarCallbackManager = _ProgressBarCallbackManager(
+        length=100, label="Uploading Attachments"
+    )
+
+    deadline: BaseClient = api.get_boto3_client("deadline", config=config)
+    queue_id: str = get_setting("defaults.queue_id", config=config)
+    farm_id: str = get_setting("defaults.farm_id", config=config)
+
+    queue: dict = deadline.get_queue(
+        farmId=farm_id,
+        queueId=queue_id,
+    )
+
+    # assume queue role - session permissions
+    queue_role_session: boto3.Session = api.get_queue_user_boto3_session(
+        deadline=deadline,
+        config=config,
+        farm_id=farm_id,
+        queue_id=queue_id,
+        queue_display_name=queue["displayName"],
+    )
+
+    asset_manager: S3AssetManager = S3AssetManager(
+        farm_id=farm_id,
+        queue_id=queue_id,
+        job_attachment_settings=JobAttachmentS3Settings(**queue["jobAttachmentSettings"]),
+        session=queue_role_session,
+    )
+
+    asset_uploader: S3AssetUploader = S3AssetUploader()
+
+    # read local manifest into BaseAssetManifest object
+    asset_manifest: BaseAssetManifest = read_local_manifest(manifest=manifest_dir)
+    clear_S3_mapping(manifest=manifest_dir)
+
+    if asset_manifest is None:
+        raise NonValidInputError(
+            f"Specified manifest directory {manifest_dir} does contain valid manifest input file. "
+        )
+
+    asset_root_manifest: AssetRootManifest = AssetRootManifest(
+        root_path=asset_root_dir,
+        asset_manifest=asset_manifest,
+    )
+
+    manifest_changes: List[tuple] = diff_manifest(
+        asset_manager=asset_manager,
+        asset_root_manifest=asset_root_manifest,
+        manifest=manifest_dir,
+        update=update,
+    )
+
+    # if there are modified files, will either auto --update manifest or prompt user of file discrepancy
+    if len(manifest_changes) > 0:
+        if update:
+            asset_root_manifest.asset_manifest = update_manifest(
+                manifest=manifest_dir, new_or_modified_paths=manifest_changes
+            )
+            click.echo(f"Manifest information updated: {len(manifest_changes)} files updated. \n")
+        else:
+            raise ManifestOutdatedError(
+                f"Manifest contents in {manifest_dir} are outdated; versioning does not match local files in {asset_root_dir}. Please run with --update to fix current files. \n"
+            )
+
+    attachment_settings: dict = api.upload_attachments(
+        asset_manager=asset_manager,
+        manifests=[asset_root_manifest],
+        print_function_callback=click.echo,
+        upload_progress_callback=upload_callback_manager.callback,
+    )
+
+    full_manifest_key: str = attachment_settings["manifests"][0]["inputManifestPath"]
+    manifest_name = os.path.basename(full_manifest_key)
+    manifest_dir_name = os.path.basename(manifest_dir)
+    asset_uploader._write_local_manifest_s3_mapping(
+        manifest_write_dir=asset_root_dir,
+        manifest_name=manifest_name,
+        full_manifest_key=full_manifest_key,
+        manifest_dir_name=manifest_dir_name,
+    )
+
+    click.echo(f"Upload of {asset_root_dir} complete. \n")
 
 
 @cli_asset.command(name="diff")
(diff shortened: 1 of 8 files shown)


src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -116,11 +129,103 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
...
+        asset_root_dir = Path(manifest).parent.as_posix()
+    else:
+        if not os.path.isdir(root_dir):
+            raise NonValidInputError(f"Specified root directory {root_dir} does not exist. ")
+        asset_root_dir = root_dir
+
+    config = _apply_cli_options_to_config(required_options={"farm_id", "queue_id"}, **args)
+    upload_callback_manager = _ProgressBarCallbackManager(length=100, label="Uploading Attachments")
+
+    deadline = api.get_boto3_client("deadline", config=config)
```
leongdl: consider adding type annotation? Its very hard to read Python code that is typeless.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+    """
+    Clears manifest_s3_mapping file contents if it previously exists.
+    """
+    for filename in os.listdir(manifest):
+        if filename.endswith("manifest_s3_mapping"):
+            # if S3 mapping already exists, clear contents
+            filepath = os.path.join(manifest, filename)
+            with open(filepath, "w") as _:
+                pass
+    return
```
leongdl: not needed :)

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+
+
+def get_directory_changes():
+    """
+    TODO: Gets the file paths in specified directory if the contents of file have changed compared to a specified snapshot of the respective directory.
+    """
```
leongdl: Don't we need a pass statement here?

src/deadline/job_attachments/upload.py
```diff
@@ -179,7 +190,7 @@ def upload_assets(
 
         if manifest_write_dir:
             self._write_local_manifest(
-                manifest_write_dir, manifest_name, full_manifest_key, manifest
+                manifest_write_dir, manifest_name, full_manifest_key, manifest, None
```
leongdl: If the method now has default = None, is adding None here needed?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -103,8 +113,11 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
 
 
 @cli_asset.command(name="upload")
+@click.option("--root-dir", help="The root directory of assets to upload. ")
 @click.option(
-    "--manifest", help="The path to manifest folder of the directory specified for upload. "
+    "--manifest",
+    required=True,
+    help="The path to manifest folder of the directory specified for upload. ",
 )
```
jericht: Can the manifest directory have more than one manifest file in it? Should we consider a different name for this option? Seems to mean a manifest file rather than directory at first glance.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -116,11 +129,103 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
...
     """
     Uploads the assets in the provided manifest file to S3.
     """
-    click.echo("upload done")
+
+    if not os.path.isdir(manifest):
+        raise NonValidInputError(f"Specified manifest directory {manifest} does not exist. ")
+
+    if root_dir is None:
+        asset_root_dir = Path(manifest).parent.as_posix()
```
jericht: We should mention this default behaviour in the help text for `--root-dir`

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -116,11 +129,103 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
...
+    else:
+        if not os.path.isdir(root_dir):
+            raise NonValidInputError(f"Specified root directory {root_dir} does not exist. ")
```
jericht, line 149: This validation should also happen if `root_dir` is provided, right?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+            filepath = os.path.join(manifest, filename)
+            with open(filepath, "r") as input_file:
+                manifest_data_str = input_file.read()
+                asset_manifest = decode_manifest(manifest_data_str)
+
+                return asset_manifest
+
+    raise NonValidInputError(
+        f"Specified manifests directory {manifest} does not contain manifest file. "
+    )
```
jericht: Looks like this function reads the first manifest file returned by `os.listdir`. If the input is a directory, and we expect there to be exactly one manifest file in it, we should add that as a check in the function. If having more than a single manifest in the directory is a valid case, we should handle that and return a list of manifests here

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -116,11 +129,103 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
...
+    if asset_manifest is None:
+        raise NonValidInputError(
+            f"Specified manifest directory {manifest} does contain valid manifest input file. "
+        )
+
+    asset_root_manifest = AssetRootManifest(
+        root_path=asset_root_dir,
+        asset_manifest=asset_manifest,
+    )
+    asset_root_manifests: List[AssetRootManifest] = [asset_root_manifest]
```
jericht: Any reason we are creating this list up here? A bit further down we index into the first element to support the `--update` option but we could reference the single instance just above here directly. If we need to wrap the manifest in an array for the `api.upload_attachments` call, we can just do that directly in the function call

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+    # input_paths = []
+    # for root, dirs, files in os.walk(root_path):
+    #     if os.path.samefile(root, manifest):
+    #         dirs[:] = []
+    #         continue
+    #     for filename in files:
+    #         file_path = os.path.join(root, filename)
+    #         input_paths.append(file_path)
+
+    # return input_paths
```
jericht: Looks like an old function we don't need anymore. Let's delete this?

src/deadline/client/cli/_groups/asset_group.py
jericht: Not sure what this function is doing. Is the intention to overwrite the file with an empty file?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+
+
+def get_manifest_changes(
```
jericht: nit: instead of the word `get`, could we rename this to `diff_manifest`? It's a bit clearer that way. Right now, the "get" kind of makes it seem like a getter at first glance, when it is really a diff function

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+            # skip the manifest folder, or else every upload will need an update after a previous change
+            continue
+        input_paths.append(Path(root_path, base_manifest_path.path))
+
+    return get_file_changes(
+        asset_manager=asset_manager, input_paths=input_paths, root_path=root_path, update=update
+    )
+
+
+def get_file_changes(
```
jericht: The current name kind of looks like it would mean "get the changes in the file contents" at a glance IMO. How about changing this function to be a little more generic like `find_files_with_status` and a take in a param for the statuses to look for (and pass in `NEW` and `MODIFIED`)?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+            new_or_modified_paths: List[(tuple)] = []
+            for future in concurrent.futures.as_completed(futures):
+                (file_status, _, manifestPath) = future.result()
+                if file_status is FileStatus.NEW or file_status is FileStatus.MODIFIED:
+                    new_or_modified_paths.append((file_status, manifestPath))
+
+            return new_or_modified_paths
+
+
+def update_manifest(manifest: str, new_or_modified_paths: List[(tuple)]) -> BaseAssetManifest:
```
jericht: nit: typing for `new_or_modified_paths` is a bit weird. I don't think we need the round brackets around `tuple`.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +260,146 @@ def asset_download(**args):
...
+            manifest_info_dict[base_asset_manifest.path].hash = base_asset_manifest.hash
+        else:
+            # Add the new object if it doesn't exist
+            manifest_info_dict[base_asset_manifest.path] = base_asset_manifest
+
+    # write ot local manifest
+    updated_path_list = list(manifest_info_dict.values())
+    local_base_asset_manifest.paths = updated_path_list
+    with open(manifest_file_path, "w") as manifest_file:
+        manifest_file.write(local_base_asset_manifest.encode())
```
jericht: This looks incorrect. The `local_base_asset_manifest` file is being updated for each file from `os.listdir(manifest)` but we break out of that for loop and only operate on the last value read into the variable. Should the code underneath (starting at line 386) be indented another level?

jericht (changes requested):
Just these comments for this pass. I'll take a look at the unit tests and other files once these have been addressed.

src/deadline/client/cli/_groups/asset_group.py
seant-aws: that makes sense, I'm thinking since its --root-dir maybe --manifest-dir would be easier for user to understand

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Yea, I think that functionality is already here ?

src/deadline/client/cli/_groups/asset_group.py
seant-aws: I'm mainly doing since the manifest folder will also include another file for S3 mapping. I think we talked about how there might be more than one manifest file for windows, but thats a small edge case we will forgo for now ?

stangch: should I add just a single _input check for now ?

src/deadline/client/cli/_groups/asset_group.py
seant-aws: yea, its mainly to warp the manifest for api.upload_attachments. That makes sense, I'll change that.

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Just making sure, this means that the return statement isn't needed

seant-aws: Yes, this deletes the text content of the local S3_mapping file by overwriting with nothing

src/deadline/client/cli/_groups/asset_group.py
seant-aws: spoken offline, will include checks for 1 manifest

src/deadline/client/cli/_groups/asset_group.py
seant-aws: spoken offline, will add checks for single manifest file

leongdl (dismissed):

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +266,142 @@ def asset_download(**args):
...
+
+
+def read_local_manifest(manifest: str) -> BaseAssetManifest:
+    """
+    Read manifets specified by filepath to manifest folder, returns BaseAssetManifest Object
+    """
+    input_files = glob.glob(os.path.join(manifest, "*_input"))
+
+    if len(input_files) >= 2:
+        raise NonValidInputError(f"Does not support multiple manifest_input files in {manifest}. ")
```
jericht:
```suggestion
        raise NonValidInputError(f"Multiple input manifest files are not supported, but found {input_files}.")
```

nit: Slight improvement to the error message. Feel free to word as you want but listing the files we found is helpful

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +266,142 @@ def asset_download(**args):
     Downloads input manifest of previously submitted job.
     """
     click.echo("download complete")
+
+
+def read_local_manifest(manifest: str) -> BaseAssetManifest:
+    """
+    Read manifets specified by filepath to manifest folder, returns BaseAssetManifest Object
```
jericht:
```suggestion
    Read manifests specified by filepath to manifest folder, returns BaseAssetManifest Object
```

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +266,142 @@ def asset_download(**args):
...
+    for filepath in input_files:
+        with open(filepath, "r") as input_file:
+            manifest_data_str = input_file.read()
+            asset_manifest = decode_manifest(manifest_data_str)
+
+            return asset_manifest
+
+    raise NonValidInputError(
+        f"Specified manifests directory {manifest} does not contain manifest file. "
+    )
```
jericht: nit: We have confirmed there is no more than one input file. This `for` loop with the error happening after the loop completes is awkward IMO. The error case is when there are no elements in the `input_files` list, right? We can probably change the code to just check for that, raise the error, otherwise just continue on with the exactly one input file we have confirmed.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +266,142 @@ def asset_download(**args):
...
+def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAssetManifest:
+    """
+    Updates the local manifest file to reflect modified or new files
+    """
+    input_files = glob.glob(os.path.join(manifest, "*_input"))
+
+    if not input_files:
+        raise ValueError(f"No manifest files found in {manifest}")
+    elif len(input_files) >= 2:
+        raise NonValidInputError(f"Does not support multiple manifest_input files in {manifest}. ")
```
jericht:
```suggestion
        raise NonValidInputError(f"Multiple input manifest files are not supported, but found {input_files}.")
```
Same suggestion here for the error message to be consistent

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -42,6 +52,45 @@ def upload_group_mock(asset_group_mock):
...
 
 
+@pytest.fixture
+def mock_asset_manifest():
+    return AssetManifest(paths=[], hash_alg=HashAlgorithm("xxh128"), total_size=0)
+
+
+@pytest.fixture
+def mock_attachment_settings():
+    return Attachments(manifests=[], fileSystem="").to_dict
```
jericht: Tip for the future. If the fixtures are not actually mocking anything, I think you should remove the `mock` from the fixture name so it is not confusing. Maybe these could be like `empty` or `basic` `<object name>`instead.

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -42,6 +52,45 @@ def upload_group_mock(asset_group_mock):
...
+def mock_init_objects():
+    with patch.object(S3AssetManager, "__init__", lambda self, *args, **kwargs: None), patch.object(
+        S3AssetUploader, "__init__", lambda self, *args, **kwargs: None
+    ), patch.object(JobAttachmentS3Settings, "__init__", lambda self, *args, **kwargs: None):
+        yield
+
+
+@pytest.fixture
+def mock_update_manifest():
+    with patch.object(asset_group, "update_manifest", return_value=mock_asset_manifest) as mock:
```
jericht:
```suggestion
def mock_update_manifest(mock_asset_manifest):
    with patch.object(asset_group, "update_manifest", return_value=mock_asset_manifest) as mock:
```

We need to take in the fixture here in order to use the return value. I believe this code is setting the `return_value` as the fixture function (Callable), not its return value (AssetManifest).

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -155,3 +204,212 @@ def test_asset_snapshot_recursive(
...
+            manifest_dir_name = os.path.basename(MOCK_MANIFEST_DIR)
+
+            mock_write_manifest_mapping.assert_called_once_with(
+                manifest_write_dir=MOCK_ROOT_DIR,
+                manifest_name=manifest_name,
+                full_manifest_key=full_manifest_key,
+                manifest_dir_name=manifest_dir_name,
+            )
+            mock_upload_attachments.assert_called_once()
+            assert result.exit_code == 0
```
jericht: nit: We don't need this assertion code inside the `with` block

Same for other assertion code below that is inside the `with` block unnecessarily

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -155,3 +204,212 @@ def test_asset_snapshot_recursive(
...
+                manifest_write_dir=MOCK_ROOT_DIR,
+                manifest_name=manifest_name,
+                full_manifest_key=full_manifest_key,
+                manifest_dir_name=manifest_dir_name,
+            )
+            mock_upload_attachments.assert_called_once()
+            assert "Manifest information updated:" in result.output
+            assert result.exit_code == 0
+
+    def testpload_with_modified_files_without_update(fresh_deadline_config, mock_init_objects):
```
jericht:
```suggestion
    def test_upload_with_modified_files_without_update(fresh_deadline_config, mock_init_objects):
```

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -155,3 +204,212 @@ def test_asset_snapshot_recursive(
...
+
+            mock_write_manifest_mapping.assert_called_once_with(
+                manifest_write_dir=MOCK_ROOT_DIR,
+                manifest_name=manifest_name,
+                full_manifest_key=full_manifest_key,
+                manifest_dir_name=manifest_dir_name,
+            )
+            mock_upload_attachments.assert_called_once()
+            assert "Manifest information updated:" in result.output
+            assert result.exit_code == 0
```
jericht: Should we also verify that `update_manifest` was called?
```
mock_update_manifest.assert_called_once_with(...args...)
```

jericht (commented):
Just some last minor comments, then let's merge this in!

src/deadline/client/cli/_groups/asset_group.py
seant-aws: right, I'm not sure why I didn't catch this the first time

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -155,3 +266,145 @@ def asset_download(**args):
...
+            f"Multiple input manifest files are not supported, found: {input_files}."
+        )
+
+    manifest_file_path = input_files[0]
+
+    with open(manifest_file_path, "r") as input_file:
+        manifest_data_str = input_file.read()
+        asset_manifest = decode_manifest(manifest_data_str)
+
+        return asset_manifest
```
jericht, line 290: Can be address in a later code cleanup pass, but we don't need to decode the manifest inside the `with` block here.

jericht (approved):
LGTM! Only had one minor comment that we can address in a future PR :)

src/deadline/client/cli/_groups/asset_group.py
leongdl: Yes.

src/deadline/client/cli/_groups/asset_group.py
leongdl: Yes

leongdl (approved):
Looks good, thanks for adding the type definitions.

PR #395, opened 2024-07-05
feat: support skipping prompts for installing GUI dependencies

crowecawcaw:
DCM should be able to invoke `deadline bundle gui-submit` non-CLI users can still submit bundles. But if the GUI dependencies aren't already installed, the command will hang until the prompt about installing GUI dependencies is answered.

Support a `--install-gui` flag that skips the prompt.

DCM can invoke the bundle submitter more reliably.

By uninstalling PySide and then calling `deadline bundle gui-submit` and `deadline config gui` with and without the `--install-gui` flag.

Yes - in the CLI help.

No

----

### DIFF
--- a/src/deadline/client/ui/__init__.py
+++ b/src/deadline/client/ui/__init__.py
@@ -45,19 +45,22 @@ def gui_error_handler(message_title: str, parent: Any = None):
 
 
 @contextmanager
-def gui_context_for_cli():
+def gui_context_for_cli(automatically_install_dependencies: bool):
     """
     A context manager that initializes a Qt GUI context for
     the CLI handler to use.
 
     For example:
 
-    with gui_context_for_cli() as app:
+    with gui_context_for_cli(automatically_install_dependencies) as app:
         from deadline.client.ui.cli_job_submitter import show_cli_job_submitter
 
         show_cli_job_submitter()
 
         app.exec()
+
+    If automatically_install_dependencies is true, dependencies will be installed without prompting the user. Useful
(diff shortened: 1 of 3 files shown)


epmog (changes requested):
We have some lessons learned with the regular `deadline bundle submit` command that a `--yes`/`-y` for these bigger workflows that can have multiple prompts does not work well. It's not obvious what the `--yes` should be choosing if a default is not affirmative. It begins to become arbitrary as you force your way through the flow and you lose fine-grained control of those choices.

After the first-use, the -y isn't even used for anything since the user won't be prompted again.

Instead, what are your thoughts about doing something like `--ensure-gui`? This way we can still bypass this by choosing ahead of time that we will accept the "do you want to install the gui" prompt without confusing it with other prompts/options

lucaseck (approved):

src/deadline/client/ui/__init__.py
```diff
@@ -45,19 +45,22 @@ def gui_error_handler(message_title: str, parent: Any = None):
 
 
 @contextmanager
-def gui_context_for_cli():
+def gui_context_for_cli(automatically_install_dependencies: bool):
```
epmog, line 48: Should this have a default value of False? I know the flags we pass in have a default value of False, ~~but adding that here would assist in some safety of a new CLI being added that uses the gui but does not provide this CLI arg.~~

Actually, that should all be fine, since this would force it to be specified if they add a new one that uses this context. All good!

epmog (approved):

PR #399, opened 2024-07-10
test: Squish tests for Submitter GUI settings dialogue

rena-cheng:

We are missing automated user-facing tests for the Submitters dialogue which will catch regressions/bugs as a result of new changes being committed, and prevent them from being released to customers (when a CI/CD pipeline is setup). 

Implement/add tests for the Submitters dialogue using Squish Automated Test Framework. This is a first set of automated tests for the Submitters, and this is PR focuses on the Setting dialogue (from running `deadline config gui`). 

These tests are written from a user-facing perspective, so they would do actions that a user/artist would do. As this dialogue is a 'Settings' dialogue, the actions in this test selects various settings/configs and helps to ensure that the Farm information/resources load (if they do not load as expected, the tests will fail), opens the Job history directory file browser (and verifies that the file browser opens successfully), that the checkboxes can be checked, and also any default settings/file paths. 

Before running, this test assumes that you will already be logged into the Deadline Cloud Monitor (as you need to be authenticated first - note that logging in will be on the test automation roadmap), and also already have test data/resources set up prior, including (adding any additional resources to the below is fine, as well):
- Deadline Cloud Monitor with a Profile named: `deadlinecloud_squish`
- A farm named: `Deadline Cloud Squish Farm`
- A queue named: `Squish Automation Queue`
- A storage profile named: `Squish Storage Profile`

Additionally, to run the tests/launch the AUT ('application under test'), `LD_LIBRARY_PATH` environment variable in the `envvars` file will need to be updated to reflect where the QT libraries live on your system. 

There is an existing bug with the 'Apply' button, so that test has been disabled for now:  https://sim.amazon.com/issues/Bea-28289

The tests will catch any regressions/bugs introduced by new code changes. We would need a CI/CD pipeline setup though, in order for these tests to run and catch new issues introduced. These tests can still be run locally (either using the Squish command line or Squish IDE), but you will need to set up/install Squish first and acquire a license.  

I ran my tests locally and ensured that the tests passed with no issues. I ran some error cases as well and ensured that the tests failed as expected. 

I primarily developed these tests on Rocky Linux, using a farm on `us-east-1` region, and using latest `deadline --version 0.48.7`. 

I will have a quip doc/draft of Squish Automated framework, how to get set up with it, debug, etc. I can add any additional documentation anywhere else, including a ReadME, if needed. 

No. While the code can be merged into the repo, the tests will not run at all [both locally and CI/CD] unless you are set up with Squish IDE and licensing. 

----


test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
```diff
@@ -0,0 +1,181 @@
...
+    test.log("Set Profile settings")
+
+
+def openAndSetDefaultJobHistDirectory():
+    # hit '...' button to open Choose Job history directory file browser
+    squish.clickButton(squish.waitForObject(gui_locators.profile_settings_QPushButton))
+    test.compare(
+        str(squish.waitForObjectExists(gui_locators.qFileDialog_QFileDialog).windowTitle),
+        "Choose Job history directory",
+        "Expect Choose Job history directory dialogue to be open.",
```
hindleym: If this or the next compare fail, it will be hard to discern which was the issue. Would probably be worth updating the output text.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_locators.py
hindleym, line 1: It would probably be useful to take the locators that are automatically created, which are used during the tests, and separate them into their own [page]_locator.py files with more meaningful variable names.

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
hindleym, line 1: May want to rename the file, so we don't end up with multiple `test.py` files throughout the suite.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
hindleym, line 1: If there are plans to test more of the GUI and the other tabs, it may be worth separating the helper functions into separate files.

hindleym (dismissed):
Most of these comments are suggestions to help with future tests. Separating helpers and locators into their own respective files with more meaningful names will make life easier, but that can be done on a follow-up project. Renaming the `test.py` file can be handled at that point too, as things get more cleaned up.

If there are other reasons this PR gets held up, I'd definitely suggest fixing the compares with the same output on line 67 of the `gui_helpers.py` file. Otherwise, that can be fixed when the helpers are separated with the changes above.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: Thanks. Will make a new PR for this minor change.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_locators.py
rena-cheng: I agree with renaming the locators so that they're more meaningful/friendlier. 

Each new test will have it's own locator page. For example, all of the locators on this page test the Settings dialogue, which I don't believe will be re-used in the other tests/Submitter dialogues.

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
rena-cheng: Sounds good, will make a new PR for this update.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
```diff
@@ -0,0 +1,181 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# -*- coding: utf-8 -*
+# mypy: disable-error-code="attr-defined"
+
+import gui_locators
+import squish
+import test
+
+snoozeTimeout = 3  # seconds
```
justinsaws: nit: If this is a constant we should be using standard Python style for it.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def launchDeadlineConfigGUI():
+    squish.startApplication("deadline config gui")
+    test.log("Launching Deadline Config GUI")
+    test.log(
+        "Sleep for " + str(snoozeTimeout) + " seconds to allow GUI authentication to fully load."
+    )
+    squish.snooze(snoozeTimeout)
+    test.compare(
```
justinsaws, line 19: If these are helper functions, why are they testing states of the application?

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
```diff
@@ -0,0 +1,28 @@
...
+farmName = "Deadline Cloud Squish Farm"
+queueName = "Squish Automation Queue"
+jobAttachments = "COPIED"
+storageProfile = "Squish Storage Profile"
+conflictResOption = "NOT\\_SELECTED"
+loggingLevel = "WARNING"
+
+
+def main():
+    gui_helpers.launchDeadlineConfigGUI()
```
justinsaws: Is there a way to structure these in their own individual tests? What does the log output look like when a test run is done? Is it easy to see which tests have failed at a glance?

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
```diff
@@ -0,0 +1,28 @@
...
+    gui_helpers.setGlobalSettings(profileName)
+    gui_helpers.setProfileSettings(jobHistDir, farmName)
+    gui_helpers.openAndSetDefaultJobHistDirectory()
+    gui_helpers.setFarmSettings(queueName, storageProfile, jobAttachments)
+    gui_helpers.setGeneralSettings(conflictResOption, loggingLevel)
+    # disabling due to https://sim.amazon.com/issues/Bea-28289
+    # gui_helpers.hitApplyButton()
+    test.log("All settings config have been applied.")
+    gui_helpers.closeDeadlineConfigGUI()
```
justinsaws: If one of the tests above fail, does this line still get called? Is the machine that was running the tests left in a weird state?

rena-cheng: No - Currently, if a test fails (at least for the Settings dialogue), Squish will shut down the GUI (and any new tests would pretty much be started up again and any failed states won't affect it). But I do agree we should add 'beforeAll' or 'afterAll' blocks to set the test up and help clean things up. 

Edit - Did some research and it looks like the equivalent of 'beforeAll' and 'afterAll' is 'setUp' and 'tearDown' in Python. I tried incorporating these into the test but it caused Squish to error out ...

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: Do you have an example? (Is this comment referring to lines 6-9?) 
I ran linters, etc, but there were issues since some of the Python libraries are named the same as Squish libraries and we need to point correctly to the Squish libraries (hence adding `# mypy: disable-error-code="attr-defined"`). I do not know a way around this currently.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: By 'state of the application', do you mean to whether the application is opened, or not?

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
rena-cheng: Each line in def main() function is it's own individual test (or what do you mean by this)? 

Squish will print out what test failed (or passed) as well-- attached screenshot when running the Squish test
Screenshot 2024-07-15 at 3 29 18 PM
 from the Command Line. I've also written these tests so that it should be easy to deduce what specific test failed when the test run is over. Also open to any specific suggestions, etc.

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: Note for self: I decided to make a revision; will incorporate this change.

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
rena-cheng: It actually doesn't look like Squish allows you to rename the `test.py` file (and if you try to rename it locally, it looks like Squish crashes when you try to rename the test...)

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: Ok - I think I misunderstood the comment. I guess when I worked in Selenium, these functions were called Page Objects. I don't know what's the equivalent name for GUI QT testing.....

justinsaws: Since we are running tests in these functions, I wouldn't consider them helpers but the tests themselves. I expect utility functions to be in a file called `gui_helpers.py`

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
justinsaws: It likely won't be a Python test setup thing but something specific to Squish. I'm pretty sure they have the concept of test suites that you can setup and have standard setup and teardown. 

If these tests are run as is on a developers laptop, will they leave state behind say the next time the just run the deadline CLI on their own?

test/squish/suite_VerifySettingsDialog/shared/scripts/gui_helpers.py
rena-cheng: Just adding a note here that I will make a change for this (and it is a fairly easy change as it just involves re-organizing some lines of code). While tests/'verifications' are allowed to be within the helper functions, it does make sense to separate out some of the tests/'verifications' and put them on the actual test file itself. 

Note that tests/'verifications' can be added within helper functions as helper functions assist in 'setting up' the test, the tests/'verifications' ensure that the 'setup' is correct as well before proceeding further in the tests themselves.

test/squish/suite_VerifySettingsDialog/tst_VerifySettingsDialog/test.py
rena-cheng: Yes - It looks like Squish does have some concepts on setup/teardown. Reference is here: https://doc.qt.io/squish/tester-created-special-functions.html

I was thinking about this, and I can't think of any setup/teardown function we should add other than login and logout functions (for the dialogue). The farm resources that appear in the dialogue pretty much depend on what AWS profile you are logged into, so the best thing to do would be to logout of the dialogue when the test finishes (and then the developer would need to log back in themselves and just re-select the profile they want to use). Additionally, when the test re-runs, the test would need to log back in as well (Once logged back in, the test would proceed and select the correct AWS profile, farm resources, verify those resources, etc). Let me know what you think, or if you had any other ideas. 

With that being said, I tried importing the Deadline CLI API into the Squish framework and tested out the login/logout api/functions from the Deadline API. The login api/function had some issues. There were errors saying that the AWS Profile had to be created by DCM. Additionally, I also had to modify some code in the login api/function so to get it working when importing the api so it makes me think there is a bug with the login api/function since it didn't work right off the bat. But the error saying that the 'AWS Profile had to be created by DCM' makes me think getting the login api/function to work won't be that simple. 

I tried the logout api/function and it worked right off the bat and was usable. 

I'll need to look into this further once I am back from leave.

rena-cheng:
Closing this PR as this ended up being old, and I have a new one here (which is the final PR): https://github.com/aws-deadline/deadline-cloud/pull/434

PR #403, opened 2024-07-10
fix: Add better log when JA hit Windows path length limit

AlexTranAmz:
If we have file path that is longer than 260 chars on Windows, JA will throw an error of file not found which is not necessary correct and do not bring any value to track down the problem.

Note: This only apply for Windows that have path length registry disable and/or path start with drive letter (i.e. `C:\`)
Add comparison of file path and if longer than 260 chars, we will throw a more meaningful error message
We have more meaningful logs so it's easier to detect what when wrong
Run all unit + integ tests
Manually run download job output test on Linux and Windows
Manually run Asset sync on Linux and Windows CMF
image

Yes

No

----

### DIFF
--- a/src/deadline/job_attachments/_utils.py
+++ b/src/deadline/job_attachments/_utils.py
@@ -90,7 +90,7 @@ def _is_relative_to(path1: Union[Path, str], path2: Union[Path, str]) -> bool:
         return False
 
 
-def _is_Windows_file_path_limit() -> bool:
+def _is_windows_file_path_limit() -> bool:
     if sys.platform == "win32":
         ntdll = ctypes.WinDLL("ntdll")
         ntdll.RtlAreLongPathsEnabled.restype = ctypes.c_ubyte
--- a/src/deadline/job_attachments/_utils.py
+++ b/src/deadline/job_attachments/_utils.py
@@ -5,6 +5,8 @@ from hashlib import shake_256
 from pathlib import Path
 from typing import Tuple, Union
 import uuid
+import ctypes
+import sys
 
 
 __all__ = [
@@ -86,3 +88,13 @@ def _is_relative_to(path1: Union[Path, str], path2: Union[Path, str]) -> bool:
         return True
     except ValueError:
         return False
+
+
+def _is_Windows_file_path_limit() -> bool:
+    if sys.platform == "win32":
+        ntdll = ctypes.WinDLL("ntdll")
+        ntdll.RtlAreLongPathsEnabled.restype = ctypes.c_ubyte
+        ntdll.RtlAreLongPathsEnabled.argtypes = ()
+
+        return bool(ntdll.RtlAreLongPathsEnabled())
+    return True
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -67,7 +67,7 @@ from .os_file_permission import (
     _set_fs_group_for_posix,
     _set_fs_permission_for_windows,
 )
-from ._utils import _is_relative_to, _join_s3_paths, _is_Windows_file_path_limit
+from ._utils import _is_relative_to, _join_s3_paths, _is_windows_file_path_limit
 
 download_logger = getLogger("deadline.job_attachments.download")
 
@@ -522,7 +522,7 @@ def download_file(
         # For example: file test.txt when download will be test.txt.H4SD9Ddj
         if (
             len(str(local_file_name)) + TEMP_DOWNLOAD_ADDED_CHARS_LENGTH >= WINDOWS_MAX_PATH_LENGTH
-            and not _is_Windows_file_path_limit()
+            and not _is_windows_file_path_limit()
         ):
             uncPath = str(local_file_name).startswith("\\\\?\\")
             if uncPath:
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -67,12 +67,13 @@ from .os_file_permission import (
     _set_fs_group_for_posix,
     _set_fs_permission_for_windows,
 )
-from ._utils import _is_relative_to, _join_s3_paths
+from ._utils import _is_relative_to, _join_s3_paths, _is_Windows_file_path_limit
 
 download_logger = getLogger("deadline.job_attachments.download")
 
 S3_DOWNLOAD_MAX_CONCURRENCY = 10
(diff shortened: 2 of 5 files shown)


src/deadline/job_attachments/download.py
```diff
@@ -516,7 +517,14 @@ def process_client_error(exc: ClientError, status_code: int):
             error_details=str(bce),
         ) from bce
     except Exception as e:
-        raise AssetSyncError(e) from e
+        # Add 9 to account for .Hex value when file in the middle of downloading in windows paths
+        # For example: file test.txt when download will be test.txt.H4SD9Ddj
+        if len(str(local_file_name)) + 9 < WINDOWS_MAX_PATH_LENGTH or os.name == "posix":
```
jairaws: The path size limit can vary, either by setting an env variable or by using long path format. How are we handling those cases?

nit: It would be more efficient to check first the OS. Otherwise we always check the path size for non windows OSs.

nit: It would be better to use constants instead of hardcoded values.

leongdl: This won't be the case once we enable long file path support. Also, if the file path on windows starts with \\?\ then long file paths do work. But the behavior varies depending on whether the customer has enabled long paths in the system registry. 

Is there a way to unpack and pick a specific error, and also check for the UNC path with a starts with?

test/integ/deadline_job_attachments/conftest.py
```diff
@@ -103,3 +104,13 @@ def external_bucket() -> str:
 
 def is_windows_non_admin():
     return sys.platform == "win32" and getpass.getuser() != "Administrator"
+
+
+def is_Windows_file_path_limit():
```
leongdl: I think we should use this in the code to determine the behavior!

src/deadline/job_attachments/download.py
leongdl: I think the logic should be like this:

```
if os.name == "windows" and path.length > 260:
    uncPath = path.beginsWith("\\?\")
    if not uncPath:
        print out the proposed error
    else not isLongPathRegistryEnabled():
       print UNC Path but long path registry not enabled; undefined behavior. Print Error
    else:
       # something else.
       raise e
```

leongdl (commented):
NIT: for "what" is the problem in the PR description
```
If we have file path that is longer than 260 chars on Windows, JA will throw an error of file not found which is not necessary correct and do not bring any value to track down the problem
```

This only applies on windows if the path does NOT use UNC. EG, if it starts with C:\\ (or any drive)

src/deadline/job_attachments/download.py
AlexTranAmz: Added in the new commit

test/integ/deadline_job_attachments/conftest.py
AlexTranAmz: Done in the new commit

src/deadline/job_attachments/_utils.py
```diff
@@ -86,3 +88,13 @@ def _is_relative_to(path1: Union[Path, str], path2: Union[Path, str]) -> bool:
         return True
     except ValueError:
         return False
+
+
+def _is_Windows_file_path_limit() -> bool:
```
leongdl: nit, caps `Windows` ?

src/deadline/job_attachments/download.py
```diff
@@ -516,6 +518,22 @@ def process_client_error(exc: ClientError, status_code: int):
             error_details=str(bce),
         ) from bce
     except Exception as e:
+        # Add 9 to account for .Hex value when file in the middle of downloading in windows paths
+        # For example: file test.txt when download will be test.txt.H4SD9Ddj
+        if (
```
leongdl, line 524: nice! Good set of logic paths.

test/integ/deadline_job_attachments/conftest.py
leongdl: duplicate code? can it use the definition above?

AlexTranAmz: The reason is I do not care if it's Admin or not for this particular function. Which is why I have the sys.platform check

src/deadline/job_attachments/download.py
AlexTranAmz: Tks!

test/integ/deadline_job_attachments/conftest.py
leongdl: Don't understand here, the code above is 100% the same 

```
    if sys.platform == "win32":
        ntdll = ctypes.WinDLL("ntdll")
        ntdll.RtlAreLongPathsEnabled.restype = ctypes.c_ubyte
        ntdll.RtlAreLongPathsEnabled.argtypes = ()

    return True
```

src/deadline/job_attachments/download.py
```diff
@@ -516,6 +518,22 @@ def process_client_error(exc: ClientError, status_code: int):
...
+            len(str(local_file_name)) + TEMP_DOWNLOAD_ADDED_CHARS_LENGTH >= WINDOWS_MAX_PATH_LENGTH
+            and not _is_windows_file_path_limit()
+        ):
+            uncPath = str(local_file_name).startswith("\\\\?\\")
+            if uncPath:
+                raise AssetSyncError(
+                    f"UNC notation exist, but long path registry not enabled. Undefided error\n{e}"
```
leongdl: undefined.

leongdl (dismissed):

test/integ/deadline_job_attachments/test_job_attachments.py
```diff
@@ -1487,3 +1488,53 @@ def test_download_outputs_bucket_wrong_account(
...
+            farm_id=job_attachment_test.farm_id,
+            queue_id=job_attachment_test.queue_id,
+            job_id=sync_outputs.job_id,
+            step_id=sync_outputs.step0_id,
+            task_id=sync_outputs.step0_task0_id,
+        )
+        job_output_downloader.set_root_path(
+            str(job_attachment_test.ASSET_ROOT), str(long_root_path)
+        )
+        job_output_downloader.download_job_output()
```
jericht: Which of these three function calls do we expect to fail? `download_job_output` right? We should scope the `with pytest.raises` block to only that function to not get false positives.

jericht (approved):

leongdl (approved):

PR #410, opened 2024-07-17
feat(asset-cli): asset diff subcommand to diff directory changes with manfiest

seant-aws:
Users want to be able to view directory changes since the previous capture of manifest, showing what files have been modified, added, or deleted.

Created an asset CLI subcommand diff to diff directories with their corresponding manifests

Users are now able to compare their local changes with their local manifests using a CLI command

- Added unit tests mocking snapshot functionality at test/unit/deadline_client/cli/test_cli_asset.py, passing successfully
- manual end to end testing with local directories and local manifests

Sample Output Default:
image

Sample Output Raw:
image

----

### DIFF
--- a/src/deadline/client/cli/_groups/asset_group.py
+++ b/src/deadline/client/cli/_groups/asset_group.py
@@ -237,7 +237,9 @@ def asset_upload(root_dir: str, manifest_dir: str, update: bool, **args):
 @cli_asset.command(name="diff")
 @click.option("--root-dir", help="The root directory to compare changes to. ")
 @click.option(
-    "--manifest", help="The path to manifest folder of the directory to show changes of. "
+    "--manifest",
+    required=True,
+    help="The path to manifest folder of the directory to show changes of. ",
 )
 @click.option(
     "--format",
@@ -247,13 +249,58 @@ def asset_upload(root_dir: str, manifest_dir: str, update: bool, **args):
     default=False,
 )
 @_handle_error
-def asset_diff(**args):
+def asset_diff(root_dir: str, manifest: str, format: bool, **args):
     """
-    Check file differences of a directory since last snapshot.
+    Check file differences of a directory since last snapshot, specified by manifest.
+    """
+    print("root_dir: ", root_dir)
 
-    TODO: show example of diff output
-    """
-    click.echo("diff shown")
+    if not os.path.isdir(manifest):
+        raise NonValidInputError(f"Specified manifest directory {manifest} does not exist. ")
+
+    if root_dir is None:
+        asset_root_dir = os.path.dirname(manifest)
+    else:
+        if not os.path.isdir(root_dir):
+            raise NonValidInputError(f"Specified root directory {root_dir} does not exist. ")
+        asset_root_dir = root_dir
+
+    # Placeholder Asset Manager
+    asset_manager = S3AssetManager(
+        farm_id=" ", queue_id=" ", job_attachment_settings=JobAttachmentS3Settings(" ", " ")
+    )
+
+    # get inputs of directory
+    input_paths = []
+    for root, dirs, files in os.walk(asset_root_dir):
+        # ignore manifest folder
+        # if os.path.samefile(root, manifest):
+        #     dirs[:] = []
+        #     continue
+        for filename in files:
+            file_path = os.path.join(root, filename)
+            input_paths.append(Path(file_path))
+
+    # hash and create manifest of local directory
+    cache_config = config_file.get_cache_directory()
+    with HashCache(cache_config) as hash_cache:
+        directory_manifest_object = asset_manager._create_manifest_file(
+            input_paths=input_paths, root_path=asset_root_dir, hash_cache=hash_cache
+        )
+
+    # parse local manifest
+    local_manifest_object = read_local_manifest(manifest=manifest)
+
+    # compare manifests
+    differences = compare_manifest(
+        reference_manifest=local_manifest_object, compare_manifest=directory_manifest_object
+    )
+
+    if format:
+        click.echo(f"\n{asset_root_dir}")
+        pretty_print(file_status_list=differences)
+    else:
+        click.echo(f"\nFile Diffs: {differences}")
 
 
 @cli_asset.command(name="download")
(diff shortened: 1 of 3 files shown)


src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -237,23 +239,64 @@ def asset_upload(root_dir: str, manifest_dir: str, update: bool, **args):
...
+
+    if root_dir is None:
+        asset_root_dir = os.path.dirname(manifest_dir)
+    else:
+        if not os.path.isdir(root_dir):
+            raise NonValidInputError(f"Specified root directory {root_dir} does not exist. ")
+        asset_root_dir = root_dir
+
+    # Placeholder Asset Manager
+    asset_manager = S3AssetManager(
```
leongdl, line 269: Why placeholder?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -408,3 +451,112 @@ def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAs
...
+                f"{prefix}{symbol}{color}{entry}{tooltip}{COLORS['reset']}{os.path.sep if is_dir else ''}"
+            )
+
+            if is_dir:
+                new_prefix = prefix + (SPACE if is_last_entry else PIPE + SPACE)
+                print_tree(subtree, new_prefix)
+
+        if not directory_tree:
+            symbol = ELBOW + HORIZONTAL
+            print(f"{prefix}{symbol}{COLORS['unchanged']}. {COLORS['reset']}")
```
leongdl: can we use a logger instead of print?

leongdl:
nit: for the CR description can you please include a sample execution of what the diff looks like? I think it will really help reviewers.

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Sorry, meant as a just a blank Asset Manager to access one of the asset manager functions; this was also something done in snapshot, Jericho and I talked about doing a refactor pass for this

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -237,23 +240,63 @@ def asset_upload(root_dir: str, manifest_dir: str, update: bool, **args):
...
+    # parse local manifest
+    local_manifest_object: BaseAssetManifest = read_local_manifest(manifest=manifest_dir)
+
+    # compare manifests
+    differences: List[tuple] = compare_manifest(
+        reference_manifest=local_manifest_object, compare_manifest=directory_manifest_object
+    )
+
+    if raw:
+        click.echo(f"\nFile Diffs: {differences}")
```
godobyte, line 296: You PR description has the pretty print result, maybe also add the raw prints for completeness.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -408,3 +451,125 @@ def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAs
...
+) -> List[(tuple)]:
+    """
+    Compares two manifests, reference_manifest acting as the base, and compare_manifest acting as manifest with changes.
+    Returns a list of FileStatus and BaseManifestPath
+
+    """
+    reference_dict = {
+        manifest_path.path: manifest_path for manifest_path in reference_manifest.paths
+    }
+    compare_dict = {manifest_path.path: manifest_path for manifest_path in compare_manifest.paths}
```
godobyte, line 467: nit - code format looks inconsistent here.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -408,3 +451,125 @@ def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAs
...
+    for file_path, manifest_path in compare_dict.items():
+        if file_path not in reference_dict:
+            differences.append((FileStatus.NEW, manifest_path))
+            continue
+        if file_path in reference_dict:
+            if reference_dict[file_path].hash != manifest_path.hash:
+                differences.append((FileStatus.MODIFIED, manifest_path))
+            else:
+                differences.append((FileStatus.UNCHANGED, manifest_path))
+            continue
```
godobyte: This continue looks redundant?

What do you think about using if-else (without continue) in the for loop? The conditions seem to be mutually exclusive.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -408,3 +451,125 @@ def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAs
...
+    for file_path, manifest_path in reference_dict.items():
+        if file_path not in compare_dict:
+            differences.append((FileStatus.DELETED, manifest_path))
+
+    return differences
+
+
+def pretty_print(file_status_list: List[(tuple)]):
+    """
+    Prints to command line a well formatted version of
```
godobyte: This comment looks unfinished?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -408,3 +451,125 @@ def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAs
...
+    SPACE = "    "
+
+    # ANSI escape sequences for colors
+    COLORS = {
+        "MODIFIED": "\033[93m",  # yellow
+        "NEW": "\033[92m",  # green
+        "DELETED": "\033[91m",  # red
+        "UNCHANGED": "\033[90m",  # grey
+        "reset": "\033[0m",  # base color
+        "directory": "\033[90m",  # grey
```
godobyte: Why `reset` and `directory` keys are in lowercase while others are in uppercase?

test/unit/deadline_client/cli/test_cli_asset.py
```diff
@@ -1,16 +1,23 @@
...
 
 from deadline.client.cli import main
 from deadline.client.cli._groups import asset_group
 from deadline.client import api
 from deadline.client.api import _submit_job_bundle
-from deadline.job_attachments.models import AssetRootGroup, JobAttachmentS3Settings, Attachments
-from deadline.job_attachments.upload import S3AssetManager, S3AssetUploader
+from deadline.job_attachments.models import (
+    AssetRootGroup,
+    # AssetRootManifest,
```
godobyte: Why comment it out? Let's remove it if it's not in use,

godobyte (changes requested):

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Yea this part bothers me a bit, but it's run as a result of formatting with `hatch run fmt`, unless I change the variable names to be shorter, the reference_dict line will be auto formatted I think

godobyte: Interesting. What if you manually format this part similar to `reference_dict`, would `hatch run fmt` format it back?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -408,3 +451,123 @@ def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAs
...
+    }
+    compare_dict = {manifest_path.path: manifest_path for manifest_path in compare_manifest.paths}
+
+    differences = []
+
+    # Find new files
+    for file_path, manifest_path in compare_dict.items():
+        if file_path not in reference_dict:
+            differences.append((FileStatus.NEW, manifest_path))
+        elif file_path in reference_dict:
```
godobyte: This seems an `else` to me. Having `elif` without `else` could leave to unhandled condition. If this is complementary of the prior condition, let's use `else`.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -408,3 +451,123 @@ def update_manifest(manifest: str, new_or_modified_paths: List[tuple]) -> BaseAs
...
+        "DIRECTORY": "\033[90m",  # grey
+    }
+
+    # Tooltips:
+    TOOLTIPS = {
+        "NEW": " +",  # added files
+        "DELETED": " -",  # deleted files
+        "MODIFIED": " M",  # modified files
+        "UNCHANGED": "",  # unchanged files
+    }
```
godobyte: food for thought + not blocking - The keys here (MODIFIED/NEW/DELETED/UNCHANGED) have to match FileStatus Enum for the logic to work. Could we think of a way to improve that?

There is an approach to consider
- https://realpython.com/python-enum/#using-enumerations-in-if-and-match-statements

godobyte (changes requested):

src/deadline/client/cli/_groups/asset_group.py
seant-aws: sounds good, also to reply to the above comment, unfortunately hatch run fmt will reformat it :(

src/deadline/client/cli/_groups/asset_group.py
seant-aws: This is definitely cleaner, and I've added that change for the tooltips, however since colors also contain non-filestatus colors, like directory, I don't think I can easily change that for the colors enum

leongdl (approved):
LGTM

leongdl (commented):
If you have time, please add a few integration tests which uploads to a real bucket and does some diff.

godobyte (approved):

src/deadline/client/cli/_groups/asset_group.py
godobyte: We could still use the enum for file status colors' keys. Having non-filestatus colors shouldn't make a difference.

PR #416, opened 2024-07-22
feat: support --submitter-name option for bundle GUI submitter command

crowecawcaw:
One great way of integrating Deadline Cloud into a DCC is adding a button that invokes the GUI bundle submitter. To improve the UX, we can allow the submitter name (currently always "JobBundle" for bundle GUI submissions) to be customized. The submitter name is used for some alerts (e.g. an alert title "JobBundle job submission") and for bundle directory names in the job history folder.  

Add a new `--submitter-name` option.

A process that submits a job using the GUI bundle submitter appears more like a "real submitter" in the UI and job history folder. 

Unit tests. Also manual tests:
- export a bundle without a custom submitter name. Verified the job bundle directory included "JobBundle" (previous behavior)
- export a bundle with a custom submitter name. Verified the job bundle directory included the custom name
- submit a bundle without a custom submitter name
- submit a bundle with a custom submitter name

Documented in the CLI help automatically.

No
----

### DIFF
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -257,6 +257,11 @@ def bundle_submit(
     help="Installs GUI dependencies if they are not installed already",
 )
 @click.option(
+    "--submitter-name",
+    type=click.STRING,
+    help="Name of the application submitting the bundle. If a name is specified, the GUI will automatically close after submitting the job.",
+)
+@click.option(
     "--output",
     type=click.Choice(
         ["verbose", "json"],
@@ -269,7 +274,7 @@ def bundle_submit(
     "parsed/consumed by custom scripts.",
 )
 @_handle_error
-def bundle_gui_submit(job_bundle_dir, browse, output, install_gui, **args):
+def bundle_gui_submit(job_bundle_dir, browse, output, install_gui, submitter_name, **args):
     """
     Opens GUI to submit an Open Job Description job bundle to AWS Deadline Cloud.
     """
@@ -284,7 +289,9 @@ def bundle_gui_submit(job_bundle_dir, browse, output, install_gui, **args):
             )
         output = output.lower()
 
-        submitter = show_job_bundle_submitter(input_job_bundle_dir=job_bundle_dir, browse=browse)
+        submitter = show_job_bundle_submitter(
+            input_job_bundle_dir=job_bundle_dir, browse=browse, submitter_name=submitter_name
+        )
 
         if not submitter:
             return
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -107,6 +107,11 @@ def validate_parameters(ctx, param, value):
     is_flag=True,
     help="Require all input paths to exist",
 )
+@click.option(
+    "--submitter-name",
+    type=click.STRING,
+    help="Name of the application submitting the bundle.",
+)
 @click.argument("job_bundle_dir")
 @_handle_error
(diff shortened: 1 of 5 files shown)


leongdl (dismissed):

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -85,13 +86,15 @@ def __init__(
...
         show_host_requirements_tab=False,
+        submitter_name: Optional[str] = None,
     ):
         # The Qt.Tool flag makes sure our widget stays in front of the main application window
         super().__init__(parent=parent, f=f)
         self.setWindowTitle("Submit to AWS Deadline Cloud")
         self.setMinimumSize(400, 400)
 
         self.job_settings_type = type(initial_job_settings)
+        self.submitter_name = submitter_name or self.job_settings_type.submitter_name
```
epmog: Here you're grabbing the value from the type rather than instance that it was previously doing. This is potentially bug/breaking change for consumers of this function?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -269,7 +274,7 @@ def _decide_cancel_submission(upload_group: AssetUploadGroup) -> bool:
     "parsed/consumed by custom scripts.",
 )
 @_handle_error
-def bundle_gui_submit(job_bundle_dir, browse, output, install_gui, **args):
+def bundle_gui_submit(job_bundle_dir, browse, output, install_gui, submitter_name, **args):
```
epmog, line 284: Does the non-gui version also need this? Consider an application that shows it own prompts but leverages the `deadline bundle submit` command to actually do the submission.

epmog (changes requested):

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
crowecawcaw: It took me a few iterations to sort through where `submitter_name` came from. Previously, `settings` were initialized from the settings type, not `initial_settings`. This change falls back to the previous behavior if a `submitter_name` is not provided to this function.

crowecawcaw: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py#L333

src/deadline/client/cli/_groups/bundle_group.py
crowecawcaw: The regular bundle submit command doesn't have a GUI and doesn't need the dependencies right?

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
epmog: `self.job_settings_type.submitter_name` vs. `self.job_settings_type().submitter_name`  was the difference I was calling out. Accessing class variable vs. instance variable.

src/deadline/client/cli/_groups/bundle_group.py
epmog: Correct, we do not bring in GUI deps for the CLI and should not force/assume they're there.

crowecawcaw: Sorry I got mixed up with another PR. Makes sense, will update.

epmog (approved):
LGTM! Thanks Stephen

leongdl (approved):

PR #420, opened 2024-07-25
feat(asset-cli): asset download subcommand

seant-aws:
Users want to be able to easily find the input manifest files of a previously submitted job, not just a job's output files.

Created a CLI command to allow users to download a job's corresponding input manifest, using job id's

Users are able to download manifest files off of their S3 bucket, with the S3 key mapped form job information. Once downloaded, users can now view, and share these manifests files to show the inputs of a render job.

- added unit tests mocking download functionality at test/unit/deadline_client/cli/test_cli_asset.py, passing successfully
- ran integ tests to check refactoring of existing code
- ran manual end to end test to test the refactoring of download functionalities
image

image

----

### DIFF
--- a/src/deadline/client/cli/_groups/asset_group.py
+++ b/src/deadline/client/cli/_groups/asset_group.py
@@ -20,13 +20,19 @@ import click
 
 from deadline.client import api
 from deadline.job_attachments.upload import FileStatus, S3AssetManager, S3AssetUploader
+from deadline.job_attachments.download import download_file_with_s3_key
 from deadline.job_attachments.models import (
     JobAttachmentS3Settings,
     AssetRootManifest,
+    S3_MANIFEST_FOLDER_NAME,
 )
 from deadline.job_attachments.asset_manifests.decode import decode_manifest
 from deadline.job_attachments.asset_manifests.base_manifest import BaseAssetManifest
 from deadline.job_attachments.caches import HashCache
+from deadline.job_attachments._aws.aws_clients import (
+    get_s3_client,
+    get_s3_transfer_manager,
+)
 
 from .._common import _apply_cli_options_to_config, _handle_error, _ProgressBarCallbackManager
 from ...exceptions import NonValidInputError, ManifestOutdatedError
@@ -300,15 +306,83 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
 
 
 @cli_asset.command(name="download")
-@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
-@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
+@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
+@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
 @click.option("--job-id", help="The AWS Deadline Cloud Job to get. ")
+@click.option(
+    "--manifest-out",
+    required=True,
+    help="Destination path to directory where manifest is created. ",
+)
 @_handle_error
-def asset_download(**args):
+def asset_download(manifest_out: str, **args):
     """
     Downloads input manifest of previously submitted job.
     """
-    click.echo("download complete")
+    if not os.path.isdir(manifest_out):
+        raise NonValidInputError(f"Specified destination directory {manifest_out} does not exist. ")
+
+    # setup config
+    config = _apply_cli_options_to_config(
+        required_options={"farm_id", "queue_id", "job_id"}, **args
+    )
+    deadline: BaseClient = api.get_boto3_client("deadline", config=config)
+    queue_id: str = get_setting("defaults.queue_id", config=config)
+    farm_id: str = get_setting("defaults.farm_id", config=config)
+    job_id = config_file.get_setting("defaults.job_id", config=config)
+
+    queue: dict = deadline.get_queue(
+        farmId=farm_id,
+        queueId=queue_id,
+    )
+
+    # assume queue role - session permissions
+    queue_role_session: boto3.Session = api.get_queue_user_boto3_session(
+        deadline=deadline,
+        config=config,
+        farm_id=farm_id,
+        queue_id=queue_id,
+        queue_display_name=queue["displayName"],
+    )
+
+    # get input_manifest_paths from Deadline GetJob API
+    job: dict = deadline.get_job(farmId=farm_id, queueId=queue_id, jobId=job_id)
+    attachments: dict = job["attachments"]
+    input_manifest_paths: list[str] = [
+        manifest["inputManifestPath"] for manifest in attachments["manifests"]
+    ]
+
+    # get s3BucketName from Deadline GetQueue API
+    bucket_name: str = queue["jobAttachmentSettings"]["s3BucketName"]
+
+    # get S3 prefix
+    s3_prefix: Path = Path(queue["jobAttachmentSettings"]["rootPrefix"], S3_MANIFEST_FOLDER_NAME)
+
+    s3_client: BaseClient = get_s3_client(session=queue_role_session)
+    transfer_manager = get_s3_transfer_manager(s3_client=s3_client)
+
+    # download each input_manifest_path
+    for input_manifest_path in input_manifest_paths:
+        local_file_name = Path(manifest_out, job_id + "_manifest")
+
+        result = download_file_with_s3_key(
+            s3_bucket=bucket_name,
+            s3_key=(s3_prefix / input_manifest_path).as_posix(),
+            local_file_name=local_file_name,
+            session=queue_role_session,
+            transfer_manager=transfer_manager,
+        )
+
+        if result is not None:
+            transfer_path = result.meta.call_args.fileobj  # type: ignore[attr-defined]
+            file_size = result.meta.size  # type: ignore[attr-defined]
+            click.echo(
+                f"\nDownloaded file to '{transfer_path}' ({file_size} bytes)\nWith S3 key: '{input_manifest_path}'. "
+            )
+        else:
+            click.echo(
+                f"\nFailed to download file with S3 key '{input_manifest_path}' from bucket '{bucket_name}'"
+            )
 
 
 def read_local_manifest(manifest: str) -> BaseAssetManifest:
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -372,6 +372,68 @@ def download_files_in_directory(
     )
 
 
+def download_file_with_s3_key(
+    s3_bucket: str,
+    s3_key: str,
+    transfer_manager,
+    local_file_name: Path,
+    file_bytes: Optional[int] = 0,
+    session: Optional[boto3.Session] = None,
+    progress_tracker: Optional[ProgressTracker] = None,
+    file_conflict_resolution: Optional[FileConflictResolution] = FileConflictResolution.CREATE_COPY,
+) -> Optional[concurrent.futures.Future]:
+    """
+    Helper to download a file from the S3 bucket with a specified S3 key to the local directory.
+    Returns the asynchronous result of the downloaded file, and None if otherwise
+    """
+
+    # If the file name already exists, resolve the conflict based on the file_conflict_resolution
+    if local_file_name.is_file():
+        if file_conflict_resolution == FileConflictResolution.SKIP:
+            # if file is said to be skipped, no dowload will occur, and None will be returned.
+            # This allows `download_file` to return (file_bytes, None) when this function is called.
+            return None
+        elif file_conflict_resolution == FileConflictResolution.OVERWRITE:
+            pass
+        elif file_conflict_resolution == FileConflictResolution.CREATE_COPY:
+            # This loop resolves filename conflicts by appending " (1)"
+            # to the stem of the filename until a unique name is found.
+            while local_file_name.is_file():
+                local_file_name = local_file_name.parent.joinpath(
+                    local_file_name.stem + " (1)" + local_file_name.suffix
+                )
+        else:
+            raise ValueError(
+                f"Unknown choice for file conflict resolution: {file_conflict_resolution}"
+            )
+
+    local_file_name.parent.mkdir(parents=True, exist_ok=True)
+
+    future: concurrent.futures.Future
+
+    # provides progress callback for asynchronous download from S3
+    def handler(bytes_downloaded):
+        nonlocal progress_tracker
+        nonlocal future
+
+        if progress_tracker:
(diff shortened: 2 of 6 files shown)


src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -118,6 +120,7 @@ def asset_snapshot(root_dir: str, manifest_out: str, recursive: bool, **args):
 
 
 @cli_asset.command(name="upload")
+@click.option("--root-dir", help="The root directory of assets to upload. ")
 @click.option(
     "--root-dir",
     help="The root directory of assets to upload. Defaults to the parent directory of --manifest-dir if not specified. ",
```
godobyte: Is the `--root-dir` addition intended, how is it different from the one below?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -300,15 +303,65 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
 
 
 @cli_asset.command(name="download")
-@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
-@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
-@click.option("--job-id", help="The AWS Deadline Cloud Job to get. ")
+@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
+@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
+@click.option("--job-id", required=True, help="The AWS Deadline Cloud Job to get. ")
```
godobyte: Does it have to be required? There would be `defaults.job_id` in setting no?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -300,15 +303,65 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
...
+    )
+
+    # assume queue role - session permissions
+    queue_role_session = api.get_queue_user_boto3_session(
+        deadline=deadline,
+        config=config,
+        farm_id=farm_id,
+        queue_id=queue_id,
+        queue_display_name=queue["displayName"],
+    )
```
godobyte: Some of the params are optional for `get_queue_user_boto3_session`, do they all need to be passed in here?

src/deadline/job_attachments/upload.py
```diff
@@ -191,7 +191,7 @@ def upload_assets(
 
         if manifest_write_dir:
             self._write_local_manifest(
-                manifest_write_dir, manifest_name, full_manifest_key, manifest
+                manifest_write_dir, manifest_name, full_manifest_key, manifest, None
```
godobyte: `_write_local_manifest` already defaults `root_dir` to `None`, it doesn't need to be explicitly passed in.

https://github.com/stangch/deadline-cloud-stangch/blob/d69b729ea66a6475b049718bb00d6f41d3249ddd/src/deadline/job_attachments/upload.py#L239

src/deadline/job_attachments/download.py
```diff
@@ -372,48 +372,20 @@ def download_files_in_directory(
     )
 
 
-def download_file(
-    file: RelativeFilePath,
-    hash_algorithm: HashAlgorithm,
-    local_download_dir: str,
+def download_file_with_s3_key(
```
godobyte, line 375: Could you add test for the new function?

src/deadline/job_attachments/download.py
```diff
@@ -372,48 +372,20 @@ def download_files_in_directory(
...
-    # Python will handle the path separator '/' correctly on every platform.
-    local_file_name = Path(local_download_dir).joinpath(file.path)
-
-    s3_key = (
-        f"{cas_prefix}/{file.hash}.{hash_algorithm.value}"
-        if cas_prefix
-        else f"{file.hash}.{hash_algorithm.value}"
-    )
-
+):
```
godobyte: Could you add documentation for the new function too?

src/deadline/job_attachments/download.py
```diff
@@ -372,48 +372,20 @@ def download_files_in_directory(
...
     # If the file name already exists, resolve the conflict based on the file_conflict_resolution
     if local_file_name.is_file():
         if file_conflict_resolution == FileConflictResolution.SKIP:
-            return (file_bytes, None)
+            return None
```
godobyte, line 395: Is there any change of the behavior by modifying return statement here, any tests to validate?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -300,15 +303,65 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
...
+        local_file_name = Path(manifest_out, job_id + "_manifest")
+        _ = download_file_with_s3_key(
+            s3_bucket=bucket_name,
+            s3_key="DeadlineCloud/Manifests/" + input_manifest_path,
+            local_file_name=local_file_name,
+            session=queue_role_session,
+            transfer_manager=transfer_manager,
+        )
+
+    click.echo(f"Downloaded {job_id} manifest to {manifest_out}")
```
godobyte: What do you think about providing more download visibility to customer?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -300,15 +303,65 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
...
+    # get s3BucketName from Deadline GetQueue API
+    bucket_name = queue["jobAttachmentSettings"]["s3BucketName"]
+
+    s3_client = get_s3_client(session=queue_role_session)
+    transfer_manager = get_s3_transfer_manager(s3_client=s3_client)
+
+    # download each input_manifest_path
+    for input_manifest_path in input_manifest_paths:
+        local_file_name = Path(manifest_out, job_id + "_manifest")
+        _ = download_file_with_s3_key(
```
YutongLi291: Is there any reason we need to assign a throwaway variable here?

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -300,15 +303,65 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
...
+            s3_bucket=bucket_name,
+            s3_key="DeadlineCloud/Manifests/" + input_manifest_path,
```
godobyte: I don't think we can hard-code this key prefix, which is configurable by by customer. You'll need the root prefix from Queue job attachment settings.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -300,15 +303,65 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
...
+        queue_display_name=queue["displayName"],
+    )
+
+    # get input_manifest_paths from Deadline GetJob API
+    job = deadline.get_job(farmId=farm_id, queueId=queue_id, jobId=job_id)
+    attachments = job["attachments"]
+    input_manifest_paths = [manifest["inputManifestPath"] for manifest in attachments["manifests"]]
+
+    # get s3BucketName from Deadline GetQueue API
+    bucket_name = queue["jobAttachmentSettings"]["s3BucketName"]
```
YutongLi291: nit: We should have typing for variables. It seems that typing is scarce across this file  but to me it seems that it would be good to add just for development purposes.

src/deadline/job_attachments/download.py
YutongLi291: We should add typing for the return value of the function

src/deadline/job_attachments/download.py
```diff
@@ -490,6 +516,18 @@ def process_client_error(exc: ClientError, status_code: int):
...
+            def handler(bytes_downloaded):
+                nonlocal progress_tracker
+                nonlocal future
+
+                if progress_tracker:
+                    should_continue = progress_tracker.track_progress_callback(bytes_downloaded)
+                    if not should_continue:
+                        future.cancel()
+
+            subscribers = [ProgressCallbackInvoker(handler)]
```
godobyte: What's this used for, maybe it can use some documentation too?

godobyte (changes requested):

YutongLi291 (changes requested):

src/deadline/client/cli/_groups/asset_group.py
seant-aws: apologies, mistake during merge

src/deadline/client/cli/_groups/asset_group.py
seant-aws: Yea that should work too, but I was thinking since the command itself is explicitly trying to download the input manifests of a job, it would make more sense to have the job it downloads from to be explicitly specified ?

seant-aws: looking at other CLI examples though like the job group commands, doesn't specify job-id for download. Do you think it'll be best to follow those ?

src/deadline/client/cli/_groups/asset_group.py
seant-aws: the display name can be ignored, but I believe the other params are needed to get the role of the specific queue

src/deadline/job_attachments/download.py
seant-aws: with `download_file_with_s3_key`, all of the functionality was purely refactored out of `download_file`, since I wanted to use only part of the function in the asset CLI. There are already tests for the `download_file` module, which tests the same functionalities; would I need to add additional ones ?

src/deadline/job_attachments/download.py
seant-aws: I've tested with the existing tests of `download_file`, I've only refactored the return to be elsewhere, after the `download_file_with_s3_key` helper has returned; you can find the original `(file_bytes, None)` at line 485

src/deadline/client/cli/_groups/asset_group.py
seant-aws: previously was because the variable wasn't used, but after addressing Godot's comment I use the result for adding more user visibility to customers

src/deadline/job_attachments/download.py
seant-aws: I refactored this part form the original `download_file` command, which just provides statistics for the downloading progress

src/deadline/client/cli/_groups/asset_group.py
seant-aws: I've added a the download output filename/path, as well as size. The download callback would be a nice to have, but since it's only one single manifest file, and for time's sake, I'd like to pursue other changes

src/deadline/client/cli/_groups/asset_group.py
seant-aws: thnx for this, I noticed though the root prefix is only the first part of "Deadline/" and not the "Manifests" folder. Is it still acceptable for me to hardcode the "Manifests/" part ?

src/deadline/job_attachments/download.py
godobyte: It would still need the coverage for `download_file_with_s3_key` to ensure it works as intended. For example, I don't think the signature of this function is tested.

src/deadline/client/cli/_groups/asset_group.py
godobyte: Where is this `Manifests` created, is there a constant to share? This looks like a data contract that once change, many places need to be changed to. It's likely being defined already.

godobyte: See `job_attachments/models.py`

```
S3_DATA_FOLDER_NAME = "Data"
S3_MANIFEST_FOLDER_NAME = "Manifests"
S3_INPUT_MANIFEST_FOLDER_NAME = "Inputs"
```

src/deadline/job_attachments/download.py
godobyte: If there is another return None statement with `download_file_with_s3_key`, does it always makes sense for `download_file` to return `(file_bytes, None)`? Is that explicitly stated anywhere?

src/deadline/job_attachments/download.py
godobyte: This doesn't seem to be reflected in the `How was this change tested?`?

src/deadline/client/cli/_groups/asset_group.py
godobyte: The display name is good for visibility, let's keep that too.

src/deadline/client/cli/_groups/asset_group.py
godobyte: If it's a convention for other commands, not following it may lead to customer's confusion.

src/deadline/job_attachments/download.py
```diff
@@ -372,48 +372,25 @@ def download_files_in_directory(
...
+    s3_key: str,
+    transfer_manager,
+    local_file_name: Path,
+    file_bytes: Optional[int] = 0,
     session: Optional[boto3.Session] = None,
-    modified_time_override: Optional[float] = None,
     progress_tracker: Optional[ProgressTracker] = None,
     file_conflict_resolution: Optional[FileConflictResolution] = FileConflictResolution.CREATE_COPY,
-) -> Tuple[int, Optional[Path]]:
+) -> concurrent.futures.Future | None:
```
YutongLi291: Instead of using ` | None` we generally use `Optional[concurrent.futures.Future]` instead.

YutongLi291 (dismissed):

src/deadline/job_attachments/download.py
seant-aws: I've done some extensive testing and reviewing:
- ran existing integ tests
- ran manual end to end testing by downloading outputs, cancelling downloaded outputs, and comparing these outputs with the same process on mainline
- consulted previous implementer of this part of code, which seemed good

I don't believe there is any regression with this refactoring.

src/deadline/client/cli/_groups/asset_group.py
```diff
@@ -300,15 +306,84 @@ def asset_diff(root_dir: str, manifest_dir: str, raw: bool, **args):
...
+    attachments: dict = job["attachments"]
+    input_manifest_paths: list[str] = [
+        manifest["inputManifestPath"] for manifest in attachments["manifests"]
+    ]
+
+    # get s3BucketName from Deadline GetQueue API
+    bucket_name: str = queue["jobAttachmentSettings"]["s3BucketName"]
+
+    # get S3 prefix
+    s3Prefix: Path = Path(queue["jobAttachmentSettings"]["rootPrefix"], S3_MANIFEST_FOLDER_NAME)
```
godobyte: `s3Prefix` -> `s3_prefix`

godobyte (changes requested):

godobyte (approved):

YutongLi291 (approved):

PR #422, opened 2024-07-26
test: add integ tests to test download func

AlexTranAmz:
Integ test did not cover all test cases that we wanted

Add more integ test to improve coverage.

We can now test if download with no output dir have the correct behavior
We can now test if long file path on Windows with UNC notation will not raise exception

- `hatch run test` 

- `hatch run integ:test`

Yes

No

----

### DIFF
--- a/test/integ/deadline_job_attachments/test_job_attachments.py
+++ b/test/integ/deadline_job_attachments/test_job_attachments.py
@@ -8,6 +8,7 @@ import time
 from dataclasses import dataclass
 from pathlib import Path
 from unittest.mock import MagicMock
+import sys
 
 from deadline.job_attachments.models import JobAttachmentS3Settings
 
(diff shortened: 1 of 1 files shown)


crowecawcaw (approved):

test/integ/deadline_job_attachments/test_job_attachments.py
```diff
@@ -1533,3 +1534,90 @@ def test_download_outputs_windows_max_file_path_length_exception(
         ),
     ):
         job_output_downloader.download_job_output()
+
+
+@pytest.mark.integ
+def test_download_outputs_no_outputs_dir(
+    job_attachment_test: JobAttachmentTest,
+    sync_outputs: SyncOutputsOutput,
+):
```
YutongLi291, line 1543: nit: We should add typing for this return

YutongLi291 (approved):

PR #423, opened 2024-07-27
test: Change name of Mainline Integration Tests

YutongLi291:
We should change the name of the mainline integration testsworkflow since they're ran on both merge and also as canary. 
Change the name of the integration test workflow
Better naming for the test actions
No testing required
No
No
----


leongdl (approved):
LGTM

crowecawcaw (approved):

Issue #407, opened 2024-07-16
Bug: `bundle submit` prompts for permission to upload files in a job bundle's directory


ddneilson:
### Expected Behaviour

Files whose actual true location is inside of the job bundle directory should always be allowed to upload without prompting:

```bash
$ deadline bundle submit --farm-id $FARM_ID --queue-id $QUEUE1_ID job_bundle/
Submitting to Queue: Q1
Hashing Attachments  [####################################]  100%
Hashing Summary:
    Processed 1 file totaling 39.0 B.
    Skipped re-processing 0 files totaling 0.0 B.
    Total processing time of 0.03119 seconds at 1.25 KB/s.
 ... and so on
```

Files in the job bundle directory that are symlinks to other files should, of course, be subject to regular permissions checks.

### Current Behaviour

The Deadline CLI's `bundle submit` command is asking if we're sure that we want to upload files that are located inside the job bundle's directory.

```bash
$ deadline bundle submit --farm-id $FARM_ID --queue-id $QUEUE1_ID job_bundle/
Submitting to Queue: Q1
Job submission contains 1 input files totaling 39.0 B.  All input files will be uploaded to S3 if they are not already present in the job attachments bucket.

Files were specified outside of the configured storage profile location(s).  Please confirm that you intend to submit a job that uses files from the following directories:

Under the directory '/home/cloudshell-user/job_bundle':
        1 input file

To permanently remove this warning you must only use files located within a storage profile location.

Do you wish to proceed? [y/N]: y
Hashing Attachments  [####################################]  100%
Hashing Summary:
    Processed 1 file totaling 39.0 B.
    Skipped re-processing 0 files totaling 0.0 B.
    Total processing time of 0.03119 seconds at 1.25 KB/s.
 ... and so on
```

### Reproduction Steps

See the code block in "current behavior"

### Code Snippet

N/A

PR #427, opened 2024-08-08
feat(cli)!: bundle submit - remove "--yes" and add "--trust" and "--upload"

epmog:

Starting this PR for discussion around the concept of trusting a job bundle and the UX we present around it.

Right now the deadline client lib is only concerned with job bundles uploading files a user might NOT want uploaded. This currently means that a user will be prompted if files are outside of storage profile locations returned by `GetStorageProfileForQueue`.

Fixing #407 would allow the client library to not prompt the user if the only files being uploaded are from within the job bundle. However this is all just treating the symptom of submitting untrusted bundles, rather than getting a user to confirm and understand what it means to submit a job bundle written by someone else.

Due to all of this, we've effectively blocked users from running `deadline bundle submit` as a scriptable/non-interactive command without piping "yes" into it. This especially becomes problematic for workflows that embed the usage of the deadline cli to perform the submission after they've generated a bundle. If they've just generated a job bundle in their workflow, they _should not_ be prompted about the job. They've specifically set out to create this job, we're just getting in the way.

If the workflow that generated the job bundle is malicious, then it can also just move whatever it wants to upload into the job bundle directory/storage profile locations and bypass these filepath checks. So if you trust the workflow to generate the bundle, you inherently trust the job bundle.

To start unpacking this, we really want the ability to signify that we trust a job bundle, as trusting a path to be uploaded is just 1 part of the greater problem of trusting bundles. Future changes will most likely involve configuring paths to not warn about.

So for `deadline bundle submit` we split out the existing confusing prompt into 2 smaller prompts :( so that users can answer them individually. The first one is about trusting the bundle, and the second one is about confirmation around uploading files. These two prompts have their own flags, `--trust` and `--upload` that allow users of the cli to answer both, some, or neither. 

This change intentionally does not add `--trust` to `deadline bundle gui-submit`, since that function needs quite a bit of refactoring to pass in all the options to make it have parity with `deadline bundle submit`.

I've removed the `--yes` command because `auto-accepting defaults` is useless for actually making it through these prompts. Maybe it should stay in for now, but I believe there's a bigger topic for discussion about indicating that you _cannot_ answer prompts, ie you want to use the cli non-interactively, and fail if you do not have an answer for everything.

Users can now script their usage of `deadline bundle submit` (minus concurrency) to submit jobs.

Yes

Yes

Self-documenting

You better believe it!

----

### DIFF
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -98,9 +98,14 @@ def validate_parameters(ctx, param, value):
     type=click.Choice([e.value for e in JobAttachmentsFileSystem]),
 )
 @click.option(
-    "--yes",
+    "--upload",
     is_flag=True,
-    help="Skip any confirmation prompts",
+    help="Indicates that files will be uploaded to S3 if needed",
+)
+@click.option(
+    "--trust",
+    is_flag=True,
+    help="Indicates the job bundle is trusted to not perform malicious activities",
 )
 @click.option(
     "--require-paths-exist",
@@ -118,7 +123,8 @@ def bundle_submit(
     job_bundle_dir,
     job_attachments_file_system,
     parameter,
-    yes,
+    upload,
+    trust,
     name,
     priority,
     max_failed_tasks_count,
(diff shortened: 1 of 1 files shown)


src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -98,9 +98,14 @@ def validate_parameters(ctx, param, value):
     type=click.Choice([e.value for e in JobAttachmentsFileSystem]),
 )
 @click.option(
-    "--yes",
+    "--upload",
     is_flag=True,
-    help="Skip any confirmation prompts",
+    help="Indicates that files will be uploaded to S3 if needed",
+)
+@click.option(
```
leongdl, line 105: @mwiebe explicitly said we shouldn't just have a --trust option but to add an allow list of paths.

Thoughts ?

mwiebe: Keeping the `--yes` option to mean accept default prompts seems like a good idea to me. What's our rationale for removing it?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -98,9 +98,14 @@ def validate_parameters(ctx, param, value):
...
+    "--trust",
+    is_flag=True,
+    help="Indicates the job bundle is trusted to not perform malicious activities",
```
mwiebe, line 108: I think we can do better than an option like this. The question is about where to upload data from for jobs, not about which job bundles are ok or not.

mwiebe:
What I recommend as our first step to making this experience right is to fix the bug https://github.com/aws-deadline/deadline-cloud/issues/407 by adding a list of permitted upload globs where it's relevant, and adding the job bundle directory and all paths provided from the CLI options to start. That would be a backwards-compatible change that needs no interface modification.

src/deadline/client/cli/_groups/bundle_group.py
epmog: It doesn't behave that way I or customers expect. Removal may be part of a bigger change than in this one, but this change does propose changed behaviour by switching to `--upload` (something like that).

I do think we should have an option shared amongst all (or most, can't think of exception atm) options that indicate you're operating in an session that is not interactive and thus, anything that would prompt or is not answered would cause a failure of the command. `--yes` is currently the way to get that, but that's not great

PR #431, opened 2024-08-16
fix: credential caching improvements

crowecawcaw:
The CLI and especially GUI submitter were very slow to make API calls. The root cause was poorly matched TTLs that resulted in credentials being refreshed from DCM before every API call.

DCM vends credentials which are valid for 15 minutes, and by default boto's credential handling refreshes credential when they're within 15 minutes of expiring. As a result, credential were always marked as stale and constantly refreshed. Refreshing DCM creds takes ~5 seconds, so every API call takes at least that long to complete. 

Configure the refresh mechanism to only refresh creds when they're within 2 minutes of expiring. Also strip out a custom caching mechanism we had and use the much simpler `@lru_cache` decorator instead.

DCM credentials and boto sessions are effectively reused resulting in a much more responsive GUI submitter (among other DCM functionality).

Yes

Yes

Manual tests:
- I opened the GUI submitter, waited 15 minutes for the DCM creds to expire, then opened the settings window to trigger more API calls. The client refreshed credentials as expected.
- I verified these changes work with non-DCM credentials.
- 
Not needed

No
----

### DIFF
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -75,11 +75,19 @@ def _get_boto3_session_for_profile(profile_name: str):
     # Also DCM credentials currently take several seconds to refresh. Lower the refresh timeouts so creds
     # are reused between API calls to save time.
     # See https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L342-L362
-    credentials = session.get_credentials()
-    if isinstance(credentials, RefreshableCredentials):
-        credentials._advisory_refresh_timeout = 5 * 60  # 5 minutes
-        credentials._mandatory_refresh_timeout = 2.5 * 60  # 2.5 minutes
 
+    try:
+        credentials = session.get_credentials()
+        if (
+            isinstance(credentials, RefreshableCredentials)
+            and credentials.method == "custom-process"
+        ):
+            credentials._advisory_refresh_timeout = 5 * 60  # 5 minutes
+            credentials._mandatory_refresh_timeout = 2.5 * 60  # 2.5 minutes
+    except:  # noqa: E722
+        # Attempt to patch the timeouts but ignore any errors. These patched proeprties are internal and could change
+        # without notice. Creds are functional without patching timeouts.
+        pass
     return session
 
 
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -72,13 +72,13 @@ def _get_boto3_session_for_profile(profile_name: str):
 
     # By default, DCM returns creds that expire after 15 minutes, and boto3's RefreshableCredentials
     # class refreshes creds that are within 15 minutes of expiring, so credentials would never be reused.
-    # Also DCM credentials currently take several seconds to refresh. Set the refresh timeouts to 1 and
-    # 2 minutes so they are reused between API calls to save time.
+    # Also DCM credentials currently take several seconds to refresh. Lower the refresh timeouts so creds
+    # are reused between API calls to save time.
     # See https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L342-L362
     credentials = session.get_credentials()
     if isinstance(credentials, RefreshableCredentials):
-        credentials._advisory_refresh_timeout = 2 * 60  # 2 minutes
-        credentials._mandatory_refresh_timeout = 60  # 1 minute
+        credentials._advisory_refresh_timeout = 5 * 60  # 5 minutes
+        credentials._mandatory_refresh_timeout = 2.5 * 60  # 2.5 minutes
 
     return session
 
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -5,6 +5,7 @@ Provides functionality for boto3 Sessions, Clients, and properties
 of the Deadline-configured IAM credentials.
 """
 from __future__ import annotations
+from functools import lru_cache
 import logging
 from configparser import ConfigParser
 from contextlib import contextmanager
@@ -22,12 +23,6 @@ from botocore.session import get_session as get_botocore_session
 
 from ..config import get_setting
 from ..exceptions import DeadlineOperationError
-
-__cached_boto3_session = None
-__cached_boto3_session_profile_name = None
-__cached_boto3_queue_session = None
-__cached_farm_id_for_queue_session = None
-__cached_queue_id_for_queue_session = None
 
 
 class AwsCredentialsSource(Enum):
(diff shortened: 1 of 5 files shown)


src/deadline/client/api/_session.py
```diff
@@ -67,36 +60,32 @@ def get_boto3_session(
     if profile_name in ("(default)", "default", ""):
         profile_name = None
 
-    # If a config was provided, don't use the Session caching mechanism.
-    if config:
-        return boto3.Session(profile_name=profile_name)
-
```
crowecawcaw, line 73: Unclear why this is here. It appears to be a mistake because `config` not part of the created session. The comment doesn't give more context. It was in this repo's initial commit so there's also no info in the git history. The new `lru_cache` is keyed off of profile name instead which is actually used in the session creation and also conveniently hashable.

src/deadline/client/api/_session.py
```diff
@@ -67,36 +60,32 @@ def get_boto3_session(
...
-    global __cached_queue_id_for_queue_session
+    # By default, DCM returns creds that expire after 15 minutes, and boto3's RefreshableCredentials
+    # class refreshes creds that are within 15 minutes of expiring, so credentials would never be reused.
+    # Also DCM credentials currently take several seconds to refresh. Set the refresh timeouts to 1 and
+    # 2 minutes so they are reused between API calls to save time.
+    # See https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L342-L362
+    credentials = session.get_credentials()
+    if isinstance(credentials, RefreshableCredentials):
+        credentials._advisory_refresh_timeout = 2 * 60  # 2 minutes
+        credentials._mandatory_refresh_timeout = 60  # 1 minute
```
ddneilson: This seems really aggressive to me. It doesn't leave much room for service issues before credentials expire; potentially during a long set of S3 uploads for job attachments. 

What would you think about advisory of 5 minutes and mandatory of 2.5? Still pretty aggressive, but leaves a little more leeway

src/deadline/client/api/_session.py
crowecawcaw: Huge thanks to @epmog for tracking this down in the pre-release repo. This code is a holdover from when we passed `config` to the client constructor, so it'd make sense to not reuse a session from a different config. Since the original commit, we switched to only passing the profile from `config` to the client constructor, so now we can reuse/cache a session if the profile name hasn't changed.

We can also see where the profile name is extracted from `config` here before it's passed to the client constructor: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/api/_session.py#L63

src/deadline/client/api/_session.py
crowecawcaw: Sure, seems good to me. The creds' expiration is rechecked before every API call so I think anything over a few seconds is probably sufficient. I suspect the timeout/margin is to compensate for things like processing time between the creds check and API call or network delays. But leaving extra margin seems safe and won't noticably diminish responsiveness.

src/deadline/client/api/_session.py
```diff
@@ -67,36 +60,32 @@ def get_boto3_session(
...
-    global __cached_farm_id_for_queue_session
-    global __cached_queue_id_for_queue_session
+    # By default, DCM returns creds that expire after 15 minutes, and boto3's RefreshableCredentials
+    # class refreshes creds that are within 15 minutes of expiring, so credentials would never be reused.
+    # Also DCM credentials currently take several seconds to refresh. Lower the refresh timeouts so creds
+    # are reused between API calls to save time.
+    # See https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L342-L362
+    credentials = session.get_credentials()
+    if isinstance(credentials, RefreshableCredentials):
+        credentials._advisory_refresh_timeout = 5 * 60  # 5 minutes
```
prestomation: the private prefix(_) suggests that there is a risk of boto breaking us. should we swallow exceptions here so customers don't break on a theoretical boto upgrade in the future?

src/deadline/client/api/_session.py
```diff
@@ -67,36 +60,32 @@ def get_boto3_session(
...
-    global __cached_boto3_queue_session
-    global __cached_farm_id_for_queue_session
-    global __cached_queue_id_for_queue_session
+    # By default, DCM returns creds that expire after 15 minutes, and boto3's RefreshableCredentials
+    # class refreshes creds that are within 15 minutes of expiring, so credentials would never be reused.
+    # Also DCM credentials currently take several seconds to refresh. Lower the refresh timeouts so creds
+    # are reused between API calls to save time.
+    # See https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L342-L362
+    credentials = session.get_credentials()
+    if isinstance(credentials, RefreshableCredentials):
```
prestomation: It is surprising to me that boto considers 10-15 minutes 'stale' and I wonder what the reason is. is there a risk to customers that are using other credential provider mechanisms? Should we further narrow down these credentials to just DCM-powered credentials?

crowecawcaw: Looks like it was changed from 5 minutes to 15 minutes in this PR but there's not a justification for it that I can see: https://github.com/boto/botocore/pull/286

PRs since then have improved the refreshing mechanism but carried over the 15 minute timeout without commenting on it.

crowecawcaw: I scoped this branch down to credentials whose method is "custom-process". It could include other types of process-provided creds, but I think the risk of them breaking is very small and the number of customers using other types of process-provided creds with the CLI is probably few. I looked for ways of detecting just DCM creds, but they'd require more complexity and refactoring which seemed like a bad tradeoff.

src/deadline/client/api/_session.py
```diff
@@ -67,36 +60,40 @@ def get_boto3_session(
...
+            credentials._mandatory_refresh_timeout = 2.5 * 60  # 2.5 minutes
+    except:  # noqa: E722
+        # Attempt to patch the timeouts but ignore any errors. These patched proeprties are internal and could change
+        # without notice. Creds are functional without patching timeouts.
+        pass
+    return session
+
+
+def invalidate_boto3_session_cache() -> None:
+    _get_boto3_session_for_profile.cache_clear()
```
ddneilson, line 95: Why do we clear the cache twice?

src/deadline/client/api/_session.py
```diff
@@ -67,36 +60,40 @@ def get_boto3_session(
...
+    _get_boto3_session_for_profile.cache_clear()
```
ddneilson: Is this second one supposed to be:
```
_get_queue_user_boto3_session.cache_clear()
```
?

crowecawcaw: Oof yes.

ddneilson (approved):

prestomation (approved):

PR #434, opened 2024-08-26
test: Squish `deadline config gui` settings dialogue automated tests

rena-cheng:
In the latest PR revision (08/30/2024), I've incorporated the following changes (which are now all in one commit):
- renamed/updated what was probably over 100 Squish locators into more friendly/readable names; preserved the `.names.py` file so we can still keep the old names that Squish automatically generates. Note that as more tests are recorded, the new locators will get added into the file. I also decided not to add this file into .gitignore etc since as new locators get added into the repo, they should get updated in the `.names.py` file, and they actually will not be different for those working in the suite (as mentioned in a comment below).  
- create a separate config file for the test resources (ie: profile name, farm name, etc). 
- re-organized some locators/functions into `loginout` pages as the authentication-related locators/statuses made more sense there (than in gui_locator/helpers file which is where it was previously). 
- fixed a couple typos, formatting issues, etc. 
_______________________________________________________________________________________________________

The same details from this old PR can still be referenced in this current PR: https://github.com/aws-deadline/deadline-cloud/pull/399

One notable change I incorporated was adding setup/tear down functions (in which I had login and logout functions in mind). After some discussion, we decided to use `default aws profile` (versus a DCM profile) for authentication, so that logging in is not necessary. Additionally, I added a function to clear refresh error dialogues (as part of setup) as those can occasionally appear if a DCM profile was pre-selected in the dialogue prior to starting the tests (as a developer or anyone working locally can have a different profile setup prior to running the tests.)

Other changes include refactoring the test suite as I'd previously added verifications in helper functions themselves (which can be fine if the helper functions need verifications to ensure they are working) and so I added the main test verifications into `test.py` instead. I also refactored the variable names from `camelCase` to `snake_case` (to remain consistent with the naming conventions with the rest of the repo and given that we're using Python anyway). 

All in all, there's a total of 33 tests added for the `deadline config gui` settings dialogue. 

Note that if one is working locally and developing the tests and running the test with the default aws profile, you'll need to authenticate first with credentials (or another option is to add them to the environment variable config). You'll need to specify region as well (and add it to your env var config) if you are running the tests on a different region (in my case, I was automating using a farm on `us-east-1` region. 

There are two ways to run Squish tests and where you can input these credential/env vars:
1. Using the terminal. Be sure to set any authentication credentials in the terminal, set the region using `export AWS_DEFAULT_REGION=` env var and then you can run the squish tests (which, you'll also need to be in the correect folder for) (example command: `./squishrunner --testsuite <../../../<squish_folder>/suite_VerifySettingsDialogue --local`
2. Using the Squish IDE. You'll need to set any authentication credentials which are env vars in the config file, in addition to the `export AWS_DEFAULT_PROFILE=` env var. Once set, the default aws profile should authenticate when running the Squish tests directly from the IDE. 

Additionally, as mentioned in the previous PR, you'll need to point the `LD_LIBRARY_PATH` env var to where the `PySide6/Qt` libraries live on your system in the config. 

For reference, below are some screenshots from the Logs when the tests have finished running. In this case, all the tests successfully passed: 
Screenshot 2024-08-25 at 10 27 37 PM
Screenshot 2024-08-25 at 10 28 07 PM

--See old PR.

Yes, 83% of the tests are passing (where is says at least 80% of tests is minimum for passing criteria). I'm not sure why this is but I'm looking into it so 100% of tests can pass. 

Yes, and all the tests passed. 
Screenshot 2024-08-27 at 1 39 48 AM

Additionally,
I've run a couple failure-case tests for the Squish suite itself, to ensure that things error out or fail as expected when some expected errors/failures occur:
- When selecting a DCM profile - (ensured that things error out as expected, ensured that the refresh error dialogues where checked and/or found and then closed out correctly)
Screenshot 2024-08-25 at 11 18 01 PM

- When selecting resources that are missing or don't exist - (ensured that things errored out as expected)
Screenshot 2024-08-25 at 11 08 19 PM

The only documentation that has been asked of me is a doc describing how we are planning to use this test suite (and I will have that alongside writing additional tests). 

--See old PR.

----

### DIFF
--- a/test/squish/suite_VerifySettingsDialogue/shared/scripts/gui_helpers.py
+++ b/test/squish/suite_VerifySettingsDialogue/shared/scripts/gui_helpers.py
@@ -1,0 +1,151 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# mypy: disable-error-code="attr-defined"
+
+import gui_locators
+import squish
+
+import test
+
+snooze_timeout = 1  # seconds
+
+
+def launch_deadline_config_gui():
+    squish.startApplication("deadline config gui")
+    test.log("Launched Deadline Config GUI.")
+    test.log(
+        "Sleep for " + str(snooze_timeout) + " second(s) to allow GUI authentication to fully load."
+    )
+    squish.snooze(snooze_timeout)
+    test.compare(
+        squish.waitForObjectExists(gui_locators.deadline_config_dialog).visible,
+        True,
+        "Expect the Deadline Config GUI to be open.",
+    )
+
+
+def close_deadline_config_gui():
+    test.log("Hitting `OK` button to close Deadline Config GUI.")
+    # hit 'OK' button to close Deadline Config GUI
+    squish.clickButton(squish.waitForObject(gui_locators.deadlinedialog_ok_button))
+
+
+def hit_apply_button():
+    test.log("Hitting `Apply` button to apply selected settings.")
+    # hit 'Apply' button
+    squish.clickButton(squish.waitForObject(gui_locators.deadlinedialog_apply_button))
+    test.log("Settings have been applied.")
+
+
+def set_farm_name(farm_name: str):
+    # open Default farm drop down menu
+    squish.mouseClick(squish.waitForObject(gui_locators.profilesettings_defaultfarm_dropdown))
+    test.log("Opened farm name drop down menu.")
+    test.compare(
+        squish.waitForObjectExists(gui_locators.farm_name_locator(farm_name)).text,
+        farm_name,
+        "Expect farm name to be present in drop down.",
+    )
+    # select Default farm
+    squish.mouseClick(
+        squish.waitForObjectItem(gui_locators.profilesettings_defaultfarm_dropdown, farm_name)
+    )
+    test.log("Selected farm name.")
+
+
+def open_close_job_hist_directory():
+    # hit '...' button to open Choose Job history directory file browser
+    squish.clickButton(squish.waitForObject(gui_locators.open_job_hist_dir_button))
+    test.log("Opened job history directory dialogue.")
+    # verify job history directory dialogue is open
+    test.compare(
+        str(squish.waitForObjectExists(gui_locators.choosejobhistdir_filebrowser).windowTitle),
+        "Choose Job history directory",
+        "Expect Choose Job history directory dialogue window title to be present.",
+    )
+    test.compare(
+        squish.waitForObjectExists(gui_locators.choosejobhistdir_filebrowser).visible,
+        True,
+        "Expect Choose Job history directory dialogue to be open.",
+    )
+    # hit 'choose' button to set default and close file browser
+    squish.clickButton(squish.waitForObject(gui_locators.choosejobhistdir_choose_button))
+    test.log("Closed job history directory dialogue.")
+
+
+def set_queue_name(queue_name: str):
+    # open Default queue drop down menu
+    squish.mouseClick(squish.waitForObject(gui_locators.farmsettings_defaultqueue_dropdown))
+    test.log("Opened queue name drop down menu.")
+    test.compare(
+        squish.waitForObjectExists(gui_locators.queue_name_locator(queue_name)).text,
+        queue_name,
+        "Expect queue name to be present in drop down.",
+    )
+    # select Default queue
+    squish.mouseClick(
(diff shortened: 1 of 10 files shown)


test/squish/suite_VerifySettingsDialogue/shared/scripts/names.py
hindleym, line 1: We might consider adding this file to the `.gitignore`. The locators this makes might be different for everyone writing tests.

test/squish/suite_VerifySettingsDialogue/shared/scripts/gui_locators.py
hindleym, line 1: I have a feeling that maintaining and using these locators will be a bit difficult.

My thoughts here are that the automatically created locators live in `names.py`. Then similar to how you separated the locators for `loginout_locators.py`, we can pull the used locators from `names.py` and convert them into more usable variables into separate files (shared_locators, settingspage_locators, etc).

hindleym (commented):
It will probably be better to take care of the locators sooner, rather than later. Set the standard before getting too far into writing more tests, especially if others start adding their own tests to this package.

test/squish/suite_VerifySettingsDialogue/shared/scripts/gui_locators.py
rena-cheng: Discussed on slack with Church a few notes:

- I like the idea of adding `names.py` into `.gitignore` as that file will certainly be different for everyone as you record and write more tests. We should save the original locator names there for sure. (As I didn't think to save mine, I deleted them all, and now I'll probably need to add it back now?) 
- "...with the variables at the top of test.py. Might be worth having some kind of `default_settings_config.py`" - I think this is a good idea especially if those names get re-used across tests. 
- We also discussed writing more descriptive/readable names for the locators, as the one that Squish generates for us is not very readable/understandable. So I will address this as well.

test/squish/suite_VerifySettingsDialogue/shared/scripts/gui_helpers.py
```diff
@@ -0,0 +1,217 @@
...
+    test.log("Hitting `OK` button to close Deadline Config GUI.")
+    # hit 'OK' button to close Deadline Config GUI
+    squish.clickButton(
+        squish.waitForObject(
+            gui_locators.aWS_Deadline_Cloud_workstation_configuration_OK_QPushButton
+        )
+    )
+
+
+# apply button bug: https://sim.amazon.com/issues/Bea-28289
```
amarsjac: need to remove link

test/squish/suite_VerifySettingsDialogue/shared/scripts/loginout_helpers.py
```diff
@@ -0,0 +1,42 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# mypy: disable-error-code="attr-defined"
+
+import loginout_locators
+import squish
+import test
+
+# workaround for https://sim.amazon.com/issues/Bea-30434
```
amarsjac: diddo

test/squish/suite_VerifySettingsDialogue/tst_VerifySettingsDialogue/test.py
```diff
@@ -0,0 +1,153 @@
...
+    test.compare(
+        str(
+            squish.waitForObjectExists(
+                gui_locators.general_settings_Current_logging_level_QComboBox
+            ).currentText
+        ),
+        logging_level,
+        "Expect selected current logging level to be set.",
+    )
+    # disabling due to https://sim.amazon.com/issues/Bea-28289
```
amarsjac: remove link

rena-cheng: Thanks. In previous automated tests, we added the bug link as it was helpful for QAEs (and developers) to reference the issue number, and keep track of things as they get fixed so tests can then be updated (an example is here: https://code.amazon.com/packages/MadStudioE2ETests/blobs/b42d2505354c9523b3b9c420dbbaeabe8d818608/--/src/stop-reconnect-spec.ts#L78). I assume leaving the ticket number is still OK?

rena-cheng: Update - I spoke with Marshall. As this is an open source repo, we wouldn't want to leave any Amazon confidential information in the code. So will go ahead and remove the bug ticket numbers.

amarsjac (approved):

erico-aws (approved):

PR #436, opened 2024-08-27
docs: improve doc strings text

### LINKED ISSUE #324: Docs: `deadline config show` doesn't describe all settings
### Documentation Issue

`deadline config show` is currently missing documentation about the following settings:
* `settings.conflict_resolution`
* `settings.s3_max_pool_connections`
* `settings.small_file_threshold_multiplier`

This should be added [here](https://github.com/aws-deadline/deadline-cloud/blob/72d6c26b768c09c1caa4301bebc4db61bdbb7861/src/deadline/client/cli/_groups/config_group.py#L15)

### LINKED ISSUE #365: Feature request: Improve help text for `deadline bundle submit --help`
### Use Case

Imagine a customer is trying to figure out how to provide values for their Job Parameter(s) when using the command `deadline bundle submit`. They type `deadline bundle submit --help` and see:

```
Options:
  -p, --parameter TEXT            Job template parameters
```

What is the format of "TEXT"??  (turns out that it's `-p VarName=VarValue`, but there's no way to know that)

### Proposed Solution

Improve the help text, and the README for the package. For inspiration, please see the `run` command in https://github.com/OpenJobDescription/openjd-cli. e.g.

```
% openjd run --help
...
  --job-param KEY=VALUE, file://PATH_TO_PARAMS, '{"KEY": "VALUE", ... }', -p KEY=VALUE, file://PATH_TO_PARAMS, '{"KEY": "VALUE", ... }'
                        Use these Job parameters with the provided template. Can be provided as key-value pairs, path(s) to a JSON or YAML document prefixed with
                        'file://', or inline JSON. If this option is provided more than once then the given values are all combined in the order that they appear.
```

crowecawcaw:
The client includes help text for its commands, and some of the help text could be improved.

Update help text as needed.

Fixes https://github.com/aws-deadline/deadline-cloud/issues/324
Fixes https://github.com/aws-deadline/deadline-cloud/issues/365

Someone can more easily understand how to use the client.

Yes

No

Yes

No
----

### DIFF
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -73,13 +73,22 @@ def validate_parameters(ctx, param, value):
 
 @cli_bundle.command(name="submit")
 @click.option(
-    "-p", "--parameter", multiple=True, callback=validate_parameters, help="Job template parameters"
+    "-p",
+    "--parameter",
+    multiple=True,
+    callback=validate_parameters,
+    help='The values for the job template\'s parameters. Can be provided as key-value pairs, inline JSON strings, or as paths to a JSON or YAML document. If provided more than once, the values are combined in the order that they appear. Examples: --parameter MyParam=5 -p file://parameter_file.json -p \'{"MyParam": "5"}\'',
 )
 @click.option("--profile", help="The AWS profile to use.")
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
 @click.option("--name", help="The job name to use in place of the one in the job bundle.")
-@click.option("--priority", type=int, default=50, help="The priority of the job.")
+@click.option(
+    "--priority",
+    type=int,
+    default=50,
+    help="The priority of the job. The highest priority is 100.",
+)
 @click.option(
     "--max-failed-tasks-count",
     type=int,
@@ -92,9 +101,9 @@ def validate_parameters(ctx, param, value):
 )
 @click.option(
     "--job-attachments-file-system",
-    help="The method for accessing files from job attachments. "
-    + "COPIED means to copy files to the host and "
-    + "VIRTUAL means to load files when needed with a virtual file system.",
+    help="The method workers use to access job attachments. "
+    + "COPIED means to copy files to the worker and "
+    + "VIRTUAL means to load files as needed from a virtual file system.",
     type=click.Choice([e.value for e in JobAttachmentsFileSystem]),
 )
 @click.option(
@@ -105,7 +114,7 @@ def validate_parameters(ctx, param, value):
 @click.option(
     "--require-paths-exist",
     is_flag=True,
-    help="Require all input paths to exist",
+    help="Return an error if any input files are missing.",
 )
 @click.option(
     "--submitter-name",
@@ -256,7 +265,7 @@ def bundle_submit(
 @click.option(
     "--browse",
     is_flag=True,
-    help="Allows user to choose Bundle and adds a 'Load a different job bundle' option to the Job-Specific Settings UI",
+    help="Opens a file browser to select a bundle.",
 )
 @click.option(
     "--install-gui",
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -80,14 +80,14 @@ def validate_parameters(ctx, param, value):
     help='The values for the job template\'s parameters. Can be provided as key-value pairs, inline JSON strings, or as paths to a JSON or YAML document. If provided more than once, the values are combined in the order that they appear. Examples: --parameter MyParam=5 -p file://parameter_file.json -p \'{"MyParam": "5"}\'',
 )
 @click.option("--profile", help="The AWS profile to use.")
-@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
-@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
+@click.option("--farm-id", help="The farm to use.")
+@click.option("--queue-id", help="The queue to use.")
 @click.option("--name", help="The job name to use in place of the one in the job bundle.")
 @click.option(
     "--priority",
     type=int,
     default=50,
-    help="The priority of the job. The highest priority is 100.",
+    help="The priority of the job. Jobs with a higher priority run first.",
 )
 @click.option(
     "--max-failed-tasks-count",
@@ -102,8 +102,9 @@ def validate_parameters(ctx, param, value):
 @click.option(
     "--job-attachments-file-system",
     help="The method workers use to access job attachments. "
-    + "COPIED means to copy files to the worker and "
-    + "VIRTUAL means to load files as needed from a virtual file system.",
+    "COPIED means to copy files to the worker and VIRTUAL means to load "
+    "files as needed from a virtual file system. If VIRTUAL is selected "
+    "but not supported by a worker, it will fallback to COPIED.",
     type=click.Choice([e.value for e in JobAttachmentsFileSystem]),
 )
 @click.option(
@@ -137,7 +138,7 @@ def bundle_submit(
     **args,
 ):
     """
-    Submits an Open Job Description job bundle to AWS Deadline Cloud.
+    Submits an Open Job Description job bundle.
     """
     # Get a temporary config object with the standard options handled
     config = _apply_cli_options_to_config(required_options={"farm_id", "queue_id"}, **args)
@@ -265,7 +266,7 @@ def bundle_submit(
 @click.option(
     "--browse",
     is_flag=True,
-    help="Opens a file browser to select a bundle.",
+    help="Opens a folder browser to select a bundle.",
 )
 @click.option(
     "--install-gui",
@@ -292,7 +293,7 @@ def bundle_submit(
 @_handle_error
 def bundle_gui_submit(job_bundle_dir, browse, output, install_gui, submitter_name, **args):
     """
-    Opens GUI to submit an Open Job Description job bundle to AWS Deadline Cloud.
+    Opens a GUI to submit an Open Job Description job bundle.
     """
     from ...ui import gui_context_for_cli
 
--- a/src/deadline/client/cli/_groups/config_group.py
+++ b/src/deadline/client/cli/_groups/config_group.py
@@ -14,52 +14,8 @@ from .._common import _handle_error
 @_handle_error
 def cli_config():
     """
-    Manage AWS Deadline Cloud's workstation configuration.
-
-    The available AWS Deadline Cloud settings are:
-
-    defaults.aws_profile_name:
-
-        The default AWS profile to use for AWS Deadline Cloud commands. Set to '' to use the default credentials. Other settings are saved with the profile.
-
-    defaults.farm_id:
-
-        The default farm ID to use for job submissions or CLI operations.
-
-    defaults.queue_id:
-
-        The default queue ID to use for job submissions or CLI operations.
-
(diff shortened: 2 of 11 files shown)


src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -73,13 +73,22 @@ def validate_parameters(ctx, param, value):
...
 @click.option("--profile", help="The AWS profile to use.")
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
 @click.option("--name", help="The job name to use in place of the one in the job bundle.")
-@click.option("--priority", type=int, default=50, help="The priority of the job.")
+@click.option(
+    "--priority",
+    type=int,
+    default=50,
+    help="The priority of the job. The highest priority is 100.",
```
ddneilson: What do you think about clarifying whether jobs with a higher or lower number are scheduled first?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -92,9 +101,9 @@ def validate_parameters(ctx, param, value):
 )
 @click.option(
     "--job-attachments-file-system",
-    help="The method for accessing files from job attachments. "
-    + "COPIED means to copy files to the host and "
-    + "VIRTUAL means to load files when needed with a virtual file system.",
+    help="The method workers use to access job attachments. "
+    + "COPIED means to copy files to the worker and "
+    + "VIRTUAL means to load files as needed from a virtual file system.",
```
ddneilson: We should clarify that VIRTUAL is only available on Linux SMF workers; workers where VIRTUAL is not available will fall-back to COPIED.

ddneilson (changes requested):

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -256,7 +265,7 @@ def _decide_cancel_submission(upload_group: AssetUploadGroup) -> bool:
 @click.option(
     "--browse",
     is_flag=True,
-    help="Allows user to choose Bundle and adds a 'Load a different job bundle' option to the Job-Specific Settings UI",
+    help="Opens a file browser to select a bundle.",
```
epmog:
```suggestion
    help="Opens a folder browser to select a bundle.",
```

src/deadline/client/cli/_groups/config_group.py
```diff
@@ -14,52 +14,8 @@
...
-    defaults.job_attachments_file_system:
-
-        Setting to determine if attachments are copied on to workers before a job executes
-        or fetched in realtime. Must be one of ["COPIED", "VIRTUAL"]
-
-    telemetry.opt_out:
-
-        Flag to specify opting out of telemetry data collection.
+    Manage AWS Deadline Cloud's workstation configuration. Config options are organized
+    and documented in config_file.SETTINGS.
```
epmog: What steps does a user take to then figure this information from the help text? Do they have to go code spelunking? I wonder if we can avoid that. Example:

```
(deadline) bash-3.2$ deadline config
Usage: deadline config [OPTIONS] COMMAND [ARGS]...

  Manage AWS Deadline Cloud's workstation configuration. Config options are
  organized and documented in config_file.SETTINGS.

Options:
  -h, --help  Show this message and exit.

Commands:
  get   Gets an AWS Deadline Cloud workstation configuration setting.
  gui   Open the workstation configuration settings GUI.
  set   Sets an AWS Deadline Cloud workstation configuration setting.
  show  Show AWS Deadline Cloud's current workstation configuration.
```

If we do change this, the get and set command will also have to change:

```
(deadline) bash-3.2$ deadline config set --help
Usage: deadline config set [OPTIONS] SETTING_NAME VALUE

  Sets an AWS Deadline Cloud workstation configuration setting.

  For example `deadline config set defaults.farm_id <farm-id>`. Run `deadline
  config --help` to show available settings.

Options:
  -h, --help  Show this message and exit.
```

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -151,7 +151,7 @@ def job_get(**args):
     "--mark-as",
     type=click.Choice(["CANCELED", "FAILED", "SUCCEEDED"], case_sensitive=False),
     default="CANCELED",
-    help="The run status to mark the job as.",
+    help="The task run status to mark the job as.",
```
epmog: I think we can still do better here... Is "task run status" exposed much to the users? I guess jobs that were not completed, canceled, or failed, will be set to the value they put in here. Maybe something like:

"The task status that all active tasks will be set to."

I wanted to avoid saying an actual status otherwise we would have list out all relevant status' which I also wanted to avoid. Open to ideas here though.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -608,11 +608,10 @@ def _assert_valid_path(path: str) -> None:
         ],
         case_sensitive=False,
     ),
-    help="The resolution method to use when a file already exists."
-    "Please choose one from the following options. If it is not provided, it defaults to CREATE_COPY:\n"
-    "[1] SKIP: Do not download these files\n"
-    "[2] OVERWRITE: Download these files and overwrite existing files\n"
-    "[3] CREATE_COPY: Download the file with a new name, appending '(1)' to the end",
+    help="The resolution method to use when a file already exists:\n"
```
epmog: starting with resolution feels odd to me, but I guess that's what the option is named. "How to handle existing files" might be a shorter way to put it. I think it's probably fine though. Just musing out loud.

src/deadline/client/config/config_file.py
```diff
@@ -100,7 +100,7 @@
     },
     "settings.conflict_resolution": {
         "default": FileConflictResolution.NOT_SELECTED.name,
-        "description": "How to handle duplicate files when downloading (if a file with the same path/name already exists.)",
+        "description": "How to handle duplicate files while downloading if a file with the same name already exists",
```
epmog: "How to handle downloading files that already exist" might be cleaner?

epmog (changes requested):

epmog:
Change looks good, just a few suggestions. 2 additional things that don't get captured in the diff:

1. Would help if we included the updated help text in the PR as a user would experience it. Though I checked it out locally to view the differences
2. Some additional changes we may want would be to add short help's to some of the commands to prevent some truncation. Ref: https://click.palletsprojects.com/en/8.1.x/documentation/#command-short-help

for example:

deadline isn't truncated but there's also a lot of self-referential `AWS Deadline Cloud`. Maybe we can shrink those down, since the user is already using the deadline cloud client cli.
```
$ deadline
Usage: deadline [OPTIONS] COMMAND [ARGS]...

  The AWS Deadline Cloud CLI provides functionality to interact with the AWS
  Deadline Cloud service.

Options:
  --version                       Show the version and exit.
  --log-level [ERROR|WARNING|INFO|DEBUG]
                                  Set the logging level.

Commands:
  auth            Commands to handle AWS Deadline Cloud authentication.
  bundle          Commands to work with Open Job Description job bundles.
  config          Manage AWS Deadline Cloud's workstation configuration.
  farm            Commands to work with AWS Deadline Cloud Farm resources.
  fleet           Commands to work with AWS Deadline Cloud Fleet resources.
  handle-web-url  Runs AWS Deadline Cloud commands sent from a web browser.
  job             Commands to work with AWS Deadline Cloud Jobs.
  queue           Commands for AWS Deadline Cloud Queues.
  worker          Commands to work with AWS Deadline Cloud Workers.
```
but `deadline bundle` has cut-off help text:

```
$ deadline bundle
Usage: deadline bundle [OPTIONS] COMMAND [ARGS]...

  Commands to work with Open Job Description job bundles.

Options:
  -h, --help  Show this message and exit.

Commands:
  gui-submit  Opens GUI to submit an Open Job Description job bundle to...
  submit      Submits an Open Job Description job bundle to AWS Deadline...
```
same with other ones like `deadline job`

```
$ deadline job
Usage: deadline job [OPTIONS] COMMAND [ARGS]...

  Commands to work with AWS Deadline Cloud Jobs.

Options:
  -h, --help  Show this message and exit.

Commands:
  cancel           Cancel an AWS Deadline Cloud Job from running.
  download-output  Download the output attached to an AWS Deadline Cloud...
  get              Get the details of an AWS Deadline Cloud Job.
  list             Lists the Jobs in an AWS Deadline Cloud Queue.
  trace-schedule   EXPERIMENTAL - Print statistics about how a job, and...
```

Might make sense to include that here in this change?

src/deadline/client/cli/_groups/bundle_group.py
epmog: actually kind of surprised we're not enforcing validation at the command level. Rather than when it gets to the service... Food for thought, may make sense to add it here.

src/deadline/client/cli/_groups/bundle_group.py
crowecawcaw: Nice, I'll add a note about the fallback. I don't think we should specifically call out Linux vs Windows support since that's likely to change in the future.

src/deadline/client/cli/_groups/bundle_group.py
crowecawcaw: I copied this from our API docs for indicating which number represents higher priority. But agree it could be clearer, will fix.

If someone gives a priority that's out of range, they'll quickly get back an error from the boto3 client even before the API is made, so I don't think we _need_ to mirror those validations here. I also think the help text gets a little noisy with the extra validations added. Example excerpt:
```
  --priority INTEGER RANGE        The priority of the job. Jobs with a higher
                                  priority run first.  [0<=x<=100]
```

src/deadline/client/cli/_groups/job_group.py
crowecawcaw: How about, "The status to apply to all active tasks in the job."

src/deadline/client/cli/_groups/config_group.py
crowecawcaw: Wrapping the documentation into `deadline config show`.

crowecawcaw: Excerpt of sample output:
```
 > deadline config show

AWS Deadline Cloud configuration file:
   /Users/crowest/.deadline/config

deadline-cloud-monitor.path: ...
   The filesystem path to Deadline Cloud monitor, set during login process.

defaults.aws_profile_name: ...
   The AWS profile name to use by default. Set to '' to use the default
   credentials. Other settings are saved with the profile.

settings.job_history_dir: ~/.deadline/job_history/ ... (default)
   The directory in which to place the job submission history for this AWS
   profile name.

defaults.farm_id: farm-281174...
   The Farm ID to use by default.

settings.storage_profile_id:  (default)
   The storage profile that this workstation conforms to. It specifies where
   shared file systems are mounted, and where named job attachments should go.

defaults.queue_id: queue-38906...
   The Queue ID to use by default.
```

crowecawcaw:
Sample output from `deadline`:
```
> deadline
Usage: deadline [OPTIONS] COMMAND [ARGS]...

  The AWS Deadline Cloud CLI provides functionality to interact with the AWS
  Deadline Cloud service.

Options:
  --version                       Show the version and exit.
  --log-level [ERROR|WARNING|INFO|DEBUG]
                                  Set the logging level.

Commands:
  auth            Commands to handle authentication.
  bundle          Commands to work with Open Job Description job bundles.
  config          Manage Deadline's workstation configuration.
  farm            Commands to work with farms.
  fleet           Commands to work with fleets.
  handle-web-url  Runs AWS Deadline Cloud commands sent from a web browser.
  job             Commands to work with jobs.
  queue           Commands for queues.
  worker          Commands to work with workers.
```

epmog (approved):
LGTM. Ship it!

ddneilson (approved):

PR #440, opened 2024-09-03
docs: add storage profile setup instructions

crowecawcaw:
Users must configure a storage profile when using CMF with shared storage, but the README lacked usage instructions.

Add basic storage profile usage instructions and a link to further documentation.

Users can learn how to set up the CLI with a storage profile.

n/a

Yes

No

No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ deadline config gui
 ```
 
 
-To set up shared storage for your farm, you will need to set up a customer-managed fleet with shared network storage and to create a storage profile. See [shared storage](shared-storage) for more information.
+Shared storage is possible with customer-managed fleets (CMF) but not service-managed fleets (SMF). See [shared storage](shared-storage) for more information.
 
 ## Job bundles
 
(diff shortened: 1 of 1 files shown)


erico-aws (dismissed):

README.md
```diff
@@ -77,14 +78,34 @@ api.list_farms()
...
+deadline config set settings.storage_profile_id sp-10b2e48ad6ac4fc88595dfcbef6271f2
+```
+
+Or with the configuration GUI:
+```sh
+deadline config gui
+```
+
+
+To set up shared storage for your farm, you will need to set up a customer-managed fleet with shared network storage and to create a storage profile. See [shared storage](shared-storage) for more information.
```
epmog: nit
```
To set up shared storage for your farm, you will need to set up a customer-managed fleet with shared network storage, and create a storage profile. See [shared storage](shared-storage) for detailed instructions.
```

Almost wonder if we should remove some information here because it doesn't include everything you need. Partial information might just add confusion. Maybe just:

```
For information on how to set up shared storage, see the [user guide](shared-storage).
```
I'm open to suggestions here.

epmog (changes requested):

README.md
crowecawcaw: I wanted to add enough info so that most readers could quickly tell if storage profiles are necessary for their use case. If we just add a link to shared storage, someone that's not familiar with CMF vs SMF or job attachments vs shared storage will need to read a larger amount of documentation before learning if it's relevant.

What about something like, "Shared storage is possible with CMF but not SMF. For information on how to set up..." Or have any other ideas?

erico-aws (approved):

epmog (approved):

PR #442, opened 2024-09-06
feat: adds --storage-profile-id option to the bundle submit command

ddneilson:
Fixes:  No github issue created.

Customers have expressed the need to be able to provide an override for the storage profile id when submitting jobs using `deadline bundle submit`, but that command doesn't provide a way to override that setting. The use-case is submitting many jobs, in parallel, from the same host where jobs need different storage profile ids.

Add an override. Using the override does not store the given storage profile id to the user's configuration file.

Those creating pipeline automation have one less barrier in their way.

I've run added to and run the unit tests. I have also run the new command option to ensure that the command submits jobs with the default as well as overriden storage profile id, and that the overriden id is not saved to the config. Log below:

```bash
% deadline config get settings.storage_profile_id                                    
sp-44e420f6a725493ba4067a89c60ce01c

% deadline bundle submit test_bundle
Submitting to Queue: MainQueue
Waiting for Job to be created...
Submitted job bundle:
   test_bundle
Job creation completed successfully
job-61473d1df8ba42ada64a74966171d644

% deadline job get
jobId: job-61473d1df8ba42ada64a74966171d644
...
storageProfileId: sp-44e420f6a725493ba4067a89c60ce01c
...     

% deadline config get settings.storage_profile_id                                    
sp-44e420f6a725493ba4067a89c60ce01c

% deadline bundle submit test_bundle --storage-profile-id sp-ef65fd46b42e46f19e4bb28dd2da7537
Submitting to Queue: MainQueue
Waiting for Job to be created...
Submitted job bundle:
   test_bundle
Job creation completed successfully
job-d711a6c9b1734e8f8bb826f7dffe1ea4

% deadline config get settings.storage_profile_id                                            
sp-44e420f6a725493ba4067a89c60ce01c
```

The new CLI option does have help text that indicates what it is.

No. The change is purely additive, and 100% optional.

It does not.

----

### DIFF
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -82,6 +82,7 @@ def validate_parameters(ctx, param, value):
 @click.option("--profile", help="The AWS profile to use.")
 @click.option("--farm-id", help="The farm to use.")
 @click.option("--queue-id", help="The queue to use.")
+@click.option("--storage-profile-id", help="The storage profile to use.")
 @click.option("--name", help="The job name to use in place of the one in the job bundle.")
 @click.option(
     "--priority",
@@ -140,7 +141,7 @@ def bundle_submit(
     """
     Submits an Open Job Description job bundle.
     """
-    # Get a temporary config object with the standard options handled
+    # Apply the CLI args to the config
     config = _apply_cli_options_to_config(required_options={"farm_id", "queue_id"}, **args)
 
     hash_callback_manager = _ProgressBarCallbackManager(length=100, label="Hashing Attachments")
@@ -226,6 +227,7 @@ def bundle_submit(
             and args.get("profile") is None
             and args.get("farm_id") is None
             and args.get("queue_id") is None
+            and args.get("storage_profile_id") is None
         ):
             set_setting("defaults.job_id", job_id)
 
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -34,6 +34,11 @@ def fresh_deadline_config():
         # Yield the temp file name with it patched in as the
         # AWS Deadline Cloud config file
         with patch.object(config_file, "CONFIG_FILE_PATH", str(temp_file_path)):
+            # Write a telemetry id to force it getting saved to the config file. If we don't, then
+            # an ID will get generated and force a save of the config file in the middle of a test.
+            # Writing the config file may be undesirable in the middle of a test.
+            config_file.set_setting("telemetry.identifier", "00000000-0000-0000-0000-000000000000")
+
             yield str(temp_file_path)
     finally:
         temp_dir.cleanup()
--- a/test/unit/deadline_client/cli/test_cli_bundle.py
+++ b/test/unit/deadline_client/cli/test_cli_bundle.py
@@ -15,6 +15,7 @@ import pytest
 
 from deadline.client import config
 from deadline.client.api import _queue_parameters
+import deadline.client.api as api_module
 from deadline.client.cli import main
 from deadline.client.cli._groups import bundle_group
 from deadline.client.api import _submit_job_bundle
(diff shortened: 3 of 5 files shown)


leongdl (dismissed):

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -82,6 +82,7 @@ def validate_parameters(ctx, param, value):
 @click.option("--profile", help="The AWS profile to use.")
 @click.option("--farm-id", help="The farm to use.")
 @click.option("--queue-id", help="The queue to use.")
+@click.option("--storage-profile-id", help="The storage profile to use.")
```
epmog, line 85: gui-submit implementation is being punted until they share options?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -218,6 +230,12 @@ def _decide_cancel_submission(upload_group: AssetUploadGroup) -> bool:
             submitter_name=submitter_name,
         )
 
+        # Reset the storage profile ID so that we don't save it to disk.
+        # Note: The config object is a singleton, so the config object that
+        # we may have modified above if a storage_profile_id was given is saved
+        # into that singleton. Setting the job_id writes that singleton to disk
+        set_setting(storage_profile_config_name, prev_storage_id, config=config)
```
epmog: so if I'm understanding correctly, the config is being passed in both times we're callign it for storage profile id so that nothing is saved to disk when these particular calls happens. But due to the singleton nature of the config object, setting the job id would trigger setting this to disk which is why we revert to the previous value. Correct?

ddneilson: Yes, that is 100% correct.

src/deadline/client/cli/_groups/bundle_group.py
ddneilson: Good point. I suppose I *should* add it to there as well...

ddneilson: Update... I took a look through the implementation.

Implementing it for `gui-submit` would change the config file setting 'cause gui-submit doesn't pass around a config object. It relies on the singleton nature of the config. So, I think that implementing the option for gui-submit would not be great in that the options would behave differently. `bundle submit` would treat it as a temporary override for just this submission, but `bundle gui-submit` would treat it as doing a `config set` beforehand. 

So, I'm inclined to leave this as-is. What do you think?

test/unit/deadline_job_attachments/test_download.py
```diff
@@ -2430,7 +2430,7 @@ def download_file(*args):
 
     with patch(
         f"{deadline.__package__}.job_attachments.download.download_file", side_effect=download_file
-    ):
+    ), patch(f"{deadline.__package__}.job_attachments.download.get_s3_client"):
```
ddneilson, line 2433: I noticed that this test was blowing up if the test environment doesn't have valid AWS credentials. It was missing a mock.

leongdl (dismissed):

test/unit/deadline_client/cli/test_cli_bundle.py
```diff
@@ -290,6 +291,52 @@ def test_cli_bundle_job_name(fresh_deadline_config):
...
+    session_mock().client().create_job.assert_called_once_with(
+        farmId=MOCK_FARM_ID,
+        queueId=MOCK_QUEUE_ID,
+        template=ANY,
+        templateType="JSON",
+        priority=50,
+        storageProfileId=CLI_STORAGE_PROFILE_ID,
+    )
+    assert result.exit_code == 0
+    assert get_setting("settings.storage_profile_id") == CLI_STORAGE_PROFILE_ID
```
epmog: Mmm... Should the value we be verifying against be `PRE_STORAGE_PROFILE_ID` if the config hasn't changed due to the command line arg?

ddneilson: Right you are.

leongdl (dismissed):

test/unit/conftest.py
```diff
@@ -34,6 +34,11 @@ def fresh_deadline_config():
         # Yield the temp file name with it patched in as the
         # AWS Deadline Cloud config file
         with patch.object(config_file, "CONFIG_FILE_PATH", str(temp_file_path)):
+            # Write a telemetry id to force it getting saved to the config file. If we don't, then
+            # an ID will get generated and force a save of the config file in the middle of a test.
+            # Writing the config file may be undesirable in the middle of a test.
+            config_file.set_setting("telemetry.identifier", "d9fecc89-676b-4d12-aa22-ee906fb950d9")
```
epmog:
```suggestion
            config_file.set_setting("telemetry.identifier", "00000000-0000-0000-0000-000000000000")
```

epmog (approved):
Thanks Daniel! Ship it!

leongdl (approved):

PR #443, opened 2024-09-10
fix: only override queue environment settings if empty

jflatt-gia:
Fixes: CondaPackages queue environment setting is overridden in deadline-cloud-for-blender submitter

Initial shared parameter values are populated from the CondaPackages queue environment as expected, then the blender submitter overrides with its own Conda package list.
See:
https://github.com/aws-deadline/deadline-cloud-for-blender/blob/aac1bc731302d0780b3e077e4b389befe47f969d/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py#L79
These defaults are reasonable, but if you're running a local version of blender that differs from the cloud version, or if your conda packages have different names, this override isn't always useful, and would need changing on each submission.

If the variable default value is empty (which is how it's shipped for the CondaPackages variable), allow the blender submitter to override, otherwise keep what is in the Queue Environment variable.  This is more consistent with how the CondaChannels variable is working.

Should allow the default Conda Queue Environment to work as-is.  If any value is entered into the Conda Queue Environment for CondaPackages, it will override the default/empty value.

1. In Blender, Render->Submit to AWS Deadline Cloud, find CondaPackages = blender=a.b.* blender-openjd=x.y.*, Conda Channels = deadline-cloud
2. Modify the default "Conda" Queue Environment, set a different default value for CondaPackages and CondaChannels.
3.  In Blender, Render->Submit to AWS Deadline Cloud, find CondaPackages = blender=a.b.* blender-openjd=x.y.*, Conda Channels = <new channels value>.  Regardless of what CondaPackages is set in Deadline "Conda" Queue Environment, the value in the Blender submitter will always start with this value.
4. Apply this change
5. In Blender, Render->Submit to AWS Deadline Cloud, find CondaPackages = <new packages value>, Conda Channels = <new channels value>.  Having a value set in the CondaPackages variable in the "Conda" Queue Environment will flow through to the submitter dialog box.
6. Modify the default "Conda" Queue Environment, set the default value for CondaPackages back to an empty string "".
7. In Blender, Render->Submit to AWS Deadline Cloud, find CondaPackages = blender=a.b.* blender-openjd=x.y.*, Conda Channels = <new channels value>.  The reasonable default is given for CondaPackages.

I don't have a full setup of all deadline-cloud-for-xyz software packages installed/setup and am unable to fully test.  If any assistance is available, that would be great.

Not applicable.

I would say no.  This is a change in behavior, but Queue Environment variables can still be modified.

No.

----

### DIFF
--- a/src/deadline/client/ui/widgets/shared_job_settings_tab.py
+++ b/src/deadline/client/ui/widgets/shared_job_settings_tab.py
@@ -184,7 +184,8 @@ class SharedJobSettingsWidget(QWidget):  # pylint: disable=too-few-public-method
             # Apply the initial queue parameter values
             for parameter in queue_parameters:
                 if parameter["name"] in self.initial_shared_parameter_values:
-                    parameter["value"] = self.initial_shared_parameter_values[parameter["name"]]
+                    if parameter["default"] == "":
+                        parameter["value"] = self.initial_shared_parameter_values[parameter["name"]]
             self.queue_parameters_box.rebuild_ui(parameter_definitions=queue_parameters)
 
     def _load_queue_parameters_thread_function(self, refresh_id: int, farm_id: str, queue_id: str):


src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -184,7 +184,8 @@ def _handle_queue_parameters_update(self, refresh_id, queue_parameters):
             # Apply the initial queue parameter values
             for parameter in queue_parameters:
                 if parameter["name"] in self.initial_shared_parameter_values:
-                    parameter["value"] = self.initial_shared_parameter_values[parameter["name"]]
+                    if parameter["default"] == "":
```
mwiebe, line 187: I see what you're doing here, and the intent is great. But, it introduces an ambiguity between an intentional default "" and the lack of a default. It could be in another case we don't want the empty string to be special like here. We didn't think very deeply about "null" values for the open jd types. Can you think of other ways we might fix this up? If not what you've got here may be the pragmatic choice even with the inconsistency.

PR #445, opened 2024-09-11
chore: add client version and submitter name to user agent

crowecawcaw:
It'd be useful to have the Deadline Client version and submitter name as part of the user agent string for API calls. The user agent string appears in CloudTrail and can help track down the source of API calls.

Add the client version and submitter name to the user agent string for API requests.

Better data on the source of API calls.

I submitted bundles using the GUI submitter and CLI submitter and observed the user agent strings both in debug logs and in the actual requests. Specifically I tried:
- GUI submit without a custom submitter name
- GUI submit with a custom submitter name. I also used special characters to make sure boto sanitized them as expected
- CLI submit with and without custom submitter names

Example user agent string from testing (relevant info at the end):
```
Boto3/1.34.138 md/Botocore#1.34.138 ua/2.0 os/macos#23.6.0 md/arch#arm64 lang/python#3.12.4 md/pyimpl#CPython cfg/retry-mode#legacy Botocore/1.34.138 app/deadline-client#0.48.8.post25+g61df452 submitter/CLI
```

- Has the README.md been updated? If you modified CLI arguments, for instance. n/a

No

No

----

### DIFF
--- a/src/deadline/client/api/_submit_job_bundle.py
+++ b/src/deadline/client/api/_submit_job_bundle.py
@@ -42,6 +42,7 @@ from ...job_attachments.models import (
 )
 from ...job_attachments.progress_tracker import SummaryStatistics, ProgressReportMetadata
 from ...job_attachments.upload import S3AssetManager
+from ._session import session_context
 
 logger = logging.getLogger(__name__)
 
@@ -65,7 +66,7 @@ def create_job_from_job_bundle(
     upload_progress_callback: Optional[Callable[[ProgressReportMetadata], bool]] = None,
     create_job_result_callback: Optional[Callable[[], bool]] = None,
     require_paths_exist: bool = False,
-    submitter_name: str = "CLI",
+    submitter_name: Optional[str] = None,
 ) -> Union[str, None]:
     """
     Creates a job in the farm/queue configured as default for the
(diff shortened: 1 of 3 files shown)


src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -134,6 +135,11 @@ def create_job_from_job_bundle(
                 ProgressReport as a parameter, which can be used for projecting remaining time, as in done in the CLI.
     """
 
+    if not submitter_name:
```
ddneilson, line 138: How come the switch from required with default to optional with hand-coded default?

src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -52,6 +53,11 @@ def show_job_bundle_submitter(
     to stay on top.
     """
 
+    if not submitter_name:
```
ddneilson, line 56: Same question as above.

src/deadline/client/api/_submit_job_bundle.py
crowecawcaw: This fixes a bug from a previous change I made to this code. Previously we added the ability to customize the submitter name as a CLI argument (`--submitter-name`) and the intention was the if one isn't provided, the default value would be used. The defaults were set as function parameter defaults. What I noticed was actually happening was that the if the CLI argument is not present, this function gets called with `None` as the `submitter_name` value and the default doesn't get applied. This change makes it so `None` values are replaced by the submitter name default.

ddneilson (approved):

epmog (approved):

PR #446, opened 2024-09-16
[WIP] feat(JA CLI): Job Attachment Manifest CLIs. Snapshot, Diff, Download,…

leongdl:
This PR is being broken down into smaller chunks to make submissions and reviews easier:

Part 1: Glob utilities - https://github.com/aws-deadline/deadline-cloud/pull/450
Part 2: Internal refactoring - https://github.com/aws-deadline/deadline-cloud/pull/451/commits
Part 3: Snapshot API https://github.com/aws-deadline/deadline-cloud/pull/463
Part 4: diff API. - https://github.com/aws-deadline/deadline-cloud/pull/465
Part 5: Upload / Download API

PR #450, opened 2024-09-20
feat: Glob utilities and misc functions

leongdl:
- We are introducing Job Attachment Manifest and Attachments APIs.
- There will be snapshot, diff, upload and download functionalities to create and manage manifests.
- There will be upload and download attachment of assets captured by manifests.
 
- This PR is the first part of submitting
 - https://github.com/aws-deadline/deadline-cloud/pull/446
- In this PR, I am introducing utilities for Globbing file systems, processing a glob config file and misc utility functions.
 
- Utilities to get Job Attachments user interfaces launched.

- hatch build, hatch run fmt, hatch run lint, hatch run test.
- Integration tests are not applicable here, and will be added with subsequent CLI / API tests.

- These are internal classes.

No.

- No, these are internal utilities. Standard AWS security review best practices.

### DIFF
--- a/src/deadline/job_attachments/_glob.py
+++ b/src/deadline/job_attachments/_glob.py
@@ -1,0 +1,70 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import json
+from pathlib import Path
+from typing import List, Optional
+from deadline.client.exceptions import NonValidInputError
+from deadline.job_attachments.models import GlobConfig
+
+
+def _process_glob_inputs(glob_arg_input: str) -> GlobConfig:
+    """
+    Helper function to process glob inputs.
+    glob_input: String, can represent a json, filepath, or general include glob syntax.
+    """
+
+    # Default Glob config.
+    glob_config = GlobConfig()
+    if glob_arg_input is None or len(glob_arg_input) == 0:
+        # Not configured, or not passed in.
+        return glob_config
+
+    try:
+        input_as_path = Path(glob_arg_input)
+        if input_as_path.is_file():
+            # Read the file so it can be parsed as JSON.
+            with open(glob_arg_input) as f:
+                glob_arg_input = f.read()
+    except Exception:
+        # If this cannot be processed as a file, try it as JSON.
+        pass
+
+    try:
+        # Parse the input as JSON, default to Glob Config defaults.
+        input_as_json = json.loads(glob_arg_input)
+        glob_config.include_glob = input_as_json.get(GlobConfig.INCLUDE, glob_config.include_glob)
+        glob_config.exclude_glob = input_as_json.get(GlobConfig.EXCLUDE, glob_config.exclude_glob)
+    except Exception:
+        # This is not a JSON blob, bad input.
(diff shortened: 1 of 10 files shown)


benl-2023 (approved):

src/deadline/job_attachments/_glob.py
```diff
@@ -0,0 +1,70 @@
...
+from pathlib import Path
+from typing import List, Optional
+from deadline.client.exceptions import NonValidInputError
+from deadline.job_attachments.models import GlobConfig
+
+
+def _process_glob_inputs(glob_arg_input: str) -> GlobConfig:
+    """
+    Helper function to process glob inputs.
+    glob_input: String, can represent a json, filepath, or general include glob syntax.
```
mwiebe, line 13: Is the plan here to accept JSON as a CLI parameter input? If so, maybe we can adjust that to accept multiple arguments of include & exclude instead? I believe CLI interfaces that don't use JSON that way are generally more comfortable to use.

src/deadline/job_attachments/_glob.py
```diff
@@ -0,0 +1,70 @@
...
+        raise NonValidInputError(f"Glob input {glob_arg_input} cannot be deserialized as JSON")
+
+    return glob_config
+
+
+def _glob_paths(
+    path: str, include: List[str] = ["**/*"], exclude: Optional[List[str]] = None
+) -> List[str]:
+    """
+    Glob routine that supports Unix style pathname pattern expansion for includes and excludes.
```
mwiebe, line 48: This description doesn't say what the function does. What is the purpose of this function, what does it produce as output?

godobyte (approved):

src/deadline/job_attachments/_glob.py
leongdl: Do you mean something like 
```--include abc/*.xyz,def/*.123```. ?

I was going by the OpenJD suggestion for globs, but happy to go about it either ways.

src/deadline/job_attachments/_glob.py
leongdl: Will add more details!

src/deadline/job_attachments/_glob.py
leongdl: I was thinking of something like this example: https://github.com/aws-deadline/deadline-cloud/pull/446/files#diff-33ab9cc270d578c6903c5e1cf7eb1e25e518cfc7222717f4599df29a368d0cc5R8

leongdl: For flexibility, I can also add `--include` and `--exclude` as comma separated inputs w/ the CLI version too. This function here specifically supports JSON file and JSON string.

leongdl: We chatted internally within the team. This code provides support for a JSON based glob syntax and can be used for large lists of include and exclude. For CLI use cases, I will also add `--include` and `--exclude` so both basis are covered.

src/deadline/job_attachments/_glob.py
leongdl: Updated in the next PR of this feature series.

PR #451, opened 2024-09-20
feat(JA): Code refactor and moving functions around for code sharing …

leongdl:
- We are introducing Job Attachment Manifest and Attachments APIs.
- There will be snapshot, diff, upload and download functionalities to create and manage manifests.
- There will be upload and download attachment of assets captured by manifests.
 
- This PR is the second part of submitting
 - https://github.com/aws-deadline/deadline-cloud/pull/446
- In this PR, I am doing small internal refactoring changes to make upcoming changes easier to consume. Just moving code around. No logical changes.
 
- Utilities to get Job Attachments user interfaces launched.

- hatch build, hatch run fmt, hatch run lint, hatch run test, hatch run all:test
- Integration tests
```
==================================================================== slowest 5 durations ====================================================================
8.02s setup    test/integ/deadline_job_attachments/test_job_attachments.py::test_upload_input_files_all_assets_in_cas[2023-03-03]
5.91s setup    test/integ/deadline_job_attachments/test_job_attachments.py::test_sync_inputs_with_step_dependencies[2023-03-03]
4.91s call     test/integ/deadline_job_attachments/test_job_attachments.py::test_sync_outputs_no_job_attachment_settings_in_job[2023-03-03]
4.35s teardown test/integ/deadline_job_attachments/test_job_attachments.py::test_download_outputs_windows_file_path_UNC[2023-03-03]
4.00s call     test/integ/deadline_job_attachments/test_job_attachments.py::test_sync_inputs_with_step_dependencies[2023-03-03]
======================================================== 15 passed, 2 skipped, 4 warnings in 54.42s =========================================================
```

- These are internal classes.

No.

- No, these are internal utilities. Standard AWS security review best practices.

### DIFF
--- a/src/deadline/client/api/__init__.py
+++ b/src/deadline/client/api/__init__.py
@@ -61,7 +61,10 @@ from ._telemetry import (
     TelemetryClient,
     record_success_fail_telemetry_event,
 )
-from ._submit_job_bundle import create_job_from_job_bundle, wait_for_create_job_to_complete
+from ._submit_job_bundle import (
+    create_job_from_job_bundle,
+    wait_for_create_job_to_complete,
+)
 from ._get_storage_profile_for_queue import get_storage_profile_for_queue
 
 logger = getLogger(__name__)
--- a/src/deadline/client/api/_job_attachment.py
+++ b/src/deadline/client/api/_job_attachment.py
@@ -1,0 +1,47 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from deadline.client import api
+from deadline.client.config import config_file
+from deadline.job_attachments.models import AssetRootGroup, AssetRootManifest
+from deadline.job_attachments.upload import S3AssetManager, SummaryStatistics
+
+
+import textwrap
+from configparser import ConfigParser
+from typing import Callable, Dict, List, Optional, Tuple
+
+
+def _hash_attachments(
+    asset_manager: S3AssetManager,
+    asset_groups: List[AssetRootGroup],
+    total_input_files: int,
+    total_input_bytes: int,
+    print_function_callback: Callable = lambda msg: None,
+    hashing_progress_callback: Optional[Callable] = None,
+    config: Optional[ConfigParser] = None,
+) -> Tuple[SummaryStatistics, List[AssetRootManifest]]:
+    """
+    Starts the job attachments hashing and handles the progress reporting
+    callback. Returns a list of the asset manifests of the hashed files.
+    """
+
+    def _default_update_hash_progress(hashing_metadata: Dict[str, str]) -> bool:
+        return True
+
+    if not hashing_progress_callback:
+        hashing_progress_callback = _default_update_hash_progress
+
+    hashing_summary, manifests = asset_manager.hash_assets_and_create_manifest(
+        asset_groups=asset_groups,
+        total_input_files=total_input_files,
+        total_input_bytes=total_input_bytes,
+        hash_cache_dir=config_file.get_cache_directory(),
+        on_preparing_to_submit=hashing_progress_callback,
+    )
+    api.get_deadline_cloud_library_telemetry_client(config=config).record_hashing_summary(
+        hashing_summary
+    )
+    print_function_callback("Hashing Summary:")
+    print_function_callback(textwrap.indent(str(hashing_summary), "    "))
+
+    return hashing_summary, manifests
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -5,25 +5,25 @@ Provides functionality for boto3 Sessions, Clients, and properties
 of the Deadline-configured IAM credentials.
 """
 from __future__ import annotations
-from functools import lru_cache
+
 import logging
 from configparser import ConfigParser
 from contextlib import contextmanager
 from enum import Enum
+from functools import lru_cache
 from typing import Optional
+
 import boto3  # type: ignore[import]
+import botocore
 from botocore.client import BaseClient  # type: ignore[import]
 from botocore.credentials import CredentialProvider, RefreshableCredentials
 from botocore.exceptions import (  # type: ignore[import]
     ClientError,
     ProfileNotFound,
 )
-
 from botocore.session import get_session as get_botocore_session
-import botocore
+
 from .. import version
-
-
 from ..config import get_setting
 from ..exceptions import DeadlineOperationError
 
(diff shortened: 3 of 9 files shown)


src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -403,7 +403,7 @@ def wait_for_create_job_to_complete(
     )
 
 
-def _hash_attachments(
+def hash_attachments(
```
leongdl: This method will be shared with other CLIs. I'm thinking if we should move it to `_common`, or another common file.

I think I will just move it.

src/deadline/client/api/_job_attachment.py
```diff
@@ -0,0 +1,47 @@
...
+from deadline.job_attachments.models import AssetRootGroup, AssetRootManifest
+from deadline.job_attachments.upload import S3AssetManager, SummaryStatistics
+
+
+import textwrap
+from configparser import ConfigParser
+from typing import Callable, Dict, List, Optional, Tuple
+
+
+def _hash_attachments(
```
leongdl, line 14: Migrated from the bundle submission CLI.

src/deadline/client/api/_session.py
```diff
@@ -104,6 +104,18 @@ def invalidate_boto3_session_cache() -> None:
     _get_queue_user_boto3_session.cache_clear()
 
 
+def get_default_client_config() -> botocore.config.Config:
```
leongdl, line 107: Refactoring so this code can be shared.

src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -403,42 +404,6 @@ def wait_for_create_job_to_complete(
     )
 
 
-def _hash_attachments(
```
leongdl, line 406: This function is used in a few CLIs moving forward, so to reduce confusion, I moved it to its own file.

src/deadline/client/cli/_common.py
```diff
@@ -182,3 +189,42 @@ def _cli_object_repr(obj: Any):
     # strings to end with "\n".
     obj = _fix_multiline_strings(obj)
     return deadline_yaml_dump(obj)
+
+
+class _ProgressBarCallbackManager:
```
leongdl, line 194: Re-using for multiple APIs going forward, moved from `bundle_group.py`

src/deadline/client/api/__init__.py
```diff
@@ -4,6 +4,7 @@
     "login",
     "logout",
     "create_job_from_job_bundle",
+    "_hash_attachments",
```
mwiebe: This `api` namespace is generally about calling the web APIs. Hashing job attachments files seems like a more local operation, maybe there's another place it would fit better?

leongdl: It was originally in the CLI `_submit_job_bundle.py`

One of my ideas was to completely move it to job attachments sub folder. but the current definition and usage with CLI elements here does not lend itself easily to be in JA logic only. I can also move it to a `utils` like location?

mwiebe: Yes, it was there as an implementation detail. What's the purpose to move it? Could it stay an internal detail the way it is now, or does it need to be a function that the library exposes for external use?

leongdl: Good point - I'm not sure why we added it to `__all__`, now that I think about it, its `_` means it is internal. I'll remove this from `__all__` altogether.

Its one of the things I picked up from the original implementation branch while cutting it up for merging. Great catch!

I will still extract this method from the bundle submission file to a separate file. There will be a few places that leverage this shared code.

src/deadline/client/api/_job_attachment.py
jericht, line 1: nit: should filename be `_job_attachments.py` to align with the actual package name? or is this intentionally different to avoid confusion?

src/deadline/client/cli/_common.py
```diff
@@ -9,20 +9,27 @@
...
 
 from ..config import config_file
 from ..exceptions import DeadlineOperationError
 from ..job_bundle import deadline_yaml_dump
+from ._groups._sigint_handler import SigIntHandler
 
 _PROMPT_WHEN_COMPLETE = "PROMPT_WHEN_COMPLETE"
 
+# Set up the signal handler for handling Ctrl + C interruptions.
+sigint_handler = SigIntHandler()
```
jericht, line 31: nit: should we prefix this with `_` since it's only needed within this file?

src/deadline/client/cli/_groups/click_logger.py
```diff
@@ -0,0 +1,39 @@
...
+        self,
+        message: t.Optional[dict] = None,
+        file: t.Optional[t.IO[t.Any]] = None,
+        nl: bool = True,
+        err: bool = False,
+        color: t.Optional[bool] = None,
+        indent=None,
+    ):
+        if self._is_json:
+            click.echo(json.dumps(obj=message, indent=indent), file, nl, err, color)
```
jericht, line 39: Depending on what `self._is_json` is set to, each of these functions can be a no-op. Is that the desired behavior? or do we want to throw an exception or log an error instead? or maybe try to print the message anyway?

jericht (approved):

src/deadline/client/api/_job_attachment.py
```diff
@@ -0,0 +1,47 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
godobyte, line 1: Do you plan for this file to hold customer facing APIs as well, or only job attachments util functions?

src/deadline/client/cli/_common.py
godobyte: nit - Should we remove the prefix `_` considering it's a common util now?

src/deadline/client/api/_job_attachment.py
leongdl: Yeah I named it specifically for "Job Attachment" not "Attachments" because the former is a feature, this is a helper.

src/deadline/client/cli/_common.py
leongdl: Its common util, but still internal to this package. Its not public for external users so `_` should stay.

src/deadline/client/cli/_groups/click_logger.py
leongdl: Ah - the intention is to allow users to call `.json(message)` or `.echo` without thinking about it. Just log as you see fit. It does not need to raise in either case. 

The output is either "human readable" or "json" only, cannot be both, which is what this class helps to do.

We do not want developers to actively think if a CLI is in JSON mode or plain text mode.

src/deadline/client/cli/_common.py
leongdl: This was also `signint_handler` in the bundle_group code, so I would keep it for consistency.

godobyte (approved):

src/deadline/client/api/_job_attachment.py
leongdl: No - this is the job attachments util only.

The API will be in the `job_attachment` sub-folder and therefore sub-package. It'll e clearly deadline.job_attachment.api

PR #452, opened 2024-09-25
feat: Add a simple backoff and retry utility helper

leongdl:
- We had to revert this [PR](https://github.com/aws-deadline/deadline-cloud/commit/be5761765dc122fe8c9742ec185f6309c834b1df) because it added a new dependency.
- Our goal is to minimize external 3rd party library dependencies in deadline-cloud's python lib.

- Add a basic retry decorator to support use cases that need retry. 

- We can decorate @retry now with exponential back off and retry

- hatch build
- hatch run all:lint
- hatch run fmt
- hatch run test

- Yes

- Not applicable.

- Not applicable

Not applicable. New internal API function docs added.

No

No

----

### DIFF
--- a/src/deadline/job_attachments/_utils.py
+++ b/src/deadline/job_attachments/_utils.py
@@ -1,9 +1,12 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
 import datetime
+from functools import wraps
 from hashlib import shake_256
 from pathlib import Path
-from typing import Tuple, Union
+import random
+import time
+from typing import Any, Callable, Optional, Tuple, Type, Union
 import uuid
 import ctypes
 import sys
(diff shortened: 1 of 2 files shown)


src/deadline/job_attachments/_utils.py
```diff
@@ -98,3 +101,56 @@ def _is_windows_file_path_limit() -> bool:
 
         return bool(ntdll.RtlAreLongPathsEnabled())
     return True
+
+
+def _retry(
+    ExceptionToCheck: Union[Type[Exception], Tuple[Type[Exception], ...]] = AssertionError,
+    tries: int = 4,
+    delay: Union[int, float, Tuple[Union[int, float], Union[int, float]]] = 3.0,
+    backoff: float = 2.0,
```
crowecawcaw: Nit: these defaults would lead to very slow failure in the worst case. Would it make sense to tighten them up?

crowecawcaw (dismissed):

src/deadline/job_attachments/_utils.py
leongdl: Sure, I can adjust them in a follow up (i have many PRs on going)

how about 1 retry, delay of 1s, back off 1s?

crowecawcaw (approved):

jericht (approved):

PR #456, opened 2024-09-26
docs: developer guidance on adding dependencies

leongdl:
- It is not clear what deadline cloud's guidance was to using 3rd party dependencies.
 
- Adding documentation and expectations on PRs with new dependencies.
- Update Pull request template, and also `DEVELOPMENT.md`.

- Documentation, improve developer experience.

NA.

NA

NA

- This PR is documentation.

- No.

- No

----

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,6 +20,13 @@ See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainlin
 - Are relevant docstrings in the code base updated?
 - Has the README.md been updated? If you modified CLI arguments, for instance.
 
+### Does this PR introduce new dependencies?
+
+This library is designed to be integrated into third-party applications that have bespoke and customized deployment environments. Adding dependencies will increase the chance of library version conflicts and incompatabilities. Please evaluate the addition of new dependencies. See the [Dependencies](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies) section of DEVELOPMENT.md for more details.
+
+*   [ ] This PR adds one or more new dependency Python packages. I acknowledge I have reviewed the considerations for adding dependencies in [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies).
+*   [ ] This PR does not add any new dependencies.
+
 ### Is this a breaking change?
 
 A breaking change is one that modifies a public contract in a way that is not backwards compatible. See the 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -12,6 +12,7 @@ Table of Contents:
    * [Integration tests](#integration-tests)
 * [Things to Know](#things-to-know)
    * [Public contracts](#public-contracts)
+   * [Library Dependencies](#dependencies)
    * [Qt and Calling AWS APIs](#qt-and-calling-aws-including-aws-deadline-cloud-apis)
 
 ## Development Environment Setup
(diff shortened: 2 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,6 +20,12 @@ See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainlin
 - Are relevant docstrings in the code base updated?
 - Has the README.md been updated? If you modified CLI arguments, for instance.
 
+### Does this PR introduce new dependencies?
+
+- Does the PR add new package dependencies to `dependencies` section of `pyproject.toml`?
+- Please consider why a new dependency is needed? Dependency libraries need to be updated or maintained.
```
crowecawcaw:
```suggestion
- Please consider why a new dependency is needed. Dependency libraries need to be updated or maintained.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,6 +20,12 @@ See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainlin
 - Are relevant docstrings in the code base updated?
 - Has the README.md been updated? If you modified CLI arguments, for instance.
 
+### Does this PR introduce new dependencies?
+
+- Does the PR add new package dependencies to `dependencies` section of `pyproject.toml`?
```
mwiebe: A short description of why it's better to minimize dependencies could help with reasoning about it. The main reason is that this library is designed to be integrated into third-party applications by customers who have highly customized deployments. Every additional dependency increases the risk of version conflicts with something they've already set up.

DEVELOPMENT.md
```diff
@@ -201,6 +201,13 @@ For the Python library interface:
 * Changing the location that a file or directory is created should be considered to be a breaking change. These locations have a tendancy to become
   de-facto parts of the public contract as users build automation that assumes these locations is unchanged.
 
+### Library Dependencies
```
jusiskin, line 205: Other things that should be considered when adding new dependencies:

### quality of the dependency

- Is the dependency reputable or maintained by a reputable source? Not sure how we can quantify this, but some ideas include:
    -  PyPI download stats?
    - GitHub stars?
    - GitHub dependency graph showing downstream consumers?
- Is it well-maintained? does it release regularly/recently?

### version pinning

How should we pin this new dependency? Does the dependency document their versioning scheme (e.g. semver)?

`deadline-cloud` is both a python library and an application in the case of the CLI. Libraries should strive to accommodate a range of versions to allow consumers to have some flexibility in their closures. Applications should be more opinionated to enforce more reproducible behavior.

In the current structure, we need to balance between both. This is probably a "use your best judgement" decision, but we should prompt the contributor to think about this.

### licensing

*   We also need to ensure the license of the dependency meets our needs
*   We also aim to attribute our dependencies in https://github.com/aws-deadline/deadline-cloud/blob/mainline/THIRD_PARTY_LICENSES

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,6 +20,12 @@ See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainlin
 - Are relevant docstrings in the code base updated?
 - Has the README.md been updated? If you modified CLI arguments, for instance.
 
+### Does this PR introduce new dependencies?
```
jusiskin, line 23: Another option is to have two checkboxes:

*   [ ] This PR adds one or more new dependency Python packages. I acknowledge I have reviewed the considerations for adding dependencies in `DEVELOPMENT.md`.

jusiskin (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
leongdl: Sure thing, let me change

.github/PULL_REQUEST_TEMPLATE.md
leongdl: Nice! I'll add a framing into here.

.github/PULL_REQUEST_TEMPLATE.md
leongdl: NIce! I like the checkbox idea.

DEVELOPMENT.md
leongdl: Totally agree - these are all good things to consider when selecting 3rd party libs.

.github/PULL_REQUEST_TEMPLATE.md
leongdl: Adopted this checkbox idea, also changed the link to `DEVELOPMENT.md`

DEVELOPMENT.md
leongdl: Added - I reframed the sections The Version pinning one I think needs more work.

crowecawcaw (approved):

epmog (approved):

jusiskin (approved):

PR #461, opened 2024-10-05
feat: add attachment upload command implementation

godobyte:
We'd like to introduce `attachment upload` CLI to improve asset manage workflow.

Add a CLI to upload assets based specified in manifests.

Help customer to better manage and potentially automate asset upload workflow.

#### Manual Testing

Upload to given directory
```
(attachment-cli)> deadline attachment upload --manifests "/Users/yuanbian/Downloads/a027
1fe0c8b1c1f99b82b442cd878122_input" -r "/tmp" --upload-manifest-path Test
Uploaded assets from /tmp, to s3 s3://smf-queue/DeadlineCloud key Test/a0271fe0c8b1c1f99b82b442cd878122_input, hashed data 5e3a52261d2d5d542713f832c57ce197
```
Upload with path mapping
```
(attachment-cli) [0|1]> deadline attachment upload --manifest
s "/Users/XXXXXXXX/Downloads/0bb7eb91fdf8780c4a7e6174de6dfc5e_input" --path-mapping-rules "/Users/XXXXXXXX/Library/CloudS
torage/WorkDocsDrive-Documents/code/thinkbox/deadline-clients/deadline-cloud/path_mapping.json" --upload-manifest-path Tes
t
Uploaded assets from /local/home/XXXXXXXX/bea/deadline-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example, to s3 s3://smf-queue/DeadlineCloud key Test/0bb7eb91fdf8780c4a7e6174de6dfc5e_input, hashed data 5e3a52261d2d5d542713f832c57ce197
```

Error Handling - directory not valid
```
(attachment-cli)> deadline attachment upload --manifests "/Us
ers/XXXXXXXX/Downloads/0bb7eb91fdf8780c4a7e6174de6dfc5e_input" -r "/local/home/XXXXXXXX/bea/deadline-clients/deadline-clou
d-worker-agent/scripts/submit_jobs/asset_example" --upload-manifest-path Test
The AWS Deadline Cloud CLI encountered the following exception:
Specified root dir ('/local/home/XXXXXXXX/bea/deadline-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example',) contains nonvalid directory.
  File "/Users/XXXXXXXX/.local/share/mise/installs/python/3.11/lib/python3.11/site-packages/deadline/client/cli/_common.py", line 54, in wraps
    func(*args, **kwargs)
  File "/Users/XXXXXXXX/.local/share/mise/installs/python/3.11/lib/python3.11/site-packages/deadline/client/cli/_groups/attachment_group.py", line 174, in attachment_upload
    attachment_api.attachment_upload(
  File "/Users/XXXXXXXX/.local/share/mise/installs/python/3.11/lib/python3.11/site-packages/deadline/job_attachments/api/attachment.py", line 108, in attachment_upload
    raise NonValidInputError(f"Specified root dir {root_dirs} contains nonvalid directory.")
deadline.client.exceptions.NonValidInputError: Specified root dir ('/local/home/XXXXXXXX/bea/deadline-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example',) contains nonvalid directory.
```

Error Handling - no such file
```
(attachment-cli)> deadline attachment upload --manifests "/Us
ers/XXXXXXXX/Downloads/0bb7eb91fdf8780c4a7e6174de6dfc5e_input" --path-mapping-rules "/Users/XXXXXXXX/deadline-cloud/path_mapping.json" --upload-manifest-path Test
The AWS Deadline Cloud CLI encountered the following exception:

FileNotFoundError: [Errno 2] No such file or directory: '/Users/XXXXXXXX/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example/files/file1.txt'
````

- Are relevant docstrings in the code base updated? [Yes]
  - The CLI is still beta under feature flag, will collect feedback and improve before formally document them.

*   [v] This PR does not add any new dependencies.

No

No

----

### DIFF
--- a/src/deadline/client/cli/_groups/attachment_group.py
+++ b/src/deadline/client/cli/_groups/attachment_group.py
@@ -12,6 +12,7 @@ import boto3
 
 from typing import Optional
 
+from .click_logger import ClickLogger
 from .._common import _apply_cli_options_to_config, _handle_error
 from ...config import config_file
 
@@ -26,7 +27,7 @@ from deadline.job_attachments.models import JobAttachmentS3Settings
 @_handle_error
 def cli_attachment():
     """
-    Commands to work with AWS Deadline Cloud Job Attachments.
+    Commands to work with Job Attachments.
     """
 
 
@@ -35,23 +36,24 @@ def cli_attachment():
     help="BETA",
 )
 @click.option(
+    "-m",
     "--manifests",
     multiple=True,
     required=True,
-    help="Comma separated file paths to manifest formatted files.",
+    help="File path(s) to manifest formatted file(s).",
 )
 @click.option(
-    "--s3-root-path", help="Job Attachments S3 root path including bucket name and root prefix."
+    "--s3-root-uri", help="Job Attachments S3 root uri including bucket name and root prefix."
 )
-@click.option("--path-mapping-rules", help="Path to a file with the path mapping rules to use")
+@click.option("--path-mapping-rules", help="Path to a file with the path mapping rules to use.")
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
 @click.option("--profile", help="The AWS profile to use.")
-@click.option("--json", default=None, is_flag=True, help="Output is printed as JSON for scripting")
+@click.option("--json", default=None, is_flag=True, help="Output is printed as JSON for scripting.")
 @_handle_error
 def attachment_download(
     manifests: list[str],
-    s3_root_path: str,
+    s3_root_uri: str,
     path_mapping_rules: str,
     json: bool,
     **args,
@@ -60,6 +62,7 @@ def attachment_download(
     Synchronize files of manifest root(s) to a machine.
     The input of the CLI is a path to a Job Attachments manifest file to download assets.
     """
+    logger: ClickLogger = ClickLogger(is_json=json)
 
     # Setup config
     config = _apply_cli_options_to_config(**args)
@@ -85,14 +88,95 @@ def attachment_download(
         if not s3_settings:
             raise MissingJobAttachmentSettingsError(f"Queue {queue_id} has no attachment settings")
 
-        s3_root_path = s3_settings.to_root_path()
+        s3_root_uri = s3_settings.to_root_path()
 
-    if not s3_root_path:
+    if not s3_root_uri:
         raise MissingJobAttachmentSettingsError("No valid s3 root path available")
 
     attachment_api.attachment_download(
         manifests=manifests,
-        s3_root_path=s3_root_path,
+        s3_root_uri=s3_root_uri,
         boto3_session=boto3_session,
         path_mapping_rules=path_mapping_rules,
+        logger=logger,
+    )
+
+
+@cli_attachment.command(name="upload", help="BETA")
+@click.option(
+    "-m",
+    "--manifests",
+    multiple=True,
+    required=True,
+    help="File path(s) to manifest formatted file(s).",
+)
+@click.option(
+    "-r",
+    "--root-dirs",
+    multiple=True,
+    help="The root directory of assets to upload.",
+)
+@click.option("--path-mapping-rules", help="Path to a file with the path mapping rules to use.")
+@click.option(
+    "--s3-root-uri", help="Job Attachments S3 root uri including bucket name and root prefix."
+)
+@click.option(
+    "--upload-manifest-path", default=None, help="File path for uploading the manifests to CAS."
+)
+@click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
+@click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
+@click.option("--profile", help="The AWS profile to use.")
+@click.option("--json", default=None, is_flag=True, help="Output is printed as JSON for scripting")
+@_handle_error
+def attachment_upload(
+    manifests: list[str],
+    root_dirs: list[str],
+    path_mapping_rules: str,
+    s3_root_uri: str,
+    upload_manifest_path: str,
+    json: bool,
+    **args,
+):
+    """
+    Upload output files to s3. The files always include data files, optionally upload manifests prefixed by given path.
+    """
+    logger: ClickLogger = ClickLogger(is_json=json)
+
+    # Setup config
+    config = _apply_cli_options_to_config(**args)
+
+    # Assuming when passing with config, session constructs from the profile id
+    # TODO - add type for profile, if queue type, get queue sesson directly
+    boto3_session: boto3.session = api.get_boto3_session(config=config)
+
+    if not args.pop("profile", None):
+        queue_id: str = config_file.get_setting("defaults.queue_id", config=config)
+        farm_id: str = config_file.get_setting("defaults.farm_id", config=config)
+
+        deadline_client = boto3_session.client("deadline")
+        boto3_session = api.get_queue_user_boto3_session(
+            deadline=deadline_client,
+            config=None,
+            farm_id=farm_id,
+            queue_id=queue_id,
+        )
+        s3_settings: Optional[JobAttachmentS3Settings] = get_queue(
+            farm_id=farm_id, queue_id=queue_id
+        ).jobAttachmentSettings
+        if not s3_settings:
+            raise MissingJobAttachmentSettingsError(f"Queue {queue_id} has no attachment settings")
+
+        s3_root_uri = s3_settings.to_s3_root_uri()
+
+    if not s3_root_uri:
+        raise MissingJobAttachmentSettingsError("No valid s3 root path available")
+
+    attachment_api.attachment_upload(
+        root_dirs=root_dirs,
+        manifests=manifests,
+        s3_root_uri=s3_root_uri,
+        boto3_session=boto3_session,
+        path_mapping_rules=path_mapping_rules,
+        upload_manifest_path=upload_manifest_path,
+        logger=logger,
     )
--- a/src/deadline/job_attachments/api/__init__.py
+++ b/src/deadline/job_attachments/api/__init__.py
@@ -1,5 +1,5 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-__all__ = ["attachment_download"]
+__all__ = ["attachment_download", "attachment_upload"]
 
-from .attachment import attachment_download
+from .attachment import attachment_download, attachment_upload
--- a/src/deadline/job_attachments/api/attachment.py
+++ b/src/deadline/job_attachments/api/attachment.py
@@ -6,6 +6,8 @@ import json
 
 from contextlib import ExitStack
 from typing import Optional, List, Dict
+from pathlib import Path
+from dataclasses import asdict
 
 from deadline.job_attachments.asset_manifests.base_manifest import BaseAssetManifest
 from deadline.job_attachments.asset_manifests.decode import decode_manifest
@@ -13,13 +15,16 @@ from deadline.job_attachments.download import download_files_from_manifests
 from deadline.job_attachments.models import JobAttachmentS3Settings, PathMappingRule
 from deadline.job_attachments.progress_tracker import DownloadSummaryStatistics
 from deadline.client.exceptions import NonValidInputError
+from deadline.job_attachments.upload import S3AssetUploader
+from deadline.client.cli._groups.click_logger import ClickLogger
 
 
 def attachment_download(
     manifests: List[str],
-    s3_root_path: str,
+    s3_root_uri: str,
     boto3_session: boto3.Session,
     path_mapping_rules: Optional[str] = None,
+    logger: ClickLogger = ClickLogger(False),
 ):
     """
     BETA API - This API is still evolving but will be made public in the near future.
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by godobyte** [comment]
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=aws-deadline_deadline-cloud&pullRequest=461) **Quality Gate failed**
> 
> Failed conditions  [1 Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=aws-deadline_deadline-cloud&pullRequest=461&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [5.7% Duplication on New Code](https://sonarcloud.io/component_measures?id=aws-deadline_deadline-cloud&pullRequest=461&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)  [B Reliability Rating on New Code](https://sonarcloud.io/dashboard?id=aws-deadline_deadline-cloud&pullRequest=461) (required ≥ A)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=aws-deadline_deadline-cloud&pullRequest=461)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarLint](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Security Hotspot is not applicable, the `/tmp` folder is not created.

Will fix the reliability issue along with addressing comments.

File: src/deadline/job_attachments/api/attachment.py [hunk 1]
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+        # Uploads all files to a CAS in the manifest, optionally upload manifest file
+        key, data = asset_uploader.upload_assets(
+            job_attachment_settings=s3_settings,
+            manifest=manifest,
+            partial_manifest_prefix=upload_manifest_path,
+            source_root=Path(rule.source_path),
+            asset_root=Path(rule.destination_path),
+        )
+        logger.echo(
+            f"Uploaded assets from {rule.source_path}, to {s3_settings.to_s3_root_uri()} key {key}, hashed data {data}"
```
jericht:
```suggestion
            f"Uploaded assets from {rule.source_path}, to {s3_settings.to_s3_root_uri()}/{key}, hashed data {data}"
```

nit: technically the `{key}` part is the last part of the key. From the PR description, the `s3_settings.to_s3_root_uri` returns an S3 bucket + key prefix. I think we can just remove the `key` part from the message and just directly combine `{s3_settings.to_s3_root_uri()}/{key}` to log the full S3 URI. 

What do you think?

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+
+    path_mapping_rule_list.extend(
+        PathMappingRule(source_path_format="", source_path=root, destination_path=root)
+        for root in root_dirs
+    )
+
+    return path_mapping_rule_list
+
+
+def _get_file_name_manifest_dict(manifests: List[str]) -> Dict[str, BaseAssetManifest]:
```
jericht: nit: Can we rename the function or give it a docstring? Name could be more informative IMO. Maybe it can just be `read_manifests` since it is already taking in a list of manifest paths and return value is dict of paths and manifests.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
 
-    # path validation
-    if not all([os.path.isfile(manifest) for manifest in manifests]):
-        raise NonValidInputError(f"Specified manifests {manifests} contain invalid file.")
+    file_name_manifest_dict: Dict[str, BaseAssetManifest] = _get_file_name_manifest_dict(
+        manifests=manifests
+    )
+    path_mapping_rule_list: List[PathMappingRule] = _process_path_mapping(
+        path_mapping_rules=path_mapping_rules
+    )
```
jericht, line 49: nit on variable naming: maybe the parameter can be renamed to `path_mapping_rules_file`, then this variable with the processed rules can just be `path_mapping_rules`. How does that sound?

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+        # File name is supposed to be prefixed by a hash of source path in path mapping, use that to determine destination
+        # If it doesn't appear in path mapping or mapping doesn't exist, download to current directory instead
+        destination = next(
+            (
+                rule.destination_path
+                for rule in path_mapping_rule_list
+                if rule.get_hashed_source_path(manifest.get_default_hash_alg()) in file_name
+            ),
+            # Write to current directory partitioned by manifest name when no path mapping defined
+            f"{os.getcwd()}/{file_name}",
```
jericht, line 63: I think we should log a warning message in this case. I think it will be helpful since this is not a typical case. Maybe we can follow the same pattern we have in the `attachment_upload` loop where we do `None` as default and handle that accordingly.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+        )
+        # Assuming the manifest is already aggregated and correspond to a single destination
+        merged_manifests_by_root[destination] = manifest
```
jericht: Should we have error handling in case there is a collision? Even though the current assumption is that it is not possible, might be good idea to code defensively here.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+        manifests (List[str]): File Path to the manifest file for upload.
+        s3_root_uri (str): S3 root uri including bucket name and root prefix.
+        boto3_session (boto3.Session): Boto3 session for interacting with customer s3.
+        root_dirs (List[str]): List of root directories holding attachments. Defaults to empty.
+        path_mapping_rules (Optional[str], optional): Optional file path to a JSON file contains list of path mapping. Defaults to None.
+        upload_manifest_path (Optional[str], optional): Optional prefix for uploading given manifests. Defaults to None.
+        logger (ClickLogger, optional): Logger to provide visibility. Defaults to ClickLogger(False).
+
+    Raises:
+        NonValidInputError: _description_
```
jericht: empty _description_

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
 
     if path_mapping_rules:
         if not os.path.isfile(path_mapping_rules):
             raise NonValidInputError(
                 f"Specified path mapping file {path_mapping_rules} is not valid."
             )
-
         with open(path_mapping_rules, encoding="utf8") as f:
-            parsed_mappings = [PathMappingRule(**mapping) for mapping in json.load(f)]
+            path_mapping_rule_list = [PathMappingRule(**mapping) for mapping in json.load(f)]
```
jericht:
```suggestion
            path_mapping_rule_list.extend([PathMappingRule(**mapping) for mapping in json.load(f)])
```

Since we are initializing the list above now, we should extend here instead of replacing it with a new list

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+
+    if not all(os.path.isdir(root) for root in root_dirs):
+        raise NonValidInputError(f"Specified root dir {root_dirs} contains nonvalid directory.")
```
jericht: Could we improve the error message by saying which directories were nonvalid? Maybe we can change the condition to get a list of nonvalid dirs and use that here

Same comment for similar checks

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -35,23 +36,24 @@ def cli_attachment():
...
 @click.option(
+    "-m",
     "--manifests",
     multiple=True,
     required=True,
     help="Comma separated file paths to manifest formatted files.",
 )
 @click.option(
-    "--s3-root-path", help="Job Attachments S3 root path including bucket name and root prefix."
+    "--s3-root-uri", help="Job Attachments S3 root uri including bucket name and root prefix."
```
leongdl, line 46: nice! I like this, uri is a good type to use here. It can be generic like File or S3

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -85,14 +88,95 @@ def attachment_download(
...
+    )
+
+
+@cli_attachment.command(name="upload", help="BETA")
+@click.option(
+    "-m",
+    "--manifests",
+    multiple=True,
+    required=True,
+    help="Comma separated file paths to manifest formatted files.",
```
leongdl: Not comma separated now right? Since we support `multiple=True`

src/deadline/job_attachments/api/__init__.py
```diff
@@ -1,5 +1,5 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-__all__ = ["attachment_download"]
+__all__ = ["attachment_download", "attachment_upload"]
```
leongdl, line 3: Thanks! This is awesome.

leongdl: I'll add snapshot here separately.

src/deadline/job_attachments/models.py
```diff
@@ -230,9 +230,25 @@ def from_root_path(root_path: str) -> JobAttachmentS3Settings:
 
         return JobAttachmentS3Settings(path_split[0], "/".join(path_split[1:]))
 
+    @staticmethod
+    def from_s3_root_uri(uri: str) -> JobAttachmentS3Settings:
+
+        try:
+            bucket, key = uri.replace("s3://", "").split("/", maxsplit=1)
```
leongdl: Nit, urllib.parse will make this code work better:
https://docs.python.org/3/library/urllib.parse.html

```
ParseResult(scheme='scheme', netloc='netloc', path='/path;parameters', params='',
            query='query', fragment='fragment')
o = urlparse("http://docs.python.org:80/3/library/urllib.parse.html?"
             "highlight=params#url-parsing")
o
ParseResult(scheme='http', netloc='docs.python.org:80',
            path='/3/library/urllib.parse.html', params='',
            query='highlight=params', fragment='url-parsing')
```

leongdl (commented):
Mostly ok - need to take another pass at this

src/deadline/job_attachments/upload.py
```diff
@@ -141,12 +141,14 @@ def upload_assets(
         self,
         job_attachment_settings: JobAttachmentS3Settings,
         manifest: BaseAssetManifest,
-        partial_manifest_prefix: str,
         source_root: Path,
+        partial_manifest_prefix: Optional[str] = None,
         file_system_location_name: Optional[str] = None,
         progress_tracker: Optional[ProgressTracker] = None,
         s3_check_cache_dir: Optional[str] = None,
         manifest_write_dir: Optional[str] = None,
```
leongdl, line 149: nit, this isn't in the args description., can be a follow up.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+
+    path_mapping_rule_list.extend(
```
leongdl, line 185: I think its safer to have either root_dir OR path_mapping, not both. Thoughts? Having both can make the resolution very complicated if they somehow collide.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,156 @@ def attachment_download(
...
+    for file_name in file_name_manifest_dict:
+        manifest: BaseAssetManifest = file_name_manifest_dict[file_name]
+
+        # File name is supposed to be prefixed by a hash of source path in path mapping or provided root dirs
+        rule: Optional[PathMappingRule] = next(
+            # search in path mapping to determine source and destination
+            (
+                rule
+                for rule in path_mapping_rule_list
+                if rule.get_hashed_source_path(manifest.get_default_hash_alg()) in file_name
```
leongdl, line 133: This implies the file name of the manifest is always going to contain the hash. 

If we are doing this, we should also change snapshot to output the file name as ```{user-supplied-name}-{hash}-{timestamp}``` This will allow us to obfuscate the use of the hash to the root dir transparently. Right now, snapshot will NOT be easy to use with this attachment upload.

I can make this change in the morning to snapshot if we agree.

godobyte: This is symmetrical to how `attachment upload` maps path mapping rule to its manifest.

I think we would need the hash in the file name, otherwise the download input data cannot maintain this correlation. The caller could pass in the hash as supplied name, but I think having the hash explicitly in the name is good for consistency. This is important for the snapshot generated manifest files being used by upload and download out of the box.

src/deadline/job_attachments/api/attachment.py
godobyte: I don't have a strong preference here, depending on how customer would like to use them. If collision happens, the current code path would always takes path mapping rule file over directly passed dir. Maybe we can gather some feedback before we change?

src/deadline/job_attachments/upload.py
godobyte: This arg already exist, I only changed from required to optional with default. That's why the description didn't show up as diff.

src/deadline/job_attachments/api/attachment.py
leongdl: OK I will add it now.

src/deadline/job_attachments/api/attachment.py
leongdl: For forwards compatibility then an or is better, we can make it "and" later?

src/deadline/client/cli/_groups/attachment_group.py
godobyte: Sure I can reword the message.

src/deadline/job_attachments/api/attachment.py
godobyte: I'd imagine this is a normal case for customer workstation use case, who doesn't always pass in a path mapping file. 

For upload we handle `None` as an error since if the asset root is missing, the CLI can find no file to upload. It's different in this case since we don't ask for a directory to write the downloaded files too, while we do for the download case as `root-dirs`.

src/deadline/job_attachments/api/attachment.py
godobyte: It's a good call. This pattern has been used for more than one CLIs. The manifests also represent path(s) to manifest files. We'll be auditing all JA CLIs we're proposing, and rename them all if that's more intuitive.

src/deadline/job_attachments/api/attachment.py
godobyte: Sure, added a check here.

src/deadline/job_attachments/api/attachment.py
godobyte: TY!

src/deadline/job_attachments/api/attachment.py
godobyte: Added this check in the api prior calling this helper.

jericht (approved):

src/deadline/job_attachments/api/attachment.py
```diff
@@ -28,56 +33,186 @@ def attachment_download(
...
+        NonValidInputError: raise when any of the input is not valid.
     """
 
-    # path validation
-    if not all([os.path.isfile(manifest) for manifest in manifests]):
-        raise NonValidInputError(f"Specified manifests {manifests} contain invalid file.")
+    file_name_manifest_dict: Dict[str, BaseAssetManifest] = _read_manifests(manifests=manifests)
 
-    parsed_mappings: List[PathMappingRule] = list()
+    if bool(path_mapping_rules) == bool(root_dirs):
```
leongdl, line 115: thanks!

src/deadline/job_attachments/api/attachment.py
leongdl: thx, looks good on the validation above.

leongdl (approved):

PR #468, opened 2024-10-10
test: Squish test suite for Blender gui-submit dialogue

rena-cheng:

Preface: 
I am working on implementing an automated test suite using Squish framework to test the GUI interface of the Deadline GUI Submitter dialogues. The goal is to eventually achieve at least 95% automation coverage of the GUI interface. 

This PR focuses on testing the Blender gui-submitter dialogue using the `deadline bundle gui-submit --browse` command (where no user interface automation exists right now). It utilizes test files/job bundles grabbed from the [deadline-cloud-samples repo](https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/job_bundles) (rather than creating my own for the time being) as we need existing job bundles to open/test these dialogues, and these files from the deadline-cloud-samples are already vetted/public (so I am not planning to create any of my own at the moment). These files must live in the same repo as the Squish tests as the tests need to pull those files while running (both locally and CI/CD). 

I identified four test cases for the Blender GUI Submit dialogue utilizing the "Blender Render" file from [deadline-cloud-samples repo](https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/job_bundles). Those test cases contain numerous verifications (with a total of 43 tests/verifications). 

As I have several test cases and Squish follows a framework where tests are only run in a single "main()" block, each test case has been organized into a single function/line in the main() block, in which if tests need to be disabled, they can be easily commented out. Additionally, this allows us to easily organize the test suites efficiently (so that we don't have dozens of separate test suites running the same GUI interface. This follows E2E test fundamentals you see in other frameworks where "it" blocks allow you to run multiple test cases in a single file (which Squish doesn't allow and hence each test case being a single line of code in main()) -- hopefully that makes sense.

The four test cases I decide to automate (and more can be added):
1. Verify defaults are correct on shared job settings tab based on selected Blender job bundle.
2. Verify defaults are correct on job-specific settings tab based on selected Blender job bundle.
3. Verify a few tooltip texts (and that they contain correct text). 
4. Verify 'Load a different job bundle' button works. 

Additionally, to implement/write this suite, I followed the similar/same test conventions I had written for the `deadline config gui` tests. This includes the following:
- Authenticating using default AWS profile in test setup, and test cleanup (this also includes clearing out any refresh error dialogues)
- Renaming all pre-named locators created by Squish, as they are less-user friendly. (This amounted to having to rename probably over 100 locators, again).

Note that the `deadline config gui` tests (which is in the `tst_VerifySettingsDialogue` folder) does not currently follow the above convention for the main() block, however, that is due to that test being one test case itself (and more test cases will need to be added). 

Currently none as the tests cannot be run in a CI/CD pipeline, and you will need a Squish licenses in order to run these tests. 

Running locally on my Linux Workstation. 

See attached logs. 

Screenshot 2024-10-15 at 9 45 14 PM
Screenshot 2024-10-15 at 9 45 41 PM

- Have you run the unit tests? N/A

Yes. 

No. 

No. 

----

### DIFF
--- a/test/squish/deadline_cloud_samples/blender_render/template.yaml
+++ b/test/squish/deadline_cloud_samples/blender_render/template.yaml
@@ -1,0 +1,105 @@
+specificationVersion: 'jobtemplate-2023-09'
+name: Blender Render
+
+parameterDefinitions:
+
+# Render Parameters
+- name: BlenderSceneFile
+  type: PATH
+  objectType: FILE
+  dataFlow: IN
+  userInterface:
+    control: CHOOSE_INPUT_FILE
+    label: Blender Scene File
+    groupLabel: Render Parameters
+    fileFilters:
+    - label: Blender Scene Files
+      patterns: ["*.blend"]
+    - label: All Files
+      patterns: ["*"]
+  description: >
+    Choose the Blender scene file you want to render. Use the 'Job Attachments' tab
+    to add textures and other files that the job needs.
+- name: Frames
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Frames
+    groupLabel: Render Parameters
+  default: 1-10
+- name: OutputDir
+  type: PATH
+  objectType: DIRECTORY
+  dataFlow: OUT
+  userInterface:
+    control: CHOOSE_DIRECTORY
+    label: Output Directory
+    groupLabel: Render Parameters
+  default: "./output"
+  description: Choose the render output directory.
+- name: OutputPattern
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Output File Pattern
+    groupLabel: Render Parameters
+  default: "output_####"
+  description: Enter the output filename pattern (without extension).
+- name: Format
+  type: STRING
+  userInterface:
+    control: DROPDOWN_LIST
+    label: Output File Format
+    groupLabel: Render Parameters
+  description: Choose the file format to render as.
+  default: JPEG
+  allowedValues: [TGA, RAWTGA, JPEG, IRIS, IRIZ, PNG, HDR, TIFF, OPEN_EXR, OPEN_EXR_MULTILAYER, CINEON, DPX, DDS, JP2, WEBP]
+
+# Software Environment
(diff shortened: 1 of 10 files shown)


test/squish/deadline_cloud_samples/blender_render/template.yaml
```diff
@@ -0,0 +1,105 @@
+specificationVersion: 'jobtemplate-2023-09'
```
leongdl, line 1: Can / Should this be in a different repository?

rena-cheng: I thought about it but the tests need to be able to pull/use these files when running (both CI/CD and locally) and AFAIK it would have to live in the same repository as the code itself. 

Note that I am pulling test files from [deadline-cloud-samples](url) repo -- I am mainly doing this because I don't know if there is some standard for test file creation before we offer them to our users (and I am also worried about open source contributors modifying existing test files causing security issues), and everything in deadline-cloud-samples is already public and vetted already.

leongdl: Yeah, and this is why I'm asking because we now have blender related test assets / files in different places. Would a DCC agnostic test make more sense? I am not sure how we can make it "generic"

rena-cheng: Off the top of my head, we would create a "generic file" that could reflect "generic" settings that exist in all DCCs (such as frame range, file output, output directory, etc). 

As far as how to create that file, I'm not entirely sure (as test files are usually created in the DCC itself), but I would need to create a template.yaml file itself (which is the job bundle being used).  

My reasoning (which I'm sure is obvious) to automate the different DCC dialogues is to provide test coverage as the DCC dialogues have some differences between each other (such as conda package versions). I would worry about losing test coverage in creating a "generic" test/test file. But maybe there's a way to still test the different specific settings? **Thinking on this**

rena-cheng:
I will be closing this PR in favor of a new PR as I've taken this suite in a new direction since then (where I have written brand new job bundles).

PR #471, opened 2024-10-11
feat(JA): Add manifest upload and download to complete the JA standalo…

leongdl:
- Customers have requested APIs and CLI interfaces to use Job Attachments.
- We designed  new CLI + APIs:
   - manifest snapshot
   - manifest diff
   - manifest download
   - manifest upload
   - attachment upload
   - attachment download.

- In this PR, we are adding the implementation for `manifest upload` and `manifest download`

- The new CLIs + APIs will allow users to directly upload and download manifests to and from the CAS.
- `manifest upload` will upload a snapshot manifest file to the CAS manifest directory structure. 
- `manifest download` will download a manifest file for a job (all inputs), or job and task (all inputs and outputs), where manifests are aggregated by the root path.

    - Yes, `hatch run test` passes 100%
```
=================================================================== 1381 passed, 51 skipped, 1 xfailed, 2 warnings in 22.11s ===================================================================
```

    - Yes, `hatch run integ:test` passes 100%
```
===================================== 24 passed, 2 skipped, 5 warnings in 72.33s (0:01:12) ======================================
```

   - No, this CR only uploads and download manifests. It re-uses existing code paths without modification.

   - Yes, new APIs and CLIs have been documented.

   - WIP, pending implementation and sign off.

- No.

- Yes, we are working on threat modelling and release.

----


src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    manifest_file: str,
+    s3_bucket_name: str,
+    manifest_path: str,
+    s3_metadata: Dict[str, Any],
+    boto_session: boto3.Session,
+    logger: ClickLogger = ClickLogger(False),
+):
+    """
+    BETA API - This API is still evolving but will be made public in the near future.
+    API to upload a job attachment manifest to the Content Address Storage. Manifests will be
```
jericht: nit: should this be `Content Addressable Storage` to be consistent with CLI help text/docstring?
```suggestion
    API to upload a job attachment manifest to the Content Addressable Storage. Manifests will be
```

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
             process_output(fast_diff_item[1], fast_diff_item[0], output)
 
     return output
+
+
+def _manifest_upload(
+    manifest_file: str,
+    s3_bucket_name: str,
+    manifest_path: str,
```
jericht: nit: at first glance, this looks like it should be a local file path, but it's really an S3 key prefix for the asset root in the S3 bucket, right? How about renaming to something like `s3_key_prefix`?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    uploaded to s3://{s3_bucket_name}/Manifest/{manifest_path} as per the Deadline CAS folder structure.
+    manifest_file: File Path to the manifest file for upload.
+    s3_bucket_name: S3 bucket name.
+    manifest_path: S3 path for the manifest.
+    s3_metadata: Additional S3 file metadata tagged on upload.
+    boto_session: Boto3 session.
+    logger: Click Logger instance to print to CLI as test or JSON.
+    """
+    # Always upload the manifest file to case root /Manifest with the original file name.
+    manifest_path = manifest_path + "/Manifests/" + Path(manifest_file).name
```
jericht: Should we reuse the same variable for `Manifests` folder here as well?

```suggestion
    manifest_path = "/".join([manifest_path, S3_MANIFEST_FOLDER_NAME, Path(manifest_file).name])
```

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -228,28 +234,98 @@ def manifest_download(
...
+        farm_id = config_file.get_setting("defaults.farm_id", config=config)
+        queue_id = config_file.get_setting("defaults.queue_id", config=config)
+
+        deadline = api.get_boto3_client("deadline", config=config)
+        queue = deadline.get_queue(
+            farmId=farm_id,
+            queueId=queue_id,
+        )
+        bucket_name = queue["jobAttachmentSettings"]["s3BucketName"]
+        manifest_path = queue["jobAttachmentSettings"]["rootPrefix"]
```
godobyte: We should utilize `JobAttachmentS3Settings` class to extract bucket name and prefix here.

```
s3_settings: Optional[JobAttachmentS3Settings] = get_queue(
    farm_id=farm_id, queue_id=queue_id
).jobAttachmentSettings
```

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -228,28 +234,98 @@ def manifest_download(
...
+            farm_id=farm_id,
+            queue_id=queue_id,
+            queue_display_name=queue["displayName"],
+        )
+
+    else:
+        # Self supplied cas path.
+        url_fragments = urlparse(s3_cas_uri)
+        bucket_name = url_fragments.netloc
+        manifest_path = url_fragments.path
```
godobyte: We could reuse existing function to parse the uri.

```suggestion
        s3_settings: JobAttachmentS3Settings = JobAttachmentS3Settings.from_s3_root_uri(s3_root_uri)
        bucket_name = s3_settings.s3BucketName
        root_prefix = s3_settings.rootPrefix
```

If this uri contains more than bucket name and root prefix, maybe the path logic can be added to the class.

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    # Get input_manifest_paths from Deadline GetJob API
+    job: dict = deadline.get_job(farmId=farm_id, queueId=queue_id, jobId=job_id)
+    attachments: dict = job["attachments"]
+    input_manifest_paths: List[Tuple[str, str]] = [
+        (manifest["inputManifestPath"], manifest["rootPath"])
+        for manifest in attachments["manifests"]
+    ]
+
+    # Download each input_manifest_path
+    for input_manifest in input_manifest_paths:
```
jericht: nit: for readability, it would be nice if we split up the tuple into individual variables here
```suggestion
    for input_manifest_path, root_path in input_manifest_paths:
```

src/deadline/job_attachments/api/manifest.py
godobyte: Could you check `JobAttachmentS3Settings` class, there should be function to construct Manifest path.

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+
+    # Save the manifest files to disk.
+    for root in merged_manifests.keys():
+        # Save the merged manifest as {root}_{hash}_timestamp.
+        root_hash: str = hash_data(
+            root.encode("utf-8"), merged_manifests[root].get_default_hash_alg()
+        )
+        timestamp = datetime.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
+        manifest_name = root.replace("/", "_")
+        manifest_name = manifest_name[1:] if manifest_name[0] == "_" else manifest_name
```
jericht, line 437: nit: can use built-in str function for this too
```suggestion
        manifest_name = manifest_name.removeprefix("_")
```

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    logger: ClickLogger = ClickLogger(False),
+):
+    """
+    BETA API - This API is still evolving but will be made public in the near future.
+    API to upload a job attachment manifest to the Content Address Storage. Manifests will be
+    uploaded to s3://{s3_bucket_name}/Manifest/{manifest_path} as per the Deadline CAS folder structure.
+    manifest_file: File Path to the manifest file for upload.
+    s3_bucket_name: S3 bucket name.
+    manifest_path: S3 path for the manifest.
+    s3_metadata: Additional S3 file metadata tagged on upload.
```
godobyte: This is inconsistent with the command, I found it awkward from AWS CLI/SDK perspective since they're always consistent. Thoughts?

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -228,28 +234,98 @@ def manifest_download(
...
 
 @cli_manifest.command(
     name="upload",
-    help="BETA - Uploads a job attachment manifest file to a Content Addressable Storage's Manifest store. If calling via --s3-cas-path, it is recommended to use with --profile for a specific AWS profile with CAS S3 bucket access.",
+    help="BETA - Uploads a job attachment manifest file to a Content Addressable Storage's Manifest store. If calling via --s3-cas-path, it is recommended to use with --profile for a specific AWS profile with CAS S3 bucket access. Check exit code for success or failure.",
 )
 @click.argument("manifest_file")
 @click.option("--profile", help="The AWS profile to use.")
-@click.option("--s3-cas-path", help="The path to the Content Addressable Storage root.")
+@click.option("--s3-cas-uri", help="The URI to the Content Addressable Storage S3 bucket and root.")
```
godobyte, line 272: From reading the code below, this uri contains more than bucket name and root prefix?

test/integ/cli/test_cli_manifest_upload.py
```diff
@@ -0,0 +1,140 @@
...
+        manifest_s3_path = f"/DeadlineCloud/Manifests/{manifest_file_name}"
+        s3_client = boto3.client("s3")
+        s3_client.head_object(Bucket=s3_bucket, Key=manifest_s3_path)
+
+        # Cleanup.
+        s3_client.delete_object(Bucket=s3_bucket, Key=manifest_s3_path)
+
+    def test_manifest_upload_by_farm_queue(self, temp_dir):
+        """
+        Simple test to generate a manifest, and then call the upload CLI to uplaod to S3.
```
jericht:
```suggestion
        Simple test to generate a manifest, and then call the upload CLI to upload to S3.
```

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -228,28 +234,98 @@ def manifest_download(
...
+
+    else:
+        # Self supplied cas path.
+        url_fragments = urlparse(s3_cas_uri)
+        bucket_name = url_fragments.netloc
+        manifest_path = url_fragments.path
+
+    logger.echo(f"Uploading Manifest to {bucket_name} {manifest_path}")
+    _manifest_upload(manifest_file, bucket_name, manifest_path, metadata, session, logger)
+    logger.echo("Uploading successful!")
```
godobyte: There is no additional logging for json format?

test/integ/cli/test_cli_manifest_upload.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+        assert result.exit_code == 0, f"Non-Zeo exit code, CLI output {result.output}"
+
+        # Then validate the Manifest file is uploaded to S3 by checking the file actually exists.
+        manifest_s3_path = f"DeadlineCloud/Manifests/{manifest_file_name}"
+        s3_client = boto3.client("s3")
+        try:
+            s3_client.head_object(Bucket=s3_bucket, Key=manifest_s3_path)
+        except ClientError:
+            assert 0, f"File not found at {s3_bucket}, {manifest_s3_path}"
```
jericht: nit: we could also just use `pytest.fail` here to always fail: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-fail

jericht (dismissed):
Just some non-blocking comments, otherwise LGTM!

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    boto_session: Boto3 session.
+    step_id: Optional[str]: Optional, download manifest for a step
+    logger: Click Logger instance to print to CLI as test or JSON.
+    return ManifestDownloadResponse Downloaded Manifest data. Contains source S3 key and local download path.
+    """
+
+    # Deadline Client and get the Queue to download.
+    deadline = boto3_session.client("deadline", config=get_default_client_config())
+
+    queue: dict = deadline.get_queue(
```
godobyte, line 334: This returns `Queue` no?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+        queueId=queue_id,
+    )
+
+    # assume queue role - session permissions
+    queue_role_session: boto3.Session = _get_queue_user_boto3_session(
+        deadline=deadline,
+        base_session=boto3_session,
+        farm_id=farm_id,
+        queue_id=queue_id,
+        queue_display_name=queue["displayName"],
```
godobyte, line 345: Could be `queue.displayName`, same as other queue attribute data access.

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+
+    def add_manifest_by_root(
+        manifests_by_root: Dict[str, list], root: str, manifest: BaseAssetManifest
+    ):
+        if root not in manifests_by_root:
+            manifests_by_root[root] = []
+        manifests_by_root[root].append(manifest)
+
+    # Get input_manifest_paths from Deadline GetJob API
+    job: dict = deadline.get_job(farmId=farm_id, queueId=queue_id, jobId=job_id)
```
godobyte, line 368: `get_job` returns `Job` type. Data unboxing could directly use the job data class definition rather than dict key access.

godobyte (changes requested):

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    attachments: dict = job["attachments"]
+    input_manifest_paths: List[Tuple[str, str]] = [
+        (manifest["inputManifestPath"], manifest["rootPath"])
+        for manifest in attachments["manifests"]
+    ]
```
godobyte: There should be defined manifest `ManifestProperties`  from model.

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,182 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+        # Save the merged manifest as {root}_{hash}_timestamp.
+        root_hash: str = hash_data(
+            root.encode("utf-8"), merged_manifests[root].get_default_hash_alg()
+        )
+        timestamp = datetime.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
+        manifest_name = root.replace("/", "_")
+        manifest_name = manifest_name[1:] if manifest_name[0] == "_" else manifest_name
+        manifest_name = f"{manifest_name}-{root_hash}-{timestamp}.manifest"
+
+        local_manifest_file_path = os.path.join(download_dir, manifest_name)
```
godobyte, line 440: Could we reuse some code from here?

https://github.com/aws-deadline/deadline-cloud/blob/2f4ceb6128af89241b9005a2a1df9c40e3f7ab13/src/deadline/job_attachments/api/manifest.py#L169-L180

leongdl:
> Just some non-blocking comments, otherwise LGTM!

Thanks! I'll get on them.

src/deadline/client/cli/_groups/manifest_group.py
leongdl: How so? I only use the bucket and prefix from it.

src/deadline/client/cli/_groups/manifest_group.py
leongdl: Good idea - will clean it up w/ the data structures.

src/deadline/client/cli/_groups/manifest_group.py
leongdl: Let me try, I think that will work.

src/deadline/client/cli/_groups/manifest_group.py
leongdl: Nope, success / fail all prints out to the output root: manifest file.

src/deadline/job_attachments/api/manifest.py
leongdl: Great idea, will rename. Yes, this is the `s3_key_prefix` (which we join with the CAS root)

src/deadline/job_attachments/api/manifest.py
leongdl: Agreed!

src/deadline/job_attachments/api/manifest.py
leongdl: What do you mean? Will check with you offline

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -228,28 +234,98 @@ def manifest_download(
     """
     Downloads input manifest of previously submitted job.
     """
-    raise NotImplementedError("This CLI is being implemented.")
+    logger: ClickLogger = ClickLogger(is_json=json)
+    if not os.path.isdir(download_dir):
+        raise NonValidInputError(f"Specified destination directory {download_dir} does not exist. ")
+
+    # setup config
+    config = _apply_cli_options_to_config(required_options={"farm_id", "queue_id"}, **args)
```
jairaws: nit: Would be better to type it as `Optional[ConfigParser]`.

jairaws (dismissed):

jairaws (commented):
The test coverage is below the required 80%.

src/deadline/client/cli/_groups/manifest_group.py
godobyte: From L327 and L330 below, seems the manifest path is also included?

src/deadline/client/cli/_groups/manifest_group.py
leongdl: sure thing.

src/deadline/client/cli/_groups/manifest_group.py
leongdl: Actually no that won't work, think of this case:

`s3://bucketname/prefix1/prefix2/some/path/i/want`

it is hard to capture `prefix1` and `prefix2` separate from `/some/path/i/want`

I'll just migrate it to the shared function.

src/deadline/client/cli/_groups/manifest_group.py
leongdl: See below, it is just the bucket and root.

src/deadline/job_attachments/api/manifest.py
leongdl: Great idea

leongdl: Yeah, this is literally
```
def _join_s3_paths(root: str, *args: str):
    return "/".join([root, *args])
```

It has little to no value! I generally do not use functions that just adds indirection. If it ever gets modified the blast radius is big. If there is an actual if / else / loop we can consider.

src/deadline/job_attachments/api/manifest.py
leongdl: Nope, this is a boto3 client as per 322, so its just returning a `dict`, not strongly typed.

src/deadline/job_attachments/api/manifest.py
leongdl: This is a dict, and not the `get_queue` helper.

If the function does need all of the other properties, then we can use `get_queue` with the `Queue` structure. Just saving some code when not more than 50% is necessary

src/deadline/job_attachments/api/manifest.py
leongdl: Yes, but have you seen how much code `get_job` runs! 

https://github.com/aws-deadline/deadline-cloud/blob/61fb59581cd4c5d548077612721f7ee1b7527049/src/deadline/job_attachments/_aws/deadline.py#L67-L108

src/deadline/job_attachments/api/manifest.py
leongdl: Same as above, not using these utilities.

src/deadline/job_attachments/api/manifest.py
leongdl: Great idea! (one of these pythonic things I'll never fully get used to)

src/deadline/job_attachments/api/manifest.py
leongdl: Done!

src/deadline/job_attachments/api/manifest.py
leongdl: Yes and No - I was thinking about it but I am un-decided about the downloaded name. I used the same name for now for consistency, but was also thinking about tagging them differently like "aggregated".

If no one has suggestions after the test session I can refactor.

test/integ/cli/test_cli_manifest_upload.py
leongdl: done!

test/integ/cli/test_cli_manifest_upload.py
leongdl: sure

src/deadline/job_attachments/api/manifest.py
godobyte: What do you mean, I thought you've already calling this function?

godobyte: Code reuse brings benefits in terms of efficiency, consistency, and maintainability. The additional code is boxing the data and error handling, which can be useful here unless we want to directly surface the error. For example, `attachments` is not required in GetJobResponse, L362 could lead to KeyError, same goes for `inputManifestPath` in L364.

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,184 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    s3_bucket_name: str,
+    s3_key_prefix: str,
+    s3_metadata: Dict[str, Any],
+    boto_session: boto3.Session,
+    logger: ClickLogger = ClickLogger(False),
+):
+    """
+    BETA API - This API is still evolving but will be made public in the near future.
+    API to upload a job attachment manifest to the Content Addressable Storage. Manifests will be
+    uploaded to s3://{s3_bucket_name}/{root_prefix}/Manifest/{s3_key_prefix} as per the Deadline CAS folder structure.
```
godobyte: From 
https://github.com/aws-deadline/deadline-cloud/blob/b307b21e6a4cd04025d1ead8d8f232114be06983/src/deadline/client/cli/_groups/manifest_group.py#L316-L317

and 
https://github.com/aws-deadline/deadline-cloud/blob/b307b21e6a4cd04025d1ead8d8f232114be06983/src/deadline/client/cli/_groups/manifest_group.py#L333-334

s3_key_bucket name is bucket name, s3_key_prefix is root prefix, where is this s3_key_prefix from? Did I miss anything?

godobyte: This is also related to my confusion in an earlier comment 

https://github.com/aws-deadline/deadline-cloud/pull/471/files#r1809364159

src/deadline/job_attachments/api/manifest.py
godobyte: I meant this s3_metadata is not available via the command, while required for the API. I thought our plan is to have feature parity between the CLI and API, is that no longer true? I'm free if you'd like to sync offline :)

src/deadline/job_attachments/api/manifest.py
leongdl: Cannot - this is not Py 3.8 compatible.

test/integ/cli/conftest.py
```diff
@@ -0,0 +1,171 @@
...
+from unittest.mock import MagicMock
+import pytest
+from deadline.job_attachments import upload
+from deadline.job_attachments._aws.deadline import get_queue
+from deadline.job_attachments.asset_manifests.hash_algorithms import HashAlgorithm, hash_file
+from deadline.job_attachments.asset_manifests.versions import ManifestVersion
+from .test_utils import JobAttachmentTest, UploadInputFilesOneAssetInCasOutputs
+
+
+@pytest.fixture(scope="session")
```
leongdl, line 12: This code is refactored from the old `sync_output` integration test. Eventually, we should delete the old sync_output test. It has no dependencies to the old test implementation so we can cleanly remove the old tests.

leongdl: This includes the test data below, once our migration is complete, we have an easy clean cut.

test/integ/conftest.py
```diff
@@ -10,3 +11,144 @@ def external_bucket() -> str:
     Return a bucket that all developers and test accounts have access to, but isn't in the testers account.
     """
     return os.environ.get("INTEG_TEST_JA_CROSS_ACCOUNT_BUCKET", "job-attachment-bucket-snipe-test")
+
+
+@pytest.fixture(scope="session")
+def default_job_template() -> str:
```
leongdl, line 17: Moved from below, now sharable for all JA tests!

test/integ/conftest.py
```diff
@@ -10,3 +11,144 @@ def external_bucket() -> str:
...
+                        ],
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.fixture(scope="session")
+def default_job_template_step_step_dependency() -> str:
```
leongdl, line 70: New step-step dependency template.

src/deadline/job_attachments/api/manifest.py
leongdl: Ah - ok this is a documentation error. I will change this.

src/deadline/job_attachments/api/manifest.py
leongdl: Got it - yes I will move s3_metadata tagging from the CLI into the API.

src/deadline/job_attachments/api/manifest.py
leongdl: But isn't that case regardless? All the code has done is add an error message instead of where it actually failed. Overtime, as get_job changes, it runs more and more code that may or may not have tangible benefits. 

If it was a simple typed dict class I would be more interested. No better or worse solution here.

test/integ/cli/test_utils.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+
+from deadline.job_attachments._aws.deadline import get_queue
+from deadline.job_attachments.asset_manifests.hash_algorithms import HashAlgorithm, hash_file
+from deadline.job_attachments.asset_manifests.versions import ManifestVersion
+from deadline.job_attachments.models import Attachments, JobAttachmentS3Settings
+from deadline_test_fixtures.deadline import DeadlineClient
+
+
+class JobAttachmentTest:
```
leongdl, line 16: All re-factored code from the general integ test. This has no dependencies to the existing tests and it will allow us a clean cut when we delete the old `sync_input` and `sync_output` code.

src/deadline/job_attachments/api/manifest.py
leongdl: OK now i remember - I was in the middle of adding this,  `s3://{s3_bucket_name}/{cas_prefix}/Manifest/{s3_key_prefix}/{manifest_file_name}`

The CAS `Manifest` folder can have subfolder paths, eg job/step/task that we use today.

This is now properly fixed!

src/deadline/job_attachments/api/manifest.py
leongdl: Chatted offline, I am against using helpers that swallow up the real error under the hood. Either we use generate boto3 stubs, or none at alll.

src/deadline/job_attachments/api/manifest.py
leongdl: Moved the [] accessors to .get() with default handling as discussed offline.

jericht (approved):

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -228,28 +238,113 @@ def manifest_download(
...
+        deadline = api.get_boto3_client("deadline", config=config)
+        queue = deadline.get_queue(
+            farmId=farm_id,
+            queueId=queue_id,
+        )
+        queue_ja_settings: JobAttachmentS3Settings = JobAttachmentS3Settings(
+            **queue["jobAttachmentSettings"]
+        )
+        bucket_name = queue_ja_settings.s3BucketName
+        cas_path = queue_ja_settings.rootPrefix
```
godobyte, line 320: nit - is there specific reasons to rename `rootPrefix` to cas_pass? Could we directly access them via JobAttachmentS3Settings instance?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,191 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
             process_output(fast_diff_item[1], fast_diff_item[0], output)
 
     return output
+
+
+def _manifest_upload(
+    manifest_file: str,
+    s3_bucket_name: str,
+    s3_cas_prefix: str,
```
godobyte, line 268: WDYT about using `s3_cas_uri` to be consistent with the CLI option?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,191 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    manifest_file: File Path to the manifest file for upload.
+    s3_bucket_name: S3 bucket name.
+    boto_session: S3 Content Addressable Storage prefix.
+    s3_key_prefix: [Optional] S3 prefix path to the Content Addressable Storge.
+    boto_session: Boto3 session.
+    logger: Click Logger instance to print to CLI as test or JSON.
+    """
+    # S3 metadata
+
+    # Upload settings:
```
godobyte, line 286: super nit -  is this intended?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -247,3 +260,191 @@ def process_output(status: FileStatus, path: str, output_diff: ManifestDiff):
...
+    # Always upload the manifest file to case root /Manifest with the original file name.
+    manifest_path: str = "/".join(
+        [s3_cas_prefix, S3_MANIFEST_FOLDER_NAME, s3_key_prefix, Path(manifest_file).name]
+        if s3_key_prefix
+        else [s3_cas_prefix, S3_MANIFEST_FOLDER_NAME, Path(manifest_file).name]
+    )
+
+    # S3 uploader.
+    upload = S3AssetUploader(session=boto_session)
+    with open(manifest_file) as manifest:
```
godobyte, line 299: This API doesn't have the path check like in the CLI, is this intended?

src/deadline/job_attachments/models.py
```diff
@@ -399,16 +399,15 @@ class ManifestDiff:
 class ManifestDownload:
     """Data structure to store the S3 and local paths of a manifest"""
 
-    s3_key: str = field(default_factory=str)
-    local: str = field(default_factory=str)
+    manifest_root: str = field(default_factory=str)
+    local_manifest_path: str = field(default_factory=str)
```
godobyte, line 403: This looks like a breaking change, should the commit be `feat(JA)!` instead of `feat(JA)`?

src/deadline/client/cli/_groups/manifest_group.py
leongdl: None, purely a variable naming thing.

src/deadline/job_attachments/api/manifest.py
leongdl: No - because at the API level this is split to bucket name and cas prefix. This small difference is due to customers not always using Queue Job Attachment settings. Programatically, they are 2 things, just like the Queue settings's attributes.

src/deadline/job_attachments/api/manifest.py
leongdl: ? What is intended? The comment? Yes

src/deadline/job_attachments/api/manifest.py
leongdl: Yeah, let it fail with normal file errors, this is an API afterall. Why wrap errors when the `open` error is very specific of what is wrong. I don't like wrapping errors :)

src/deadline/job_attachments/models.py
leongdl: No, because this model is not used anywhere until this PR.

godobyte (approved):
Synced offline, the comments are not blocking. Let's ship it as beta, gather user feedback and improve before release.

PR #483, opened 2024-10-26
feat!: refactor manifest aggregation, add helper to persist manifests and check disk capacity

godobyte:
`attachment download` takes in list of manifest paths and download based on the inputs. 

Added helper and refactor code in preparation for worker agent integration for sync input.

AssetSync now has helper to prepare files for running `attachment download`

No

No

----

### DIFF
--- a/src/deadline/job_attachments/asset_sync.py
+++ b/src/deadline/job_attachments/asset_sync.py
@@ -166,7 +166,7 @@ class AssetSync:
                 f"No path mapping rule found for the source path {manifest_properties.rootPath}"
             )
 
-    def aggregate_asset_root_manifests(
+    def _aggregate_asset_root_manifests(
         self,
         session_dir: Path,
         s3_settings: JobAttachmentS3Settings,
@@ -176,7 +176,7 @@ class AssetSync:
         step_dependencies: Optional[list[str]] = None,
         dynamic_mapping_rules: dict[str, PathMappingRule] = {},
         storage_profiles_path_mapping_rules: dict[str, str] = {},
-    ) -> DefaultDict[str, list[BaseAssetManifest]]:
+    ) -> dict[str, BaseAssetManifest]:
         """
          Args:
             session_dir: the directory that the session is going to use.
@@ -187,7 +187,7 @@ class AssetSync:
             step_dependencies: the list of Step IDs whose output should be downloaded over the input job attachments.
             dynamic_mapping_rules: manifest root path to worker host destination mapping relative to local session.
             storage_profiles_path_mapping_rules: manifest root path to worker host destination mapping given storage profile.
-        Returns: a dictionary of manifest files stored in the session directory.
+        Returns: a dictionary of manifest file stored in the session directory.
         """
         grouped_manifests_by_root: DefaultDict[str, list[BaseAssetManifest]] = DefaultDict(list)
 
@@ -225,7 +225,15 @@ class AssetSync:
                     local_root = str(session_dir.joinpath(dir_name))
                     grouped_manifests_by_root[local_root].extend(manifests)
 
-        return grouped_manifests_by_root
+        # Merge the manifests in each root into a single manifest
+        merged_manifests_by_root: dict[str, BaseAssetManifest] = dict()
+        for root, manifests in grouped_manifests_by_root.items():
+            merged_manifest = merge_asset_manifests(manifests)
+
+            if merged_manifest:
+                merged_manifests_by_root[root] = merged_manifest
+
+        return merged_manifests_by_root
 
     def _launch_vfs(
         self,
@@ -242,7 +250,7 @@ class AssetSync:
             fs_permission_settings: An instance defining group ownership and permission modes
                                     to be set on the downloaded (synchronized) input files and directories.
             merged_manifests_by_root: Merged manifests produced by
-                                    aggregate_asset_root_manifests()
+                                    _aggregate_asset_root_manifests()
         Returns: None
         Raises: VFSExecutableMissingError If VFS is not startable.
         """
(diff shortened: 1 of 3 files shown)


src/deadline/job_attachments/asset_sync.py
```diff
@@ -166,7 +166,7 @@ def get_local_destination(
                 f"No path mapping rule found for the source path {manifest_properties.rootPath}"
             )
 
-    def aggregate_asset_root_manifests(
+    def _aggregate_asset_root_manifests(
```
mwiebe, line 169: This is changing the public interface of the library. If you do this, your commit and CR need to be `feat!` instead of `feat`, and you need to describe what the backwards incompatible change is.

src/deadline/job_attachments/asset_sync.py
```diff
@@ -225,7 +225,15 @@ def aggregate_asset_root_manifests(
                     local_root = str(session_dir.joinpath(dir_name))
                     grouped_manifests_by_root[local_root].extend(manifests)
 
-        return grouped_manifests_by_root
+        # Merge the manifests in each root into a single manifest
+        merged_manifests_by_root: dict[str, BaseAssetManifest] = dict()
+        for root, manifests in grouped_manifests_by_root.items():
+            merged_manifest = merge_asset_manifests(manifests)
+
+            if merged_manifest:
```
mwiebe, line 233: Under what condition is this not true? If it's when `manifests` is an empty list, I think it's more clear to put the if higher up:

```
if len(manifests) > 0:
    merged_manifests_by_root[root] = merge_asset_manifests(manifests)
```

src/deadline/job_attachments/asset_sync.py
```diff
@@ -313,6 +326,41 @@ def copied_download(
             else:
                 raise
 
+    def _check_and_write_local_manifests(
+        self, merged_manifests_by_root: dict[str, BaseAssetManifest], manifest_write_dir: str
+    ) -> list[str]:
+        """Write manifests to the session directory and check disk capacity is sufficient for the assets
```
mwiebe: nit: I think this function could write manifests in locations other than the session directory, right? The location `manifest_write_dir` is a parameter. The docstring is a bit confusing because of that, I was looking for where it gets the session directory.

src/deadline/job_attachments/asset_sync.py
```diff
@@ -313,6 +326,41 @@ def copied_download(
...
+
+        Args:
+            merged_manifests_by_root (dict[str, BaseAssetManifest]): manifest file to its stored root.
+            session_dir (str): local directory to write to
```
mwiebe: nit: Parameter name is incorrect here.

src/deadline/job_attachments/asset_sync.py
godobyte: Good catch! The caller I had in mind was meant for session dir, then I realized this can be generalized for any dir. Updated the function without the doc, will push a fix for this.

src/deadline/job_attachments/asset_sync.py
godobyte: Same as previous, will push a fix for this.

src/deadline/job_attachments/asset_sync.py
leongdl: Seems to have come from the `asset_sync` code at the bottom of `upload.py` which is now deleted.

src/deadline/job_attachments/asset_sync.py
```diff
@@ -313,6 +326,41 @@ def copied_download(
...
+                manifest=manifest,
+            )
+
+            total_input_size += manifest.totalSize  # type: ignore[attr-defined]
+            manifest_paths.append(local_manifest_file.as_posix())
+
+        self._ensure_disk_capacity(Path(manifest_write_dir), total_input_size)
+        return manifest_paths
+
     def attachment_sync_inputs(
```
leongdl, line 364: Curious - why are we modifying this existing code instead of having an A / B old code path to new code path clean cut ? 

We already deemed the existing code need to be broken down.

src/deadline/job_attachments/asset_sync.py
```diff
@@ -225,7 +225,15 @@ def aggregate_asset_root_manifests(
                     local_root = str(session_dir.joinpath(dir_name))
                     grouped_manifests_by_root[local_root].extend(manifests)
 
-        return grouped_manifests_by_root
+        # Merge the manifests in each root into a single manifest
```
leongdl, line 228: Are we doing this refactoring so Worker Agent works on a similar input code path? EG: new code path uses merged manifests and this `aggregate_asset_root_manifests ` function, and the old code path use `sync_input` as a monolithic function ? 

(Just providing context)

src/deadline/job_attachments/asset_sync.py
godobyte: Yes it's moving existing code around for refactoring. `merge_asset_manifests` returns `BaseAssetManifest | None`, that's why there is a check before adding to `dict[str, BaseAssetManifest]`. Even checks length at a higher level, it still needs a check for the merged manifest.

src/deadline/job_attachments/asset_sync.py
godobyte: Makes sense, updated the title to reflect the breaking change. Will update the commit message during squash and merge.

src/deadline/job_attachments/asset_sync.py
godobyte: Right the new code path for WA only uses the decomposed functions including this `aggregate_asset_root_manifests`, making sure it has the logic for the feature parity.

src/deadline/job_attachments/asset_sync.py
godobyte: Right we are not going to use this function. There are tests around this function to make sure the parity between the decomposed functions and the monolithic `sync_input`. As I'm refactoring  `_aggregate_asset_root_manifests`, the tests would fail if the caller don't update in this function.

leongdl (approved):

jericht (approved):

PR #485, opened 2024-10-29
chore(deps): update to black 24.10

ddneilson:
Fixes: None

Our `black` dependency is out of date, but `black` dropped Python 3.8 support in version 24.9.

Conditionally use the newer black in newer python versions. This *may* cause issues in future if newer black versions deviate from 24.8. We'll cross that bridge when we get to it; right now, it's fine.

Updated deps.

Running the linter. See the CI.

N/A

No

No

----

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -9,7 +9,8 @@ freezegun == 1.*
 types-pyyaml == 6.*
 twine == 4.*; python_version == '3.7'
 twine == 5.*; python_version > '3.7'
-black == 24.8.*; python_version > '3.7'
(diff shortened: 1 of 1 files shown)


epmog (approved):

jusiskin (approved):
According to this documentation:

https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy

The project uses calendar versioning, and they aim to maintain stable styling for the year. We should be able to pin to `24.*` and pip will pick the latest-available minor version supported by the Python runtime version.

What we have here works too, so approved!

PR #492, opened 2024-11-04
fix: The hash_cache fails with filenames including surrogates

### LINKED ISSUE #481: Bug: Special Characters cannot be included in job bundles
### Describe the bug

Many non-English speaking users will have special characters like ñ in their filenames. This can cause errors when submitting to Deadline Cloud.

To be clear, special characters can be included in job attachments, but if any strings, like the input scene path or the output path, end up in the template.yaml or template.json file, it will fail to submit.

### Expected Behaviour

We should support any string that can be a valid path.

### Current Behaviour

When you submit a job with invalid characters you get the following error:
image

### Reproduction Steps

Make a job bundle and make either the input scene path or the output path include a special character. Try to submit it through `deadline bundle gui-submit`

### Environment

At minimum:
1. Operating system: Windows
2. Output of `python3 --version`: 3.11.1
3. Output of `deadline --version`: version 0.48.4
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? The version that came out in September.

mwiebe:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/481

Cases encountered in practice show that filenames can include unicode surrogates. This changes the hash_cache functionality to preserve them by using the "surrogatepass" handler for them.

* Modify the hash_cache sqlite schema to use a BLOB instead of TEXT for the file_path, and then explicitly encode/decode it using the 'utf-8' encoding and 'surrogatepass' error handler.
* In the canonical json code, use the 'surrogatepass' error handler in the canonical file path sorting.
* In the error logging of the exception that revealed this bug, include the exception context to improve debugging of future similar issues.
* Add a test case for the hash_cache based on an example that was failing in practice.

Jobs that could not be submitted, because their filenames include surrogates, will work now.

  Yes
  No
  Yes, but have not run these tests.

N/A

No. In increases the table version of the sqlite hash cache, but that does not break compatibility.

No

----


src/deadline/job_attachments/caches/hash_cache.py
```diff
@@ -45,12 +47,12 @@ class HashCache(CacheDB):
     """
 
     CACHE_NAME = "hash_cache"
-    CACHE_DB_VERSION = 2
+    CACHE_DB_VERSION = 3
```
epmog, line 50: Does this impact the user in anyway? ie. Forcing a purge of the existing cache? Or will everything continue to work for previous users and new users alike?

mwiebe: It does effectively purge the cache, but users do not have to take any action.

src/deadline/job_attachments/asset_manifests/_canonical_json.py
```diff
@@ -15,7 +15,9 @@ def canonical_path_comparator(path: BaseManifestPath):
     """
     # Sort by UTF-16 values as per the spec
     # https://www.rfc-editor.org/rfc/rfc8785.html#name-sorting-of-object-propertie
-    return path.path.encode("utf-16_be")
+    # Use the "surrogatepass" error handler because filenames encountered in the wild
+    # include surrogates.
+    return path.path.encode("utf-16_be", errors="surrogatepass")
```
leongdl, line 20: And I thought this was some magic keyword :)  
https://docs.python.org/3/library/codecs.html

src/deadline/job_attachments/asset_sync.py
```diff
@@ -527,7 +528,16 @@ def _upload_output_manifest_to_s3(
             full_output_prefix,
             f"{manifest_name_prefix}_output",
         )
-        metadata = {"Metadata": {"asset-root": root_path}}
+        metadata = {"Metadata": {"asset-root": json.dumps(root_path, ensure_ascii=True)}}
```
leongdl, line 531: I will need to amend this to the `manifest upload` CLI since it has similar mechanisms.

src/deadline/job_attachments/download.py
```diff
@@ -670,7 +671,10 @@ def _get_asset_root_from_s3(
     except Exception as e:
         raise AssetSyncError(e) from e
 
-    return head["Metadata"].get("asset-root", None)
+    if "asset-root-json" in head["Metadata"]:
```
leongdl, line 674: Same here, I need to port this to `manifest download`. (Will do after this PR is merged)

leongdl (approved):

epmog (approved):

PR #515, opened 2024-11-22
feat: allow attachments commands to read openjd path mapping, allow upload with given manifest name and attach metadata

godobyte:
1. attachments commands should be able to read path mapping files generated by openjd, customer can directly grab the file from worker.
2. attachment upload now upload manifest files with name `_input` suffix by root hash, this doesn't fit well for output attachment upload use case.
3. attachment upload doesn't attach metadata, which prevent download job-output by job id from working.
4. fix: AssetSync `_check_and_write_local_manifests` currently hashes worker root rather than source root as file name, this doesn't work with attachment upload since hash in the file name should be from hashing submission local source root.

1. additionally allow attachments commands to read openjd path mapping file format
2. use given manifest name as s3 object name
3. attach manifest metadata
4. keeping class variable for lookup and hash source root for manifest name

more flexibility for attachments commands and make sure they work as expected

```
============== 28 passed, 2 skipped, 5 warnings in 31.39s ===============
```
```
%deadline job download-output --job-id job-ff3c4a430ce74c0397d38fbe7e194fac --farm-i
d $FARM_ID --queue-id $QUEUE_ID
Downloading output from Job 'AssetsExample'

Summary of files to download:
    /bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example (6 files)

You are about to download files which may come from multiple root directories. Here are a list of the current root directories:
[0] bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example
> Please enter the index of root directory to edit, y to proceed without changes, or n to cancel the download (0, y, n) [y]: y
The following files already exist in your local directory:
        /bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example/files/file1.txt
        bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example/files/file1.txt
        bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example/files/file2.txt
       bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example/files/file2.txt
        bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example/template.json
        bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example/template.json
You have three options to choose from:
[1] Skip: Do not download these files
[2] Overwrite: Download these files and overwrite existing files
[3] Create a copy: Download the file with a new name, appending '(1)' to the end
> Please enter your choice (1, 2, 3, or n to cancel the download) (1, 2, 3, n) [3]: 2

Download Summary:
    Downloaded 6 files totaling 4.51 KB.
    Total download time of 0.14941 seconds at 30.2 KB/s.
        /local/home/yuanbian/bea/bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example (6 files)
```

- Are relevant docstrings in the code base updated? [Yes]

----

### DIFF
--- a/src/deadline/job_attachments/api/attachment.py
+++ b/src/deadline/job_attachments/api/attachment.py
@@ -104,7 +104,7 @@ def attachment_upload(
         boto3_session (boto3.Session): Boto3 session for interacting with customer s3.
         root_dirs (List[str]): List of root directories holding attachments. Defaults to empty.
         path_mapping_rules (Optional[str], optional): Optional file path to a JSON file contains list of path mapping. Defaults to None.
-        upload_manifest_path (Optional[str], optional): Optional prefix for uploading given manifests. Defaults to None.
+        upload_manifest_path (Optional[str], optional): Optional path prefix for uploading given manifests. Defaults to None.
         logger (ClickLogger, optional): Logger to provide visibility. Defaults to ClickLogger(False).
 
     Raises:
@@ -140,17 +140,34 @@ def attachment_upload(
                 f"No valid root defined for given manifest {file_name}, please check input root dirs and path mapping rule."
             )
 
+        metadata = {"Metadata": {"asset-root": json.dumps(rule.source_path, ensure_ascii=True)}}
+        # S3 metadata must be ASCII, so use either 'asset-root' or 'asset-root-json' depending
+        # on whether the value is ASCII.
+        try:
+            # Add the 'asset-root' metadata if the path is ASCII
+            rule.source_path.encode(encoding="ascii")
+            metadata["Metadata"]["asset-root"] = rule.source_path
+        except UnicodeEncodeError:
+            # Add the 'asset-root-json' metadata encoded to ASCII as a JSON string
+            metadata["Metadata"]["asset-root-json"] = json.dumps(
+                rule.source_path, ensure_ascii=True
+            )
+        if rule.source_path_format:
+            metadata["Metadata"]["file-system-location-name"] = rule.source_path_format
+
         # Uploads all files to a CAS in the manifest, optionally upload manifest file
         key, data = asset_uploader.upload_assets(
             job_attachment_settings=s3_settings,
             manifest=manifest,
             partial_manifest_prefix=upload_manifest_path,
+            manifest_file_name=file_name,
+            manifest_metadata=metadata,
             source_root=Path(rule.source_path),
             asset_root=Path(rule.destination_path),
             s3_check_cache_dir=config_file.get_cache_directory(),
         )
         logger.echo(
-            f"Uploaded assets from {rule.source_path}, to {s3_settings.to_s3_root_uri()}/{key}, hashed data {data}"
+            f"Uploaded assets from {rule.destination_path}, to {s3_settings.to_s3_root_uri()}/Manifests/{key}, hashed data {data}"
         )
 
 
@@ -179,7 +196,12 @@ def _process_path_mapping(
                 f"Specified path mapping file {path_mapping_rules} is not valid."
             )
         with open(path_mapping_rules, encoding="utf8") as f:
-            path_mapping_rule_list.extend([PathMappingRule(**mapping) for mapping in json.load(f)])
+            data = json.load(f)
+            if "path_mapping_rules" in data:
+                data = data["path_mapping_rules"]
+
+            assert isinstance(data, list), "Path mapping rules have to be a list of dict."
+            path_mapping_rule_list.extend([PathMappingRule(**mapping) for mapping in data])
 
     if nonvalid_dirs := [root for root in root_dirs if not os.path.isdir(root)]:
         raise NonValidInputError(f"Specified root dir {nonvalid_dirs} are not valid.")
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -183,7 +183,7 @@ def _get_tasks_manifests_keys_from_s3(
             Prefix=manifest_prefix,
         )
 
-        # 1. Find all files that match the pattern: task-{any}/{any}/{any}-output
+        # 1. Find all files that match the pattern: task-{any}/{any}/{any}output{any}
         task_prefixes = defaultdict(list)
         for page in page_iterator:
             contents = page.get("Contents", None)
@@ -192,7 +192,7 @@ def _get_tasks_manifests_keys_from_s3(
                     f"Unable to find asset manifest in s3://{s3_bucket}/{manifest_prefix}"
                 )
             for content in contents:
-                if re.search(r"task-.*/.*/.*_output", content["Key"]):
+                if re.search(r"task-.*/.*/.*output.*", content["Key"]):
                     parts = content["Key"].split("/")
                     for i, part in enumerate(parts):
                         if "task-" in part:
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -149,6 +149,8 @@ class S3AssetUploader:
         s3_check_cache_dir: Optional[str] = None,
         manifest_write_dir: Optional[str] = None,
         manifest_name_suffix: str = "input",
+        manifest_metadata: dict[str, dict[str, str]] = dict(),
+        manifest_file_name: Optional[str] = None,
         asset_root: Optional[Path] = None,
     ) -> tuple[str, str]:
         """
@@ -163,6 +165,8 @@ class S3AssetUploader:
             job_attachment_settings: The settings for the job attachment configured in Queue.
             progress_tracker: Optional progress tracker to track progress.
             manifest_name_suffix: Suffix for given manifest naming.
+            manifest_metadata: File metadata for given manifest to be uploaded.
+            manifest_file_name: Optional file name for given manifest to be uploaded, otherwise use default name.
             asset_root: The root in which asset actually in to facilitate path mapping.
 
         Returns:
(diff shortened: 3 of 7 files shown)


src/deadline/job_attachments/api/attachment.py
```diff
@@ -90,6 +90,7 @@ def attachment_upload(
     root_dirs: List[str] = [],
     path_mapping_rules: Optional[str] = None,
     upload_manifest_path: Optional[str] = None,
+    manifest_name_suffix: str = "input",
```
leongdl: nit - what do you think if we just had "upload file name", and the customer or caller can provide the full name? 

Or better yet, what if we just uploaded it with the same name as the manifest file? Users can rename files before running the API or CLI

jericht: Yeah, wondering why do we need/want the suffix? What happens if a customer provides an empty suffix?

src/deadline/job_attachments/api/attachment.py
```diff
@@ -145,12 +147,13 @@ def attachment_upload(
...
             manifest=manifest,
             partial_manifest_prefix=upload_manifest_path,
+            manifest_name_suffix=manifest_name_suffix,
             source_root=Path(rule.source_path),
             asset_root=Path(rule.destination_path),
             s3_check_cache_dir=config_file.get_cache_directory(),
         )
         logger.echo(
-            f"Uploaded assets from {rule.source_path}, to {s3_settings.to_s3_root_uri()}/{key}, hashed data {data}"
+            f"Uploaded assets from {rule.destination_path}, to {s3_settings.to_s3_root_uri()}/Manifests/{key}, hashed data {data}"
```
jericht, line 170: Is changing the message to `rule.destination_path` intentional? Aren't we uploading from `source_path`?

src/deadline/job_attachments/api/attachment.py
```diff
@@ -179,7 +182,11 @@ def _process_path_mapping(
                 f"Specified path mapping file {path_mapping_rules} is not valid."
             )
         with open(path_mapping_rules, encoding="utf8") as f:
-            path_mapping_rule_list.extend([PathMappingRule(**mapping) for mapping in json.load(f)])
+            data = json.load(f)
+            if "path_mapping_rules" in data:
+                data = data["path_mapping_rules"]
```
jericht, line 201: Is it guaranteed that the data is a list of `PathMappingRule` if the JSON doesn't have a `path_mapping_rules` attribute? In what cases can this happen that we support?

src/deadline/job_attachments/api/attachment.py
godobyte: I don't see if `_input` suffix is being used anywhere, that's probably for visibility only. However, for `_output`, it's used for download. Without the `_output` suffix, job download cannot find the manifests and download corresponding files.

https://github.com/godobyte/deadline-cloud/blob/mainline/src/deadline/job_attachments/download.py#L174-L175

Job Attachment relies on file name implicitly for some logic (mapping with root path, discover output manifests in s3 for download). The current approach I'm following is to keep it the same as what we currently have. If we'd like to allow any customer defined manifest name, we should consider different customer use case and make sure they work with the current/future logic. I'm happy to sync more on this.

godobyte: Apart from that, the `_input` prefix is already there (whether passing in or not) to maintain feature parity with the previous Job Attachment implementation. 
https://github.com/godobyte/deadline-cloud/blob/mainline/src/deadline/job_attachments/upload.py#L151

src/deadline/job_attachments/api/attachment.py
godobyte: For openjd generated path mapping rule, it's guaranteed to be a list as long as the file exist.

https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run#path-mapping

godobyte: I'm still thinking about whether we should support if the JSON doesn't have a path_mapping_rules, i.e. non openjd format, keeping it here for now for backward compatibility. Either way, the supported format(s) should be documented clearly.

src/deadline/job_attachments/api/attachment.py
godobyte: The `source_path` is the submission path on user's workstation, the `destination_path` is the mapped path on worker, or asset root in general, hence the destination path should be the correct path. (edit - if caller doesn't provide path mapping, this 2 paths would be the same).

jericht: Right, but I feel like this would be awkward since the log happens on attachment upload. Is this code shared between both submitter and worker flows? (uploading inputs and uploading outputs)?

I think the change makes sense for the worker flow only, but awkward for submission flow

src/deadline/job_attachments/api/attachment.py
jericht: Sounds good, if the behaviour accepted a list directly before, I think it's fine to keep it here.

For that case, I was considering whether we could do an extra check to at least make sure `data` is actually a `list` of `dict`, but it's probably not worth the code if the error message is clear enough and would be similar anyway (`TypeError`)

src/deadline/job_attachments/api/attachment.py
godobyte: No, this code is not used for submitter, it's purely for worker/customer CLI flow. The CLI can be used for both input and output upload, if no path mapping provided, the path would be local root, if path mapping provided, the path would be the mapped path, which should be the expected behavior.

src/deadline/job_attachments/api/attachment.py
godobyte: Currently the error would be from L189 below as `TypeError: deadline.job_attachments.models.PathMappingRule() argument after ** must be a mapping, not str`. That can be a bit confusing, I can add some checks to provide explicit error message in an upcoming PR I'm currently working on.

jericht (dismissed):

src/deadline/job_attachments/api/attachment.py
leongdl: But that's the question - feature parity for the heck of it is not the goal of creating the API. 

The name of the file on S3 simply by naming / moving the file and using it. Adding a new parameter to accomplish this seems odd.

If we want to have a parameter, I would rather make it generic like "Upload file name", which lets the caller customize the entire name instead of just the postfix.

leongdl: See here: https://github.com/aws-deadline/deadline-cloud/blob/43cd3623f3bb6cd2c0295205fc61f5677975dd51/src/deadline/job_attachments/upload.py#L231

All the `postfix` does is a string format. For the CLI and API we should just have the input as the right name to begin with.

godobyte: My concern with being able to upload any customer-defined manifest file names are 
1. Violating the file name hash to root mapping Job Attachment relies on to get upload source and download destination.
2.  Breaking the existing logic we have for [output download that relies on file suffix](https://github.com/godobyte/deadline-cloud/blob/mainline/src/deadline/job_attachments/download.py#L195).

If we allow any name for manifests, the above 2 cases need to be proper thought out with error handling.

godobyte: Synced offline, the `_input` suffix shouldn't be used anywhere, and we could update the output suffix check for job output download based on job id. Added a rev to us the manifest file name for upload.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -140,17 +140,34 @@ def attachment_upload(
                 f"No valid root defined for given manifest {file_name}, please check input root dirs and path mapping rule."
             )
 
+        metadata = {"Metadata": {"asset-root": json.dumps(rule.source_path, ensure_ascii=True)}}
+        # S3 metadata must be ASCII, so use either 'asset-root' or 'asset-root-json' depending
+        # on whether the value is ASCII.
+        try:
+            # Add the 'asset-root' metadata if the path is ASCII
+            rule.source_path.encode(encoding="ascii")
```
leongdl, line 148: nice catch! We totally buried some hidden details

src/deadline/job_attachments/download.py
```diff
@@ -192,7 +192,7 @@ def _get_tasks_manifests_keys_from_s3(
                     f"Unable to find asset manifest in s3://{s3_bucket}/{manifest_prefix}"
                 )
             for content in contents:
-                if re.search(r"task-.*/.*/.*_output", content["Key"]):
+                if re.search(r"task-.*/.*/.*output.*", content["Key"]):
```
leongdl, line 195: *output.* ? this means any part of the string contains output right?

leongdl (approved):

src/deadline/job_attachments/api/attachment.py
jericht: Why do we need to support this case if the path is just ASCII? Could we just always use `asset-root-json`? or is this because that is how Job Attachments was already working?

src/deadline/job_attachments/upload.py
```diff
@@ -163,6 +165,8 @@ def upload_assets(
             job_attachment_settings: The settings for the job attachment configured in Queue.
             progress_tracker: Optional progress tracker to track progress.
             manifest_name_suffix: Suffix for given manifest naming.
+            manifest_metadata: File metadata for given manifest to be uploaded.
+            manifest_file_name: Optional file name for given manifest to be uploaded, otherwise use default name.
```
jericht, line 169: nit: can we add details on what the `default` name is and where it comes from? This is the one derived from the hash of the asset root?

src/deadline/job_attachments/api/attachment.py
godobyte: That's a good question, it comes from here https://github.com/aws-deadline/deadline-cloud/pull/492

src/deadline/job_attachments/upload.py
godobyte: Do you think this is blocking? I'd like to take it in a follow-up if possible.

jericht (approved):

PR #523, opened 2024-11-29
fix: Custom Host Requirements ui now accepts decimals.

### LINKED ISSUE #94: Bug: Fleet Worker Requirements lets values to be set in AWS console that cannot be set in the submitter UI
### Expected Behaviour

When adding custom worker requirements to a fleet in the AWS console I can add values that are not able to be added to the DCC submitter UI

For example in the AWS console I can create a custom amount:
Amount name = amount.version.blender
Minimum amount = 3.6
Maximum amount = 4.1

### Current Behaviour

When in the submitter UI I am unable to add the "." character so the values end up as
Amount name = version
Minimum amount = 3
Maximum amount = 4

### Reproduction Steps

n/a

### Code Snippet

n/a

erico-aws:
Fixes: [*Bug: Fleet Worker Requirements lets values to be set in AWS console that cannot be set in the submitter UI*](https://github.com/aws-deadline/deadline-cloud-for-blender/issues/94)

Through the command line, Custom Host Requirements allow for decimal values. Through the UI it wasn't possible to enter a decimal value.

Switched the `OptionalSpinBox` control to inherit from `QDoubleSpinBox`. It also defaults to allowing values with 2 decimal places. 

The hardware requirements use the same OptionalSpinBox control. They don't need decimals so I set them to have 0 decimal places.

I also checked the [OpenJD spec](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#331-amountrequirement) to make sure that host requirements could have floating point values. 

The submitters that use this UI (like Blender) will behave more like their cli counterparts.

Built locally and installed into the Blender submitter. Opened Blender and submitted Jobs with Host Requirements with the integrated submitter.

No.

No

No

Screenshot 2024-12-18 at 10 30 56 AM

----


miabatta (dismissed):

leongdl (dismissed):

leongdl:
Nit; please add a screenshot to the PR description next time?

miabatta (dismissed):

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -859,19 +864,27 @@ def has_input(self) -> bool:
...
     """
 
-    def __init__(self, min: int = MIN_INT_VALUE, max: int = MAX_INT_VALUE, parent=None) -> None:
+    def __init__(
+        self,
+        min: int = MIN_INT_VALUE,
+        max: int = MAX_INT_VALUE,
+        decimal: int = DECIMAL_VALUE,
+        parent=None,
+    ) -> None:
```
epmog, line 955: OptionalSpinBox appears to be a public class and this change will break anyone using positional args to create it, so I think we'll either have to mark this as a breaking change or add decimal after parent.

Other option is to still make this a breaking change, but only accept keyword args (and maybe apply this to the other gui bits): 

```python
def func(self, *, key="word", only="args"):
```

epmog (changes requested):

src/deadline/client/ui/widgets/host_requirements_tab.py
erico-aws: I could make it non-breaking by not including decimal in the constructor, defaulting it to `DECIMAL_VALUE` and then using `setDecimals()` where I need it changed?

epmog: Yea that works as well

erico-aws: This doesn't stop it from being a breaking change. The `valueFromText()` function changing from an int to float is still breaking.

I'm going to make a new `OptionalDoubleSpinbox` instead.

miabatta (approved):

epmog (approved):

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -933,3 +939,88 @@ def stepBy(self, steps: int) -> None:
...
+
+class OptionalDoubleSpinBox(QDoubleSpinBox):
+    """
+    A custom QDoubleSpinBox that set min - 1 value as "-" to represent value not set.
+    """
+
+    def __init__(
+        self,
+        min: int = MIN_INT_VALUE,
+        max: int = MAX_INT_VALUE,
```
epmog, line 952: Should the double spin box have ints for the min/max? Does openjd specify what they do and if they should be float min/max?

erico-aws: min needs to be 0 (non-negative). Max could be float max.

PR #538, opened 2024-12-16
feat: Allow submitters to prepopulate HostRequirementsWidget

### LINKED ISSUE #535: Bug: Unable to set default Host Requirements in `SubmitJobToDeadlineDialog.__init__`
### Describe the bug

Exposing the default settings for the HostRequirementsWidget would be useful for dcc submitters to have the ability to predefine requirements, either based on the scene file loaded (requiring GPUs) or for Operating System requirements

### Expected Behaviour

I would love it to expose host requirements as an `Optional[dict[str,Any]]` (Or a custom type) that when provided (in combination with the `show_host_requirements_tab` setting) will setup defaults for the `HostRequirementsWidget`

The code change will have to happen in `submit_job_to_deadline_dialog.py` and `host_requirements_tab.py`

`host_requirements_tab.py`
```diff
  
          return requirements
  
+     def set_requirements(self, requirements: Dict[str, Any]):
+         custom_requirements = {
+             "attributes": [],
+             "amounts": [],
+         }
+         removed_requirements = {
+             "attributes": [],
+             "amounts": [],
+         }
+         os_requirements = self.os_requirements_box.get_requirements()
+         hardware_requirements = self.hardware_requirements_box.get_requirements()
+ 
+         if requirements.get("attributes"):
+             for requirement in requirements.get("attributes"):
+                 if requirement not in os_requirements:
+                     custom_requirements.get("attributes").append(requirement)
+             for existing_requirement in os_requirements:
+                 if existing_requirement not in requirements.get("attributes"):
+                     removed_requirements.get("attributes").append(existing_requirement)
+             for requirement in removed_requirements.get("attributes"):
+                 os_requirements.remove(requirement)
+ 
+         if requirements.get("amounts"):
+             for requirement in requirements.get("amounts"):
+                 if requirement not in hardware_requirements:
+                     custom_requirements.get("amounts").append(requirement)
+             for existing_requirement in hardware_requirements:
+                 if existing_requirement not in requirements.get("amounts"):
+                     removed_requirements.get("amounts").append(existing_requirement)
+             for requirement in removed_requirements.get("amounts"):
+                 hardware_requirements.remove(requirement)
+ 
+         self.custom_requirements_box.set_requirements(custom_requirements)
+         self.os_requirements_box.set_requirements(os_requirements)
+         self.hardware_requirements_box.set_requirements(hardware_requirements)
+ 
  
  class OverrideRequirementsWidget(QGroupBox):  # pylint: disable=too-few-public-methods
      """
...
  class OSRequirementsWidget(QGroupBox):
      """
      UI elements to hold OS settings.
  
      Args:
          parent: The parent Qt Widget.
      """
+     OS_ROW_OPTIONS = ["linux", "macos", "windows"]
+     CPU_ROW_OPTIONS = ["x86_64", "arm64"]
  
      def __init__(self, parent=None):
          super().__init__("", parent=parent)
          self.layout = QVBoxLayout(self)
          self._build_ui()
  
      def _build_ui(self):
!         self.os_row = OSRequirementRowWidget("Operating system", self.OS_ROW_OPTIONS)
!         self.cpu_row = OSRequirementRowWidget("CPU architecture", self.CPU_ROW_OPTIONS)
  
          self.layout.addWidget(self.os_row)
          self.layout.addWidget(self.cpu_row)
...
          return requirements
  
+     def set_requirements(self, requirements: List[Dict[str, Any]]):
+         for requirement in requirements:
+             if requirement.get("name") == "attr.worker.os.family":
+                 if requirement.get("anyOf") not in self.OS_ROW_OPTIONS:
+                     raise ValueError(
+                         "Invalid value {} for option attr.worker.os.family".format(requirement.get("anyOf"))
+                     )
+                 self.os_row.combo_box.setCurrentText(requirement.get("anyOf"))
+             if requirement.get("name") == "attr.worker.cpu.arch":
+                 if requirement.get("anyOf") not in self.CPU_ROW_OPTIONS:
+                     raise ValueError(
+                         "Invalid value {} for option attr.worker.cpu.arch".format(requirement.get("anyOf"))
+                     )
+                 self.os_row.combo_box.setCurrentText(requirement.get("anyOf"))
+             else:
+                 raise ValueError("Unknown OS Requirement option of {}".format(requirement.get("name")))
+ 
  
  class HardwareRequirementsWidget(QGroupBox):  # pylint: disable=too-few-public-methods
...
          return requirements
  
+     def set_requirements(self, requirements: List[Dict[str, Any]]):
+         for requirement in requirements:
+             if requirement.get("name") == "amount.worker.vcpu":
+                 self.cpu_row.update_requirement(requirement)
+ 
+             elif requirement.get("name") == "amount.worker.memory":
+                 self.memory_row.update_requirement(requirement, scaling_factor=1024)
+ 
+             elif requirement.get("name") == "amount.worker.gpu":
+                 self.gpu_row.update_requirement(requirement)
+ 
+             elif requirement.get("name") == "amount.worker.gpu.memory":
+                 self.gpu_memory_row.update_requirement(requirement, scaling_factor=1024)
+ 
+             elif requirement.get("name") == "amount.worker.disk.scratch":
+                 self.scratch_space_row.update_requirement(requirement)
+ 
+             else:
+                 raise ValueError("Unknown Hardware Requirement option of {}".format(requirement.get("name")))
+ 
  
  class CustomRequirementsWidget(QGroupBox):
...
  
          return requirements
  
+     def set_requirements(self, requirements: Dict[str, List]):
+         amount_widgets = {}
+         attribute_widgets = {}
+         for i in range(self.list_widget.count()):
+             widget = self.list_widget.itemWidget(self.list_widget.item(i))
+             widget_requirement = widget.get_requirement()
+             if widget_requirement:
+                 if isinstance(widget, CustomAmountWidget):
+                     amount_widgets[widget_requirement.get("name")] = widget
+                 elif isinstance(widget, CustomAttributeWidget):
+                     attribute_widgets[widget_requirement.get("name")] = widget
+                 else:
+                     logger.warning(
+                         f"Widget requirement is not a valid expected type: {type(widget)}"
+                     )
+ 
+         for requirement in requirements.get("amounts"):
+             if requirement.get("name") not in amount_widgets:
+                 logger.warning(
+                     "Amount requirement is not tied to a widget: {}".format(requirement.get("name"))
+                 )
+             else:
+                 widget = amount_widgets.get(requirement.get("name"))
+                 widget.set_requirement(requirement)
+         for requirement in requirements.get("attributes"):
+             if requirement.get("name") not in attribute_widgets:
+                 logger.warning(
+                     "Attribute requirement is not tied to a widget: {}".format(requirement.get("name"))
+                 )
+             else:
+                 widget = attribute_widgets.get(requirement.get("name"))
+                 widget.set_requirement(requirement)
+ 
  
  class CustomCapabilityWidget(QGroupBox):
...
          return requirement
  
+     def set_requirement(self, requirement: Dict[str, Any]):
+         if requirement.get("min"):
+             self.min_spin_box.setValue(requirement.get("min"))
+         if requirement.get("max"):
+             self.max_spin_box.setValue(requirement.get("max"))
+ 
  
  class CustomAttributeWidget(CustomCapabilityWidget):
...
  
          return requirement
  
+     def set_requirement(self, requirement: Dict[str, Any]):
+         option = ""
+         if requirement.get("anyOf") and not self.any_of_button.isChecked():
+             self.any_of_button.setChecked(True)
+             option = "anyOf"
+         elif requirement.get("allOf") and self.any_of_button.isChecked():
+             self.any_of_button.setChecked(False)
+             option = "allOf"
+ 
+         values = requirement.get(option)
+         if len(values) > self.value_list_widget.count():
+             # Added additional values:
+             for _ in range(len(values) - self.value_list_widget.count()):
+                 self._add_value()
+         elif len(values) < self.value_list_widget.count():
+             # Removed values:
+             for i in range(self.value_list_widget.count() - len(values)):
+                 value = self.value_list_widget.itemWidget(self.value_list_widget.item(i + len(values)))
+                 self.remove_value_item(value.value_list_item)
+ 
+         # We now have the correct number of value widgets. Let us just go fill all the values in.
+         for i in range(self.value_list_widget.count()):
+             value = self.value_list_widget.itemWidget(self.value_list_widget.item(i))
+             value.line_edit.setText(values[i])
+ 
  
  class CustomAttributeValueWidget(QWidget):
...
              requirements.append(requirement)
  
+     def update_requirement(self, requirement: Dict[str, Any], scaling_factor: int = 1):
+         if requirement.get("min"):
+             self.min_spin_box.setValue(requirement.get("min") / scaling_factor)
+         if requirement.get("max"):
+             self.max_spin_box.setValue(requirement.get("max") / scaling_factor)
+ 
  
  class OptionalComboBox(QComboBox):
```

### Current Behaviour

No functionality exposed

### Reproduction Steps

N/A

### Environment

N/A

Ahuge:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/535

Submitters were unable to prepopulate the Host Requirements and had to resort to forcing the values in the job template (cinema-4d). This caused issues if the "default" workflow wasn't the one that the artist wanted as they had no idea of the defaults in the job template
Implement setters on the HostRequirementsWidget and allow a `host_requirements` parameter to the SubmitToDeadlineDialog
New functionality in the submitter dialog
Ran local testing to confirm it is working.
I have added a few unit tests for the dataclasses as well.

I don't believe it is breaking

----

### DIFF
--- a/src/deadline/client/ui/dataclasses/__init__.py
+++ b/src/deadline/client/ui/dataclasses/__init__.py
@@ -7,8 +7,11 @@ from __future__ import annotations
 
 import os
 from dataclasses import dataclass, field
+from typing import Literal, Dict, List, Union
 
 from ...job_bundle.parameters import JobParameter
+
+MAX_INT_VALUE = (2**31) - 1
 
 
 @dataclass
@@ -66,3 +69,292 @@ echo "Content for generated file {{Task.Param.Index}}" > task_output_file_{{Task
     array_parameter_values: str = field(default="1-5")
     data_dir: str = field(default=os.path.join("~", "CLIJobData"))
     file_format: str = field(default="YAML")
+
+
+@dataclass
+class OsRequirements:
+    """
+    Settings for a OSRequirementsWidget.
+    """
+
+    LINUX = "linux"
+    MACOS = "macos"
+    WINDOWS = "windows"
+    OPERATING_SYSTEMS = [LINUX, MACOS, WINDOWS]
+    X86_64 = "x86_64"
+    ARM_64 = "arm64"
+    CPU_ARCHITECTURES = [X86_64, ARM_64]
+    operating_systems: List[Literal[LINUX, MACOS, WINDOWS]] = field(default_factory=list)  # type: ignore
+    cpu_archs: List[Literal[X86_64, ARM_64]] = field(default_factory=list)  # type: ignore
+
+    def __post_init__(self):
+        for operating_system in self.operating_systems:
+            if operating_system not in self.OPERATING_SYSTEMS:
+                raise ValueError(
+                    f"Operating system {operating_system} is not in supported list: {self.OPERATING_SYSTEMS}."
+                )
+        for cpu in self.cpu_archs:
+            if cpu not in self.CPU_ARCHITECTURES:
+                raise ValueError(
+                    f"CPU architecture {cpu} is in supported list: {self.CPU_ARCHITECTURES}."
+                )
+
+    def serialize(self) -> list:
+        requirements: List[dict] = []
+        if self.operating_systems:
+            for operating_system in self.operating_systems:
+                if operating_system not in self.OPERATING_SYSTEMS:
+                    raise ValueError(
+                        f"Operating system {operating_system} is not in supported list: {self.OPERATING_SYSTEMS}."
+                    )
+            requirements.append(
+                {
+                    "name": "attr.worker.os.family",
+                    "anyOf": self.operating_systems,
+                }
+            )
+        if self.cpu_archs:
+            for cpu in self.cpu_archs:
+                if cpu not in self.CPU_ARCHITECTURES:
+                    raise ValueError(
+                        f"CPU architecture {cpu} is not in supported list: {self.CPU_ARCHITECTURES}."
+                    )
+            requirements.append(
+                {
+                    "name": "attr.worker.cpu.arch",
+                    "anyOf": self.cpu_archs,
+                }
+            )
+        return requirements
+
+
+@dataclass
+class HardwareRequirements:
+    """
+    Settings for a HardwareRequirementsWidget.
+    """
+
+    DEFAULT_VALUE = -1
+    cpu_min: int = field(default=DEFAULT_VALUE)
+    cpu_max: int = field(default=DEFAULT_VALUE)
+    memory_min: int = field(default=DEFAULT_VALUE)
+    memory_max: int = field(default=DEFAULT_VALUE)
+    acceleration_min: int = field(default=DEFAULT_VALUE)
+    acceleration_max: int = field(default=DEFAULT_VALUE)
+    acceleration_memory_min: int = field(default=DEFAULT_VALUE)
+    acceleration_memory_max: int = field(default=DEFAULT_VALUE)
+    scratch_space_min: int = field(default=DEFAULT_VALUE)
+    scratch_space_max: int = field(default=DEFAULT_VALUE)
+
+    def __post_init__(self):
+        self._validate(self.cpu_min, self.cpu_max, "CPU")
+        self._validate(self.memory_min, self.memory_max, "Memory")
+        self._validate(self.acceleration_min, self.acceleration_max, "Acceleration")
+        self._validate(
+            self.acceleration_memory_min, self.acceleration_memory_max, "Acceleration Memory"
+        )
+        self._validate(self.scratch_space_min, self.scratch_space_max, "Scratch Space")
+
+    def _validate(self, minimum: int, maximum: int, name: str):
+        if minimum != self.DEFAULT_VALUE and maximum != self.DEFAULT_VALUE and minimum > maximum:
+            raise ValueError(
+                f"{name} Minimum cannot be higher than {name} Maximum. {minimum} > {maximum}"
+            )
+
+    def _serialize(self, minimum, maximum, name):
+        requirements: List[dict] = []
+        if minimum != self.DEFAULT_VALUE or maximum != self.DEFAULT_VALUE:
+            _requirement = {
+                "name": name,
+            }
+            if minimum != self.DEFAULT_VALUE:
+                _requirement["min"] = max(minimum, self.DEFAULT_VALUE)
+            if maximum != self.DEFAULT_VALUE:
+                _requirement["max"] = min(maximum, MAX_INT_VALUE)
+            requirements.append(_requirement)
+        return requirements
+
+    def serialize(self) -> list:
+        requirements: List[dict] = []
+        requirements.extend(self._serialize(self.cpu_min, self.cpu_max, "amount.worker.vcpu"))
+        requirements.extend(
+            self._serialize(self.memory_min, self.memory_max, "amount.worker.memory")
+        )
+        requirements.extend(
+            self._serialize(self.acceleration_min, self.acceleration_max, "amount.worker.gpu")
+        )
+        requirements.extend(
+            self._serialize(
+                self.acceleration_memory_min,
+                self.acceleration_memory_max,
+                "amount.worker.gpu.memory",
+            )
+        )
+        requirements.extend(
+            self._serialize(
+                self.scratch_space_min, self.scratch_space_max, "amount.worker.disk.scratch"
+            )
+        )
+        return requirements
+
+
+@dataclass
+class CustomAmountRequirement:
+    """
+    Settings for a CustomAmountWidget
+    """
+
+    DEFAULT_VALUE = -(2**31) + 1
+    name: str = field(default="")
+    min: int = field(default=DEFAULT_VALUE)
+    max: int = field(default=DEFAULT_VALUE)
+
+    def __post_init__(self):
+        if not self.name:
+            raise ValueError(f"Custom Amount {self} has no name")
+        elif self.min and self.max and self.min > self.max:
+            raise ValueError(f"Custom Amount {self} has min higher than max")
+
+    def serialize(self) -> dict:
+        _requirement = {"name": "amount.worker." + self.name}
+        if self.min != self.DEFAULT_VALUE:
+            _requirement["min"] = max(self.min, self.DEFAULT_VALUE)
+        if self.max != self.DEFAULT_VALUE:
+            _requirement["max"] = min(self.max, MAX_INT_VALUE)
+        return _requirement
+
+
+@dataclass
+class CustomAttributeRequirement:
+    """
+    Settings for a CustomAttributeWidget
+    """
+
+    ANY_OF = "anyOf"
+    ALL_OF = "allOf"
+    OPTIONS = [ANY_OF, ALL_OF]
+    name: str = field(default="")
+    option: Literal[ANY_OF, ALL_OF] = field(default=ALL_OF)  # type: ignore
+    values: List[str] = field(default_factory=list)
+
+    def __post_init__(self):
+        if not self.name:
+            raise ValueError(f"Custom Amount {self} has no name")
+        elif not self.option or self.option not in CustomAttributeRequirement.OPTIONS:
+            raise ValueError(
+                f"Custom Amount {self} option is not in {CustomAttributeRequirement.OPTIONS}"
+            )
+
+    def serialize(self) -> dict:
+        if not self.name:
+            raise ValueError(f"Custom Attribute {self} has no name")
+        _requirement = {
+            "name": "attr.worker." + self.name,
+        }
+        if not self.option:
+            raise ValueError(f"Custom Attribute {self} has no option")
+        elif self.option not in ["anyOf", "allOf"]:
+            raise ValueError(
+                f'Custom Attribute {self} has option {self.option} which is not in list: ["anyOf", "allOf"]'
+            )
+        if not self.values:
+            raise ValueError(f"Custom Attribute {self} has no values")
+        _requirement[self.option] = self.values
+        return _requirement
+
+
+@dataclass
+class CustomRequirements:
+    """
+    Settings for a CustomRequirementsWidget.
+    """
+
+    amounts: List[CustomAmountRequirement] = field(default_factory=list)
+    attributes: List[CustomAttributeRequirement] = field(default_factory=list)
+
+    def __post_init__(self):
+        amounts = []
+        attributes = []
+        for amount in self.amounts:
+            amounts.append(self._validate_amount(amount))
+        for attribute in self.attributes:
+            attributes.append(self._validate_attribute(attribute))
+        self.amounts = amounts
+        self.attributes = attributes
+
+    @staticmethod
+    def _validate_amount(amount: Union[dict, CustomAmountRequirement]) -> CustomAmountRequirement:
+        if isinstance(amount, dict):
+            amount = CustomAmountRequirement(**amount)
+        return amount
+
+    @staticmethod
+    def _validate_attribute(
+        attribute: Union[dict, CustomAttributeRequirement]
+    ) -> CustomAttributeRequirement:
+        if isinstance(attribute, dict):
+            attribute = CustomAttributeRequirement(**attribute)
+        return attribute
+
+    def __iter__(self):
+        if self.amounts:
+            yield from self.amounts
+        if self.attributes:
+            yield from self.attributes
+
+    def _serialize_amounts(self):
+        requirements = []
+        for amount in self.amounts:
+            requirements.append(amount.serialize())
+        return requirements
+
+    def _serialize_attributes(self):
+        requirements = []
+        for attribute in self.attributes:
+            requirements.append(attribute.serialize())
+        return requirements
+
+    def serialize(self) -> Dict[str, List]:
+        requirements: Dict[str, List] = {}
+
+        if self.amounts:
+            requirements.setdefault("amounts", []).extend(self._serialize_amounts())
+
+        if self.attributes:
+            requirements.setdefault("attributes", []).extend(self._serialize_attributes())
+        return requirements
+
+
+@dataclass
+class HostRequirements:
+    """
+    Settings for a HostRequirementsWidget.
+    """
+
+    os_requirements: OsRequirements = field(default=None)
+    hardware_requirements: HardwareRequirements = field(default=None)
+    custom_requirements: CustomRequirements = field(default=None)
+
+    def __post_init__(self):
+        if isinstance(self.os_requirements, dict):
+            self.os_requirements = OsRequirements(**self.os_requirements)
+        if isinstance(self.hardware_requirements, dict):
+            self.hardware_requirements = HardwareRequirements(**self.hardware_requirements)
+        if isinstance(self.custom_requirements, dict):
+            self.custom_requirements = CustomRequirements(**self.custom_requirements)
+
+    def serialize(self) -> dict:
+        requirements: Dict[str, Union[List[str], List[int]]] = {}
+        if self.os_requirements is not None:
+            requirements.setdefault("attributes", []).extend(self.os_requirements.serialize())
+
+        if self.hardware_requirements:
+            requirements.setdefault("amounts", []).extend(self.hardware_requirements.serialize())
+
+        custom_requirements = self.custom_requirements.serialize()
+        if custom_requirements.get("amounts", []):
+            requirements.setdefault("amounts", []).extend(custom_requirements["amounts"])
+        if custom_requirements.get("attributes", []):
+            requirements.setdefault("attributes", []).extend(custom_requirements["attributes"])
+
+        return requirements
--- a/src/deadline/client/ui/widgets/host_requirements_tab.py
+++ b/src/deadline/client/ui/widgets/host_requirements_tab.py
@@ -3,11 +3,20 @@
 """
 UI widgets for the host requirements tab.
 """
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union, Literal
 from pathlib import Path
 
 from qtpy.QtCore import Qt  # type: ignore
-from qtpy.QtGui import QFont, QValidator, QIntValidator, QDoubleValidator, QBrush, QIcon, QRegularExpressionValidator  # type: ignore
+from qtpy.QtGui import (  # type: ignore
+    QFont,
+    QValidator,
+    QIntValidator,
+    QDoubleValidator,
+    QBrush,
+    QIcon,
+    QRegularExpressionValidator,
+    QStandardItem,
+)
 from qtpy.QtWidgets import (  # type: ignore
     QComboBox,
     QGroupBox,
@@ -29,6 +38,7 @@ from qtpy.QtWidgets import (  # type: ignore
 )
 
 from deadline.client.exceptions import NonValidInputError
+from ..dataclasses import HostRequirements, CustomRequirements, OsRequirements, HardwareRequirements
 
 from logging import getLogger
 
@@ -83,7 +93,7 @@ class HostRequirementsWidget(QWidget):  # pylint: disable=too-few-public-methods
         parent: The parent Qt Widget.
     """
 
-    def __init__(self, parent=None):
+    def __init__(self, requirements: Optional[HostRequirements] = None, parent=None):
         super().__init__(parent)
         layout = QVBoxLayout(self)
 
@@ -107,6 +117,8 @@ class HostRequirementsWidget(QWidget):  # pylint: disable=too-few-public-methods
         self.mode_selection_box.use_custom_button.toggled.connect(
             self._on_mode_selection_use_custom_button_toggled
         )
+        if requirements is not None:
+            self.set_requirements(requirements)
 
     def _on_mode_selection_use_custom_button_toggled(self, state):
         """
@@ -147,6 +159,16 @@ class HostRequirementsWidget(QWidget):  # pylint: disable=too-few-public-methods
             requirements.setdefault("attributes", []).extend(custom_requirements["attributes"])  # type: ignore
 
         return requirements
+
+    def set_requirements(self, requirements: HostRequirements):
+        if requirements.custom_requirements:
+            self.custom_requirements_box.set_requirements(requirements.custom_requirements)
+        if requirements.os_requirements:
+            self.os_requirements_box.set_requirements(requirements.os_requirements)
+        if requirements.hardware_requirements:
+            self.hardware_requirements_box.set_requirements(requirements.hardware_requirements)
+        self.mode_selection_box.use_default_button.setChecked(False)
+        self.mode_selection_box.use_custom_button.setChecked(True)
 
 
 class OverrideRequirementsWidget(QGroupBox):  # pylint: disable=too-few-public-methods
@@ -201,14 +223,17 @@ class OSRequirementsWidget(QGroupBox):
         parent: The parent Qt Widget.
     """
 
+    OS_ROW_OPTIONS = ["linux", "macos", "windows"]
+    CPU_ROW_OPTIONS = ["x86_64", "arm64"]
+
     def __init__(self, parent=None):
         super().__init__("", parent=parent)
         self.layout = QVBoxLayout(self)
         self._build_ui()
 
     def _build_ui(self):
-        self.os_row = OSRequirementRowWidget("Operating system", ["linux", "macos", "windows"])
-        self.cpu_row = OSRequirementRowWidget("CPU architecture", ["x86_64", "arm64"])
+        self.os_row = OSRequirementRowWidget("Operating system", self.OS_ROW_OPTIONS)
+        self.cpu_row = OSRequirementRowWidget("CPU architecture", self.CPU_ROW_OPTIONS)
 
         self.layout.addWidget(self.os_row)
         self.layout.addWidget(self.cpu_row)
@@ -225,21 +250,25 @@ class OSRequirementsWidget(QGroupBox):
 
         # TODO: currently only supports "AnyOf" from the UI
         requirements: List[dict] = []
-        if self.os_row.combo_box.has_input():
+        if self.os_row.combo_box.checkedItems():
             requirements.append(
                 {
                     "name": "attr.worker.os.family",
-                    "anyOf": [self.os_row.combo_box.currentText()],
+                    "anyOf": self.os_row.combo_box.checkedItems(),
                 }
             )
-        if self.cpu_row.combo_box.has_input():
+        if self.cpu_row.combo_box.checkedItems():
             requirements.append(
                 {
                     "name": "attr.worker.cpu.arch",
-                    "anyOf": [self.cpu_row.combo_box.currentText()],
+                    "anyOf": self.cpu_row.combo_box.checkedItems(),
                 }
             )
         return requirements
+
+    def set_requirements(self, requirements: OsRequirements):
+        self.os_row.combo_box.setChecked(requirements.operating_systems)  # type: ignore
+        self.cpu_row.combo_box.setChecked(requirements.cpu_archs)  # type: ignore
 
 
 class HardwareRequirementsWidget(QGroupBox):  # pylint: disable=too-few-public-methods
@@ -292,6 +321,28 @@ class HardwareRequirementsWidget(QGroupBox):  # pylint: disable=too-few-public-m
         # Disk Scratch capability has unit in GiB
         self.scratch_space_row.add_requirement(requirements, "amount.worker.disk.scratch")
         return requirements
+
+    def set_requirements(self, requirements: HardwareRequirements):
+        self.cpu_row.clear()
+        self.memory_row.clear()
+        self.gpu_row.clear()
+        self.gpu_memory_row.clear()
+        self.scratch_space_row.clear()
+
+        self.cpu_row.minimum = requirements.cpu_min
+        self.cpu_row.maximum = requirements.cpu_max
+
+        self.memory_row.minimum = requirements.memory_min // 1024
+        self.memory_row.maximum = requirements.memory_max // 1024
+
+        self.gpu_row.minimum = requirements.acceleration_min
+        self.gpu_row.maximum = requirements.acceleration_max
+
+        self.gpu_memory_row.minimum = requirements.acceleration_memory_min // 1024
+        self.gpu_memory_row.maximum = requirements.acceleration_memory_max // 1024
+
+        self.scratch_space_row.minimum = requirements.scratch_space_min
+        self.scratch_space_row.maximum = requirements.scratch_space_max
 
 
 class CustomRequirementsWidget(QGroupBox):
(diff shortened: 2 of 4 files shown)


src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -239,6 +277,23 @@ def get_requirements(self) -> List[Dict[str, Any]]:
...
 
+    def set_requirements(self, requirements: List[Dict[str, Any]]):
+        for requirement in requirements:
+            if requirement.get("name") == "attr.worker.os.family":
+                if requirement.get("anyOf") not in self.OS_ROW_OPTIONS:
+                    raise ValueError(
+                        "Invalid value {} for option attr.worker.os.family".format(requirement.get("anyOf"))
+                    )
+                self.os_row.combo_box.setCurrentText(requirement.get("anyOf"))
+            if requirement.get("name") == "attr.worker.cpu.arch":
```
joel-wong-aws: I believe this should be `elif` instead of `if`? Otherwise the `else` condition will be taken if the requirement name is `"attr.worker.os.family"`?

joel-wong-aws (changes requested):

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: You are correct! Fixing now

miabatta (changes requested):
Thanks so much for posting! I have a few notes while this is still WIP.
- Running `hatch run fmt` and `hatch run lint` shows some quick fixes to be made, particularly making sure `host_requirements` and its members aren't `None` before operating on them.
- Could you describe how you've tested this so far? If I'm reading your description right, you intend to add unit tests, which will be great! If you've been testing manually in the meantime, I'm curious about you've set that up.

Ahuge:
> Thanks so much for posting! I have a few notes while this is still WIP.
> 
> * Running `hatch run fmt` and `hatch run lint` shows some quick fixes to be made, particularly making sure `host_requirements` and its members aren't `None` before operating on them.
> * Could you describe how you've tested this so far? If I'm reading your description right, you intend to add unit tests, which will be great! If you've been testing manually in the meantime, I'm curious about you've set that up.

Thanks for the feedback
You are right there are quite a lot of typehinting linter errors.
  
I think I am going to look into implementing a dataclass instead of building typehinting for what is essentially a `Dict[str,List[str]]`
Give me a day or two to dig into implementing that and I will come back, hopefully with something a bit more structured.

The testing I have done has been in an internal "Callbacks feature" that I have been developing and discussing with Morgan on.
It is not exhaustive which is why I want to add some formal tests to it as well.

Thanks!

miabatta:
Interesting- sounds good to me! I think a dataclass is a solid plan; it'll be much easier to work with in the individual DCCs' submitters.  
Looking forward to it, and thank _you_ for the contribution!

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -146,6 +146,42 @@ def get_requirements(self) -> Optional[Dict[str, Any]]:
 
         return requirements
 
+    def set_requirements(self, requirements: Dict[str, Any]):
```
joel-wong-aws: I think you've generally got the right approach with all of these values flowing down into subcomponents and being handled there.

Here's what I _think_ we want as our logic for OS requirements. Does this align with your thoughts?
1. If there is no input OS attribute value (e.g. `"attr.worker.os.family"` is not input), clear the field
2. If there is an input OS attribute value (e.g. `"attr.worker.os.family"` is `"linux"`), use that value

Currently, I don't think this code block quite captures all the logic for that. For one, the `os_requirements` returned by `self.os_requirements_box.get_requirements()` would be empty when first creating the host requirements, right? So this might miss those.

I think what we need to do is make constants for the attr names that are special e.g. `"attr.worker.os.family", "attr.worker.cpu.arch"` and filter those out specifically.

Similar note for the hardware requirements.

I think it's doing the right thing for custom requirements, awesome work there :)

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -431,6 +506,39 @@ def get_requirements(self) -> Dict[str, List]:
 
         return requirements
 
+    def set_requirements(self, requirements: Dict[str, List]):
+        amount_widgets = {}
+        attribute_widgets = {}
+        for i in range(self.list_widget.count()):
+            widget = self.list_widget.itemWidget(self.list_widget.item(i))
+            widget_requirement = widget.get_requirement()
+            if widget_requirement:
```
joel-wong-aws: Similar to my other comment, I think we've generally got the right idea here, but there are some small updates needed.

I think our desired behaviour should be:
1. If there is an input custom requirement, we should have a custom widget for it, even if it doesn't already exist in the UI
2. If there is a custom requirement in the UI that isn't in the input, then we should remove the widget for it.

Based on my understanding of this code, if a widget does not already exist for a custom capability (which would be the case for all custom capabilities when the host requirements are first created), then we'll just get a warning and no widget will be created, right? And vice versa - if there is a widget that already exists for a custom attribute, it won't be deleted, even if it's not in the input.

Thoughts?

joel-wong-aws:
Thank you so much for your contribution @Ahuge! I've left a few comments. I think this is generally is the right approach and just needs a few tweaks to work optimally :)

Ahuge:
Ok weirdly I cannot reproduce the linter errors locally but I fixed them based off the error logs

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,287 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+    """
+
+    DEFAULT_VALUE = -(2**31) + 1
+    MAX_INT_VALUE = (2**31) - 1
+    name: str = field(default="")
+    min: int = field(default=DEFAULT_VALUE)
+    max: int = field(default=DEFAULT_VALUE)
+
+    def serialize(self) -> dict:
+        return {"name": self.name, "min": self.min, "max": self.max}
```
joel-wong-aws: Does this need the same default exists logic as for other amount attributes?

```
requirement = {
    "name": self.name,
}
if self.min != self.DEFAULT_VALUE:
    requirement["min"] = max(self.min, self.DEFAULT_VALUE)
if self.max != self.DEFAULT_VALUE:
    requirement["max"] = min(self.max, self.MAX_INT_VALUE)
return requirement
```

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,287 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+class CustomAttributeRequirement:
+    """
+    Settings for a CustomAttributeWidget
+    """
+
+    ANY_OF = "anyOf"
+    ALL_OF = "allOf"
+    OPTIONS = [ANY_OF, ALL_OF]
+    name: str = field(default="")
+    option: Literal[ANY_OF, ALL_OF] = field(default=ANY_OF)  # type: ignore
```
joel-wong-aws: nit: can we make this match the UI default of `ALL_OF`?

Ahuge:
It looks like on MacOS the linter errors are more strict around the PySide.Slot decorator.
However I don't believe I need an explicit slot decorator as we are just in pure python so I have removed them.

I believe the linters should be happy now

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,274 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+    ) -> CustomAttributeRequirement:
+        if isinstance(attribute, dict):
+            attribute = CustomAttributeRequirement(**attribute)
+
+        if not attribute.name:
+            raise ValueError(f"Custom Amount {attribute} has no name")
+        elif not attribute.option or attribute.option not in CustomAttributeRequirement.OPTIONS:
+            raise ValueError(
+                f"Custom Amount {attribute} option is not in {CustomAttributeRequirement.OPTIONS}"
+            )
```
joel-wong-aws: Would it ve better to do this validation in the `CustomAttributeRequirement`?

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,274 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+
+    @staticmethod
+    def _validate_amount(amount: Union[dict, CustomAmountRequirement]) -> CustomAmountRequirement:
+        if isinstance(amount, dict):
+            amount = CustomAmountRequirement(**amount)
+
+        if not amount.name:
+            raise ValueError(f"Custom Amount {amount} has no name")
+        elif amount.min and amount.max and amount.min > amount.max:
+            raise ValueError(f"Custom Amount {amount} has min higher than max")
```
joel-wong-aws: Would it be better to add this validation in the `CustomAmountRequirement`?

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,274 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+            if not amount.name:
+                raise ValueError(f"Custom Amount {amount} has no name")
+            _requirement = {
+                "name": "amount.worker." + amount.name,
+            }
+            if amount.min != CustomAmountRequirement.DEFAULT_VALUE:
+                _requirement["min"] = max(amount.min, CustomAmountRequirement.DEFAULT_VALUE)
+            if amount.max != CustomAmountRequirement.DEFAULT_VALUE:
+                _requirement["max"] = min(amount.max, CustomAmountRequirement.MAX_INT_VALUE)
+            requirements.append(_requirement)
```
joel-wong-aws: This looks like a duplicate of the logic in `CustomAmountRequirement`, can we reuse that?

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,274 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+            }
+            if not attribute.option:
+                raise ValueError(f"Custom Attribute {attribute} has no option")
+            elif attribute.option not in ["anyOf", "allOf"]:
+                raise ValueError(
+                    f'Custom Attribute {attribute} has option {attribute.option} which is not in list: ["anyOf", "allOf"]'
+                )
+            if not attribute.values:
+                raise ValueError(f"Custom Attribute {attribute} has no values")
+            _requirement[attribute.option] = attribute.values
```
joel-wong-aws: This looks like a duplicate of the logic in `CustomAmountRequirement`, can we reuse that?

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -527,6 +612,28 @@ def _build_ui(self):
         # LineEdit  / LineEdit / Optional [X]
         self.layout.addLayout(self.columns)
 
+    def name(self) -> str:
```
joel-wong-aws, line 606: Nice to have: could we use `@property` and `@setter` methods here for all these properties ([ref](https://docs.python.org/3/library/functions.html#property))? Then we can just do
`amount.name = "amount.custom.test"`, and retrieve the value via `amount.name`. Similar note for the min and max values

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -693,6 +802,67 @@ def _set_remove_button_for_first_item(self):
             first_item = self.value_list_widget.itemWidget(self.value_list_widget.item(0))
             first_item.remove_button.setEnabled(True)
 
+    def name(self) -> str:
```
joel-wong-aws, line 802: Similar note here about `@property` and `@setter` methods

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -843,6 +1010,18 @@ def add_requirement(self, requirements: List, name: str, scaling_factor: int = 1
...
+            self.min_spin_box.setValue(self.min_spin_box.no_input_value)
+        else:
+            self.min_spin_box.setValue(minimum_value // scaling_factor)
+
+    def set_maximum(self, maximum_value: int, scaling_factor: int = 1):
+        if maximum_value == self.max_spin_box.no_input_value:
+            self.max_spin_box.setValue(self.max_spin_box.no_input_value)
+        else:
+            self.max_spin_box.setValue(maximum_value // scaling_factor)
+
```
joel-wong-aws: Could we use `@setter` here?

test/unit/deadline_client/ui/test_host_requirements_dataclass_serialize.py
```diff
@@ -0,0 +1,178 @@
...
+    ],
+    "amounts": [
+        {"name": "amount.worker.vcpu", "min": 8, "max": 64},
+        {"name": "amount.worker.memory", "min": 16384, "max": 131072},
+        {"name": "amount.worker.Bugs", "min": 1, "max": 10},
+    ],
+}
+
+
+def _compare_requirements(first, second):
```
joel-wong-aws, line 32: A cool, simple way to do this could be:
```
sorted_first_attributes = sorted(first.get("attributes". []), key=lambda x: x["name"])
sorted_second_attributes = sorted(second.get("attributes". []), key=lambda x: x["name"])
assert sorted_first_attributes == sorted_second_attributes

sorted_first_amounts = sorted(first.get("amounts". []), key=lambda x: x["name"])
sorted_second_amounts = sorted(second.get("amounts". []), key=lambda x: x["name"])
assert sorted_first_amounts == sorted_second_amounts
```

test/unit/deadline_client/ui/test_host_requirements_dataclass_serialize.py
```diff
@@ -0,0 +1,178 @@
...
+                    CustomAttributeRequirement(
+                        name="pipelineFeatures",
+                        option=CustomAttributeRequirement.ANY_OF,
+                        values=["feature1", "feature2"],
+                    ),
+                ],
+            ),
+        )
+        requirements.serialize()
+    except ValueError:
```
joel-wong-aws: Instead of this try-except, does `pytest.raises(ValueError):` work? It's a little bit simpler and has better messaging if it fails.

Similar note for the next test.

src/deadline/client/ui/dataclasses/__init__.py
Ahuge: Good point!

src/deadline/client/ui/dataclasses/__init__.py
Ahuge: Good point!

src/deadline/client/ui/dataclasses/__init__.py
Ahuge: Good point!

src/deadline/client/ui/dataclasses/__init__.py
Ahuge: Good point!

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: Good idea, working on that now

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: Good idea, working on that now

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: I think a setter here could be possible, but we'd have to do the scaling factor work outside of the setter. How do we feel about that?

test/unit/deadline_client/ui/test_host_requirements_dataclass_serialize.py
Ahuge: Good catch, I am less familiar with pytest so I wasn't aware of that

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,294 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+    ARM_64 = "arm64"
+    CPU_ARCHITECTURES = [X86_64, ARM_64]
+    operating_systems: List[Literal[LINUX, MACOS, WINDOWS]] = field(default_factory=list)  # type: ignore
+    cpus: List[Literal[X86_64, ARM_64]] = field(default_factory=list)  # type: ignore
+
+    def __post_init__(self):
+        for operating_system in self.operating_systems:
+            if operating_system not in self.OPERATING_SYSTEMS:
+                raise ValueError(
+                    f"Operating system {operating_system} is in supported list: {self.OPERATING_SYSTEMS}."
```
miabatta: nit: "Operating system is not* in supported list"

I noticed this a few times in other functions, so just a quick once-over to correct them would be great!

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +67,294 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+
+
+@dataclass
+class HardwareRequirements:
+    """
+    Settings for a HardwareRequirementsWidget.
+    """
+
+    DEFAULT_VALUE = -1
+    MAX_INT_VALUE = (2**31) - 1
```
miabatta: We can punt this comment to a later PR, but I'd be in favour of moving some of these dataclasses out of `__init__.py` for cleanliness/readability. That way we could also just define `MAX_INT_VALUE` at the top of some file rather than defining it multiple times in each of these classes.

src/deadline/client/ui/widgets/host_requirements_tab.py
miabatta: I think that's fine. If the calling method is already defining the scaling factor itself, it shouldn't be an issue to do the scaling there instead- using `@setter` is also better for consistency

miabatta (changes requested):
A few comments, but looking good overall! I'm also going to try testing this with one of the integrated submitters to verify it plays nicely; if anything comes up I'll leave another comment.

edit: Also needs a rebase but we can do that afterwards

src/deadline/client/ui/dataclasses/__init__.py
Ahuge: Good catch thanks

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: Sounds good

src/deadline/client/ui/dataclasses/__init__.py
Ahuge: I agreed, some sort of `constants.py` where we can reference them.

I've moved them to the top of the init

miabatta (dismissed):
Nothing further from me! Thanks so much for getting this done.

We'll have to rebase it onto mainline before merging, then squash & make sure the final commit message uses conventional commits. Once Joel gives his feedback we can get more into that.

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -431,6 +475,47 @@ def get_requirements(self) -> Dict[str, List]:
...
+            elif requirement.name in widget_requirements_by_name and not isinstance(
+                widget_requirements_by_name[requirement.name], CustomAmountWidget
+            ):
+                raise ValueError(
+                    'Custom requirement with name {name} already exists and is not an "amount" requirement.'
+                )
+            else:
+                amount_widget = widget_requirements_by_name[requirement.name]
+                amount_widget.minimum = requirement.min
+                amount_widget.maximum = requirement.max
```
joel-wong-aws: Instead of this approach which retains existing host requirements (if they aren't overwritten by new requirements), can we just `self.list_widget.clear()` and then always add new custom values as new widgets? That will provide a clean slate between different job bundles (e.g. if we use the existing "load job bundle" functionality).

Similar comment for the attributes logic below.

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -431,6 +475,47 @@ def get_requirements(self) -> Dict[str, List]:
 
         return requirements
 
+    def set_requirements(self, requirements: CustomRequirements):
```
joel-wong-aws, line 490: Type annotation should be `def set_requirements(self, requirements: Optional[CustomRequirements]):` as that's the input passed in when `host_requirements = HostRequirements(os_requirements=OsRequirements(operating_systems=["windows"]))`

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -431,6 +475,47 @@ def get_requirements(self) -> Dict[str, List]:
...
+        widget_requirements_by_name = {}
+        for i in range(self.list_widget.count()):
+            widget = self.list_widget.itemWidget(self.list_widget.item(i))
+            widget_requirements_by_name[widget.name] = widget
+
```
joel-wong-aws, line 492: When I input `HostRequirements(os_requirements=OsRequirements(operating_systems=["windows"]))` (like I would hope to do for Cinema 4D, for example), I get the following error:

```
...
self.custom_requirements_box.set_requirements(requirements.custom_requirements)
for requirement in requirements.amounts
AttributeError: 'NoneType' object has no attribute 'amounts'
```

We are assuming here that `requirements` is provided, but it could be `None`, in which case this block and subsequent blocks should be skipped.

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -291,6 +316,22 @@ def get_requirements(self) -> List[Dict[str, Any]]:
         self.scratch_space_row.add_requirement(requirements, "amount.worker.disk.scratch")
         return requirements
 
+    def set_requirements(self, requirements: HardwareRequirements):
```
joel-wong-aws, line 325: When I input `HostRequirements(os_requirements=OsRequirements(operating_systems=["windows"]))` (like I would hope to do for Cinema 4D, for example), I get the following error:
```
...
self.hardware_requirements_box.set_requirements(requirements.hardware_requirements)
...
AttributeError: 'NoneType' object has no attribute 'cpu_min'
```

I think we need an
```
if not requirements:
    self.cpu_row.clear()
    self.memory_row.clear()
    self.gpu_row.clear()
    self.gpu_memory_row.clear()
    self.scratch_space_row.clear()
    return
```
statement here

and then we can define `clear()` on `HardwareRequirementsWidget` as
```
    def clear(self):
        self.min_spin_box.setValue(self.min_spin_box.no_input_value)
        self.max_spin_box.setValue(self.max_spin_box.no_input_value)
        
```


and then the type definition should be
`def set_requirements(self, requirements: Optional[HardwareRequirements]):`

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -223,22 +244,26 @@ def get_requirements(self) -> List[Dict[str, Any]]:
...
                 {
                     "name": "attr.worker.cpu.arch",
-                    "anyOf": [self.cpu_row.combo_box.currentText()],
+                    "anyOf": self.cpu_row.combo_box.checkedItems(),
                 }
             )
         return requirements
 
+    def set_requirements(self, requirements: OsRequirements):
+        self.os_row.combo_box.setChecked(requirements.operating_systems)  # type: ignore
```
joel-wong-aws, line 270: Similar to the `set_requirements` of the `HardwareRequirements` and `CustomRequirements`, we should have an
```
if not requirements:
    self.os_row.combo_box.setChecked([])
    self.cpu_row.combo_box.setChecked([])
    return
```
statement at the top of this

and the type annotation should be
```
def set_requirements(self, requirements: Optional[OsRequirements]):
```

joel-wong-aws (changes requested):
I gave this a spin and found a few bugs - added some comments :)

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -223,22 +244,30 @@ def get_requirements(self) -> List[Dict[str, Any]]:
...
             requirements.append(
                 {
                     "name": "attr.worker.cpu.arch",
-                    "anyOf": [self.cpu_row.combo_box.currentText()],
+                    "anyOf": self.cpu_row.combo_box.checkedItems(),
                 }
             )
         return requirements
 
+    def set_requirements(self, requirements: Optional[OsRequirements] = None):
```
joel-wong-aws: Can we require the `requirements` argument? Otherwise we allow `os_requirements_widget.set_requirements()`, which could be confusing. Same note for the `HardwareRequirementsWidget` and `CustomRequirementsWidget` `set_requirements` functions

joel-wong-aws (changes requested):

joel-wong-aws:
One small comment and then it looks like the DCO approval workflow is failing, so you'll need to sign off all commits. Otherwise, LGTM!

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: Can do!

Ahuge:
Fixed, Rebased, and Squashed.

Please let me know if the commit messages need changing

Thanks!

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +69,292 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+
+    LINUX = "linux"
+    MACOS = "macos"
+    WINDOWS = "windows"
+    OPERATING_SYSTEMS = [LINUX, MACOS, WINDOWS]
+    X86_64 = "x86_64"
+    ARM_64 = "arm64"
+    CPU_ARCHITECTURES = [X86_64, ARM_64]
+    operating_systems: List[Literal[LINUX, MACOS, WINDOWS]] = field(default_factory=list)  # type: ignore
+    cpus: List[Literal[X86_64, ARM_64]] = field(default_factory=list)  # type: ignore
```
joel-wong-aws: Small ask: can we rename this to `cpu_archs`? When I saw `cpus` in the operating system options, I was confused because I thought it was referring to the number of cpus.

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -710,18 +876,15 @@ def get_requirement(self) -> Dict[str, Any]:
...
+            try:
+                values = self.values
+            except ValueError:
+                values = []
+                requirements_are_valid = False
             if values:
                 requirement = {
                     "name": ATTRIBUTE_CAPABILITY_PREFIX + self.name_line_edit.text(),
-                    f"{option}": values,
+                    f"{self.option()}": values,
```
joel-wong-aws: This should be `self.option` instead of `self.option()`

src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -433,6 +487,21 @@ def get_requirements(self) -> Dict[str, List]:
...
         return requirements
 
+    def set_requirements(self, requirements: CustomRequirements):
+        self.list_widget.clear()
+
+        for requirement in requirements.amounts:
+            amount_widget: CustomAmountWidget = self._add_new_custom_amount()
+            amount_widget.name = requirement.name
+            amount_widget.minimum = requirement.min
+            amount_widget._maximum = requirement.max
```
joel-wong-aws: Typo: `amount_widget._maximum` -> `amount_widget.maximum`

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +69,292 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+    max: int = field(default=DEFAULT_VALUE)
+
+    def __post_init__(self):
+        if not self.name:
+            raise ValueError(f"Custom Amount {self} has no name")
+        elif self.min and self.max and self.min > self.max:
+            raise ValueError(f"Custom Amount {self} has min higher than max")
+
+    def serialize(self) -> dict:
+        _requirement = {"name": "amount.worker." + self.name}
```
joel-wong-aws, line 219: This is mostly an existing bug, but can you replace the 2 references of `"amount.worker."` -> `"amount."`? The `amount.worker` namespace is reserved in the OpenJD standard, so this current prefix is not really usable - I can't really add `amount.custom`, for example.

src/deadline/client/ui/dataclasses/__init__.py
```diff
@@ -66,3 +69,292 @@ class CliJobSettings:  # pylint: disable=too-many-instance-attributes
...
+        elif not self.option or self.option not in CustomAttributeRequirement.OPTIONS:
+            raise ValueError(
+                f"Custom Amount {self} option is not in {CustomAttributeRequirement.OPTIONS}"
+            )
+
+    def serialize(self) -> dict:
+        if not self.name:
+            raise ValueError(f"Custom Attribute {self} has no name")
+        _requirement = {
+            "name": "attr.worker." + self.name,
```
joel-wong-aws, line 252: Similar to the comment for `amount.worker.`, we should replace the 2 references to `"attr.worker."` -> `"attr."` for custom attributes.

joel-wong-aws (changes requested):
Sorry for stepping back a bit, but I did some more testing today and spotted a few more bugs - do you mind adding those fixes in?

src/deadline/client/ui/dataclasses/__init__.py
Ahuge: Good idea

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: Thanks!

src/deadline/client/ui/widgets/host_requirements_tab.py
Ahuge: Thanks!

Ahuge:
Updated with latest fixes, thanks @joel-wong-aws 

Once you give the go ahead I'll squash and signoff

miabatta:
> Updated with latest fixes, thanks @joel-wong-aws
> 
> Once you give the go ahead I'll squash and signoff

As a heads-up, I believe signing off changes the commit messages, so you may need to force-push to update them which would require another round of approvals. I'll be on the lookout for notifs so I don't mind- just letting you know!

miabatta (dismissed):

joel-wong-aws:
Looks like with the change from `cpus` -> `cpu_archs` we'll have to update the unit tests too :(

Thank you so much for your patience with all the revisions!

joel-wong-aws (approved):

joel-wong-aws:
Awesome job with all of this @Ahuge! Ship it!

miabatta (approved):
Awesome- super excited to see this go in! Thanks so much.

PR #540, opened 2024-12-19
fix: improve response time on windows when using deadline config GUI

AWS-Samuel:

On Windows, there was a 10+ second hang whenever updating settings through the `deadline config gui` GUI when `apply` or `OK` were clicked. 

This hang was caused by opening a subprocess to call `icacls` in order to set and read directory permissions. 

This lead to discovering that we were erroneously resetting the permissions on that directory on each config change.

We only need to set directory permissions when creating the directory, so add logic to detect if we are creating the directory.

Also switch to using the `win32security` module to set permissions instead of opening a subprocess that calls `icacls`. This is faster and doesn't rely on interpreting the STDOUT of another process. 

The GUI is more responsive when applying or saving settings. 

Yes
Yes
No

I also tested manually by comparing the response time when using the config GUI before and after this change. 
  - Before the change there was a 10ish second hang when clicking `Apply` before the GUI could be interacted with again. 
  - After the change the GUI was immediately responsive (at least, as fast as I could move my mouse from `apply` to another GUI element)

For the permissions change, I added unit tests, and also manually tested by:
- Deleting my `~/.deadline` folder, and running `deadline config gui` to initialize the directory + the config file.
   - Confirmed that the directory had permissions for the `SYSTEM` account, domain Administrators, and my logged in username
- Modified permissions on the directory, and re-ran `deadline config gui`. 
   - Confirmed that the directory permissions had not changed, and that my modified permissions remained intact.  

  - No relevant docstrings
  - No README.md changed to make
  
This change is not a breaking change

Yes-ish. 

It modifies a part of the code where we set file permissions, and removes the part where we always overwrite existing file permissions. 

The actual permissions which are set have not been changed. I manually validated this by creating the directory through `deadline config gui`, and confirming that the permission on the folder where identical upon first creation both before and with this change. 
----

### DIFF
--- a/test/unit/deadline_client/config/test_config_file.py
+++ b/test/unit/deadline_client/config/test_config_file.py
@@ -6,8 +6,9 @@ tests the deadline.client.config settings
 
 import os
 import platform
-import sys
-from unittest.mock import patch, MagicMock, call
+import getpass
+import tempfile
+from unittest.mock import patch, MagicMock
 from pathlib import Path
 
 import boto3  # type: ignore[import]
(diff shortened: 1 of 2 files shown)


test/unit/deadline_client/config/test_config_file.py
```diff
@@ -273,32 +273,84 @@ def test_default_log_level():
...
+    """
+    Asserts the _reset_directory_permissions_windows configures the provided
+    folder with access only to the active user, the domain admin, and SYSTEM.
+    """
+    # GIVEN
+    import ntsecuritycon
 
-    config_file.set_setting("defaults.aws_profile_name", "goodguyprofile")
+    win32security_mock = MagicMock()
+    sys.modules["win32security"] = win32security_mock
```
mwiebe: Nit: I don't think this functionality needs mocking-based tests at all. It could test within a temporary directory to create a few directories with the wrong permissions, and validate that the function resets them successfully. This would decouple the tests from the implementation details and let those change more easily later.

mwiebe (dismissed):

test/unit/deadline_client/config/test_config_file.py
AWS-Samuel: Changed the test to remove the file system mocks and instead validate that the permissions are what we expect.

epmog (approved):

mwiebe (approved):

PR #547, opened 2025-01-09
feat: allow windows long paths during manifest snapshot, manifest upload, and download

YutongLi291:
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later

There are 3 closely related edge case issues.

When customers:
* Use the deadline CLI to snapshot a directory and create a manifest: there is a possibility that due to the length of the resulting manifest path exceeds Windows path limits, causing the file to not be able to be created
* Upload the manifest to the CAS storage: there is a possibility that the manifest path exceeds Windows path limits, causing the file not to be able to be uploaded
* Download outputs of a job: the download path on the customer's computer could exceed Windows path limits due to the length of the resulting file path, causing the file not to be able to be downloaded

In all 3 of these cases, the customer has an existing path that barely does not exceed the path limit, which is why these are edge cases. If they already have an existing path that exceeds the path limit, then they would have already had the windows long path registry turned on allowing long paths across their entire workstation.

Prepend `\\?\` during manifest snapshot, manifest upload, and job output download when needed so that when the path exceedes the windows path limit (260 characters), the path still works properly and we are able to snapshot/upload/download.
Manifest snapshot, manifest upload, and file download will now work on Windows workstations even if the file exceeds 260 characters.

  that you ensure that the docker-based unit tests pass. yes

Also tested on a windows local computer to make sure that with long paths, manifest upload/snapshot and download now work. Furthermore, the warning message shows up for manifest snapshot and job output download.

```
WARNING: Manifest file path <path hidden in PR> exceeds Windows path length limit. This may cause unexpected issues.
For details and a fix using the registry, see: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
```

No

No

No

----


src/deadline/job_attachments/_utils.py
```diff
@@ -22,6 +22,11 @@
 ]
 
 
+TEMP_DOWNLOAD_ADDED_CHARS_LENGTH = 9
```
leongdl, line 24: Why is this "9"? Can we do a length of some string?

src/deadline/job_attachments/_utils.py
```diff
@@ -105,6 +110,28 @@ def _is_windows_long_path_registry_enabled() -> bool:
...
+    # Given a Path or string representing a path,
+    # make it long path compatible if needed on Windows and return the Path object
+    # https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+
+    original_path_string = str(original_path)
+    if sys.platform != "win32":
+        return Path(original_path_string)
+
+    # Add 9 to account for .Hex value when file is in the middle of downloading in windows paths
+    # For example: file test.txt when downloaded will become test.txt.H4SD9Ddj
```
leongdl: This comment would be great for the constant above to explain "9"

src/deadline/job_attachments/_utils.py
```diff
@@ -105,6 +110,28 @@ def _is_windows_long_path_registry_enabled() -> bool:
...
+    if (
+        len(original_path_string) + TEMP_DOWNLOAD_ADDED_CHARS_LENGTH >= WINDOWS_MAX_PATH_LENGTH
+        and not original_path_string.startswith("\\\\?\\")
+        and not _is_windows_long_path_registry_enabled()
+    ):
+        # Prepend \\?\ to the file name to treat it as an UNC path
```
leongdl, line 147: If we had to do this, maybe we should also print a log? This is a helper and ideally the customer has enabled the registry?

src/deadline/job_attachments/_utils.py
```diff
@@ -105,6 +110,28 @@ def _is_windows_long_path_registry_enabled() -> bool:
     return bool(ntdll.RtlAreLongPathsEnabled())
 
 
+def _get_long_path_compatible_path(original_path: Union[str, Path]) -> Path:
+    # Given a Path or string representing a path,
+    # make it long path compatible if needed on Windows and return the Path object
+    # https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+
+    original_path_string = str(original_path)
+    if sys.platform != "win32":
```
jairaws, line 139: nit: Would be better to use a constant.

src/deadline/job_attachments/_utils.py
```diff
@@ -105,6 +110,28 @@ def _is_windows_long_path_registry_enabled() -> bool:
...
+        return Path(original_path_string)
+
+    # Add 9 to account for .Hex value when file is in the middle of downloading in windows paths
```
jairaws: Add 9 characters?  IMO, this comment is not clear.

src/deadline/job_attachments/download.py
```diff
@@ -520,25 +525,6 @@ def process_client_error(exc: ClientError, status_code: int):
...
-                    + "This could be the error if you do not enable longer file path in Windows"
-                )
-            elif not _is_windows_long_path_registry_enabled():
-                # Path start with \\?\ but do not enable registry -> Undefined error
-                raise AssetSyncError(
-                    f"{e}\nUNC notation exist, but long path registry not enabled. Undefined error"
-                ) from e
-
-        # Path start with \\?\ and registry is enable, something else cause this error
         raise AssetSyncError(e) from e
```
leongdl, line 529: What is `e` here? Do we want to customize the message?

src/deadline/job_attachments/_utils.py
```diff
@@ -105,6 +110,28 @@ def _is_windows_long_path_registry_enabled() -> bool:
...
+    # For example: file test.txt when downloaded will become test.txt.H4SD9Ddj
+    if (
+        len(original_path_string) + TEMP_DOWNLOAD_ADDED_CHARS_LENGTH >= WINDOWS_MAX_PATH_LENGTH
+        and not original_path_string.startswith("\\\\?\\")
+        and not _is_windows_long_path_registry_enabled()
+    ):
+        # Prepend \\?\ to the file name to treat it as an UNC path
+
+        return Path("\\\\?\\" + original_path_string)
```
jairaws: We use `\\\\?\\` many times. It should be a constant.

src/deadline/job_attachments/_utils.py
YutongLi291: This constant was kept over from previous PRs that were merged regarding long paths. I can change it

src/deadline/job_attachments/download.py
YutongLi291: It just wraps the error in an AssetSyncError. Since long path is supported now we don't really want to customize the message anymore

src/deadline/job_attachments/_utils.py
YutongLi291: I'm going to pass on that for now in this PR since there are actually a lot of strings that could be made into constants, but we don't have a common constants file in this repo.

jusiskin:
I would like to clarify the intention of the change here. This change is supposed to let someone using job attachments to upload and download job attachments on a Windows machine with long file path support not turned on, right?

Will those same people have issues later accessing the downloaded output job attachments or cleaning up manifest files? If we are creating files that people will have difficulty accessing, wouldn't that create confusion? The people using this might rightfully believe that the feature is not working properly.

Would it make more sense to detect long file path is not turned on on the host machine and fail fast with a helpful actionable error message?

YutongLi291:
> I would like to clarify the intention of the change here. This change is supposed to let someone using job attachments to upload and download job attachments on a Windows machine with long file path support not turned on, right?
> 
> Will those same people have issues later accessing the downloaded output job attachments or cleaning up manifest files? If we are creating files that people will have difficulty accessing, wouldn't that create confusion? The people using this might rightfully believe that the feature is not working properly.
> 
> Would it make more sense to detect long file path is not turned on on the host machine and fail fast with a helpful actionable error message?

Yes, that is exactly the intention. The interesting thing is that although the downloaded file exceeds the long path limit, users are still able to view/modify/delete the file contents in their file explorer properly. So it should not be an issue with allowing this.

jusiskin:
> > I would like to clarify the intention of the change here. This change is supposed to let someone using job attachments to upload and download job attachments on a Windows machine with long file path support not turned on, right?
> > Will those same people have issues later accessing the downloaded output job attachments or cleaning up manifest files? If we are creating files that people will have difficulty accessing, wouldn't that create confusion? The people using this might rightfully believe that the feature is not working properly.
> > Would it make more sense to detect long file path is not turned on on the host machine and fail fast with a helpful actionable error message?
> 
> Yes, that is exactly the intention. The interesting thing is that although the downloaded file exceeds the long path limit, users are still able to view/modify/delete the file contents in their file explorer properly. So it should not be an issue with allowing this.

I suppose it makes sense that Windows would let you interact with files beyond the path limit in the file explorer regardless of whether long path support is turned on. But still, if the customer then opens the output job attachment files in another application, it is likely that application will fail to access those files unless that application uses the same `\\?\` prefixing behaviour. Depending on how the error is surfaced in the application, the person experiencing this may believe that Deadline Cloud is corrupting the files.

Are we convinced this is the right behaviour? It seems to be allowing customers to get into confusing territory. It might be helpful to warn (and/or prompt for confirmation) in this situation, particularly for output job attachments, where we can anticipate customers getting into trouble. What do you think?

YutongLi291:
> > > I would like to clarify the intention of the change here. This change is supposed to let someone using job attachments to upload and download job attachments on a Windows machine with long file path support not turned on, right?
> > > Will those same people have issues later accessing the downloaded output job attachments or cleaning up manifest files? If we are creating files that people will have difficulty accessing, wouldn't that create confusion? The people using this might rightfully believe that the feature is not working properly.
> > > Would it make more sense to detect long file path is not turned on on the host machine and fail fast with a helpful actionable error message?
> > 
> > 
> > Yes, that is exactly the intention. The interesting thing is that although the downloaded file exceeds the long path limit, users are still able to view/modify/delete the file contents in their file explorer properly. So it should not be an issue with allowing this.
> 
> I suppose it makes sense that Windows would let you interact with files beyond the path limit in the file explorer regardless of whether long path support is turned on. But still, if the customer then opens the output job attachment files in another application, it is likely that application will fail to access those files unless that application uses the same `\\?\` prefixing behaviour. Depending on how the error is surfaced in the application, the person experiencing this may believe that Deadline Cloud is corrupting the files.
> 
> Are we convinced this is the right behaviour? It seems to be allowing customers to get into confusing territory. It might be helpful to warn (and/or prompt for confirmation) in this situation, particularly for output job attachments, where we can anticipate customers getting into trouble. What do you think?

That does make sense. I've added a warning for customers when downloading output files or when creating manifests, to let them know that they now have a file over the long path limit and there might be unexpected behaviour. 

The warning message in question is 

```
WARNING: File path {original_path_string} exceeds Windows path length limit. This may cause unexpected issues.
For details and a fix using the registry, see: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
```

Hopefully this should clear up any confusion if customers have issues with the resulting file.

src/deadline/job_attachments/_utils.py
YutongLi291: Added alongside Josh's comment

src/deadline/job_attachments/_utils.py
```diff
@@ -105,6 +124,43 @@ def _is_windows_long_path_registry_enabled() -> bool:
...
+    if (
+        len(original_path_string) + TEMP_DOWNLOAD_ADDED_CHARS_LENGTH >= WINDOWS_MAX_PATH_LENGTH
+        and not original_path_string.startswith(WINDOWS_UNC_PATH_STRING_PREFIX)
+        and not _is_windows_long_path_registry_enabled()
+    ):
+        # Prepend \\?\ to the file name to treat it as an UNC path
+
+        if show_long_path_warning:
+            # Show warning to make sure customer knows that the resulting file is in a long path
+            click.secho(
```
leongdl: This can break the --json view, can we use the ClickLogger which wraps click ?

src/deadline/job_attachments/_utils.py
```diff
@@ -105,6 +124,43 @@ def _is_windows_long_path_registry_enabled() -> bool:
     return bool(ntdll.RtlAreLongPathsEnabled())
 
 
+def _get_long_path_compatible_path(
+    original_path: Union[str, Path], show_long_path_warning: Optional[bool] = False
```
leongdl: With a default, does it need to be `Optional` ?

leongdl (commented):
nit comments only. 

Double check on the click logger - if its running in `--json` mode, the warning will now break json output.

src/deadline/job_attachments/_utils.py
YutongLi291: It can be an optional parameter only because it has a default

godobyte:
Could you also check `attachment download` and make sure it works for the long path change?

YutongLi291:
> Could you also check `attachment download` and make sure it works for the long path change?

I have checked this and it does work properly

src/deadline/job_attachments/_utils.py
YutongLi291: ClickLogger also does not support putting the warning into the resulting json, so this will not work either as it would just return two json objects which is still hard to parse. I'm looking into other solutions, or just keeping it as is. Since if the customer used the --json option, we probably do want to stop them from doing anything automatically after the long path warning, to prevent errors.

src/deadline/job_attachments/_utils.py
leongdl: OK you mean just mypy fun, got it.

jairaws (dismissed):

jairaws:
SonarCloud is failing due to high code duplication. Are you aware of this?

YutongLi291:
> SonarCloud is failing due to high code duplication. Are you aware of this?

Yes, I am aware of this. It's due to the tests.

leongdl (dismissed):

src/deadline/job_attachments/download.py
```diff
@@ -23,6 +23,7 @@
 from boto3.s3.transfer import ProgressCallbackInvoker
 from botocore.client import BaseClient
 from botocore.exceptions import BotoCoreError, ClientError
+import click
```
epmog: Adding click usage directly to the library doesn't seem right. That's specifically a CLI tool.

src/deadline/job_attachments/download.py
```diff
@@ -1106,13 +1095,35 @@ def __init__(
             session=session,
         )
 
-    def get_output_paths_by_root(self) -> dict[str, list[str]]:
+    def get_output_paths_by_root(self, is_json: bool = False) -> dict[str, list[str]]:
```
epmog: This ties to the usage of click in this file, but shouldn't whatever calls this function determine what to print and how?

test/integ/cli/test_cli_manifest_download.py
```diff
@@ -181,9 +181,9 @@ def test_manifest_download_job(
         result = runner.invoke(main, args)
 
         # Then
-        assert result.exit_code == 0, (
-            f"{result.output}, {job_attachment_test.farm_id}, {job_attachment_test.queue_id}"
-        )
+        assert (
+            result.exit_code == 0
+        ), f"{result.output}, {job_attachment_test.farm_id}, {job_attachment_test.queue_id}"
```
epmog: This may require a rebase and a `hatch env prune` for this change to disappear? We recently swapped from black to ruff for formatting.

test/unit/deadline_job_attachments/test_download.py
```diff
@@ -2375,7 +2373,7 @@ def test_download_files_from_manifests(
 
     assert merged_manifest
 
-    downloaded_files: list[str] = []
+    downloaded_files: List[str] = []
```
epmog: nit: `list[str]` is preferred, then we don't need to import List from typing.

epmog (changes requested):

src/deadline/job_attachments/download.py
YutongLi291: Addressed in new revision

src/deadline/job_attachments/download.py
YutongLi291: Addressed in newest code revision.

epmog (approved):

leongdl (approved):

PR #557, opened 2025-01-14
fix(download): expand ~ to user home dir

### LINKED ISSUE #549: Bug: `deadline job download-output` does not expand ~ when prompting for new root path
### Describe the bug

When downloading job output on a system that does not have the root path format used in the job output files, the CLI prompts the user to enter a new root path. I wanted to save it to `~/tmp` but it ended up saving it to `$PWD/~/tmp`.

```
This root path format does not match the operating system you're using. Where would you like to save the files?
The location was C:\Users\Administrator\test-resources\maya\ClientReleaseTesting, on Windows.
> Please enter a new root path: ~/tmp        

Download Summary:
    Downloaded 48 files totaling 63.37 MB.
    Total download time of 5.63655 seconds at 11.24 MB/s.
        /Users/jericht/~/~/tmp (48 files)
```

### Expected Behaviour

I expected the `~` to get expanded so that it would save the files to `/Users/jericht/tmp`, not `/Users/jericht/~/tmp`

### Current Behaviour

It looks like it treated the `~/tmp` as a relative path and prepended the current working directory

### Reproduction Steps

Run `deadline job download-output` on a job where it causes this prompt (e.g. a job submitted on Windows and downloaded to Mac)

### Environment

At minimum:
1. Operating system: Windows Server 2022 and Mac
2. Output of `python3 --version`: Python 3.9.6
3. Output of `deadline --version`: deadline, version 0.49.3
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? No

jericht:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/549

When running `deadline job download-output` on a machine that has a different root path format than the machine a job was submitted from, the CLI prompts the user to enter a new root path to download output files to. If you use a `~`, it does not properly expand that to the user home directory.

Expand `~`

CLI prompt now expands `~` to user home directory

    - PASSED
    - PASSED
    - PASSED

Manual testing on Mac. Verified it expands the `~` correctly:

```
Downloading output from Job 'AnimatedSphere.ma (some)'
This root path format does not match the operating system you're using. Where would you like to save the files?
The location was C:\Users\Administrator\test-resources\maya\ClientReleaseTesting, on Windows.
> Please enter a new root path: ~/tmp/files/new

Download Summary:
    Downloaded 24 files totaling 31.01 MB.
    Total download time of 4.42551 seconds at 7.01 MB/s.
        /Users/jericht/tmp/files/new (24 files)
```

No

No

No

----

### DIFF
--- a/src/deadline/client/cli/_groups/job_group.py
+++ b/src/deadline/client/cli/_groups/job_group.py
@@ -315,7 +315,7 @@ def _download_job_output(
                 )[0]
                 _assert_valid_path(new_root)
 
-            job_output_downloader.set_root_path(asset_root, new_root)
+            job_output_downloader.set_root_path(asset_root, os.path.expanduser(new_root))
 
     output_paths_by_root = job_output_downloader.get_output_paths_by_root()
 
--- a/test/unit/deadline_client/cli/test_cli_job.py
+++ b/test/unit/deadline_client/cli/test_cli_job.py
@@ -1003,3 +1003,93 @@ Within-session Overhead Duration Per Action: 0:00:30
 """
         )
         assert result.exit_code == 0
+
+
+@pytest.mark.usefixtures("fresh_deadline_config")
+def test_cli_job_download_output_with_different_asset_root_path_format_than_job(tmp_path: Path):
+    """
+    Tests whether the output messages printed to stdout match expected messages
+    when `download-output` command is executed.
+    """
+    config.set_setting("defaults.farm_id", MOCK_FARM_ID)
+    config.set_setting("defaults.queue_id", MOCK_QUEUE_ID)
+
+    with patch.object(api, "get_boto3_client") as boto3_client_mock, patch.object(
+        job_group, "OutputDownloader"
+    ) as MockOutputDownloader, patch.object(job_group, "round", return_value=0), patch.object(
+        job_group, "_get_conflicting_filenames", return_value=[]
+    ), patch.object(job_group, "_assert_valid_path", return_value=None), patch.object(
+        api, "get_queue_user_boto3_session"
+    ), patch.object(
+        job_group.os.path,
+        "expanduser",
+        return_value=tmp_path,
+    ) as mock_expanduser:
+        mock_download = MagicMock()
+        MockOutputDownloader.return_value.download_job_output = mock_download
+        windows_root_path = "C:\\Users\\username"
+        not_windows_root_path = "/root/path"
+        mock_root_path = not_windows_root_path if sys.platform == "win32" else windows_root_path
+        mock_files_list = ["outputs/file1.txt", "outputs/file2.txt", "outputs/file3.txt"]
+        MockOutputDownloader.return_value.get_output_paths_by_root.side_effect = [
+            {
+                f"{mock_root_path}": mock_files_list,
+            },
+            {
+                f"{tmp_path}": mock_files_list,
+            },
+        ]
+
+        mock_host_path_format = PathFormat.POSIX if sys.platform == "win32" else PathFormat.WINDOWS
+
+        boto3_client_mock().get_queue.side_effect = [MOCK_GET_QUEUE_RESPONSE]
+        boto3_client_mock().get_job.return_value = {
+            "name": "Mock Job",
+            "attachments": {
+                "manifests": [
+                    {
+                        "rootPath": f"{mock_root_path}",
+                        "rootPathFormat": mock_host_path_format,
+                        "outputRelativeDirectories": ["."],
+                    },
+                ],
+            },
+        }
+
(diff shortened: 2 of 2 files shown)


src/deadline/job_attachments/download.py
```diff
@@ -1120,7 +1120,7 @@ def set_root_path(self, original_root: str, new_root: str) -> None:
         (It will store the new root path as an absolute path.)
         """
         # Need to use absolute to not resolve symlinks, but need normpath to get rid of relative paths, i.e. '..'
-        new_root = str(os.path.normpath(Path(new_root).absolute()))
+        new_root = str(os.path.normpath(Path(new_root).expanduser().absolute()))
```
mwiebe: This place will call `expanduser()` on all calls to `set_root_path`, but I doubt that's what we want to do. The original issue is about interactive user input, I think it's better to put this call immediately after that input is collected instead of inside this setter.

mwiebe (changes requested):

sakshie95 (dismissed):

src/deadline/job_attachments/download.py
jericht: I think we do want it in here since we make a call to `pathlib.Path.absolute()` which will prepend the current working directory to the path in cases where the path starts with `~`, resulting in a root path like `/home/jericht/~/tmp/123`, which is something we don't want in my opinion.

We are applying `os.normpath` and `pathlib.Path.absolute` to the path, so we're already not setting the `new_root` parameter as is to the `self.outputs_by_root` dict. To me, this means the code is intended to "make sure the path is a valid root path". However, transforming a path like `~/tmp/123` into something like `/the/current/working/dir/~/tmp/123`, for example, seems like a bug. If the caller really wants to set a root path like `/the/current/working/dir/~/tmp/123` by just passing in `~/tmp/123` while the cwd is `/the/current/working/dir/`, I think it's reasonable to have the caller pass in the absolute path in this case.

I've updated the function docstring to make it clear that this happens along with the other transformations it applies to `new_root`.

I tested this locally to make sure `expanduser` handles some edge cases I could think of correctly. If the path is already has the `~` expanded, it won't attempt to expand it again creating an invalid path. Here are a few tests I ran locally.

```
>>> def test_expanduser(path: str):
...     print(f"os.path.expanduser: {os.path.expanduser(path)}")
...     print(f"pathlib.Path.expanduser: {pathlib.Path(path).expanduser()}")
... 

>>> test_expanduser("~/tmp/123")
os.path.expanduser: /home/jericht/tmp/123
pathlib.Path.expanduser: /home/jericht/tmp/123

>>> test_expanduser("~/~/tmp/123")
os.path.expanduser: /home/jericht/~/tmp/123
pathlib.Path.expanduser: /home/jericht/~/tmp/123

>>> test_expanduser("/tmp/~/123")
os.path.expanduser: /tmp/~/123
pathlib.Path.expanduser: /tmp/~/123

>>> test_expanduser("/home/jericht/~/tmp/123")
os.path.expanduser: /home/jericht/~/tmp/123
pathlib.Path.expanduser: /home/jericht/~/tmp/123

>>> test_expanduser("~/home/jericht/~/tmp/123")
os.path.expanduser: /home/jericht/home/jericht/~/tmp/123
pathlib.Path.expanduser: /home/jericht/home/jericht/~/tmp/123
```

leongdl (commented):
Does this change also affect `deadline attachment download` ? I hope the same logic (or the same code path) will apply

src/deadline/job_attachments/download.py
leongdl: side comment: Does this change also work for `deadline attachment download` ?

mwiebe: If we do want this expansion to happen always in the setter, I think we should look more closely at where the value comes from currently, and where it could come from. Will this be surprising in any case?

godobyte:
> Does this change also affect `deadline attachment download` ? I hope the same logic (or the same code path) will apply

`deadline attachment download` does not use this `set_root_path` API. It downloads to specified destination in path mapping file, or the working directory if it's not provided. Maybe it's worth checking the path mapping approach by defining `~` as part of destination in the input file. Upon checking with Jericho, this path normalization issue is related to click, whereas `attachment download` doesn't use user prompt path for download.

src/deadline/job_attachments/download.py
jericht: After thinking about it, I want to take a safer approach and only fix what the issue called out originally. I've pushed a change to only expand the `~` in that specific case.

mwiebe (dismissed):

mwiebe (approved):

leongdl (approved):

PR #559, opened 2025-01-15
fix: manifest destination directory not being used when path uses backslashes

YutongLi291:
…

When using the `deadline manifest snapshot` CLI, if we are use backslashes for the path separators, the --destination CLI option is not respected if we don't pass in the `--name` option and instead uses the `--root` directory as fallback.

This is because during the result manifest name creation, we only replaced forward slashes with `_` in the file name, and not also backslashes. This caused the resulting file name to be an absolute path, overwriting the destination CLI option.
Replace backslashes and colons with `_` in the result file name. This fixes the issue with the result manifest name being an absolute path.
Also replace colons with `_` as this fixes Windows drive name being excluded from the resulting manifest name.
Users that use backslashes for path separators (usually on Windows) can get the manifest file output in the correct destination directory.
Using `deadline manifest snapshot`, confirmed that after this fix, the manifest output file gets saved in the specified destination directory on both Linux and Windows correctly without `--name` option specified

`hatch run test`
`hatch run integ:test`

```
hatch run deadline manifest snapshot --root C:\\Users\\ryan\\path -d C:\Users\ryan\abcdef
Hashing Summary:
    Processed 0 files totaling 0.0 B.
    Skipped re-processing 16 files totaling 25.36 KB.
    Total processing time of 0.02656 seconds at 0.0 B/s.

Manifest Generated at C:\Users\ryan\abcdef\C___Users__ryan__path-2fcc508f47a4c41ff65363f8120d3fab-2025-01-15T13-56-33.manifest

```

- Are relevant docstrings in the code base updated? No
- Has the README.md been updated? If you modified CLI arguments, for instance. No

No

No

----

### DIFF
--- a/src/deadline/job_attachments/api/manifest.py
+++ b/src/deadline/job_attachments/api/manifest.py
@@ -170,7 +170,9 @@ def _manifest_snapshot(
         # Encode the root path as
         root_hash: str = hash_data(root.encode("utf-8"), output_manifest.get_default_hash_alg())
         timestamp = datetime.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
-        manifest_name = name if name else root.replace("/", "_")
+        manifest_name = (
+            name if name else root.replace("/", "_").replace("\\", "_").replace(":", "_")
+        )
         manifest_name = manifest_name[1:] if manifest_name[0] == "_" else manifest_name
         manifest_name = f"{manifest_name}-{root_hash}-{timestamp}.manifest"
 


src/deadline/job_attachments/api/manifest.py
```diff
@@ -170,7 +170,9 @@ def _manifest_snapshot(
         # Encode the root path as
         root_hash: str = hash_data(root.encode("utf-8"), output_manifest.get_default_hash_alg())
         timestamp = datetime.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
-        manifest_name = name if name else root.replace("/", "_")
+        manifest_name = (
+            name if name else root.replace("/", "_").replace("\\", "_").replace(":", "_")
```
leongdl, line 174: There must be a better way than chanining replace? https://stackoverflow.com/questions/3411771/best-way-to-replace-multiple-characters-in-a-string

I guess not?

leongdl: What about re.sub?  https://docs.python.org/3/library/re.html#re.sub

leongdl (approved):

src/deadline/job_attachments/api/manifest.py
leongdl: Think about it, but I won't block the fix PR.

leongdl (approved):

src/deadline/job_attachments/api/manifest.py
YutongLi291: Yeah, as mentioned in the post you linked the .replace() is faster, and I think (IMO) it looks more clear than res.sub

src/deadline/job_attachments/api/manifest.py
```diff
@@ -170,7 +170,9 @@ def _manifest_snapshot(
         # Encode the root path as
         root_hash: str = hash_data(root.encode("utf-8"), output_manifest.get_default_hash_alg())
         timestamp = datetime.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
-        manifest_name = name if name else root.replace("/", "_")
+        manifest_name = (
```
sakshie95, line 173: Non blocking for merge but can we pls add a small comment above it saying why we did this?

src/deadline/job_attachments/api/manifest.py
sakshie95: +1, I think ``.sub`` looks more readable than chain replace

sakshie95 (approved):

src/deadline/job_attachments/api/manifest.py
godobyte: I'm not usually particular about naming, but we should provide customer visibility what happens when `--name` is not provided. The generated file name here is `C___Users__ryan__path__-2fcc508f47a4c41ff65363f8120d3fab-2025-01-15T13-56-33.manifest`, this should be documented somewhere.

PR #560, opened 2025-01-15
fix: remove default host requirements tab from DCCs that don't support it

joel-wong-aws:
I discovered that https://github.com/aws-deadline/deadline-cloud/pull/558 was not compatible with some of our DCCs. Removed the default setting - it will now show up in `deadline bundle gui-submit` but not in our DCCs by default since some DCCs can't handle the `host_requirements` argument.

https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py#L357-L377

`hatch clean && hatch build && hatch shell`
`deadline bundle gui-submit`
image

No, N/A

No

No

No

----


rena-cheng (approved):

miabatta (approved):
Thanks for the thoroughness Joel!

PR #561, opened 2025-01-17
Reduce job history dir length

joel-wong-aws:
Related to: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/105

When submitting jobs from DCCs which don't have long path support, the job history bundle will not be able to be created if the path is too long due to a long job name.

Truncated the job history dir for the submitted bundle so that the job history bundle, including asset manifest, can be created.

Previously failing submissions caused by long paths in the job history dir will succeed for DCCs which don't support long paths.

`hatch build && hatch  shell`

Confirmed that paths were truncated when submitting a job with a long name but the job as still able to submit successfully.

Edited my `settings.job_history` dir to a very long directory and got the expected error

image

No, N/A

No

No

----

### DIFF
--- a/src/deadline/client/job_bundle/__init__.py
+++ b/src/deadline/client/job_bundle/__init__.py
@@ -11,6 +11,7 @@ __all__ = [
 import datetime
 import glob
 import os
+import sys
 
 from ..config import get_setting
 from ._yaml import deadline_yaml_dump
@@ -34,10 +35,6 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
         char for char in submitter_name if char.isalnum() or char in " -_"
     )
 
-    # Clean the job_name's characters and truncate for the filename
-    job_name_cleaned = "".join(char for char in job_name if char.isalnum() or char in " -_")
-    job_name_cleaned = job_name_cleaned[:128]
-
     timestamp = datetime.datetime.now()
     month_tag = timestamp.strftime("%Y-%m")
     date_tag = timestamp.strftime("%Y-%m-%d")
@@ -53,8 +50,22 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
         latest_dir = existing_dirs[-1]
         number = int(os.path.basename(latest_dir)[len(date_tag) + 1 :].split("-", 1)[0]) + 1
 
-    result = os.path.join(
-        month_dir, f"{date_tag}-{number:02}-{submitter_name_cleaned}-{job_name_cleaned}"
-    )
+    job_dir_prefix = f"{date_tag}-{number:02}-{submitter_name_cleaned}-"
+
+    max_job_name_prefix = 128  # max job name from OpenJD spec
+
+    # max path length - manifest file name
+    # 256 - len("\manifests\d2b2c3102af5a862db950a2e30255429_input")
+    # = 207
+    if sys.platform in ["win32", "cygwin"]:
+        max_job_name_prefix = min(
+            207 - len(os.path.abspath(os.path.join(month_dir, job_dir_prefix))), max_job_name_prefix
+        )
+
+    # Clean the job_name's characters and truncate for the filename
+    job_name_cleaned = "".join(char for char in job_name if char.isalnum() or char in " -_")
+    job_name_cleaned = job_name_cleaned[:max_job_name_prefix]
+
+    result = os.path.join(month_dir, f"{job_dir_prefix}{job_name_cleaned}")
     os.makedirs(result)
     return result
--- a/src/deadline/client/job_bundle/__init__.py
+++ b/src/deadline/client/job_bundle/__init__.py
@@ -61,6 +61,10 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
         max_job_name_prefix = min(
             207 - len(os.path.abspath(os.path.join(month_dir, job_dir_prefix))), max_job_name_prefix
         )
+        if max_job_name_prefix < 1:
+            raise RuntimeError(
+                "Job history directory is too long. Please update your 'settings.job_history_dir' to a shorter path."
+            )
 
     # Clean the job_name's characters and truncate for the filename
     job_name_cleaned = "".join(char for char in job_name if char.isalnum() or char in " -_")
--- a/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
+++ b/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
@@ -417,7 +417,8 @@ class SubmitJobToDeadlineDialog(QDialog):
             deadline = api.get_boto3_client("deadline")
 
             self.job_history_bundle_dir = create_job_history_bundle_dir(
-                self.submitter_name, settings.name
+                self.submitter_name,
+                settings.name,
             )
 
             if self.show_host_requirements_tab:
--- a/test/unit/deadline_client/job_bundle/test_job_history_folders.py
+++ b/test/unit/deadline_client/job_bundle/test_job_history_folders.py
@@ -105,7 +105,7 @@ def test_job_history_dir_truncation_from_job_name_with_129_chars(fresh_deadline_
     # Use a temporary directory for the job history
     if sys.platform in ["win32", "cygwin"]:
         # Windows has a long tmp dir path so it will truncate anyway - we only want to test job name truncation
-        tmp_parent_dir = "C:\ProgramData"
+        tmp_parent_dir = "C:\\ProgramData"
     else:
         tmp_parent_dir = None
     with tempfile.TemporaryDirectory(dir=tmp_parent_dir) as tmpdir, freeze_time(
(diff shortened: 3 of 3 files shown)


src/deadline/client/job_bundle/__init__.py
```diff
@@ -53,8 +49,17 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
...
 
-    result = os.path.join(
-        month_dir, f"{date_tag}-{number:02}-{submitter_name_cleaned}-{job_name_cleaned}"
-    )
+    job_dir_prefix = f"{date_tag}-{number:02}-{submitter_name_cleaned}-"
+
+    # max path length - manifest file name
+    # 256 - len("\manifests\d2b2c3102af5a862db950a2e30255429_input")
+    # = 207
+    max_job_name_prefix = 207 - len(os.path.abspath(os.path.join(month_dir, job_dir_prefix)))
```
leongdl: the path of `\manifests\d2b2c3102af5a862db950a2e30255429_input` may become longer in the near future. Can we have more reserved characters so the structure does not change? Say 160 or 180 instead of "207" ?

src/deadline/client/job_bundle/__init__.py
```diff
@@ -53,8 +49,17 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
...
+
+    # Clean the job_name's characters and truncate for the filename
+    job_name_cleaned = "".join(char for char in job_name if char.isalnum() or char in " -_")
+    job_name_cleaned = job_name_cleaned[: min(128, max_job_name_prefix)]
+
+    result = os.path.join(month_dir, f"{job_dir_prefix}{job_name_cleaned}")
```
leongdl, line 73: While this solves its for DCCs that do not support long paths, should we not truncate for DCCs that DO support long paths? 

Can we pass in a boolean or something to indicate long path is supported? Truncating is by its nature losing some data.

leongdl (commented):
Discussion on how to handle long paths?

src/deadline/client/job_bundle/__init__.py
joel-wong-aws: Even with DCCs that do support long paths, Windows users would still have to enable long paths on their system in order to use long paths. I think a reasonable default is to use 256 characters across the board.

If we think about Linux/Mac users who typically have a default max path of 4096, we have a choice to be consistent with Windows or not limit the max path length at all. My immediate thought is to allow them to be unlimited on these platforms. Updated the code accordingly.

No information is lost because the job name still exist in the template, albeit a little less accessible.

src/deadline/client/job_bundle/__init__.py
joel-wong-aws: We can always shorten this more later if it becomes an issue :)

src/deadline/client/job_bundle/__init__.py
leongdl: Yeah I understand, you can see Ryan's PR. He added `\\?\` for some of these client / customer facing code paths and also a warning when long paths registry is off. Consider having the same or similar experience?

joel-wong-aws: I'm not sure I follow. Even if we were to add a warning, there's nothing actionable about the warning for users since some DCCs don't support long paths, right?

crowecawcaw: The job history folder is an implementation detail for the library and not intended as a user interface. I think it makes sense to prioritize robustness over user-friendliness here. Especially since the cost is minimal - only long job names are truncated, and they likely still have enough context to be identifiable.

crowecawcaw (dismissed):

src/deadline/client/job_bundle/__init__.py
leongdl: Sure - as long as the customer is aware.

leongdl (dismissed):

src/deadline/client/job_bundle/__init__.py
```diff
@@ -53,8 +50,22 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
...
-    result = os.path.join(
-        month_dir, f"{date_tag}-{number:02}-{submitter_name_cleaned}-{job_name_cleaned}"
-    )
+    job_dir_prefix = f"{date_tag}-{number:02}-{submitter_name_cleaned}-"
+
+    max_job_name_prefix = 128  # max job name from OpenJD spec
+
+    # max path length - manifest file name
+    # 256 - len("\manifests\d2b2c3102af5a862db950a2e30255429_input")
+    # = 207
```
epmog, line 59: nit: MAX_PATH is actually 260: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

> `D:\some 256-character path string<NUL>`

src/deadline/client/job_bundle/__init__.py
```diff
@@ -53,8 +50,22 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
...
+    if sys.platform in ["win32", "cygwin"]:
+        max_job_name_prefix = min(
+            207 - len(os.path.abspath(os.path.join(month_dir, job_dir_prefix))), max_job_name_prefix
+        )
```
epmog, line 63: Have we considered what happens if this ends up being 0 or a negative number?

src/deadline/client/job_bundle/__init__.py
epmog: As far as I understand, there's no requirement that the manifest folder and subsequent files to exist in the job history bundle. Though then you're trying to balance the usefulness of having the manifest already created vs. length of the folder name. Don't have an answer off-hand on what's more important.

I'm in agreement with David that my preferred solution would be to not truncate on applications that are longPathAware, is that something we can easily check when being used as a library?

This may be one of the cases (job history) where we do want to use the `\\?\` prefix considering those bundles could work fine on a machine that's longPathAware

epmog:
Do we have a stack trace or anything of the experience when being unable to create a history directory? Was it the job bundle history creation that failed? Perhaps for a quick fix, we could just be catching/warning on that.

joel-wong-aws:
> Do we have a stack trace or anything of the experience when being unable to create a history directory? Was it the job bundle history creation that failed? Perhaps for a quick fix, we could just be catching/warning on that.

The error message:
```
Error occurred: [Errno 2] No such file or directory: 'C:\\Users\\test-user\\.deadline\\job_history\\test-user-gamma-package-build\\2025-01\\2025-01-23-01-Cinema4D-GSG_REDSHIFT_StarterScene_April_2024 sdkjfhsakjdfghhagwefkjhawvefkjhavwfjkhavwefjhawvfjhvasdjkhfvsdjfvajwfjkhwevfjhwevfjhawevfjh\\manifests\\7512c012abaae0a151ca32f96e07ace5_input'
```
In Cinema 4D, the console has the stack trace
```
[Errno 2] No such file or directory: 'C:\\Users\\test-user\\.deadline\\job_history\\test-user-gamma-package-build\\2025-01\\2025-01-23-01-Cinema4D-GSG_REDSHIFT_StarterScene_April_2024 sdkjfhsakjdfghhagwefkjhawvefkjhavwfjkhavwefjhawvfjhvasdjkhfvsdjfvajwfjkhwevfjhwevfjhawevfjh\\manifests\\7512c012abaae0a151ca32f96e07ace5_input'
  File "C:\Users\test-user\AppData\Roaming\DeadlineCloudSubmitter\deadline\client\ui\dialogs\submit_job_progress_dialog.py", line 386, in _upload_background_thread
    ).upload_assets(
      ^^^^^^^^^^^^^^
  File "C:\Users\test-user\AppData\Roaming\DeadlineCloudSubmitter\deadline\job_attachments\upload.py", line 1288, in upload_assets
    (partial_manifest_key, asset_manifest_hash) = self.asset_uploader.upload_assets(
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\test-user\AppData\Roaming\DeadlineCloudSubmitter\deadline\job_attachments\upload.py", line 195, in upload_assets
    self._write_local_manifest(
  File "C:\Users\test-user\AppData\Roaming\DeadlineCloudSubmitter\deadline\job_attachments\upload.py", line 253, in _write_local_manifest
    self._write_local_input_manifest(manifest_write_dir, manifest_name, manifest, root_dir_name)
  File "C:\Users\test-user\AppData\Roaming\DeadlineCloudSubmitter\deadline\job_attachments\upload.py", line 274, in _write_local_input_manifest
    with open(local_manifest_file, "w") as file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\test-user\\.deadline\\job_history\\test-user-gamma-package-build\\2025-01\\2025-01-23-01-Cinema4D-GSG_REDSHIFT_StarterScene_April_2024 sdkjfhsakjdfghhagwefkjhawvefkjhavwfjkhavwefjhawvfjhvasdjkhfvsdjfvajwfjkhwevfjhwevfjhawevfjh\\manifests\\7512c012abaae0a151ca32f96e07ace5_input'
```

src/deadline/client/job_bundle/__init__.py
joel-wong-aws: While the official MAX_PATH is 260, there may be some DCC constraints that limit this to 256 characters, so I decided to go with that. This also gives us a few characters of wiggle room.

src/deadline/client/job_bundle/__init__.py
joel-wong-aws: Added error handling for this case!

crowecawcaw (approved):

leongdl (approved):

src/deadline/client/job_bundle/__init__.py
```diff
@@ -53,8 +50,26 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
...
+        if max_job_name_prefix < 1:
+            raise RuntimeError(
+                "Job history directory is too long. Please update your 'settings.job_history_dir' to a shorter path."
+            )
```
epmog, line 67: Sorry for the delay, unfortunately due to the customer experience impact we can't do this change.

A valid windows submission that works with long paths will now fail to submit because we won't create the job history dir (to be used as input to the create job bundle callback for where to create the submission files).

In my mind, there's 2 part of this at play:
1. job bundle history directory right now is a required portion of submission or export
    * in my opinion, it is an internal implementation detail that we _must_ create a history entry to submit, rather than populate it after submission with best effort.
2. using the library in a runtime that doesn't support long paths will have submission issues based off the job bundle directory location + job name length

I think short-term we can make the second issue less likely and do the truncation but only if it would fail otherwise. We can't have previously working setups failing due to this. If we can successfully create the history directory and a file whose path matches the longest path therein, then no truncation should happen.

Longer-term I think we should absolutely remove the truncation and:
1. use a local temp folder to allow submission and then move it to the history folder after submission, and/or
2. leverage `\\?\` for this specific case of the internal implementation detail of having a history directory and how we populate it.

Let me know your thoughts

epmog (changes requested):

PR #562, opened 2025-01-17
feat: add sticky settings for deadline bundle gui-submit

joel-wong-aws:
Sticky settings did not work on `deadline bundle gui-submit`.

Sticky settings are read from `parameter_values.yaml`. However, we don't save them to a job bundle directly after submission, only the job history directory! Saved the parameter settings after submission.

Note that some settings like job name and description will still not be persistent with this change.

Sticky settings work for most use-configurable fields.

`hatch build && hatch shell`
`deadline bundle gui-submit`

No, N/A

*  [X] This PR does not add any new dependencies.

No

No

----

### DIFF
--- a/src/deadline/client/ui/job_bundle_submitter.py
+++ b/src/deadline/client/ui/job_bundle_submitter.py
@@ -139,6 +139,7 @@ def show_job_bundle_submitter(
             AssetReferences(),
         )
 
+        # save to job history dir
         save_yaml_or_json_to_file(
             bundle_dir=job_bundle_dir, filename="template", file_type=file_type, data=template
         )
@@ -150,6 +151,14 @@ def show_job_bundle_submitter(
         )
         save_yaml_or_json_to_file(
             bundle_dir=job_bundle_dir,
+            filename="parameter_values",
+            file_type=file_type,
+            data={"parameterValues": parameters_values},
+        )
+
+        # save parameter values to input job bundle dir for sticky settings
+        save_yaml_or_json_to_file(
+            bundle_dir=input_job_bundle_dir,
             filename="parameter_values",
             file_type=file_type,
             data={"parameterValues": parameters_values},


karthikbekalp (approved):

leongdl (approved):

src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -155,6 +156,14 @@ def on_create_job_bundle_callback(
             data={"parameterValues": parameters_values},
         )
 
+        # save parameter values to input job bundle dir for sticky settings
+        save_yaml_or_json_to_file(
+            bundle_dir=input_job_bundle_dir,
+            filename="parameter_values",
+            file_type=file_type,
+            data={"parameterValues": parameters_values},
+        )
```
epmog, line 165: I think we'll have to chat about this change a bit more. I'm not certain we want it as is.

This turns the input job bundle from being "immutable" into "mutable". TDs that write job bundle for users will now have that job bundle update every time it's used. On a per-user basis this could be fine, although you've now lost what the original template was. But I think we end up with undesirable behaviour when you expand this to more users. ie. A bundle repository on a shared filesystem. Now everyone is overriding everyone's values by default.

It's also tricky for people that leverage `deadline bundle gui-submit` workflows. If these bundles are installed system-wide, like in a DCC, then they may not even have the permissions to modify the bundle and this will crash.

epmog (changes requested):

src/deadline/client/ui/job_bundle_submitter.py
joel-wong-aws: Discussed with Morgan offline. For job bundles specifically from `deadline bundle gui-submit` (manually invoked by the user), we don't want to save sticky settings because job bundles are often in shared storage and we wouldn't want an individual's settings to override the job bundle's settings shared by multiple users.

epmog: Yea, I could be holding it wrong but I think there's a way to balance automation (DCC submitters) leveraging gui-submit to get sticky settings automatically vs. other usage of bundles.

Perhaps an optional update, rather than forced. Or per-user sticky settings for bundles, etc.

DCCs with sticky settings layer it over-top of the built-in job bundle, rather than modify it directly, or are doing throwaway bundles using a bundled template. For a system install it still affects all users, but only for that specific usage (scene file), rather than all bundles for that DCC.

crowecawcaw: +1 for making this an optional feature or one that's enabled by default but can be opted out of.

KeyShot's submitter builds on top of GUI submit and the bulk of its code is dedicated to implementing sticky settings. 

Anecdotally, when I've used the GUI submitter before, I've found it frustrating that settings aren't saved so I end up hardcoding values into the job template itself so they're persisted between submissions.

Agree that saving the values to the original job bundle might be problematic, but I hope this feature doesn't stall.

joel-wong-aws: This discussion continued offline. We decided that saving the sticky settings to a generic `parameter_values.yaml` file is not the right choice for the `deadline-cloud` repository because job bundles submitted from `deadline bundle gui-submit` are often shared between multiple artists and who will have separate settings and wouldn't want to overwrite each others settings. It's different than DCCs, which have sticky settings on a per-scene basis and are usually stored on an individuals workstations rather than shared storage.

PR #563, opened 2025-01-18
feat: add manifest merge and pass in suffix for local manifest write

godobyte:
Worker Agent needs an API to merge job input and step dependency manifests.

Add the functionality to merge given manifests for the same root.

There is an internal API to merge manifests for the same root and write to local. This can serve as a conversation starter to see whether a customer facing API is needed for merge.

- Have you run the unit tests? [Y]
- Have you run the integration tests? [Y]

- Are relevant docstrings in the code base updated? [Y]

No

No

----


src/deadline/job_attachments/api/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        Dict[str, BaseAssetManifest]: File name to encoded manifest mapping
+    """
+
+    if nonvalid_files := [manifest for manifest in manifests if not os.path.isfile(manifest)]:
+        raise NonValidInputError(f"Specified manifests {nonvalid_files} are not valid.")
+
+    with ExitStack() as stack:
+        file_name_manifest_dict: Dict[str, BaseAssetManifest] = {
+            os.path.basename(file_path): decode_manifest(
+                stack.enter_context(open(file_path)).read()
```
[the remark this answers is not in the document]
godobyte: https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack

src/deadline/job_attachments/api/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import List, Dict
+
+from deadline.client.exceptions import NonValidInputError
+from deadline.job_attachments.asset_manifests.base_manifest import BaseAssetManifest
+from deadline.job_attachments.asset_manifests.decode import decode_manifest
+
+
+def _read_manifests(manifests: List[str]) -> Dict[str, BaseAssetManifest]:
+    """
+    Read in manfiests from the give file path list, and produce file name to manifest mapping.
```
jericht:
```suggestion
    Read in manfiests from the given file path list, and produce file name to manifest mapping.
```

src/deadline/job_attachments/api/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+from contextlib import ExitStack
+from typing import List, Dict
+
+from deadline.client.exceptions import NonValidInputError
+from deadline.job_attachments.asset_manifests.base_manifest import BaseAssetManifest
+from deadline.job_attachments.asset_manifests.decode import decode_manifest
+
+
+def _read_manifests(manifests: List[str]) -> Dict[str, BaseAssetManifest]:
```
jericht: Can we put this function into `manifest.py`?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -450,3 +468,35 @@ def add_manifest_by_root(
...
+    name: Optional[str],
+    logger: ClickLogger = ClickLogger(False),
+) -> Optional[ManifestMerge]:
+    """
+    BETA API - API to merge multiple manifests into one.
+    root: Root path for the manifest.
+    manifest_files: List of manifest files to merge.
+    destination: Destination directory for the merged manifest.
+    name: Name of the merged manifest.
+    logger: Click Logger instance to print to CLI as test or JSON.
```
jericht:
```suggestion
    logger: Click Logger instance to print to CLI as text or JSON.
```

test/unit/deadline_job_attachments/api/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Fixture to provide a temporary job bundle directory.
+    """
+
+    with tempfile.TemporaryDirectory() as job_bundle_dir:
+        yield job_bundle_dir
+
+
+@pytest.fixture(scope="function")
+def temp_assets_dir():
```
jericht: Are these being used anywhere?

test/unit/deadline_job_attachments/api/test_manifest_merge.py
```diff
@@ -0,0 +1,61 @@
...
+        manifest_merge: Optional[ManifestMerge] = _manifest_merge(
+            root=temp_dir, manifest_files=[manifest_one], destination=merge_dir, name="merge"
+        )
+
+        # Then
+        assert manifest_merge is not None
+        assert manifest_merge.manifest_root == temp_dir
+        assert merge_dir in manifest_merge.local_manifest_path
+
+    def test_diff_new_files(self, temp_dir, test_manifest_one, test_manifest_two):
```
jericht:
```suggestion
    def test_merge_different_files(self, temp_dir, test_manifest_one, test_manifest_two):
```

nit: name is a bit confusing/misleading. Sounds like we are testing a `diff` functionality

src/deadline/job_attachments/api/_utils.py
godobyte: This is used by `attachment.py` too, that's why I extracted from `attachment.py` to utils for both manifest and attachment.

leongdl: I like this, its a great small primitive to have.

Question or nit; should `manifests` be named `manifest_paths` ? Or `manifests: List[Path]` so it is obvious that it is a file path?

src/deadline/job_attachments/api/manifest.py
godobyte: Good catch, this is pasted from elsewhere, will update them all.

src/deadline/job_attachments/api/manifest.py
```diff
@@ -167,29 +169,45 @@ def _manifest_snapshot(
...
-        logger.echo(f"Manifest Generated at {local_manifest_file}\n")
+        logger.echo(f"Manifest generated at {local_manifest_file}\n")
         return ManifestSnapshot(manifest=local_manifest_file)
     else:
         # No manifest generated.
         logger.echo("No manifest generated")
         return None
 
 
+def _write_manifest(
```
leongdl, line 187: Do we want to move this also to `_utils` ?

godobyte: This function is currently only used by `manifest.py`, that's why I kept it here.

test/unit/deadline_job_attachments/api/conftest.py
godobyte: I must have committed this by accident. Removed..

src/deadline/job_attachments/api/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        Dict[str, BaseAssetManifest]: File name to encoded manifest mapping
+    """
+
+    if nonvalid_files := [manifest for manifest in manifest_paths if not os.path.isfile(manifest)]:
+        raise NonValidInputError(f"Specified manifests {nonvalid_files} are not valid.")
+
+    with ExitStack() as stack:
+        file_name_manifest_dict: Dict[str, BaseAssetManifest] = {
+            os.path.basename(file_path): decode_manifest(
+                stack.enter_context(open(file_path)).read()
```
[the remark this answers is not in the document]
godobyte: https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack

src/deadline/job_attachments/api/_utils.py
jericht: Not a blocker, but IMO it would be cleaner if we had this in the `manifest.py` file and just import it in `attachment.py` if we need to use it. This function is responsible for reading manifest files, so it feels like it belongs with all the other manifest code in `manifest.py`.

jericht (approved):

src/deadline/job_attachments/api/manifest.py
leongdl: Yeah, but it has good utility usage for the customer. 

Think if we wrote this in Java, there would be a class manifest, and read, write would be methods of this class right?

godobyte: I agree this can be a good utility for customers, but this helper is not generalized enough to be a good utility function yet. For example, the file name contains very specific hatch logic for attachments. I'd track the `good utility usage` in a task I'm creating to properly model this as a util.

leongdl (approved):

PR #565, opened 2025-01-21
fix: Allow Identifier's in custom host requirements.

### LINKED ISSUE #553: Bug: custom requirements are broken in the GUI submitter host requirements
### Describe the bug

Currently, custom requirements do not work in the GUI submitter. All custom requirements are prepended with `amount.worker.` or `attr.worker.`, which are [reserved namespaces](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#3311-amountcapabilityname) and hence explicitly not "custom".

Additionally, the GUI submitter does not support periods in custom requirement names. e.g. `attr.custom.myattr` is not allowed.

### Expected Behaviour

Instead of prepending `amount.worker.`, we should prepend either `amount.` or have no prepended value at all.

Custom requirement names should allow periods.

### Current Behaviour

`amount.worker`/`attr.worker` is prepended.

Custom requirement names do not allow periods.

### Reproduction Steps

1. Open the bundle GUI submitter in a DCC that support host requirements (like Cinema 4D) or by manually editing the GUI submitter to have `show_host_requirements_tab=True`, then `hatch build && hatch shell` -> `deadline bundle gui-submit`
2. Open the host requirements tab
3. Enter a custom amount requirement or customer attribute requirement
4. Submit the job
5. The job will fail to create with an error like 
```
steps[0] -> hostRequirements -> amounts[0] -> name:
	Only Open Job Description defined capabilities may start with 'amount.worker': amount.worker.custom
```

### Environment

At minimum:
1. Operating system: (e.g. Windows Server 2022; Amazon Linux 2023; etc.): Affects all operating systems, tested on Windows 10 Enterprise
2. Output of `python3 --version`: Affects all python versions, tested in Python 3.12.8
3. Output of `deadline --version`: `deadline, version 0.49.2`

erico-aws:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/553

Users could not enter identifiers in the name field of custom host requirements. It was restricted to just letters, numbers, dashes and underscores up to 99 characters. This doesn't match the attributes of an Identifier as outlined in the OpenJD specification. 

1. I removed the reserved "worker" identifier from the attribute and amount prefixes.
2. I added an Identifier regex. The 2 name line edits now use this to check that the value entered is an Identifier or a series of Identifiers separated by periods. 
3. Check the first Identifier entered to make sure that it's not one of the reserved first Identifiers.
4. Check that the name doesn't end with a period.

Names for custom host requirements should work now and should be validated correctly.

Added Unit tests for the changes.

Deployed changes to Blender for testing to make sure it worked as expected.

No.

No.

No.
----


src/deadline/client/ui/widgets/host_requirements_tab.py
```diff
@@ -72,11 +73,19 @@
 )
 
 CUSTOM_CAPABILITY_NAME_REGEX = "^(\\.[a-zA-Z][a-zA-Z0-9]{0,63})+$"
```
epmog, line 75: nit: this wasn't and still isn't actually used anywhere so we can actually remove this.

epmog (approved):
one little nit, took it for a spin, looks good! Ship it!

lucaseck (approved):

PR #570, opened 2025-01-23
fix: limit number of hashing threads

joel-wong-aws:
Deadline Cloud GUI submission was freezing during the hashing step on some platforms.

The [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) has a different number of workers by default depending on the Python version, but as of Python 3.11, the default number of workers is `min(32, os.cpu_count() + 4)`. Hashing is a CPU-intensive operation, and we've previously encountered hanging processes when using the default number of workers. So we limit it to most N - 2 threads, where N is the number of CPUs on the computer.

Fixes frozen GUI submissions.

Tested this on a Windows workstation, submissions passed.

In-code comments were added

No

No

----

### DIFF
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -853,7 +853,14 @@ class S3AssetManager:
         }:
             paths: list[base_manifest.BaseManifestPath] = []
 
-            with concurrent.futures.ThreadPoolExecutor() as executor:
+            # The [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)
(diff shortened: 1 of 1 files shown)


karthikbekalp (approved):

MarinMaksutaj (approved):

leongdl (approved):
This will work, but I am not 100% convinced this is the root cause of the bug.

joel-wong-aws:
> This will work, but I am not 100% convinced this is the root cause of the bug.

Agreed. To unblock customers, we will merge this in and can continue investigating the root cause.

PR #573, opened 2025-01-24
feat: add maxWorkerCount support for jobs

shaurysh:
* The requirement is to support the new `maxWorkerCount` feature of AWS Deadline Cloud which enables customers to specify a maximum worker count on a per job basis. This will allow customers to control how many instances at any given point of time would be dedicated to processing of that job. 

* Add support for the `maxWorkerCount` to the CLI
* Add support for the `maxWorkerCount` to the Deadline Cloud GUI.
* Screenshot 2025-01-24 at 11 25 20 AM

* Allows customers to specify `maxWorkerCount` as a parameter when using the Deadline Cloud CLI to create jobs
* Allows customers to specify `maxWorkerCount` in their job bundles and appropriately passes it into the Deadline Cloud CreateJob API call
* Adds support for the `maxWorkerCount` parameter to the Deadline Cloud CLI GUI.

- Have you run the unit tests? - YES
- Have you run the integration tests? - Will work on running them now, but I dont think my changes should be breaking anything here. 
  that you ensure that the docker-based unit tests pass. - NO

- Ran manual tests
  - Verified that using `deadline bundle submit <path_to_job_bundle>` maxWorkerCount is correctly passed to Deadline
  - Verified that using `deadline bundle submit <path_to_job_bundle> --max-worker-count <number>` correctly overrode the maxWorkerCount specified in the job bundle and the number specified in the CLI call was passed to Deadline.
  - Verified that maxWorkerCount in the GUI is correctly populated when specified in the job bundle
  - Verified that changing maxWorkerCount in the GUI and clicking 'Export Bundle' results in the maxWorkerCount value being correctly populated in exported `parameter_values.yaml`
  - Verified that specifying `maxWorkerCount` in the GUI and submitting results in the maxWorkerCount parameter being passed to Deadline by calling GetJob API and verifying that maxWorkerCount was returned correctly.

- Are relevant docstrings in the code base updated? - YES
- Has the README.md been updated? If you modified CLI arguments, for instance. - NO

*   [ ✅ ] This PR does not add any new dependencies.

* We are adding support for a new optional parameter, so this should not be considered backward incompatible. 

* No
----

### DIFF
--- a/src/deadline/client/ui/widgets/shared_job_settings_tab.py
+++ b/src/deadline/client/ui/widgets/shared_job_settings_tab.py
@@ -17,6 +17,7 @@ from qtpy.QtWidgets import (  # type: ignore
     QHBoxLayout,
     QLabel,
     QLineEdit,
+    QRadioButton,
     QSpinBox,
     QVBoxLayout,
     QWidget,
@@ -291,6 +292,27 @@ class SharedJobPropertiesWidget(QGroupBox):  # pylint: disable=too-few-public-me
         self.max_retries_per_task_box.setRange(0, 2147483647)
         self.layout.addRow(self.max_retries_per_task_box_label, self.max_retries_per_task_box)
 
+        self.max_worker_count_box_label = QLabel("Maximum worker count")
+        self.max_worker_count_box_label.setToolTip("Maximum worker count of job.")
+        self.max_worker_count_box = QSpinBox()
+        self.max_worker_count_box.setRange(1, 2147483647)
+        self.unlimited_max_worker_count = QRadioButton("No max worker count")
+        self.limited_max_worker_count = QRadioButton("Set max worker count")
+        self.limited_max_worker_count.toggled.connect(
+            self.limited_max_worker_count_radio_button_toggled
+        )
+        self.max_worker_count_layout = QVBoxLayout(self)
+        self.max_worker_count_layout.addWidget(self.unlimited_max_worker_count)
+        self.max_worker_count_layout.addWidget(self.limited_max_worker_count)
+        self.max_worker_count_layout.addWidget(self.max_worker_count_box)
+        self.layout.addRow(self.max_worker_count_box_label, self.max_worker_count_layout)
+
+    def limited_max_worker_count_radio_button_toggled(self, state):
+        """
+        Enable the max worker count text box when limited max worker count radio button is enabled.
+        """
+        self.max_worker_count_box.setHidden(not state)
+
     def refresh_ui(self, settings: Any):
         self.sub_name_edit.setText(settings.name)
         self.desc_edit.setText(settings.description)
@@ -298,6 +320,9 @@ class SharedJobPropertiesWidget(QGroupBox):  # pylint: disable=too-few-public-me
         self.max_failed_tasks_count_box.setValue(20)
         self.max_retries_per_task_box.setValue(5)
         self.priority_box.setValue(50)
+        self.unlimited_max_worker_count.setChecked(True)
+        self.limited_max_worker_count.setChecked(False)
+        self.max_worker_count_box.setHidden(True)
 
     def set_parameter_value(self, parameter: dict[str, Any]):
         """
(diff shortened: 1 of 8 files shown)


src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -289,13 +290,37 @@ def _build_ui(self):
         self.max_retries_per_task_box.setRange(0, 2147483647)
         self.layout.addRow(self.max_retries_per_task_box_label, self.max_retries_per_task_box)
 
+        self.max_worker_count_box_label = QLabel("Maximum worker count")
+        self.max_worker_count_box_label.setToolTip("Maximum worker count of job.")
+        self.max_worker_count_box = QSpinBox()
+        self.max_worker_count_box.setRange(1, 2147483647)
```
epmog, line 298: What's the default value for the UI here? Is it 1? Is that a reasonable default to have?

~The other options get away with 0 being the default since it's used to signify "unlimited", but we're using -1 was the underlying value here which is different in the UI~

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -355,6 +390,15 @@ def get_parameters(self):
                 "value": self.max_retries_per_task_box.value(),
             },
             {"name": "deadline:priority", "type": "INT", "value": self.priority_box.value()},
+            {
+                "name": "deadline:maxWorkerCount",
+                "type": "INT",
+                "value": (
+                    -1
```
epmog: The request to the API should be -1 to specify unlimited? Is that the same behaviour as not including it? What happens if 0 is sent?

mwiebe: Instead of populating it with a -1 value, I think we should exclude the item entirely from the list in this case. That will keep better compatibility of job bundles it creates for deadline-cloud library versions that don't have this feature included yet.

mwiebe (changes requested):

src/deadline/client/ui/widgets/shared_job_settings_tab.py
shaurysh: The 'default' if not specified in the job bundle is unlimited. The 'no maximum worker count' toggle would be selected by default. 

This is in keeping with line how the CreateJob API of Deadline Cloud will behave. If a customer calls CreateJob but doesnt specify a `maxWorkerCount` - no limit would be applied by the service on the number of worker instances which can be used to process the job. Hence, tried to model the same behavior here by selecting 'No maximum worker count' by default or if no value is specified in the job bundle.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
shaurysh: Yes, it is the same behavior as not including it. 0 is not a valid input for this parameter. We have validation logic in the backend to check for this. 

As the UI has been designed today, customers can either select 'No maximum worker count' (which will send -1) or will select 'Set max worker count' (where the valid input range is 1 to INT_MAX). It is therefore not possible to send in 0 using the UI for this parameter.

shaurysh: This does sound like a good idea. So if the unlimited max worker count box is checked, we simply don't populate the -1 at all and since its anyway the default - we should be good with auto-selecting the 'No max worker count' toggle by default when someone imports the job bundle.

mwiebe: Yes, think about the case where an existing job bundle is submitted - the behavior should not change with the new code. That means if it's not specified, "No max worker count" must be selected. Then the other direction is if you update the version but are not using the feature, the job bundles it creates are the same as before. That means to exclude the new property name.

mwiebe (dismissed):

epmog (dismissed):

epmog (approved):

mwiebe (approved):

PR #577, opened 2025-01-28
feat: add latency telemetry

justinsaws:
We don't have any telemetry around how long our calls to the service take on the local client. This information is useful to understand our local performance and if any regressions have been introduced with other changes.

Add telemetry to our calls to the main service.

Increased telemetry coverage.

  - Yes.
  - Yes.
  - No.

Yes.

No.

No.

----

### DIFF
--- a/src/deadline/client/api/_telemetry.py
+++ b/src/deadline/client/api/_telemetry.py
@@ -139,6 +139,7 @@ class TelemetryClient:
                 f"{get_deadline_endpoint_url(config=config)}/2023-10-12/telemetry",
                 TelemetryClient.ENDPOINT_PREFIX,
             )
+
             # Some environments might not have SSL, so we'll use the vendored botocore SSL context
             from botocore.httpsession import create_urllib3_context, get_cert_path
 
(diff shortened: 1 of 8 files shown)


src/deadline/client/api/_telemetry.py
```diff
@@ -384,3 +385,35 @@ def wrapper(*args: List[Any], **kwargs: Dict[str, Any]) -> Any:
...
+            :param ** Python variable argument. See https://docs.python.org/3/glossary.html#term-parameter
+            """
+            start_t = time.perf_counter_ns()
+            ret_val = function(*args, **kwargs)
+            end_t = time.perf_counter_ns()
+
+            latency = end_t - start_t
+
+            event_name = decorator_kwargs.get("metric_name", function.__name__)
+            get_deadline_cloud_library_telemetry_client().record_event(
```
erico-aws, line 410: Do we want to time the function even when there are exceptions in them or only successful executions?

justinsaws: That's a good question. I feel like if there is an exception we don't care to capture the performance. In this case, we are just worrying about E2E calls that are successful so we can always make an apples to apples comparison.

I think it would be difficult to get meaningful information from unsuccessful calls without a lot more context.

erico-aws: Maybe we could use this as some metric for impact? We definitely shouldn't rely on it to notify us of a problem or anything like that.

justinsaws: Impact would be tough to determine without other information and we already have telemetry around call success/failure. This is more to see what the customer experience is like when people are using our client.

AWS-Samuel (approved):

src/deadline/client/api/_telemetry.py
leongdl: Or what if we `try-except` the function call on 404? Then log a different metric also for failed cases?

justinsaws: We already have success/failure metrics that do exactly this so I'm not sure what additional information we would get from logging the failure.

crowecawcaw:
How does this perform if:
1. The CLI command is fast, and the process ends almost immediately. Does the telemetry event still get sent out?
2. The telemetry endpoint is slow, or someone is geographically far from it. Does the CLI get hung up on sending the telemetry event?

I saw that AWS SAM has flag for it's telemetry sending function to not wait for a response. I think it's used when the CLI process is ending so that telemetry issues don't slow down the CLI: https://github.com/aws/aws-sam-cli/blob/70ad4f78f64bd5a2906af1d7e90fef65026ec50b/samcli/lib/telemetry/telemetry.py#L69

crowecawcaw (commented):
How does this perform if:
1. The CLI command is fast, and the process ends almost immediately. Does the telemetry event still get sent out?
2. The telemetry endpoint is slow, or someone is geographically far from it. Does the CLI get hung up on sending the telemetry event?

I saw that AWS SAM has flag for it's telemetry sending function to not wait for a response. I think it's used when the CLI process is ending so that telemetry issues don't slow down the CLI: https://github.com/aws/aws-sam-cli/blob/70ad4f78f64bd5a2906af1d7e90fef65026ec50b/samcli/lib/telemetry/telemetry.py#L69

justinsaws:
The telemetry is sent the same way that our other telemetry events are. No matter how fast the function is, we will get some metrics for it since we are using Python's [time.perf_counter_ns()](https://docs.python.org/3/library/time.html#time.perf_counter) to measure the time taken. The telemetry function is blocking and not asynchronous so it will always go out.

erico-aws (approved):

PR #589, opened 2025-02-05
fix: Priority box now has correct range.

### LINKED ISSUE #575: Bug: Gui-Submit window limits job priority to 99
### Describe the bug

when using `deadline bundle gui-submit` I can only set the priority to 99, not 100.

### Expected Behaviour

The monitor lets me change the priority to 100, so I should be allowed to submit jobs with priority 100.

### Current Behaviour

When I type "100" into the priority field, it truncates the number to "10" and if I use the arrow buttons to bump the value up, the value only goes to 99.

### Reproduction Steps

Try to submit a job with priority 100 using `deadline bundle gui-submit`

### Environment

At minimum:
1. Operating system: Windows 11
3. Output of `python3 --version` Python 3.13.1
4. Output of `deadline --version` deadline, version 0.49.3
5. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? It was probably installed from the installer, but I am unsure.

erico-aws:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/575

The Priority spinbox was capped at 99, but the documentation for Deadline says that it should be 100.

I updated the range for the spinbox to match what's in the Deadline documentation. https://docs.aws.amazon.com/deadline-cloud/latest/userguide/deadline-cloud-jobs.html

It's set to 1-100 now.

Screenshot 2025-02-05 at 2 37 59 PM

Priority is now correctly limited.

Updated unit test and ran deadline bundle gui-submit to check that the spinbox limits was updated correctly. 

No.

No.

No.

----


ttblanchard (dismissed):

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -269,6 +269,7 @@ def _build_ui(self):
 
         self.priority_box_label = QLabel("Priority")
         self.priority_box = QSpinBox(parent=self)
+        self.priority_box.setRange(1, 100)
```
epmog: minimum should still be 0, no?

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/deadline/client/create_job.html

epmog (changes requested):
Just the one comment

lucaseck (approved):

epmog (approved):

PR #600, opened 2025-02-12
feat: Add helper function to add default timeouts in job templates.

karthikbekalp:

DCCs that depend on `deadline-cloud` needed an easy way to patch their job templates with default timeouts for Env enter/exit/task run. 

Discussed with @AWS-Samuel about the simplest way to provide this capability. 

- Create a helper function which can be used by different DCCs directly. [This PR]
- Add GUI components that adds the timeouts to the SubmitJobToDeadlineDialog with the render settings that can be applied when there is `on_create_job_bundle_callback`.  [Upcoming PR]

Helps provide an easy solution for DCCs to start patching their job templates with timeouts. 

### How was this change tested? [I'm currently testing this but want some quick feedback]

I updated this helper function and used it in Cinema 4D and the job template contained the timeout settings. 

Yes

Ran it. There were 3 failures and all of them were job attachment tests which this code does not change. 

```
FAILED test/integ/cli/test_cli_attachment.py::TestAttachment::test_attachment_s3_cross_account_access_denied - 
FAILED test/integ/deadline_job_attachments/test_job_attachments.py::test_upload_bucket_wrong_account[2023-03-03] - AssertionError: Regex pattern did not match.
FAILED test/integ/deadline_job_attachments/test_job_attachments.py::test_sync_outputs_bucket_wrong_account[2023-03-03] - AssertionError: Regex pattern did not match.
 
============================ 3 failed, 29 passed, 2 skipped, 5 warnings in 64.57s (0:01:04) =============================
```

No

Yes

No

No

No

----

### DIFF
--- a/src/deadline/client/job_bundle/timeouts.py
+++ b/src/deadline/client/job_bundle/timeouts.py
@@ -1,6 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-from typing import Any, Optional
+from typing import Any, Optional, Dict
 from dataclasses import dataclass
 
 SECONDS_IN_A_MINUTE = 60
@@ -27,7 +27,7 @@ class TimeoutSettings:
 
 
 def add_timeouts_to_job_template(
-    template: dict[str, Any], timeout_settings: Optional[TimeoutSettings] = None
+    template: Dict[str, Any], timeout_settings: Optional[TimeoutSettings] = None
 ) -> None:
     """
     Adds timeout values to actions in a job template.
@@ -37,7 +37,7 @@ def add_timeouts_to_job_template(
     timeouts must be hard-coded in the job template.
 
     Args:
-        template (dict[str, Any]): The job template to modify.
+        template (Dict[str, Any]): The job template to modify.
         timeout_settings (Optional[TimeoutSettings]): Configuration for timeout values. If None,
             default timeout settings will be used. If is_deactivated is True, no timeouts will be added.
 
@@ -70,14 +70,14 @@ def add_timeouts_to_job_template(
     if not timeout_settings.is_activated:
         return
 
-    def _apply_timeouts_to_environment(environment: dict):
+    def _apply_timeouts_to_environment(environment: Dict):
         if "script" in environment:
             actions = environment["script"]["actions"]
             actions["onEnter"]["timeout"] = timeout_settings.on_enter_timeout_seconds
             if "onExit" in actions:
                 actions["onExit"]["timeout"] = timeout_settings.on_exit_timeout_seconds
 
-    def _apply_timeouts_to_step(step: dict):
+    def _apply_timeouts_to_step(step: Dict):
         for environment in step.get("stepEnvironments", []):
             _apply_timeouts_to_environment(environment)
 
--- a/src/deadline/client/job_bundle/timeouts.py
+++ b/src/deadline/client/job_bundle/timeouts.py
@@ -39,10 +39,7 @@ def add_timeouts_to_job_template(
     Args:
         template (Dict[str, Any]): The job template to modify.
         timeout_settings (Optional[TimeoutSettings]): Configuration for timeout values. If None,
-            default timeout settings will be used. If is_deactivated is True, no timeouts will be added.
-
-    Raises:
-        ValueError: If timeout values in TimeoutSettings are negative.
+            default timeout settings will be used. If activated is False, no timeouts will be added.
 
     Example:
         >>> template = {
--- a/src/deadline/client/job_bundle/timeouts.py
+++ b/src/deadline/client/job_bundle/timeouts.py
@@ -1,0 +1,90 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from typing import Any, Optional
+from dataclasses import dataclass
+
+SECONDS_IN_A_MINUTE = 60
+SECONDS_IN_AN_HOUR = 60 * 60
+SECONDS_IN_A_DAY = SECONDS_IN_AN_HOUR * 24
+
+
+@dataclass
+class TimeoutSettings:
+    is_activated: bool = True
+    on_enter_timeout_seconds: int = SECONDS_IN_A_DAY
+    on_exit_timeout_seconds: int = SECONDS_IN_A_DAY
+    on_run_timeout_seconds: int = 5 * SECONDS_IN_A_DAY
+
+    def __post_init__(self):
+        # Validate timeout values
+        for timeout in (
+            self.on_enter_timeout_seconds,
+            self.on_exit_timeout_seconds,
+            self.on_run_timeout_seconds,
+        ):
+            if timeout <= 0:
+                raise ValueError(f"Timeout value cannot be negative or zero: {timeout}")
+
+
+def add_timeouts_to_job_template(
+    template: dict[str, Any], timeout_settings: Optional[TimeoutSettings] = None
+) -> None:
+    """
+    Adds timeout values to actions in a job template.
+
+    This function performs an in-place modification of the job template by adding or updating timeout values
+    for all action types: onEnter, onExit, and onRun. According to OpenJD specification 2023-09,
+    timeouts must be hard-coded in the job template.
+
+    Args:
+        template (dict[str, Any]): The job template to modify.
+        timeout_settings (Optional[TimeoutSettings]): Configuration for timeout values. If None,
+            default timeout settings will be used. If is_deactivated is True, no timeouts will be added.
+
+    Raises:
+        ValueError: If timeout values in TimeoutSettings are negative.
+
+    Example:
+        >>> template = {
+        ...     "steps": [{
+        ...         "stepEnvironments": [{
+        ...             "script": {
+        ...                 "actions": {
+        ...                     "onEnter": {"command": "start"},
+        ...                     "onExit": {"command": "stop"}
+        ...                 }
+        ...             }
+        ...         }],
+        ...         "script": {
+        ...             "actions": {
+        ...                 "onRun": {"command": "Do something"}
+        ...             }
+        ...         }
+        ...     }]
+        ... }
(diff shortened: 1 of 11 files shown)


src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,88 @@
...
+from dataclasses import dataclass
+
+SECONDS_IN_A_MINUTE = 60
+SECONDS_IN_AN_HOUR = 60 * 60
+SECONDS_IN_A_DAY = SECONDS_IN_AN_HOUR * 24
+
+
+@dataclass
+class TimeoutSettings:
+    is_deactivated: bool = False
```
joel-wong-aws: This is a double negative, would `is_activated = True` be better? Usually we write conditionals in the affirmative and then explicitly negate them to avoid double negatives.

src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,88 @@
...
+    on_run_timeout_seconds: int = 5 * SECONDS_IN_A_DAY
+
+    def __post_init__(self):
+        # Validate timeout values
+        for timeout in (
+            self.on_enter_timeout_seconds,
+            self.on_exit_timeout_seconds,
+            self.on_run_timeout_seconds,
+        ):
+            if timeout < 0:
```
joel-wong-aws: We should reject timeouts of 0 as well

src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,88 @@
...
+        ... }
+        >>> add_timeouts_to_job_template(template, TimeoutSettings())
+    """
+    if timeout_settings is None:
+        timeout_settings = TimeoutSettings()
+
+    if timeout_settings.is_deactivated:
+        return
+
+    def _handle_environment(environment: dict):
```
joel-wong-aws: nit: could this be more descriptive: `_apply_timeouts_to_environment`. Similar comment for other function names below.

test/unit/deadline_client/job_bundle/test_timeouts.py
```diff
@@ -0,0 +1,326 @@
...
+                                    ],
+                                    "cancelation": {"mode": "NOTIFY_THEN_TERMINATE"},
+                                    "timeout": 7200,
+                                }
+                            }
+                        },
+                    }
+                ],
+            },
+            id="Add custom timeouts to sample job template",
```
joel-wong-aws: Can we add tests for multi-step environment, multi-job environment, and multi-step?

joel-wong-aws (changes requested):

src/deadline/client/job_bundle/timeouts.py
karthikbekalp: Sure, I can change that.

src/deadline/client/job_bundle/timeouts.py
karthikbekalp: Makes sense. But I wonder if we want to add a threshold that is greater than say 60 seconds or so? If customers provide <60 seconds timeouts (say 1 or 2 seconds) then their jobs would all fail.

src/deadline/client/job_bundle/timeouts.py
karthikbekalp: That makes sense. I'll update it.

test/unit/deadline_client/job_bundle/test_timeouts.py
karthikbekalp: I could. Let me try finding sample job templates for this.

src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,88 @@
...
+        for timeout in (
+            self.on_enter_timeout_seconds,
+            self.on_exit_timeout_seconds,
+            self.on_run_timeout_seconds,
+        ):
+            if timeout < 0:
+                raise ValueError(f"Timeout value cannot be negative: {timeout}")
+
+
+def add_timeouts_to_job_template(
```
karthikbekalp, line 29: This is similar to https://github.com/AWS-Samuel/deadline-cloud-for-nuke/blob/e97afec40e8a102fa1a54b8165231d9882c489e4/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py#L54. The idea is that we would replace this function with the one declared in Nuke in upcoming PRs.

src/deadline/client/job_bundle/timeouts.py
AWS-Samuel: I don't think we should try to establish a lower bound for acceptable timeouts other than what the openjd spec allows (which in this case is any `int > 0`)

test/unit/deadline_client/job_bundle/test_timeouts.py
```diff
@@ -0,0 +1,326 @@
...
+        TimeoutSettings(on_enter_timeout_seconds=-1)
+
+    with pytest.raises(ValueError):
+        TimeoutSettings(on_exit_timeout_seconds=-1)
+
+    with pytest.raises(ValueError):
+        TimeoutSettings(on_run_timeout_seconds=-1)
+
+
+SAMPLE_JOB_TEMPLATE = {
```
AWS-Samuel: I think we should store this along with the expected templates as actual job template `yml` files in the test directory to make this file more readable

karthikbekalp: > I don't think we should try to establish a lower bound for acceptable timeouts other than what the openjd spec allows (which in this case is any int > 0)

Make sense. 

> I think we should store this along with the expected templates as actual job template yml files in the test directory to make this file more readable

How about making a simple job template generator that can create job templates for us? Because SonarCloud would just fail the checks because of code duplication?

src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,90 @@
...
+from dataclasses import dataclass
+
+SECONDS_IN_A_MINUTE = 60
+SECONDS_IN_AN_HOUR = 60 * 60
+SECONDS_IN_A_DAY = SECONDS_IN_AN_HOUR * 24
+
+
+@dataclass
+class TimeoutSettings:
+    is_activated: bool = True
```
joel-wong-aws, line 13: If we are applying these settings by default, this is a breaking change and should be prefixed with `feat!`, right?

test/unit/deadline_client/job_bundle/test_timeouts.py
```diff
@@ -0,0 +1,141 @@
...
+    settings = TimeoutSettings()
+    assert settings.is_activated
+    assert settings.on_enter_timeout_seconds == 86400  # 1 day in seconds
+    assert settings.on_exit_timeout_seconds == 86400
+    assert settings.on_run_timeout_seconds == 432000  # 5 days in seconds
+
+
+def test_timeout_settings_custom_values():
+    settings = TimeoutSettings(
+        is_activated=True,
```
joel-wong-aws: supernit: for more test coverage, we could set this as `False` since the default is `True`

test/unit/deadline_client/job_bundle/test_timeouts.py
```diff
@@ -0,0 +1,141 @@
...
+        TimeoutSettings(on_enter_timeout_seconds=0)
+
+    with pytest.raises(ValueError):
+        TimeoutSettings(on_exit_timeout_seconds=0)
+
+    with pytest.raises(ValueError):
+        TimeoutSettings(on_run_timeout_seconds=0)
+
+
+SAMPLE_JOB_TEMPLATE = create_sample_job_template(
```
joel-wong-aws: nit: can we put these within the `@pytest.mark.parametrize(` test cases? That way, we can see the expected output directly beside the input and compare them easier

test/unit/deadline_client/job_bundle/test_timeouts.py
AWS-Samuel: SonarCloud recommendations can be ignored in certain situations where it makes sense. While a template generator could also work, it would involve more or less reimplementing the original logic which isn't necessarily the purpose of the test. The purpose is, given X template, do you get Y template after applying whatever transformation.

For an example, this is a pattern is employed in our worker agent repository here: https://github.com/aws-deadline/deadline-cloud-worker-agent/tree/mainline/test/integ/config/data

test/unit/deadline_client/testing_utilities.py
```diff
@@ -126,3 +126,84 @@ def program_that_prints_output(
         yield temp.name
     finally:
         os.remove(temp.name)
+
+
+def create_sample_job_template(
+    num_of_job_env: int,
+    num_of_step_env_per_step: int,
+    num_of_steps: int,
+    num_of_tasks_per_step: int,
```
joel-wong-aws: supernit: we could simplify `num_of` to `num` i.e. `num_job_envs`, `num_step_envs_per_step`, `num_steps`, `num_tasks_per_step`

joel-wong-aws (dismissed):

src/deadline/client/job_bundle/timeouts.py
joel-wong-aws: I've updated the PR title to reflect this.

test/unit/deadline_client/job_bundle/test_timeouts.py
karthikbekalp: IMO its more readable to understand what the tests do by reading the variables. But I'll be creating separate files anyway as per other review's recommendations so I think this will be kept as it were. :(

test/unit/deadline_client/job_bundle/test_timeouts.py
karthikbekalp: Sure, will update.

test/unit/deadline_client/testing_utilities.py
karthikbekalp: This function is removed now.

joel-wong-aws (approved):

AWS-Samuel (approved):

PR #603, opened 2025-02-14
fix: Fixed manifest download cli to download output for selected step …

### LINKED ISSUE #602: Bug: Manifest download cli does not download all output manifests
### Describe the bug

Manifest download cli does not download output manifests as it is intended to do for the provided step-id or the entire job if no step id is provided. 

### Expected Behaviour

By default, cli should download all input & output manifests (for provided step-id or entire job if no step is provided). Additionally, customer can be provided the option to download input only or output only as per their use-case

### Current Behaviour

Only input manifests for selected step & dependent step outputs are downloaded when a step-id is provided in the input. When no step is provided only the input manifests for the job are downloaded.

### Reproduction Steps

1. Run manifest download command with job id: ```deadline manifest download --job-id <job-id> --farm-id <farm-id> --queue-id <queue-id> <local-path>``` : downloads a merged manifest with only input manifests

2. Run manifest download command with job and step ids: ```deadline manifest download --job-id <job-id> --step-id <step-id> --farm-id <farm-id> --queue-id <queue-id> <local-path>``` : downloads a merged manifest with only input manifests for step-id and outputs for its step dependencies

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.) - MacOs, but that does not matter for reproduction
2. Output of `python3 --version` - 3.10.13
3. Output of `deadline --version` - 0.49.6

sakshie95:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/602

Manifest download cli does not download output manifests as it is intended to do for the provided step-id or the entire job if no step id is provided.

Added changes to download both input & output manifests (for step & dependencies if step-id is provided or for entire job if no step-id is provided) as the default behaviour. Also, added the option for downloading only input or output if customer wants to do that. 

Manifest download clis will work as intended

Added integ tests, ran existing tests and tested the changes manually with the following steps:
``hatch env prune ``
``hatch build``
``pip install <wheel file> --force-reinstall``
``hatch run test``
``hatch run integ:test``

Manual test download cli: ``deadline manifest download --job-id <job-id> --farm-id <farm-id> --queue-id <queue-id> <download location>``

N/A, because the manifest cli & api is in beta. Also default behaviour for the cli was always intended to be to download 'all' but it downloads only `input` right now.

N/A

----

### DIFF
--- a/src/deadline/client/cli/_groups/manifest_group.py
+++ b/src/deadline/client/cli/_groups/manifest_group.py
@@ -36,6 +36,7 @@ from deadline.job_attachments.models import (
     S3_MANIFEST_FOLDER_NAME,
     JobAttachmentS3Settings,
     ManifestDiff,
+    AssetType,
 )
 
 from ...exceptions import NonValidInputError
@@ -248,6 +249,18 @@ def manifest_diff(
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
 @click.option(
+    "--asset-type",
+    default=AssetType.ALL.value,
+    help="Which asset type to download:\n"
+    "INPUT means download only input asset files for given job/step.\n"
+    "OUTPUT means download only output asset files for given job/step.\n"
+    "ALL (default) means download all input & output asset files for given job/step.\n",
+    type=click.Choice(
+        [e.value for e in AssetType],
+        case_sensitive=False,
+    ),
+)
+@click.option(
     "--json", default=None, is_flag=True, help="Output is printed as JSON for scripting. "
 )
 @_handle_error
@@ -255,11 +268,12 @@ def manifest_download(
     download_dir: str,
     job_id: str,
     step_id: str,
+    asset_type: str,
     json: bool,
     **args,
 ):
     """
-    Downloads input manifest of previously submitted job.
+    Downloads input/output manifests of a submitted job as per provided asset_type
     """
     logger: ClickLogger = ClickLogger(is_json=json)
     if not os.path.isdir(download_dir):
@@ -280,6 +294,7 @@ def manifest_download(
         queue_id=queue_id,
         job_id=job_id,
         step_id=step_id,
+        asset_type=AssetType(asset_type),
         boto3_session=boto3_session,
         logger=logger,
     )
--- a/src/deadline/job_attachments/api/manifest.py
+++ b/src/deadline/job_attachments/api/manifest.py
@@ -40,6 +40,7 @@ from deadline.job_attachments.models import (
     ManifestSnapshot,
     ManifestMerge,
     default_glob_all,
+    AssetType,
 )
 from deadline.job_attachments._utils import _get_long_path_compatible_path
 from deadline.job_attachments.upload import S3AssetManager, S3AssetUploader
@@ -340,6 +341,7 @@ def _manifest_download(
     job_id: str,
     boto3_session: boto3.Session,
     step_id: Optional[str] = None,
+    asset_type: AssetType = AssetType.ALL,
     logger: ClickLogger = ClickLogger(False),
 ) -> ManifestDownloadResponse:
     """
@@ -351,6 +353,7 @@ def _manifest_download(
     job_id: Job Id to download.
     boto_session: Boto3 session.
     step_id: Optional[str]: Optional, download manifest for a step
+    asset_type: Which asset manifests should be downloaded for given job (& optionally step), options are Input, Output, All. Default behaviour is All.
     logger: Click Logger instance to print to CLI as text or JSON.
     return ManifestDownloadResponse Downloaded Manifest data. Contains source S3 key and local download path.
     """
@@ -384,6 +387,14 @@ def _manifest_download(
     # Utility function to build up manifests by root.
     manifests_by_root: Dict[str, List[BaseAssetManifest]] = dict()
 
+    # Set the values of download input & output as per selected asset types in the api request
+    download_input: bool = (
+        True if asset_type is None or asset_type in (AssetType.INPUT, AssetType.ALL) else False
+    )
+    download_output: bool = (
+        True if asset_type is None or asset_type in (AssetType.OUTPUT, AssetType.ALL) else False
+    )
+
     def add_manifest_by_root(
         manifests_by_root: Dict[str, list], root: str, manifest: BaseAssetManifest
     ):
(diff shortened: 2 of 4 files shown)


src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -225,6 +226,21 @@ def manifest_diff(
 @click.option("--step-id", help="The AWS Deadline Cloud Step to get. ")
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
+@click.option(
+    "--asset-type",
```
leongdl, line 252: There are two --asset-type options?

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -225,6 +226,21 @@ def manifest_diff(
...
+    default=AssetType.ALL,
+    type=click.Choice([e.value for e in AssetType]),
+    help="Whether to download: input only, output only, or all.",
+)
+@click.option(
```
leongdl: Duplicate? I think this one is the correct one?

src/deadline/client/cli/_groups/manifest_group.py
sakshie95: Typo, fixed it now

src/deadline/client/cli/_groups/manifest_group.py
sakshie95: Yup, this is the correct one

src/deadline/job_attachments/api/manifest.py
```diff
@@ -332,6 +333,7 @@ def _manifest_download(
     job_id: str,
     boto3_session: boto3.Session,
     step_id: Optional[str] = None,
+    asset_type: Optional[str] = None,
```
prestomation: Using `Optional[str]` here feels like we're loosening our use of the type system for no good reason, we already used Choice with AssetType on the CLI command, why not use AssetType here too? Not doing so means we have to also check for None down below. Can we better use the type system to our advantage?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -376,69 +379,116 @@ def _manifest_download(
...
+            if asset_manifest is not None:
+                logger.echo(f"Found input manifest for root: {root_path}")
+                add_manifest_by_root(
+                    manifests_by_root=manifests_by_root, root=root_path, manifest=asset_manifest
                 )
+
+        # Now handle step-step dependencies
+        if step_id is not None:
+            # Get Step-Step dependencies.
+            nextToken = ""
```
prestomation, line 437: Why do we do this? Is the first page sufficient for what we're doing here?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -376,69 +379,116 @@ def _manifest_download(
...
+        # Merge all output manifests by root.
+        for root in output_manifests_by_root.keys():
+            for manifest in output_manifests_by_root[root]:
+                logger.echo(f"Found output manifest for root: {root}")
+                add_manifest_by_root(
+                    manifests_by_root=manifests_by_root, root=root, manifest=manifest
+                )
 
     # Finally, merge all manifest paths to create unified manifests.
-    # TODO: Filter outputs by path
```
leongdl, line 449: nit, this todo is still valid.

This is intended to allow users to "Filter" what sub paths / files to download.

src/deadline/job_attachments/api/manifest.py
```diff
@@ -376,69 +379,116 @@ def _manifest_download(
...
+
+        else:
+            # Get output manifests for all steps of the job
+            output_manifests_by_root = get_output_manifests_by_asset_root(
+                s3_settings=queue_s3_settings,
+                farm_id=farm_id,
+                queue_id=queue_id,
+                job_id=job_id,
+                session=queue_role_session,
+            )
```
godobyte: `get_output_manifests_by_asset_root` takes in `step_id` as optional and does None check already, do we have to check it here again and pass in differently?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -376,69 +379,116 @@ def _manifest_download(
...
+                for root in step_manifests_by_root.keys():
+                    for manifest in step_manifests_by_root[root]:
+                        logger.echo(f"Found step-step output manifest for root: {root}")
+                        add_manifest_by_root(
+                            manifests_by_root=manifests_by_root, root=root, manifest=manifest
+                        )
+
+    # If output manifests need to be downloaded
+    if download_output:
+        logger.echo(f"Downloading output manifests for job: {job_id}")
```
godobyte: Should we optionally log step id when it's provided to avoid confusion?

src/deadline/job_attachments/api/manifest.py
sakshie95: For downloading input for a step this cli downloads the output manifests of its step dependencies.

This is a great call out - the first page may not be sufficient for this looking at the existing logic. I haven't touched the input download in this PR though so it seems like another bug with the cli out of scope for this fix. I'll create a fresh issue & address it separately

src/deadline/job_attachments/api/manifest.py
sakshie95: I had this question when I was building the cli option so I looked at existing cli integrations and found a similar cli option ``JobAttachmentsFileSystem`` being a enum type for the cli https://github.com/aws-deadline/deadline-cloud/blob/8ee8bd1c23b464395441a955ad6db664f6b32cba/src/deadline/client/cli/_groups/bundle_group.py#L107-L113 but being an Optional[str] for the api https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/api/_submit_job_bundle.py#L59.

So I kept it the same way for consistency for the customer, since this seems like what we ended up adopting for the cli vs apis. I can change it to use the type here since it would be the same behaviour for the customer although inconsistent in terms of code of other clis? Thoughts?

src/deadline/job_attachments/api/manifest.py
```diff
@@ -376,69 +379,116 @@ def _manifest_download(
...
+                job_id=job_id,
+                session=queue_role_session,
+            )
+
+        # Merge all output manifests by root.
+        for root in output_manifests_by_root.keys():
+            for manifest in output_manifests_by_root[root]:
+                logger.echo(f"Found output manifest for root: {root}")
+                add_manifest_by_root(
+                    manifests_by_root=manifests_by_root, root=root, manifest=manifest
```
godobyte, line 498: Do we expect customer to inspect the file content and tell input and output when asset type `ALL`? With the current implementation, there is no clear indication in the file name or content whether the manifest is input or output.

src/deadline/job_attachments/api/manifest.py
sakshie95: Got it, I'll add it back

src/deadline/job_attachments/api/manifest.py
sakshie95: Yes, although ``get_output_manifests_by_asset_root`` does the null check for ``step_id`` to make it modular to use. I would rather keep the logic separation at the caller here for step vs job scoped manifests to keep the logic clear instead of relying on the null checks

src/deadline/job_attachments/api/manifest.py
sakshie95: sure I can do that!

src/deadline/job_attachments/api/manifest.py
sakshie95: The distinction for inputs/outputs is under the control of the customer basis the asset root. This is how we intended the cli to be - to make it flexible for customers to download outputs/inputs only if they need the distinction at that level & all if they don't (with the option set asset roots as per their requirements)

src/deadline/job_attachments/api/manifest.py
sakshie95: Created issue: https://github.com/aws-deadline/deadline-cloud/issues/607

leongdl: Btw, its just a while loop to fix this, so its a very small lift :)

src/deadline/job_attachments/api/manifest.py
godobyte: The output asset root in most of the cases (without storage profile) is managed by worker agent, which is not controlled by customer. The hash in the file name don't give visibility on the root itself.

With that said, customer is able to differentiate via input/output type only making this less of a concern. This is also a 2-way door where we can get feedback and quickly pivot.

src/deadline/job_attachments/api/manifest.py
godobyte: Did I miss any update? L469 doesn't seem modified since last review.

src/deadline/job_attachments/api/manifest.py
sakshie95: Discussed this offline & updated it to use the ``AssetType`` instead of str as consistency doesn't seem a good enough reason for using loosely typed str

src/deadline/job_attachments/api/manifest.py
sakshie95: That's why there are 3 options :) My thinking is asset root with "All" will be for more advance customers and other 2 for normal use-cases where they just need one of input/output. We can always take feedback & iterate

src/deadline/job_attachments/api/manifest.py
sakshie95: There's no ``step-id`` here to log, I added it to line 434 when it exists

godobyte: The reason it could be confusing is that for output, when step id is passed in, it's only downloading the step's output manifest, rather than the job's output manifest. Let's add it here for output to be clear.

godobyte (dismissed):

leongdl (dismissed):
OK as long as we fix the `next_token` in a fast follow.

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -247,6 +248,15 @@ def manifest_diff(
 @click.option("--step-id", help="The AWS Deadline Cloud Step to get. ")
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
+@click.option(
+    "--asset-type",
+    default=AssetType.ALL.value,
+    help="Which asset type to download."
+    "INPUT means download only input asset files for given job/step"
+    "OUTPUT means download only output asset files for given job/step"
+    "ALL means download all input & output asset files for given job/step",
```
epmog: The formatting here doesn't carry over to the actual help text:

```
$ deadline manifest download --help
Usage: deadline manifest download [OPTIONS] DOWNLOAD_DIR

  BETA - Download Job Attachment Manifests for a Job, or Step including
  dependencies.

Options:
  --profile TEXT                  The AWS profile to use.
  --job-id TEXT                   The AWS Deadline Cloud Job to get.
                                  [required]
  --step-id TEXT                  The AWS Deadline Cloud Step to get.
  --asset-type [input|output|all]
                                  Which asset type to download.INPUT means
                                  download only input asset files for given
                                  job/stepOUTPUT means download only output
                                  asset files for given job/stepALL means
                                  download all input & output asset files for
                                  given job/step
```

epmog: Also, the help text shows it as all caps, but the options are lowercase. We should be consistent. Other locations in the CLI use `case_sensitive=False`:

https://github.com/aws-deadline/deadline-cloud/blob/def96ba3d8334a474e89b0f1a48776212fbbac79/src/deadline/client/cli/_groups/job_group.py#L155-L160

Oof, looks like we have some mix-matching of casing for these types of options though in the same command. ie:

```
$ deadline job download-output --help
Usage: deadline job download-output [OPTIONS]
...
  --job-id TEXT                   The job to use.
  --step-id TEXT                  The step to use.
  --task-id TEXT                  The task to use.
                                  How to handle downloads if a file already
                                  exists: CREATE_COPY (default): Download the
                                  file with a new name, appending '(1)' to the
                                  end SKIP: Do not download the file
                                  OVERWRITE: Download and replace the existing
                                  file
  --output [verbose|json]         Specifies the output format of the messages
                                  printed to stdout. VERBOSE: Displays
                                  messages in a human-readable text format.
                                  JSON: Displays messages in JSON line format,
                                  so that the info can be easily
                                  parsed/consumed by custom scripts.
```

src/deadline/client/cli/_groups/manifest_group.py
```diff
@@ -247,6 +248,15 @@ def manifest_diff(
...
+    type=click.Choice([e.value for e in AssetType]),
+)
 @click.option(
     "--json", default=None, is_flag=True, help="Output is printed as JSON for scripting. "
 )
```
epmog, line 265: Not introduced here, but FYI @leongdl this is inconsistent with everywhere else that we allow a user to specify human-readable text vs. JSON. I don't like the current one, and maybe we change it later, but this should be consistent.

https://github.com/aws-deadline/deadline-cloud/blob/def96ba3d8334a474e89b0f1a48776212fbbac79/src/deadline/client/cli/_groups/job_group.py#L652-L662

src/deadline/job_attachments/api/manifest.py
```diff
@@ -384,69 +387,120 @@ def _manifest_download(
     # Utility function to build up manifests by root.
     manifests_by_root: Dict[str, List[BaseAssetManifest]] = dict()
 
+    # Set the values of download input & output as per selected asset types in the api request
+    download_input: bool = (
+        True if asset_type is None or asset_type in (AssetType.INPUT, AssetType.ALL) else False
+    )
+    download_output: bool = (
+        True if asset_type is None or asset_type in (AssetType.OUTPUT, AssetType.ALL) else False
+    )
```
epmog, line 396: We don't have to worry about None here do we? Perhaps due to a previous comment but the argument isn't Optional

epmog (changes requested):

src/deadline/client/cli/_groups/manifest_group.py
epmog: One more follow-up, do we use `asset` publicly anywhere else? I think we mostly removed that wording... Should this be named something else?

sakshie95: The formatting is a good call out thank you! I've updated it to be:
```
  --asset-type [input|output|all]
                                  Which asset type to download: INPUT means
                                  download only input asset files for given
                                  job/step. OUTPUT means download only output
                                  asset files for given job/step. ALL
                                  (default) means download all input & output
                                  asset files for given job/step.
```

> One more follow-up, do we use `asset` publicly anywhere else? I think we mostly removed that wording... Should this be named something else?

I wouldn't have enough context on this. I see a lot of "asset" references in the code but not necessarily in cli commands. Is there a word that we decided to replace it with publicly? Open to suggestions here.. I can discuss some options & we can also do this as a follow-up since the api is in BETA?

src/deadline/client/cli/_groups/manifest_group.py
sakshie95: Yeah this seems inconsistent. I thought of fixing it but on looking more I found the "--output" option to be inconsistent across more than just this new cli. For eg. ``verbose`` is the default option for some clis (bundle_group) and not the default option for others (job_group). I think we'll have to do a pass at all clis together and make the ``--output`` behaviour consistent. I'll create an issue

src/deadline/job_attachments/api/manifest.py
sakshie95: Yes you're right, as per earlier comment I changed the ``asset_type`` from Optional[str] to ``AssetType`` which alleviates risk, but I retained this none check as a safeguard.

src/deadline/client/cli/_groups/manifest_group.py
epmog: We'll be looking into this a bit before it gets out of beta.

epmog (approved):

leongdl (approved):
Done - new diffs compared to last approval was help text only.

PR #605, opened 2025-02-16
feat: Add configurable job timeouts widget.

karthikbekalp:

DCCs that depend on `deadline-cloud` needed an easy way to patch their job templates with default timeouts for Env enter/exit/task run. 

Discussed with @AWS-Samuel about the simplest way to provide this capability. Add Kudos to @AWS-Samuel for his changes in Nuke which we will be re-using to make it helpful for all DCCs. 

- Create a helper function which can be used by different DCCs directly. [PR](https://github.com/aws-deadline/deadline-cloud/pull/600)
- Add GUI components that adds the timeouts to the SubmitJobToDeadlineDialog with the render settings that can be applied when there is `on_create_job_bundle_callback`.  [This PR]

Adds GUI for the submitters. 

### How was this change tested? [I'm currently testing this but want some quick feedback]

I tried applying the changes for my locally built Cinema 4D. I added this to the shared job settings because I believe this can be useful for all the DCCs and is not specific to a single DCC.

Here's how the submitter GUI would look like :

1. When the DCC's render submissions have `timeout_settings`. 

Screenshot 2025-02-15 at 6 07 32 PM

2. When the timeouts field is unchecked. 

2.1. The alert when the field is unchecked. 

Screenshot 2025-02-15 at 6 07 43 PM

2.2. After the customer unchecks the field, the timeout fields are blocked. 

Screenshot 2025-02-15 at 6 07 54 PM

3. When 0 second timeouts are added. 

3.1 The timeout with 0 seconds is highlighted in RED. 

Screenshot 2025-02-15 at 6 08 07 PM

3.2 With 0 seconds timeouts, the submissions will not succeed. 

Screenshot 2025-02-15 at 6 08 20 PM

Yes

Currently running it. 

No

Yes

No

No

No

----

### DIFF
--- a/src/deadline/client/job_bundle/timeouts.py
+++ b/src/deadline/client/job_bundle/timeouts.py
@@ -25,6 +25,42 @@ class TimeoutSettings:
             if timeout <= 0:
                 raise ValueError(f"Timeout value cannot be negative or zero: {timeout}")
 
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Convert the object to a dict for serialization.
+        """
+        return {
+            "is_activated": self.is_activated,
+            "on_enter_timeout_seconds": self.on_enter_timeout_seconds,
+            "on_exit_timeout_seconds": self.on_exit_timeout_seconds,
+            "on_run_timeout_seconds": self.on_run_timeout_seconds,
+        }
+
+    @classmethod
+    def from_dict(
+        cls,
+        data: Dict[str, Any],
+        default_is_activated: bool = True,
+        default_on_enter_timeout_seconds: int = SECONDS_IN_A_DAY,
+        default_on_exit_timeout_seconds: int = SECONDS_IN_A_DAY,
+        default_on_run_timeout_seconds: int = 5 * SECONDS_IN_A_DAY,
+    ) -> "TimeoutSettings":
+        """
+        Convert a dict to a TimeoutSettings object.
+        """
+        return cls(
+            is_activated=data.get("is_activated", default_is_activated),
+            on_enter_timeout_seconds=data.get(
+                "on_enter_timeout_seconds", default_on_enter_timeout_seconds
+            ),
+            on_exit_timeout_seconds=data.get(
+                "on_exit_timeout_seconds", default_on_exit_timeout_seconds
+            ),
+            on_run_timeout_seconds=data.get(
+                "on_run_timeout_seconds", default_on_run_timeout_seconds
+            ),
+        )
+
 
 def add_timeouts_to_job_template(
     template: Dict[str, Any], timeout_settings: Optional[TimeoutSettings] = None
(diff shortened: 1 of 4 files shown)


src/deadline/client/job_bundle/timeouts.py
```diff
@@ -25,6 +25,42 @@ def __post_init__(self):
...
+            on_enter_timeout_seconds=data.get(
+                "on_enter_timeout_seconds", default_on_enter_timeout_seconds
+            ),
+            on_exit_timeout_seconds=data.get(
+                "on_exit_timeout_seconds", default_on_exit_timeout_seconds
+            ),
+            on_run_timeout_seconds=data.get(
+                "on_run_timeout_seconds", default_on_run_timeout_seconds
+            ),
+        )
```
AWS-Samuel, line 62: We shouldn't need either of these. 

`to_dict` can be removed as the functionality we can get for free from [`dataclasses.asdict`](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict)

`from_dict` isn't really needed either as you can initialize the TimeoutSettings using `**` to unpack the dictionary in the TimeoutSettings constructor. Ex:


```
>>> d = {
...     "is_activated": True,
...     "on_exit_timeout_seconds": 1000
... }
>>> TimeoutSettings(**d)
TimeoutSettings(is_activated=True, on_enter_timeout_seconds=86400, on_exit_timeout_seconds=1000, on_run_timeout_seconds=432000)
```

src/deadline/client/ui/widgets/shared_job_timeout_settings_widget.py
```diff
@@ -0,0 +1,192 @@
...
+    QAbstractSpinBox,
+)
+
+
+SECONDS_IN_A_MINUTE = 60
+SECONDS_IN_AN_HOUR = 60 * 60
+SECONDS_IN_A_DAY = SECONDS_IN_AN_HOUR * 24
+
+
+class SharedJobTimeoutSettingsWidget(QGroupBox):  # pylint: disable=too-few-public-methods
```
AWS-Samuel, line 28:
```suggestion
class JobTimeoutSettingsWidget(QGroupBox):  # pylint: disable=too-few-public-methods
```

nit, I would removed `shared` from the class name and path

src/deadline/client/ui/widgets/shared_job_timeout_settings_widget.py
```diff
@@ -0,0 +1,192 @@
...
+            tooltip="Maximum duration of each action which sets up the job for rendering, such as scene load.",
+            row=1,
+        )
+        hookup_zero_callback(self.on_enter_timeouts)
+
+        self.on_exit_timeouts = create_timeout_row(
+            label="Teardown Timeout",
+            tooltip="Maximum duration of action which tears down the setup required for rendering.",
+            row=2,
+        )
```
AWS-Samuel, line 106: We may want to consider providing a way for downstream consumers to configure these tooltips because the text may not apply to all integrations.

src/deadline/client/job_bundle/timeouts.py
karthikbekalp: Ah, didn't know about these. I'll make the changes for these.

src/deadline/client/ui/widgets/shared_job_timeout_settings_widget.py
karthikbekalp: Oh ok, I thought we needed it because we used in the shared settings tab. Will remove it.

src/deadline/client/ui/widgets/shared_job_timeout_settings_widget.py
karthikbekalp: Or is there a generic tooltip that I can add instead?  Providing a way for downstream consumers might not be easy as this component is around 4 or 5 layers deep.

karthikbekalp: Discussed this with others. Instead of having it in "Shared Job Settings", we will be moving it to "Job specific settings". 

Additionally, we will make this a widget that can be customizable by every DCC so I'll update this accordingly.

karthikbekalp:
Closing this PR as the changes between this PR and the current feature is way different.

PR #611, opened 2025-02-21
feat: Add configurable job timeouts widget.

karthikbekalp:

DCCs that depend on `deadline-cloud` needed an easy way to patch add timeouts GUI components to their submitters. 

Discussed with @AWS-Samuel about the simplest way to provide this capability. Kudos to @AWS-Samuel for his changes in Nuke which heavily inspires this PR. 

Once this change is merged, I'll make corresponding changes in Cinema 4D so that we can add timeouts specifically for task run, Cinema 4D launch and shutdown. (We have a [feature request for this ](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/174))

Adds GUI that can be useful for adding timeouts to submitters. 

I tried applying the changes for my locally built Cinema 4D. 

1. When the submitter is opened, the default timeouts are set. 

Screenshot 2025-02-21 at 3 13 47 PM

2. When one or more checkboxes are deactivated, there is a warning label at the bottom. 

Screenshot 2025-02-21 at 3 14 10 PM

3. When zero timeout is introduced. 

Screenshot 2025-02-21 at 3 14 33 PM

4. When zero timeout and checkboxes are deactivated. 

Screenshot 2025-02-21 at 3 14 53 PM

5. When submitting the job with zero timeout. 

Screenshot 2025-02-21 at 3 15 10 PM

Yes

Currently running it. 

No

Yes

No

No

No

----


src/deadline/client/ui/widgets/job_timeout_settings_widget.py
```diff
@@ -0,0 +1,202 @@
...
+            self.set_error_style(False)
+        self.set_enabled(is_checked)
+        self.update_suffix()
+
+
+class JobTimeoutSettingsWidget(QGroupBox):
+    def __init__(self, *, settings: Any, parent=None):
+        super().__init__("Timeouts", parent=parent)
+        self.timeout_rows: List[TimeoutRow] = []
+        self._build_ui(settings.timeout_settings)
```
karthikbekalp: Hmm, maybe this should be `settings.timeouts` instead? I'll update this.

src/deadline/client/ui/widgets/job_timeout_settings_widget.py
```diff
@@ -0,0 +1,202 @@
...
+            self.set_error_style(True)
+        else:
+            self.set_status_icon("")
+            self.set_error_style(False)
+        self.set_enabled(is_checked)
+        self.update_suffix()
+
+
+class JobTimeoutSettingsWidget(QGroupBox):
+    def __init__(self, *, settings: Any, parent=None):
```
karthikbekalp: I think docstrings will help for this class.

karthikbekalp: Updated.

src/deadline/client/ui/widgets/job_timeout_settings_widget.py
karthikbekalp: Updated.

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,308 @@
...
+        Updates the settings object with the current values from the UI.
+
+        Args:
+            settings: Settings object to update with current timeout values
+        """
+        for index in range(len(settings.timeouts)):
+            current_row = self.timeout_rows[index]
+            current_entry = settings.timeouts[index]
+
+            current_entry.is_activated = current_row.checkbox.isChecked()
```
joel-wong-aws: nice to have (can be punted): hide the timeout values altogether if unchecked.

src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,24 @@
...
+
+from dataclasses import dataclass
+
+SECONDS_IN_A_MINUTE = 60
+SECONDS_IN_AN_HOUR = 60 * 60
+SECONDS_IN_A_DAY = SECONDS_IN_AN_HOUR * 24
+
+
+@dataclass
+class TimeoutEntry:
```
joel-wong-aws: nice to have: we could have another dataclass `TimeoutSettings` which contains a list of `TimeoutEntry`s. This could also help us re-use the code for sticky settings.

src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,24 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from dataclasses import dataclass
+
+SECONDS_IN_A_MINUTE = 60
+SECONDS_IN_AN_HOUR = 60 * 60
```
joel-wong-aws: supernit: to be consistent, this would be `SECONDS_IN_AN_HOUR = SECONDS_IN_A_MINUTE * 60`

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,308 @@
...
+        self.days_box = QSpinBox(parent, minimum=0, maximum=365)
+        self.days_box.setSuffix(" days")
+
+        self.hours_box = QSpinBox(parent, minimum=0, maximum=23)
+        self.hours_box.setSuffix(" hours")
+
+        self.minutes_box = QSpinBox(parent, minimum=0, maximum=59)
+        self.minutes_box.setSuffix(" minutes")
+
+    def add_to_layout(self, layout: QGridLayout, row: int):
```
joel-wong-aws: nit: instead of `add_to_layout` function that takes a parent `layout` as input, can we `_build_ui` during initialization of the TimeoutRow and then simply add that to the parent `layout`?

joel-wong-aws (dismissed):

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Hmm, IMO I think I'd like to punt on this so that we can ship this feature quickly. It's a 2 way door and we can revisit and change it if necessary later on if we feel the UI could be improved. :)

src/deadline/client/job_bundle/timeouts.py
karthikbekalp: Hmm, good suggestion. I can add this in a follow-up PR to improve this.

src/deadline/client/job_bundle/timeouts.py
karthikbekalp: Makes sense. I can update this in the follow-up PR where I add the TimeoutSettings class.

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Yup, I could do that.

src/deadline/client/job_bundle/timeouts.py
```diff
@@ -0,0 +1,24 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mwiebe: This file is for the UI part of the library, so I think it should be within the `ui` directory as well. See https://github.com/aws-deadline/deadline-cloud/tree/mainline/src/deadline/client/ui/dataclasses for similar functionality that is already there.

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,308 @@
...
+)
+
+# UI Constants
+WARNING_ICON = "⚠️"
+ERROR_ICON = "❌"
+ERROR_BG_COLOR = "#FFE4E1"  # Light red
+WARNING_BG_COLOR = "#FFF3CD"  # Light yellow
+
+
+class TimeoutRow:
```
mwiebe: How about `TimeoutEntryWidget` for this class name?

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,308 @@
...
+    QCheckBox,
+    QGridLayout,
+    QWidget,
+)
+
+from ...job_bundle.timeouts import (
+    TimeoutEntry,
+    SECONDS_IN_A_DAY,
+    SECONDS_IN_A_MINUTE,
+    SECONDS_IN_AN_HOUR,
```
mwiebe: Instead of defining all these constants and doing custom math, I'd suggest using the Python standard library https://docs.python.org/3/library/datetime.html#timedelta-objects

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,308 @@
...
+        self.set_enabled(is_checked)
+        self.update_suffix()
+
+
+class JobTimeoutsWidget(QGroupBox):
+    """
+    UI element to hold timeout settings of the submission
+
+    The settings object should be a dataclass with:
+      - `timeouts: List[TimeoutEntry]`. The timeouts to be applied for the job.
```
mwiebe: This indirection through a dataclass doesn't appear to serve any purpose. Could you remove that and instead directly pass the `List[TimeoutEntry]` as the parameter?

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+
+        label = QLabel()
+        label.setStyleSheet(f"""
+            QLabel {{ 
+                background-color: {bg_color};
+                color: black;
+                padding: 5px;
+                border-radius: 3px;
+            }}
```
mwiebe, line 265: Does this work for both dark and light themes?

mwiebe (changes requested):

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: I briefly updated the background for Cinema 4D submitter to check and it looks like the warning and error labels are clearly visible. Although, I can darken or lighten the color if you like. :)

Screenshot 2025-02-25 at 3 43 10 PM

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Makes sense, my previous code needed the settings but now it does not. Removed it.

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Yeah, that makes sense. Updated.

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Makes sense, updated.

src/deadline/client/job_bundle/timeouts.py
karthikbekalp: Makes sense, updated.

joel-wong-aws (dismissed):

src/deadline/client/ui/dataclasses/timeouts.py
```diff
@@ -0,0 +1,99 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from typing import Dict, Any
+from dataclasses import dataclass
+from datetime import timedelta
+
+from deadline.client.exceptions import NonValidInputError
```
mwiebe: Within the library code, we want to use relative imports like `from ...exceptions import NonValidInputError`.

src/deadline/client/ui/dataclasses/timeouts.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@dataclass
+class TimeoutEntry:
+    """
+    Represents a single timeout configuration entry.
+
+    Attributes:
+        tooltip (str): Description or explanation of the timeout entry.
+        is_activated (bool): Flag indicating if the timeout is active. Defaults to True.
```
mwiebe: Would be good to make the variable name match the description, i.e. `active` instead of `activated`.

src/deadline/client/ui/dataclasses/timeouts.py
```diff
@@ -0,0 +1,99 @@
...
+        Returns:
+            Dict[str, Any]: Dictionary containing activation status and timeout duration.
+        """
+        return {"is_activated": self.is_activated, "seconds": self.seconds}
+
+
+@dataclass
+class TimeoutEntries:
+    """
+    Manages a collection of timeout entries.
```
mwiebe: It looks like the important thing is that they are *named* timeout entries. My first instinct was "this should just be a list[TimeoutEntries]" before seeing that, so it would be good to state that here.

src/deadline/client/ui/dataclasses/timeouts.py
```diff
@@ -0,0 +1,99 @@
...
+    def to_dict(self) -> Dict[str, Dict[str, Any]]:
+        """
+        Converts all timeout entries to a dictionary format.
+
+        Returns:
+            Dict[str, Dict[str, Any]]: Dictionary mapping labels to entry configurations.
+        """
+        return {label: entry.to_dict() for label, entry in self.entries.items()}
+
+    def update_from_dict(self, data: Dict[str, Dict[str, Any]]) -> None:
```
mwiebe: It probably makes sense to just call this `update`?

src/deadline/client/ui/dataclasses/timeouts.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        Updates existing timeout entries from a dictionary of configurations.
+
+        Args:
+            data (Dict[str, Dict[str, Any]]): Dictionary containing timeout configurations.
+        """
+        for label, saved_data in data.items():
+            if label in self.entries:
```
mwiebe, line 94: It looks like it will discard values from the input `data` if it's not already in `self`. This doesn't make sense for an `update_from_dict` function, but likely does for the UI context it's being used in. That might be a signal that this code is trying to be too general for the use case? It's specifically for the sticky settings mechanism, but is written as if it's for a broader purpose.

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,298 @@
...
+        else:
+            self.set_status_icon("")
+            self.set_error_style(False)
+        self.set_enabled(is_checked)
+        self.update_suffix()
+
+
+class JobTimeoutsWidget(QGroupBox):
+    """
+    UI element to hold timeout settings of the submission
```
mwiebe: The purpose of this widget is to put multiple different timeouts together into one UI control, something that's not clear from this description. Can you elaborate a bit more here? Think of someone who doesn't know anything about these widgets or timeouts, how would you explain the purpose of this widget to them?

src/deadline/client/ui/dataclasses/timeouts.py
mwiebe: Possible rename for this class: `TimeoutEntries` -> `TimeoutTableEntries`.

src/deadline/client/ui/widgets/job_timeouts_widget.py
mwiebe: Possible rename for the Widget class: `JobTimeoutsWidget` -> `TimeoutTableWidget`.

src/deadline/client/ui/widgets/job_timeouts_widget.py
mwiebe: That looks readable, yeah. It does look like our customizations do not follow the theme, but that's something to look into independently of a specific feature like this PR.

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,298 @@
...
+    def _hookup_change_callback(self, timeout_row: TimeoutEntryWidget):
+        """
+        Connects UI element signals to update handlers for a timeout row.
+
+        Args:
+            timeout_row: The TimeoutRow instance to hook up callbacks for
+        """
+        timeout_row.checkbox.clicked.connect(self._update_ui_state)
+        for spinbox in [timeout_row.days_box, timeout_row.hours_box, timeout_row.minutes_box]:
+            spinbox.valueChanged.connect(self._update_ui_state)
```
mwiebe: This outer widget should not have to reach inside the details of the inner TimeoutEntryWidget like this. The TimeoutEntryWidget itself should publish a single `changed` signal to use instead.

mwiebe (changes requested):

src/deadline/client/ui/dataclasses/timeouts.py
karthikbekalp: Oh makes sense, updated.

src/deadline/client/ui/dataclasses/timeouts.py
karthikbekalp: Yes sure, updated.

src/deadline/client/ui/dataclasses/timeouts.py
karthikbekalp: Sure, will update.

src/deadline/client/ui/dataclasses/timeouts.py
karthikbekalp: I think this is related the comment you added below. I changed the name to `update_from_sticky_settings` so that its clearer.

src/deadline/client/ui/dataclasses/timeouts.py
karthikbekalp: Yeah that makes sense. I updated the docstring and the name of this function to `update_from_sticky_settings` so that it's clearer.

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Yeah, makes sense. I'll update the name and docstring so that its more clear.

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: This is with a quick update to the background color for Cinema 4D submitter which uses its own style. So other customizations that follow light theme should work.

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Hmm, that makes sense. I'll refactor this.

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,318 @@
...
+        Calculates the total timeout in seconds from the current days, hours, and minutes values.
+
+        Returns:
+            int: Total number of seconds represented by the current time values
+        """
+        return (
+            self.days_box.value() * int(timedelta(days=1).total_seconds())
+            + self.hours_box.value() * int(timedelta(hours=1).total_seconds())
+            + self.minutes_box.value() * int(timedelta(minutes=1).total_seconds())
+        )
```
mwiebe, line 143: Another approach to using `timedelta`:

```python
>>> timedelta(days=1, hours=3, minutes=2).total_seconds()
97320.0
```

src/deadline/client/ui/widgets/job_timeouts_widget.py
```diff
@@ -0,0 +1,318 @@
...
+        Args:
+            seconds: Total number of seconds to be distributed across time fields
+        """
+        days = seconds // int(timedelta(days=1).total_seconds())
+        hours = (seconds % int(timedelta(days=1).total_seconds())) // int(
+            timedelta(hours=1).total_seconds()
+        )
+        minutes = (seconds % int(timedelta(hours=1).total_seconds())) // int(
+            timedelta(minutes=1).total_seconds()
+        )
```
mwiebe: Another possibility for this:

```python
>>> td = timedelta(seconds=97320)
>>> td.days
1
>>> td.seconds // (60 * 60)
3
>>> (td.seconds // 60) % 60
2
```

I think it's more clear to use the number 60 directly compared to making a constant or using timedelta to construct constants. It's better to to think directly about the code clarity than be overly strict about rules of thumb like never using a bare number. The simpler, shorter code is easier to visually check and think about.

mwiebe (dismissed):

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: That's so much cleaner. I'll update this right away.

src/deadline/client/ui/widgets/job_timeouts_widget.py
karthikbekalp: Makes sense. 

I tried using a similar approach before but then I thought it may be confusing as in first glance td.seconds and td.total_seconds() are not equivalent. Anyway, updated it.

mwiebe (approved):

joel-wong-aws (approved):

PR #615, opened 2025-02-26
fix: Fixed file number append to handle file name collisions while creating copy

### LINKED ISSUE #514: Maintenance: Improve Multiple File Download Collision Formatting
### Description

Currently, if you download your output on top of an existing download and use the CREATE_COPY option (i.e. you get colliding files), [it appends " (1)" to the file name](https://github.com/aws-deadline/deadline-cloud/blob/b02b0387fd32701bcf72bc882e78de77b6ad7423/src/deadline/job_attachments/download.py#L423), over and over. 

So if you download repeatedly, you could have a file like "output (1) (1) (1) (1) (1).png" which makes file names unnecessarily long.

### Solution

Ideally, we increment a number to the file name for any collisions when using the CREATE_COPY option when downloading outputs. That is, if we have a file named "output.png", downloading it multiple times would create new files called "output (1).png", "output (2).png", etc.

sakshie95:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/514

Currently, if you download your output on top of an existing download and use the CREATE_COPY option (i.e. you get colliding files), [it appends " (1)" to the file name](https://github.com/aws-deadline/deadline-cloud/blob/b02b0387fd32701bcf72bc882e78de77b6ad7423/src/deadline/job_attachments/download.py#L423), over and over.

So if you download repeatedly, you could have a file like "output (1) (1) (1) (1) (1).png" which makes file names unnecessarily long.

We increment a number to the file name for any collisions when using the CREATE_COPY option when downloading outputs. That is, if we have a file named "output.png", downloading it multiple times would create new files called "output (1).png", "output (2).png", etc. 

If a file is original named input (1).png and there is another file input.png that is being download a second time with CREATE_COPY option, it will be downloaded as input (2).png. The second copy of input (1).png will be downloaded as input (1) (1).png

Better experience with downloading files with CREATE_COPY for customers

Wrote unit tests to validate the logic works as expected. Also ran the download logic manually & verified files are created with the correct appended copy number, by using the download cli command:

```
deadline job download-output --job-id job-xxx --farm-id farm-xxx --queue-id queue-xxx --conflict-resolution CREATE_COPY
```

N/A

No

No

----

### DIFF
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -74,6 +74,7 @@ from ._utils import (
     _is_relative_to,
     _join_s3_paths,
 )
+from threading import Lock
 
 download_logger = getLogger("deadline.job_attachments.download")
 
@@ -312,6 +313,28 @@ def get_job_input_output_paths_by_asset_root(
     return combined_path_groups
 
 
+def _get_new_copy_file_path(
+    local_file_name: Path,
+    collision_lock: Lock,
+    collision_file_dict: DefaultDict[str, int],
+) -> Path:
+    with collision_lock:
+        file_str: str = str(local_file_name)
+        num: int = collision_file_dict[file_str]
+        new_file_name = local_file_name
+
+        # Iterate until we find a number we don't conflict with
+        while new_file_name.is_file():
+            num += 1
+            new_file_name = local_file_name.parent.joinpath(
+                f"{local_file_name.stem} ({num}){local_file_name.suffix}"
+            )
+
+        collision_file_dict[file_str] = num
+        local_file_name = new_file_name
+    return local_file_name
+
+
 def download_files_in_directory(
     s3_settings: JobAttachmentS3Settings,
     attachments: Attachments,
@@ -384,6 +407,8 @@ def download_file(
     file: RelativeFilePath,
     hash_algorithm: HashAlgorithm,
     local_download_dir: str,
+    collision_lock: Lock,
+    collision_file_dict: DefaultDict[str, int],
     s3_bucket: str,
     cas_prefix: Optional[str],
     s3_client: Optional[BaseClient] = None,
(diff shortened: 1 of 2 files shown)


src/deadline/job_attachments/download.py
```diff
@@ -312,6 +313,28 @@ def get_job_input_output_paths_by_asset_root(
...
+    collision_lock: Lock,
+    collision_file_dict: DefaultDict[str, int],
+) -> Path:
+    with collision_lock:
+        file_str: str = str(local_file_name)
+        num: int = collision_file_dict[file_str]
+        new_file_name = local_file_name
+
+        # Iterate until we find a number we don't conflict with
+        while new_file_name.is_file():
```
leongdl: Edge case - what happens when it now exceeds the long path on windows?

src/deadline/job_attachments/download.py
```diff
@@ -312,6 +313,28 @@ def get_job_input_output_paths_by_asset_root(
...
+            num += 1
+            new_file_name = local_file_name.parent.joinpath(
```
leongdl: This may not guarantee multi-process locks won't corrupt the file, a way to do this is:

with open(filename, mode="x") to reserve the file name and reserve it. And then the file name can increment if create failes.
```
"x" - Create - Creates the specified file, returns an error if the file exists
```

benl-2023 (dismissed):

leongdl (approved):

src/deadline/job_attachments/download.py
sakshie95: Thanks, fixed this!

src/deadline/job_attachments/download.py
sakshie95: The behaviour for long path handling doesn't change with this fix. This is a good catch though because it is an issue with the current logic for long paths - a file copy isn't handled at all. A possible fix is to call the ``_get_long_path_compatible_path`` function again after creating a unique path but we need some testing to verify it works. Created https://github.com/aws-deadline/deadline-cloud/issues/617 for this & get to it in a separate PR. Should be quick

benl-2023 (approved):

PR #618, opened 2025-02-28
fix: using operating system dependent path conversion to create file name hash

godobyte:
Converting Path to str uses OS-specific separators (\ vs /) can produce different hashes across OS. This is a problem when a job is submitted and run on different operating systems with the `attachment upload` and `attachment download` commands.

For example, a job is submitted using linux and run on windows generate different str encoding for the give path.

```
>>> from pathlib import Path
>>> path_str = "/private/var/folders/qz/djsmwst123b1b0gv040qvv900000gq/T/pytest-of/pytest-40/test_worker_uses_step_step_dep0/job_attachment_bundle_step_step_dependencies"
>>> path_path = Path("/private/var/folders/qz/djsmwst123b1b0gv040qvv900000gq/T/pytest-of/pytest-40/test_worker_uses_step_step_dep0/job_attachment_bundle_step_step_dependencies")
>>> path_str.encode()
b'/private/var/folders/qz/djsmwst123b1b0gv040qvv900000gq/T/pytest-of/pytest-40/test_worker_uses_step_step_dep0/job_attachment_bundle_step_step_dependencies'
>>> str(path_path).encode()
b'\\private\\var\\folders\\qz\\djsmwst123b1b0gv040qvv900000gq\\T\\pytest-of\\pytest-40\\test_worker_uses_step_step_dep0\\job_attachment_bundle_step_step_dependencies'
```

The hash for source root was used for comparison in to correlate path and manifest file in attachment commands. The inconsistent hash leads to no path find and default to working directory.

Use the string path directly passed in.

Fix a bug for the caller where Path were passed in that's not matching the current operating system that lead to unexpected behaviour.

- Have you run the integration tests? [N/A]
- Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended that you ensure that the docker-based unit tests pass. [Yes]
- Ran E2E tests for cross-OS use case, i.e. submit job from Linux and run on Windows via `hatch run pytest -- test/e2e/test_job_submissions.py`
```
============================= slowest 5 durations ==============================
390.88s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_success
378.04s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
121.04s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_uses_step_step_dependencies
114.68s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_fails_job_attachment_sync_when_file_does_not_exist_in_bucket
114.67s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_reports_canceled_session_actions_as_canceled[envEnter]
============ 22 passed, 5 skipped, 1 warning in 2010.04s (0:33:30) =============
```

- Are relevant docstrings in the code base updated? [N/A]

No

No

----

### DIFF
--- a/src/deadline/job_attachments/asset_sync.py
+++ b/src/deadline/job_attachments/asset_sync.py
@@ -356,9 +356,9 @@ class AssetSync:
         manifest_paths_by_root: dict[str, str] = dict()
 
         for root, manifest in merged_manifests_by_root.items():
-            (_, _, manifest_name) = S3AssetUploader._gather_upload_metadata(
+            (_, manifest_name) = S3AssetUploader._get_hashed_file_name_from_root_str(
                 manifest=manifest,
-                source_root=Path(self._local_root_to_src_map[root]),
+                source_root=self._local_root_to_src_map[root],
                 manifest_name_suffix=manifest_name_suffix,
             )
 
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -233,6 +233,7 @@ class S3AssetUploader:
         """
         hash_alg = manifest.get_default_hash_alg()
         manifest_bytes = manifest.encode().encode("utf-8")
+        # Converting Path to str uses OS-specific separators (\ vs /), which can produce different hashes across OS
         manifest_name_prefix = hash_data(str(source_root).encode(), hash_alg)
         manifest_name = f"{manifest_name_prefix}_{manifest_name_suffix}"
 
@@ -253,6 +254,21 @@ class S3AssetUploader:
         self._write_local_input_manifest(manifest_write_dir, manifest_name, manifest, root_dir_name)
 
         self._write_local_manifest_s3_mapping(manifest_write_dir, manifest_name, full_manifest_key)
+
+    @staticmethod
+    def _get_hashed_file_name_from_root_str(
+        manifest: BaseAssetManifest,
+        source_root: str,
+        manifest_name_suffix: str,
+    ) -> tuple[HashAlgorithm, str]:
+        """
+        Gathers metadata information of manifest to be used for writing the local manifest
+        """
+        hash_alg = manifest.get_default_hash_alg()
+        manifest_name_prefix = hash_data(source_root.encode(), hash_alg)
+        manifest_name = f"{manifest_name_prefix}_{manifest_name_suffix}"
+
+        return (hash_alg, manifest_name)
 
     @staticmethod
     def _write_local_input_manifest(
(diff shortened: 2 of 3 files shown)


src/deadline/job_attachments/asset_sync.py
```diff
@@ -356,9 +356,9 @@ def _check_and_write_local_manifests(
         manifest_paths_by_root: dict[str, str] = dict()
 
         for root, manifest in merged_manifests_by_root.items():
-            (_, _, manifest_name) = S3AssetUploader._gather_upload_metadata(
```
leongdl, line 359: Is `_gather_upload_metadata` still used anywhere else? Or is it a dead method now.

src/deadline/job_attachments/upload.py
```diff
@@ -254,6 +255,21 @@ def _write_local_manifest(
...
+        manifest_name_suffix: str,
+    ) -> tuple[HashAlgorithm, str]:
+        """
+        Gathers metadata information of manifest to be used for writing the local manifest
+        """
+        hash_alg = manifest.get_default_hash_alg()
+        manifest_name_prefix = hash_data(source_root.encode(), hash_alg)
+        manifest_name = f"{manifest_name_prefix}_{manifest_name_suffix}"
+
+        return (hash_alg, manifest_name)
```
leongdl, line 271: Why are we returning `hash_alg` if it is not consumed above?

src/deadline/job_attachments/upload.py
```diff
@@ -233,6 +233,7 @@ def _gather_upload_metadata(
         """
         hash_alg = manifest.get_default_hash_alg()
         manifest_bytes = manifest.encode().encode("utf-8")
+        # Converting Path to str uses OS-specific separators (\ vs /), which can produce different hashes across OS
```
leongdl, line 236: So we are not normalizing ? or does path keep things always in linux separator?

jericht (approved):

src/deadline/job_attachments/asset_sync.py
godobyte: `_gather_upload_metadata` is used by asset upload. This is not an issue for other callers as the file names are not in use. For uploaded input, the manifest file names are in the job metadata. For uploaded output, the manifest name is not used either.

src/deadline/job_attachments/upload.py
godobyte: I've tried few ways, but haven't found a way to make it consistent for cross-OS edge cases. Since the root path is passed in and static as a string, directly converting from a string seems reasonable for the worker agent use case.

src/deadline/job_attachments/upload.py
godobyte: This function somewhat followed `_gather_upload_metadata`, I thought the hash algo can be passed back if caller wants to do additional check. I don't have strong preference of returning hash_algo or not, do you think this is blocking?

leongdl (approved):

PR #622, opened 2025-03-12
fix: Deadline Cloud Job Attachments CLI did not expose file override modes to allow users to customize how files are downloaded.

leongdl:
Fixes: Deadline Cloud Job Attachments CLI did not expose file override modes that allow users to customize how files are downloaded.

- `deadline cloud` CLI supports Attachment download CLIs (beta). For attachment downloads, we also support file collision handling. Either skip, override, or download to a new name. This was not supported in `attachment download`, and only worked in `job download`.

- Connect the missing argument, and also make it available in the API. Default is to CREATE_COPY to a new name.

- Updated CLI text:
```
Usage: deadline attachment download [OPTIONS]

  BETA - Download Job Attachment data files for given manifest(s).

Options:
  -m, --manifests TEXT            File path(s) to manifest formatted file(s).
                                  File name has to contain the hash of
                                  corresponding source path.  [required]
  --s3-root-uri TEXT              Job Attachments S3 root uri including bucket
                                  name and root prefix.
  --path-mapping-rules TEXT       Path to a file with the path mapping rules
                                  to use.
  --profile TEXT                  The AWS profile to use for interacting with
                                  Job Attachments S3 bucket.
                                  How to handle downloads if a file already
                                  exists: CREATE_COPY (default): Download the
                                  file with a new name, appending '(X)' to the
                                  end. X is incremented for each duplicate
                                  SKIP: Do not download the file OVERWRITE:
                                  Download and replace the existing file
```
- Bridge the feature gap between `job download` and `attachment download`!

Yes, hatch run test 100% passes

Yes 
```
================================================================== warnings summary ==================================================================
test/integ/cli/test_cli_attachment.py:92
  /Users/leongdl/work/deadline-cloud/test/integ/cli/test_cli_attachment.py:92: PytestUnknownMarkWarning: Unknown pytest.mark.cross_account - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.cross_account

test/integ/deadline_job_attachments/test_job_attachments.py:1289
  /Users/leongdl/work/deadline-cloud/test/integ/deadline_job_attachments/test_job_attachments.py:1289: PytestUnknownMarkWarning: Unknown pytest.mark.cross_account - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.cross_account

test/integ/deadline_job_attachments/test_job_attachments.py:1339
  /Users/leongdl/work/deadline-cloud/test/integ/deadline_job_attachments/test_job_attachments.py:1339: PytestUnknownMarkWarning: Unknown pytest.mark.cross_account - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.cross_account

test/integ/deadline_job_attachments/test_job_attachments.py:1388
  /Users/leongdl/work/deadline-cloud/test/integ/deadline_job_attachments/test_job_attachments.py:1388: PytestUnknownMarkWarning: Unknown pytest.mark.cross_account - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.cross_account

test/integ/deadline_job_attachments/test_job_attachments.py:1472
  /Users/leongdl/work/deadline-cloud/test/integ/deadline_job_attachments/test_job_attachments.py:1472: PytestUnknownMarkWarning: Unknown pytest.mark.cross_account - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.cross_account

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================ slowest 5 durations =================================================================
16.81s call     test/integ/cli/test_cli_manifest_download.py::TestManifestDownload::test_manifest_download_job_asset_type_with_no_step_dependency[input-True]
13.46s call     test/integ/cli/test_cli_manifest_download.py::TestManifestDownload::test_manifest_download_job_asset_type_with_no_step_dependency[all-True]
10.86s call     test/integ/cli/test_cli_manifest_download.py::TestManifestDownload::test_manifest_download_asset_type_with_job_step_dependency[None-True]
10.76s call     test/integ/cli/test_cli_manifest_download.py::TestManifestDownload::test_manifest_download_asset_type_with_job_step_dependency[all-True]
10.72s call     test/integ/cli/test_cli_manifest_download.py::TestManifestDownload::test_manifest_download_asset_type_with_job_step_dependency[all-False]
=============================================== 49 passed, 4 skipped, 5 warnings in 274.21s (0:04:34) ================================================
```

CLI Argument is updated.

NA

No

No

----

### DIFF
--- a/src/deadline/client/cli/_groups/attachment_group.py
+++ b/src/deadline/client/cli/_groups/attachment_group.py
@@ -21,7 +21,7 @@ from deadline.client import api
 from deadline.job_attachments import api as attachment_api
 from deadline.job_attachments._aws.deadline import get_queue
 from deadline.job_attachments.exceptions import MissingJobAttachmentSettingsError
-from deadline.job_attachments.models import JobAttachmentS3Settings
+from deadline.job_attachments.models import FileConflictResolution, JobAttachmentS3Settings
 
 
 @click.group(name="attachment")
@@ -51,6 +51,21 @@ def cli_attachment():
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use. ")
 @click.option(
     "--profile", help="The AWS profile to use for interacting with Job Attachments S3 bucket."
+)
+@click.option(
+    "--conflict-resolution",
+    type=click.Choice(
+        [
+            FileConflictResolution.SKIP.name,
+            FileConflictResolution.OVERWRITE.name,
+            FileConflictResolution.CREATE_COPY.name,
+        ],
+        case_sensitive=False,
+    ),
+    help="How to handle downloads if a file already exists:\n"
+    "CREATE_COPY (default): Download the file with a new name, appending '(X)' to the end. X is incremented for each duplicate\n"
+    "SKIP: Do not download the file\n"
+    "OVERWRITE: Download and replace the existing file",
 )
 @click.option("--json", default=None, is_flag=True, help="Output is printed as JSON for scripting.")
 @_handle_error
@@ -94,12 +109,24 @@ def attachment_download(
     if not s3_root_uri:
         raise MissingJobAttachmentSettingsError("No valid s3 root path available")
 
+    # Apply conflict resolution setting from Config.
+    conflict_resolution = FileConflictResolution.CREATE_COPY
+    conflict_resolution_setting = config_file.get_setting(
+        "settings.conflict_resolution", config=config
+    )
+    if (
+        conflict_resolution_setting
+        and conflict_resolution_setting != FileConflictResolution.NOT_SELECTED.name
+    ):
+        conflict_resolution = FileConflictResolution[conflict_resolution_setting]
+
     attachment_api.attachment_download(
         manifests=manifests,
         s3_root_uri=s3_root_uri,
         boto3_session=boto3_session,
         path_mapping_rules=path_mapping_rules,
         logger=logger,
+        conflict_resolution=conflict_resolution,
     )
 
 
--- a/src/deadline/job_attachments/api/attachment.py
+++ b/src/deadline/job_attachments/api/attachment.py
@@ -11,7 +11,11 @@ from dataclasses import asdict
 from deadline.job_attachments.api._utils import _read_manifests
 from deadline.job_attachments.asset_manifests.base_manifest import BaseAssetManifest
 from deadline.job_attachments.download import download_files_from_manifests
-from deadline.job_attachments.models import JobAttachmentS3Settings, PathMappingRule
+from deadline.job_attachments.models import (
+    FileConflictResolution,
+    JobAttachmentS3Settings,
+    PathMappingRule,
+)
 from deadline.job_attachments.progress_tracker import DownloadSummaryStatistics
 from deadline.job_attachments.upload import S3AssetUploader
 from deadline.client.cli._groups.click_logger import ClickLogger
@@ -25,6 +29,7 @@ def attachment_download(
     boto3_session: boto3.Session,
     path_mapping_rules: Optional[str] = None,
     logger: ClickLogger = ClickLogger(False),
+    conflict_resolution: FileConflictResolution = FileConflictResolution.CREATE_COPY,
 ):
     """
     BETA API - This API is still evolving.
@@ -79,6 +84,7 @@ def attachment_download(
         manifests_by_root=merged_manifests_by_root,
         cas_prefix=s3_settings.full_cas_prefix(),
         session=boto3_session,
+        conflict_resolution=conflict_resolution,
     )
     logger.echo(download_summary)
     logger.json(asdict(download_summary.convert_to_summary_statistics()))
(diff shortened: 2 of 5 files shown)


benl-2023 (dismissed):

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -94,12 +109,24 @@ def attachment_download(
     if not s3_root_uri:
         raise MissingJobAttachmentSettingsError("No valid s3 root path available")
 
+    # Apply conflict resolution setting from Config.
```
sakshie95, line 112: This is the done the other way around in our download cli. If we get a file conflict resolution option (``--conflict-resolution``) from the cli option we override the config to the provided value and use it throughout: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/job_group.py#L671-L673. Let's do the same here for consistency b/w our download clis?

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -94,12 +109,24 @@ def attachment_download(
...
+    conflict_resolution = FileConflictResolution.CREATE_COPY
```
sakshie95, line 113: You can provide the default value for the enum in the cli click option which could be cleaner, eg. https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/manifest_group.py#L253

sakshie95 (changes requested):

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -52,6 +52,21 @@ def cli_attachment():
...
+    type=click.Choice(
+        [
+            FileConflictResolution.SKIP.name,
+            FileConflictResolution.OVERWRITE.name,
+            FileConflictResolution.CREATE_COPY.name,
+        ],
+        case_sensitive=False,
+    ),
+    help="How to handle downloads if a file already exists:\n"
+    "CREATE_COPY (default): Download the file with a new name, appending '(1)' to the end\n"
```
godobyte: Could we avoid using specific number `1` since it's appending incremental numbers. Maybe something like
``When a file with the same name already exists in the directory, automatically creates a copy with an incremental number in parentheses``.

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -52,6 +52,21 @@ def cli_attachment():
 @click.option(
     "--profile", help="The AWS profile to use for interacting with Job Attachments S3 bucket."
 )
+@click.option(
+    "--conflict-resolution",
```
godobyte, line 56: How does this behave with this `--conflict-resolution` flag and has a default value?

https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/job_group.py#L400

Would it still prompt customer to select again?

leongdl: https://github.com/aws-deadline/deadline-cloud/blob/69d6c04adced90a1fd892ccbdbc0ac3948511621/src/deadline/client/cli/_groups/job_group.py#L632-L646

This is no different than `deadline job download`, this adds a simple CLI option for customers to see in `--help`

In your link, the CLI also has the list of conflict files. Here we do not. This CLI only uses the overwrite setting as a guide. Since this has `--json` I am not adding more interactive elements like the job download version.

src/deadline/client/cli/_groups/attachment_group.py
leongdl: Sure - nice commend fix.

src/deadline/client/cli/_groups/attachment_group.py
leongdl: No we should not, because we have a default setting in the config. That is the default that is always applied. 

Applying a default here means, we will by default override the user's config.

src/deadline/client/cli/_groups/attachment_group.py
leongdl: I don't think that is true - the config is parsing `**args` first at like 85 above.
```
    # Setup config
```

The Config here takes care of applying the input conflict resolution over config defaults. This block here is only to handle any null cases for defensive programming.

src/deadline/client/cli/_groups/attachment_group.py
sakshie95: ah makes sense, wouldn't be applicable to this param

src/deadline/client/cli/_groups/attachment_group.py
sakshie95: synced offline - we're already overriding the cli options which is not a part of this diff :)

sakshie95 (dismissed):

benl-2023 (dismissed):

sakshie95 (approved):
Thank you for making SonarQube happy by removing the code duplications :)

godobyte (approved):

epmog (approved):

PR #638, opened 2025-03-26
fix: allow pagination over step dependencies in manifest download cli

### LINKED ISSUE #607: Bug: Manifest download cli does not paginate through step dependencies
### Describe the bug

Manifest download cli downloads inputs for a step & the outputs of it step-step dependencies if they exist. Right now, it just goes through the first page of the response for ``list_step_dependencies`` api and does not take into account pagination

### Expected Behaviour

It should paginate through the ``list_step_dependencies`` response and get all step dependencies for the step parameter provided.

### Current Behaviour

it just goes through the first page of the response for ``list_step_dependencies`` api and does not take into account pagination

### Reproduction Steps

1. Find a job template with many step-step dependencies (with 1 step having more step dependencies than the defined pagesize of 100)
2. Run manifest download command with a job and step id: ``deadline manifest download --job-id <job-id> --step-id <step-id> --farm-id <farm-id> --queue-id <queue-id> <local-path> `` 

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.) - Any, does not matter
2. Output of `python3 --version` - 3.10.13
3. Output of `deadline --version` - 0.49.6

YutongLi291:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/607

Manifest download cli downloads inputs for a step & the outputs of it step-step dependencies if they exist. Right now, it just goes through the first page of the response for list_step_dependencies api and does not take into account pagination.

I was not able to replicate this bug, as we only allow a max of 32 step-step dependencies as of right now in our job templates.

However, for future if more than 100 step dependencies for a single step is possible, I have added pagination to the list_step_dependencies API call.

Manifest download CLIs will paginate through step step dependencies if necessary.

- Have you run the unit tests? Yes, `hatch run test`
- Have you run the integration tests? Yes, `hatch run integ:test`

NA

No

No

----


jericht (approved):

src/deadline/job_attachments/api/manifest.py
```diff
@@ -433,36 +433,40 @@ def add_manifest_by_root(
...
-                        s3_settings=queue_s3_settings,
-                        farm_id=farm_id,
-                        queue_id=queue_id,
-                        job_id=job_id,
-                        step_id=dependent_step["stepId"],
-                        session=queue_role_session,
-                    )
+            # Get Step-Step dependencies with pagination
+            next_token = ""
+            while next_token is not None:
```
sakshie95, line 438: Can you point me to where the 32 step-step dependency limit is defined for the job template? Is it mutable by the customer? I would think it should be in which case we should be able to test with more than 100 dependencies for a step

sakshie95 (changes requested):
Can you rebase branch with mainline and make sure there's no conflicts

src/deadline/job_attachments/api/manifest.py
leongdl: regardless of 32, it still makes sense to support paginated list all. We cannot assume the internal implementation of the backend ,

leongdl (approved):

YutongLi291:
> Can you rebase branch with mainline and make sure there's no conflicts

There are no conflicts

src/deadline/job_attachments/api/manifest.py
sakshie95: My comment was majorly to check if we can add an integration test for it & not just a unit test as in the PR. But given this limit, we wouldn't be able to so.

sakshie95 (approved):

PR #639, opened 2025-03-28
feat!: better cache expiry behavior

shidarin:

Customers could not expire files in their S3 bucket without running into cache errors, where the local cache would insist a file should exist in the S3 job attachment bucket, and thus not upload to it. The job would then fail in Deadline Cloud, because the file had been deleted from the bucket.

This feature does 2 things:

1. Cache expiry can now be set with an environment variable, giving customers full control of how often a file is expected to stick around in S3
2. When a submission finds the file already in S3, it sets `last_seen` to the upload time of the file. Combined with the above change, customers can now use lifecycle policies on their S3 bucket and have this behave as expected.

Customers can now set an S3 lifecycle policy without causing their farm jobs to fail.
Customers who have no S3 lifecycle policy will incur more head fetches, as this library will continue to attempt to set a last seen date of the original upload, which might be much longer than 30 days. It will then keep refetching the head.....

image

It was not. The actual functionality of this function did not seem to be tested in unit tests (ಠ_ಠ) and the existing errors were not modified. There are 2 test failures- one because I adjusted the log message to be more specific about upload time, and MagicMock does not like the datetime f string formatting, and another because a dependent function is no longer asserting False (not sure why)

Updated typing. Saw no references in documentation.

API breaking, as the modified function previously returned True/False and now returns `Optional[datetime]`. If present, datetime will always return `True`, but this will be cold comfort to anyone expecting a real bool. To behave exactly like it used to, the return should be wrapped in a `bool()`

No.

----

### DIFF
--- a/docs/mcp_guide.md
+++ b/docs/mcp_guide.md
@@ -61,6 +61,8 @@ With the AWS Deadline Cloud MCP Server, you can use natural language for various
 - "Submit the render job in /path/to/my-job-bundle"
 - "Submit a job with priority 80 to my render queue"
 - "Show me the status of job job-3a907bac684841f69fc344867ee166de"
+- "Get the logs for session session-abc123 to see why my task failed"
+- "Show me the CloudWatch logs for session session-xyz789"
 - "Download output from job job-3a907bac684841f69fc344867ee166de"
 - "Download output from step step-render in job job-3a907bac684841f69fc344867ee166de"
 ```
@@ -76,7 +78,7 @@ The MCP server provides access to all allowlisted/configured Deadline Cloud API 
 - `deadline_list_storage_profiles_for_queue()`: List storage profiles for a queue
 
 - `deadline_check_authentication_status()`: Check current authentication status
-
+- `deadline_get_session_logs()`: Get CloudWatch logs for a specific session
 - `deadline_submit_job()`: Submit an Open Job Description job bundle to AWS Deadline Cloud
 - `deadline_download_job_output()`: Download job output files from AWS Deadline Cloud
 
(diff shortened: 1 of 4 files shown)


src/deadline/job_attachments/upload.py
```diff
@@ -533,11 +537,12 @@ def upload_object_to_cas(
                 progress_tracker=progress_tracker,
             )
             is_uploaded = True
+            cache_time = self._get_current_timestamp()
```
shidarin: _get_current_timestamp() is a convenience function masking the str behavior you see on line 531. If it's not used in a bunch of places, it just means we need to duplicate that behavior whenever we reference datetime strings that aren't current timestamps.

src/deadline/job_attachments/upload.py
```diff
@@ -769,16 +774,16 @@ def _is_file_within_directory(self, file_path: Path, directory_path: Path) -> bo
...
         Check whether the file has already been uploaded by doing a head-object call.
         """
         try:
-            self._s3.head_object(
+            last_modified = self._s3.head_object(
                 Bucket=bucket,
                 Key=key,
-            )
-            return True
+            ).get("LastModified")
```
shidarin: LastModified is always present in a successful return of head_object according to the documentation, but using a get here will fail safely.

However, it might lead clients of this function to think the file hasn't been uploaded, when it has, and just exists in some weird bizzaro state.

src/deadline/job_attachments/upload.py
```diff
@@ -789,7 +794,7 @@ def file_already_uploaded(self, bucket: str, key: str) -> bool:
                 raise JobAttachmentsS3ClientError(
                     "checking if object exists", error_code, bucket, key, message
                 ) from exc
-            return False
+            return
```
shidarin: Now returns None instead of False, so that we can use the `Optional[datetime]` return type rather than 2 different return types.

PR #645, opened 2025-03-31
docs: document job attachments S3 storage format

crowecawcaw:
The job attachments implementation was under documented especially about what the manifests look like and where they're stored in S3.

Add some documentation.

Easier to ramp up on job attachments.

Docs only

Yes

No

No

----

### DIFF
--- a/scripts/find_installbuilder.py
+++ b/scripts/find_installbuilder.py
@@ -15,6 +15,7 @@ from common import UnsupportedOSError
 _INSTALL_BUILDER_ARCHIVE_FILENAME = {
     "Windows": "VMware-InstallBuilder-Professional-windows.tar.gz",
     "Linux": "VMware-InstallBuilder-Professional-linux.tar.gz",
+    "Darwin": "VMware-InstallBuilder-Professional-macos.tar.gz",
 }
 
 


src/deadline/job_attachments/README.md
```diff
@@ -33,6 +33,47 @@ RootPrefix/
...
+
+Where:
+- `<file_hash>` is the hash of the file contents
+- `<hash_algorithm>` is the algorithm used to generate the hash (currently "xxh128")
+
+This content-addressable approach ensures that identical files are only stored once, regardless of their original filenames or paths.
+
+#### Manifest Files
+
+Manifests are stored for both job inputs and task outputs. Input manifests for jobs are stored under:
```
mwiebe: The statement about input manifests isn't true in general, it's just what the current job submission code does by default. Could we have this document that the location is defined by a value in the Job resource, and you can retrieve it with the Deadline CLI `deadline job get` or the AWS CLI `aws deadline get-job`? See https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_GetJob.html#API_GetJob_ResponseSyntax for the path `attachments/manifests/rootPath`.

mwiebe (changes requested):

src/deadline/job_attachments/README.md
mwiebe: Maybe it's worth linking https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/run-jobs-job-attachments.html#job-attachments-in-depth from somewhere here.

src/deadline/job_attachments/README.md
```diff
@@ -56,7 +56,7 @@ This content-addressable approach ensures that identical files are only stored o
 
 #### Manifest Files
 
-Manifests are stored for both job inputs and task outputs. Input manifests for jobs are stored under:
+Manifests are stored for both job inputs and task outputs. The S3 key for the input manifest for a job can be found by [calling GetJob](https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_GetJob.html#API_GetJob_ResponseSyntax) and looking in the response under `attachments.manifests.rootPath`. While the input manifest location is not formally defined, in the current job attachments implementation it follows the pattern:
```
mwiebe: >While the input manifest location is not formally defined

It is formally defined to be the value returned from that API. Maybe say "While the input manifest will not follow this pattern in general, ..."

PR #647, opened 2025-04-01
feat: reduce calls to S3 when fetching output asset manifests

### LINKED ISSUE #646: Feature request: Optimize manifest fetching by removing redundant HEAD requests
### Describe the problem

Currently when fetching manifests, the job attachments library [makes a HEAD request](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/download.py#L667) on each manifest to get the asset root from the object metadata. But the metadata is also returned in the GetObject requests that are already being done on the manifest. We do not need the HEAD call if we restructure our requests a bit.

### Proposed Solution

Get the metadata from the GetObject request made in this call: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/download.py#L754

Needs some light refactoring because currently manifests are fetched using the s3transfer manager which does not surface the headers. Feature request for the s3transfer library: https://github.com/boto/s3transfer/issues/340

We probably do not need the transfer manager to fetch the manifests since they'll be small and don't need multipart downloading.

### Example Use Cases

Downloading job attachments when there are many tasks.

crowecawcaw:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/646

Currently when downloading job attachments, we make 3 calls to S3:
1. HEAD request in our code to get the asset-root metadata
2. HEAD request by S3 transfer manager to determine the object size to see if multipart downloading is necessary
3. GetObject request to download the object.

Only the third is necessary since it includes the metadata and since nearly all manifests will be relatively small.

Get everything we need in a single GetObject request. In the edge case where the manifest is very large and doesn't fit in one reasonable-sized request, fall back on the previous chunked downloader.

Reduces S3 calls by 2/3 when getting output manifests.

  that you ensure that the docker-based unit tests pass. Yes, all passing.

I did a manual performance test by downloading a job with 1000 tasks  that each had a small output file. Downloading the output was about twice as fast (5:11 -> 2:38):
```
$ time deadline job download-output --farm-id farm-ff07fe0f9ea043e1a3a53310a256b090 --queue-id queue-31f0feaec56940e88b57561a5c1b71a1 --job-id job-f66d811f5b8a4adc9c8b9c7ea1c9e837                                    
Downloading output from Job 'Download Performance Test'

Download Summary:
    Downloaded 1000 files totaling 1.02 MB.
    Total download time of 35.51796 seconds at 28.83 KB/s.
        /.../output (1000 files)

deadline job download-output --farm-id farm-ff07fe0f9ea043e1a3a53310a256b090   27.06s user 7.80s system 11% cpu 5:11.73 total

$ time deadline job download-output --farm-id farm-ff07fe0f9ea043e1a3a53310a256b090 --queue-id queue-31f0feaec56940e88b57561a5c1b71a1 --job-id job-f66d811f5b8a4adc9c8b9c7ea1c9e837
Downloading output from Job 'Download Performance Test'

Download Summary:
    Downloaded 1000 files totaling 1.02 MB.
    Total download time of 35.97593 seconds at 28.46 KB/s.
        /.../output (1000 files)

deadline job download-output --farm-id farm-ff07fe0f9ea043e1a3a53310a256b090   21.73s user 4.61s system 44% cpu 59.354 total
```

n/a

No

No

----

### DIFF
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -741,17 +741,15 @@ def get_output_manifests_by_asset_root(
     except JobAttachmentsError:
         return outputs
 
-    for key in manifests_keys:
-        asset_root, asset_manifest = get_asset_root_and_manifest_from_s3(
-            manifest_key=key,
-            s3_bucket=s3_settings.s3BucketName,
-            session=session,
-        )
-        if not asset_root:
-            raise MissingAssetRootError(
-                f"Failed to get asset root from metadata of output manifest: {key}"
-            )
-        outputs[asset_root].append(asset_manifest)
+    with concurrent.futures.ThreadPoolExecutor(max_workers=S3_DOWNLOAD_MAX_CONCURRENCY) as executor:
+        futures = [executor.submit(get_asset_root_and_manifest_from_s3, key, s3_settings.s3BucketName, session) for key in manifests_keys]
+        for key, future in zip(manifests_keys, futures):
+            asset_root, asset_manifest = future.result()
+            if not asset_root:
+                raise MissingAssetRootError(
+                    f"Failed to get asset root from metadata of output manifest: {key}"
+                )
+            outputs[asset_root].append(asset_manifest)
 
     return outputs
 
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -5,7 +5,6 @@
 from __future__ import annotations
 
 import concurrent.futures
-import io
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


src/deadline/job_attachments/download.py
```diff
@@ -86,20 +86,45 @@
...
     s3_client = get_s3_client(session=session)
     try:
-        file_buffer = io.BytesIO()
-        s3_client.download_fileobj(
-            s3_bucket,
-            manifest_key,
-            file_buffer,
-            ExtraArgs={"ExpectedBucketOwner": get_account_id(session=session)},
+        # Attempt to get the manifest and metadata in a single request. If it's larger than a single 8MB
+        # chunk, fetch it in chunks using the S3 transfer manager.
```
mwiebe: It looks like 8MB gets downloaded twice in the larger case. I wonder if it would be better to not fall back to the S3 transfer manager, and instead only wrap parallelism around multi-manifest downloads?

mwiebe (dismissed):

src/deadline/job_attachments/download.py
crowecawcaw: Is the intuition there than multi-manifest downloads are unlikely to have large manifests?

S3 can download 5GBs in a single GetObject request. We could remove the fallback and have no special logic at all if we assume the manifests are less than 5GB, but I wondered if that assumption might not be great long term. Downloading the first 8MB twice is a bit inefficient, but I'm guessing large manifests are an edge case so I went with this because it seemed simpler.

mwiebe: I did some scans of manifests that I've generated, and the largest one I found was 1.3MB with about 9000 files. It was generated by the job https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/job_bundles/copy_s3_prefix_to_job_attachments. The largest output in tests I ran was 1.1MB with 7400 files.

A larger case might be a manifest of 1 million files, with data per path of 256 bytes. That would be about 256MB. Based on that, I think 5GB as an upper limit on manifest size is reasonable, and simplifying the code to not have the fallback would be good.

crowecawcaw:
I updated the code to run the manifest downloads in parallel threads. On a job with 1000 tasks, I'm now seeing download times from from 5m to 1m.

mwiebe (approved):

src/deadline/job_attachments/download.py
```diff
@@ -750,18 +723,20 @@ def get_output_manifests_by_asset_root(
...
-        asset_manifest = get_manifest_from_s3(
-            manifest_key=key,
-            s3_bucket=s3_settings.s3BucketName,
-            session=session,
-        )
-        asset_root = _get_asset_root_from_s3(key, s3_settings.s3BucketName, session)
-        if not asset_root:
-            raise MissingAssetRootError(
-                f"Failed to get asset root from metadata of output manifest: {key}"
+    with concurrent.futures.ThreadPoolExecutor(max_workers=S3_DOWNLOAD_MAX_CONCURRENCY) as executor:
```
leongdl, line 726: We'll need to change this w/ CRT changes. CRT can manage threaded downloads much better.

leongdl (approved):
Have you tested the attachment CLIs? EG: `hatch run integ:test` ?

src/deadline/job_attachments/download.py
crowecawcaw: The S3 transfer manager with the CRT client is actually less efficient than this. S3Transfer makes a HEAD request against every object to estimate its size before downloading it. For manifests, the objects will almost always be small so the HEAD request is unnecessary and we can just make a single GetObject request instead. There's this feature request for their package: https://github.com/boto/s3transfer/issues/170

Also the S3 transfer manager does not currently return headers for the objects it downloads, and we need the headers to get the asset root from the S3 object metadata. I opened a feature request to support it here: https://github.com/boto/s3transfer/issues/340

PR #665, opened 2025-04-15
feat(experimental): skeleton and locking for incremental output downloads for all jobs running on a queue

sakshie95:
Fixes: N/A. Skeleton code for a new feature in experimental state right now.

Customers have to manually download outputs of a job using CLI or DCM after a job has been completed successfully. This is painful and slow.

We want to allow customers to setup automatic downloads of outputs on their farm without having to wait for tasks to complete in the simplest way possible

Skeleton of a new feature still in flux (not ready to use), customers can setup a cron using this CLI command to setup automatic downloads on their queues. Feature is gated by environment variable ``ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD`` which needs to be exported for the command to be visible.

Added unit tests for pid locking, feature flag test and also tested it manually.

``unset ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD``
```
 deadline queue incremental-output-download -h                                              
Usage: deadline queue [OPTIONS] COMMAND [ARGS]...
Try 'deadline queue -h' for help.

Error: No such command 'incremental-output-download'.
```

``export ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD="True"  ``
```
deadline queue incremental-output-download -h   
Usage: deadline queue incremental-output-download [OPTIONS]

  BETA - Download Job Output data incrementally for all jobs running on a
  queue as session actions finish. The command bootstraps once using a
  bootstrap lookback specified in minutes and continues downloading from the
  last saved progress thereafter until bootstrap is forced

Options:
  --path-mapping-rules TEXT       Path to a file with the path mapping rules
                                  to use.
  --bootstrap-lookback-in-minutes INTEGER
                                  Downloads outputs for job-session-actions
                                  that have been completed since these many
                                  minutes at bootstrap. Default value is 0
                                  minutes.
  --download-progress-location TEXT
                                  Proceed downloading from previous progress
                                  file at this location, if it exists. If
                                  parameter not provided or file does not
                                  exist, the download will start from the
                                  provided bootstrap lookback in minutes or
                                  it's default value.   [required]
  --force-bootstrap               Ignores the previous download progress and
                                  forces command to start from the bootstrap
                                  lookback period specified in minutes.
                                  Default value is False.
                                  How to handle downloads if an output file
                                  already exists: CREATE_COPY (default):
                                  Download the file with a new name, appending
                                  '(1)' to the end SKIP: Do not download the
                                  file OVERWRITE: Download and replace the
                                  existing file. Default behaviour is to
                                  OVERWRITE.
```

- Has the README.md been updated? If you modified CLI arguments, for instance. No ReadMe so far for this feature

yes ``psutil``

N/A

Security considerations are being made. Feature is currently gated by feature flag (as an env variable - ``ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD``) and this is only an experimental draft.

----


pyproject.toml
```diff
@@ -46,6 +46,7 @@ dependencies = [
   "jsonschema == 4.17.*",
   "pywin32 >= 307,< 311; sys_platform == 'win32'",
   "QtPy == 2.4.*",
+  "psutil >= 7.0.0"
```
benl-2023, line 49: Do we need to add this to our THIRD_PARTY_LICENSES file?

src/deadline/client/_pid_utils.py
```diff
@@ -0,0 +1,77 @@
...
+                # Create a new pid file with the current process id
+                _obtain_pid_lock_atomically(pid_file_full_path, logger)
+
+    except Exception:
+        # We already checked the file exists before reading it.
+        # For any other unexpected exceptions, we should not delete the pid file and raise an error.
+        raise
+
+
+def _obtain_pid_lock_atomically(pid_file_full_path, logger):
```
benl-2023: Lets annotate the function parameter and return types.

src/deadline/client/_pid_utils.py
```diff
@@ -0,0 +1,77 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import os
+import psutil
+
+PID_FILE_NAME = "incremental_output_download.pid"
+
+
+def check_and_obtain_pid_lock_if_available(download_progress_location, logger):
```
benl-2023: Can we annotate the signature here with types?

src/deadline/client/_pid_utils.py
```diff
@@ -0,0 +1,77 @@
...
+                raise RuntimeError(
+                    f"Another download is in progress at {download_progress_location}, use --force-bootstrap or wait for previous download to finish"
+                )
+            except psutil.NoSuchProcess:
+                # No such process exists with the process id, so we can delete the pid file
+                logger.echo(f"Process with pid {pid} is not running. Deleting pid file.")
+                os.remove(pid_file_full_path)
+
+                # Create a new pid file with the current process id
+                _obtain_pid_lock_atomically(pid_file_full_path, logger)
```
benl-2023: Have we considered using a boolean return value instead of raising an exception to communicate PID lock acquisition success?

src/deadline/client/api/_queue_apis.py
```diff
@@ -0,0 +1,86 @@
...
+from typing import Optional
+import boto3
+from deadline.client.cli._groups.click_logger import ClickLogger
+from deadline.client import _pid_utils
+
+import os
+
+
+@api.record_function_latency_telemetry_event()
+def _incremental_output_download(
```
benl-2023, line 16: Should we annotate return type as None?

src/deadline/client/api/_queue_apis.py
```diff
@@ -0,0 +1,86 @@
...
+        logger.echo(f"Download failed because of error : {e}")
+        return
+    except Exception as e:
+        logger.echo(
+            f"Failed to obtain lock for download progress at {download_progress_location} due to unexpected exception : {e}"
+        )
+        return
+
+
+def _validate_file_inputs_for_incremental_output_download(
```
benl-2023, line 58: Same comment about return type annotation.

sakshie95: Sure I can add it

src/deadline/client/api/_queue_apis.py
sakshie95: Sure I can add it

src/deadline/client/_pid_utils.py
sakshie95: Sure thing, will do

src/deadline/client/_pid_utils.py
sakshie95: Sure I can do that

pyproject.toml
sakshie95: Yes good call, I think I'll have to do that.

leongdl: See: https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#library-dependencies

I think the original intent from Mark is not to "use" the 3rd party library, but to directly insource it inside of deadline cloud. See the link I provided. I believe the preference is to just take all the code.

src/deadline/client/api/_queue_apis.py
```diff
@@ -0,0 +1,86 @@
...
+    since these many minutes at bootstrap. Default value is 0 minutes.
+    :param download_progress_location: location of the download progress file
+    :param force_bootstrap: force bootstrap and ignore current download progress. Default value is False.
+    :param path_mapping_rules: path mapping rules for cross OS path mapping
+    :param boto3_session: boto3 session
+    :param logger: Click logger component
+    :return:
+    """
+
+    try:
```
leongdl, line 43: I think this function is not finished, maybe leave a comment? Right now it just creates a pid lock file.

src/deadline/client/api/_queue_apis.py
```diff
@@ -0,0 +1,86 @@
...
+        return
+    except Exception as e:
+        logger.echo(
+            f"Failed to obtain lock for download progress at {download_progress_location} due to unexpected exception : {e}"
+        )
+        return
+
+
+def _validate_file_inputs_for_incremental_output_download(
+    download_progress_location: str, path_mapping_rules: Optional[str] = None
```
leongdl: Something doesn't add up here. Especially with path mapping I would expect to have > 1 download location?

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -95,3 +104,115 @@ def queue_get(**args):
...
+
+if os.environ.get("ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD") is not None:
+
+    @cli_queue.command(
+        name="incremental-output-download",
+        help="BETA - Download Job Output data incrementally for all jobs running on a queue as session actions finish.\n"
+        "The command bootstraps once using a bootstrap lookback specified in minutes and\n"
+        "continues downloading from the last saved progress thereafter until bootstrap is forced",
+    )
+    @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use. ")
```
leongdl: nit, extra space at the end.

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -95,3 +104,115 @@ def queue_get(**args):
...
+        "--json", default=None, is_flag=True, help="Output is printed as JSON for scripting."
+    )
+    @click.option(
+        "--bootstrap-lookback-in-minutes",
+        default=0,
+        help="Downloads outputs for job-session-actions that have been completed since these many\n"
+        "minutes at bootstrap. Default value is 0 minutes.",
+    )
+    @click.option(
+        "--download-progress-location",
```
leongdl: Or is this the "cache file" of what is / was being downloaded?

If yes, maybe change the name to be something like "tool status file". Right now this reads more like "download files destination" of some kind.

leongdl:
Comment for the future, we need integration tests for this tool!!

src/deadline/client/cli/_groups/queue_group.py
sakshie95: Yep this is the state file location.. I can rename it to be more clear

src/deadline/client/api/_queue_apis.py
sakshie95: Same comment as https://github.com/aws-deadline/deadline-cloud/pull/665#discussion_r2047770281? This is the state file location and doesn't have anything to do with the actual outputs download location

src/deadline/client/cli/_groups/queue_group.py
sakshie95: I've renamed it to ``saved-progress-checkpoint-location``

src/deadline/client/cli/_groups/queue_group.py
sakshie95: fixed

src/deadline/client/api/_queue_apis.py
sakshie95: Yeah its partly implemented. Discussed this offline, feature is behind a feature flag so its not super imp but I'll add a disclaimer to the help text.

src/deadline/client/_pid_utils.py
sakshie95: returned a boolean for the positive case, but I'd like to keep the raising exception to differentiate b/w the causes for the pid lock acquisition failure

pyproject.toml
sakshie95: I'll be creating a separate PR for the third party licenses file. It requires a process to be followed

sakshie95: Separate PR for tracking this change: https://github.com/aws-deadline/deadline-cloud/pull/667

benl-2023 (dismissed):

pyproject.toml
leongdl: When? If our next release goes out before your next PR then we are in trouble.

src/deadline/client/_pid_utils.py
```diff
@@ -0,0 +1,85 @@
...
+    # Generate a tmp file for writing the pid file as a whole and making the pid locking atomic
+    tmp_file_name = pid_file_full_path + "~tmp"
+
+    # Get the current process id to write to pid file
+    text = str(os.getpid())
+
+    logger.echo(f"Creating new pid file at {pid_file_full_path} with pid {text}")
+
+    # Open tmp file in write mode and write the current process id to it
+    with open(tmp_file_name, "w+") as f:
```
leongdl, line 77: How does this make the function atomic? W+ will open an existing file and seek to 0. Replace will override the file (?)
```
If the destination exists and is a file, it will be replaced without error if the action performing user has permission. This method may fail if the source and destination are on different filesystems.
```

pyproject.toml
leongdl: OK - discussed offline as next PR with license file will be added soon.

src/deadline/client/_pid_utils.py
leongdl: Discussed offline, we need a more unique file name.

leongdl: Can be done in next PR.

leongdl (dismissed):

src/deadline/client/_pid_utils.py
sakshie95: This is a great catch. Discussed offline, I'll do this as a follow-up in the next PRs as the feature is still WIP

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -95,3 +104,117 @@ def queue_get(**args):
...
+        "--bootstrap-lookback-in-minutes",
+        default=0,
+        help="Downloads outputs for job-session-actions that have been completed since these many\n"
+        "minutes at bootstrap. Default value is 0 minutes.",
+    )
+    @click.option(
+        "--saved-progress-checkpoint-location",
+        help="Proceed downloading from previous progress file at this location, if it exists.\n"
+        "If parameter not provided or file does not exist,\n"
+        "the download will start from the provided bootstrap lookback in minutes or it's default value. \n",
```
miabatta: nit: "it's default value" -> "its default value"

miabatta (dismissed):
Looks great, thanks so much! Just had the tiniest of grammar nits but I won't hold up merging because of that :)

src/deadline/client/cli/_groups/queue_group.py
sakshie95: since I was getting a PR out for unit tests, I put this in too

leongdl (approved):

miabatta (approved):

PR #668, opened 2025-04-18
chore: allow the Deadline endpoint to be read from an AWS profile con…

crowecawcaw:
For using non-standard endpoints, we can set the `AWS_ENDPOINT_URL_DEADLINE` env var. But in some cases like using a submitter in a DCC, it's not easy to set env vars. Also it can be annoying to have to set the env var every time it's used.

Allow the Deadline endpoint to be set as a property of the AWS profile.

Easier to use non-standard endpoints when using submitters.

Manual testing and unit tests.

Yes

No

No

----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -179,8 +179,20 @@ Notes:
   3. `export INTEG_TEST_JA_CROSS_ACCOUNT_BUCKET=<your-bucket-name-in-the-second-account>`
   4. Run the integration tests.
 * AWS Developers note: If testing with a non-production deployment of AWS Deadline Cloud then you will have to
-define the `AWS_ENDPOINT_URL_DEADLINE` environment variable to the non-production endpoint URL. For example,
-production endpoints look like: `export AWS_ENDPOINT_URL_DEADLINE="https://deadline.$AWS_DEFAULT_REGION.amazonaws.com"`
+define the endpoint. There are two options:
+   * Set the `AWS_ENDPOINT_URL_DEADLINE` environment variable to the non-production endpoint URL. For example,
+production endpoints look like: `export AWS_ENDPOINT_URL_DEADLINE="https://deadline.$AWS_DEFAULT_REGION.amazonaws.com"`.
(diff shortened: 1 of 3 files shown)


DEVELOPMENT.md
```diff
@@ -179,8 +179,20 @@ Notes:
   3. `export INTEG_TEST_JA_CROSS_ACCOUNT_BUCKET=<your-bucket-name-in-the-second-account>`
   4. Run the integration tests.
 * AWS Developers note: If testing with a non-production deployment of AWS Deadline Cloud then you will have to
-define the `AWS_ENDPOINT_URL_DEADLINE` environment variable to the non-production endpoint URL. For example,
-production endpoints look like: `export AWS_ENDPOINT_URL_DEADLINE="https://deadline.$AWS_DEFAULT_REGION.amazonaws.com"`
+define the endpoint. There are two options:
+   * Set the `AWS_ENDPOINT_URL_DEADLINE` environment variable to the non-production endpoint URL. For example,
```
leongdl, line 183: I'm not sure if we want to put this out in the public?

src/deadline/client/api/_session.py
```diff
@@ -129,6 +130,18 @@ def get_boto3_client(service_name: str, config: Optional[ConfigParser] = None) -
     """
 
     session = get_boto3_session(config=config)
+
+    # If the deadline_endpoint is set as a property of the AWS profile, use it by setting
+    # the AWS_ENDPOINT_URL_DEADLINE env variable. Useful for setting a non-standard endpoint
+    # in use cases where it's not easy to set an environment variable like using a submitter.
+    if service_name == "deadline":
```
leongdl, line 137: Why this? We have `--profile` for all the CLIs

mwiebe:
You should be able to use https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html#ss-endpoints-config without any client modifications. Maybe need to document that in a readme?

crowecawcaw:
I didn't realized the SDKs already support endpoint overriding in a very similar way. Cancelling.

PR #669, opened 2025-04-18
chore: remove unused quality check

crowecawcaw:
Our code quality checks were formatting files instead of checking formatting, so it's possible for someone to check in unformatted files.

Fix the style check so that it checks the formatting instead of doing the formatting.

Enforces code formatting.

I ran `hatch run quality` with the ruff command update and incorrectly formatted code, and it failed as expected. Then I fixed the formatting nad it passed.

n/a

No
No
----

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -11,7 +11,6 @@ typing = "mypy {args:src test} --always-false=PYQT5 --always-false=PYSIDE2 --alw
 style = ["ruff check {args:.}", "ruff format --check --diff {args:.}"]
 fmt = ["ruff format {args:.}", "style"]
 lint = ["style", "typing"]
-quality = ["style", "typing"]
 
 [[envs.all.matrix]]
 python = ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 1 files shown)


hatch.toml
```diff
@@ -11,7 +11,7 @@ typing = "mypy {args:src test} --always-false=PYQT5 --always-false=PYSIDE2 --alw
 style = ["ruff check {args:.}", "ruff format --check --diff {args:.}"]
 fmt = ["ruff format {args:.}", "style"]
 lint = ["style", "typing"]
-quality = ["ruff format {args:.}", "typing", "style"]
```
leongdl, line 14: Are you sure we want to do this? Justin swapped us over to ruff.

leongdl: https://github.com/aws-deadline/deadline-cloud/commit/25b809f75f6d23b974d9b5d2aa8f4a7f0a4e9ad6

hatch.toml
```diff
@@ -11,7 +11,7 @@ typing = "mypy {args:src test} --always-false=PYQT5 --always-false=PYSIDE2 --alw
...
+quality = ["style", "typing"]
```
epmog: This is now a duplicate of `lint` which is what's used in the github workflows. I think we can just remove this command entirely, unless we want the command that combines `fmt` and `lint`

hatch.toml
epmog: This is specifically just the `quality` script, whereas `lint` is what we use in the code quality checks for PRs: https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_python_build.yml#L69-L70

We do not have any automatic runs of `hatch run quality` currently. Honestly didn't even know it existed :)

epmog (changes requested):

mwiebe (approved):

epmog (approved):

PR #671, opened 2025-05-03
feat: return output manifest paths and hashes

ryyakobe:

Currently, when uploading output manifests in the job attachments system, there's no way to retrieve information about the uploaded manifests (paths and hashes). This information is valuable for tracking and referencing the manifests after upload, especially for workflows that need to reference these manifests later.

The solution adds functionality to return manifest information (paths and hashes) from the upload process:

1. Modified `attachment_upload()` to return a dictionary mapping asset roots to manifest information (path and hash)
2. Added comprehensive unit tests to verify the new functionality and ensure backward compatibility

This change enables users to:
- Track where output manifests are stored in S3 after upload
- Reference manifest hashes for verification or comparison
- Build more sophisticated workflows that need to reference uploaded manifests
- Maintain backward compatibility with existing code that uses the original API

- Added comprehensive unit tests for the new functionality
- All unit tests pass successfully

- [] Have you run the integration tests?
- [] Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended that you ensure that the docker-based unit tests pass.

  - Added return type annotations and documentation for all modified functions
  - Updated docstrings to explain the new return values and behavior
- [x] Has the README.md been updated? If you modified CLI arguments, for instance.
  - No CLI arguments were modified, so README.md updates were not required

No, this is not a breaking change. The PR maintains backward compatibility.

No, this change does not impact security. It only enhances the return values of existing functions without changing how files are processed or stored.

----

### DIFF
--- a/src/deadline/job_attachments/api/attachment.py
+++ b/src/deadline/job_attachments/api/attachment.py
@@ -14,6 +14,7 @@ from deadline.job_attachments.download import download_files_from_manifests
 from deadline.job_attachments.models import (
     FileConflictResolution,
     JobAttachmentS3Settings,
+    UploadManifestInfo,
     PathMappingRule,
 )
 from deadline.job_attachments.progress_tracker import DownloadSummaryStatistics
@@ -98,7 +99,7 @@ def attachment_upload(
     path_mapping_rules: Optional[str] = None,
     upload_manifest_path: Optional[str] = None,
     logger: ClickLogger = ClickLogger(False),
-):
+) -> List[UploadManifestInfo]:
     """
     BETA API - This API is still evolving.
 
@@ -113,6 +114,10 @@ def attachment_upload(
         path_mapping_rules (Optional[str], optional): Optional file path to a JSON file contains list of path mapping. Defaults to None.
         upload_manifest_path (Optional[str], optional): Optional path prefix for uploading given manifests. Defaults to None.
         logger (ClickLogger, optional): Logger to provide visibility. Defaults to ClickLogger(False).
+
+    Returns:
+        List[UploadManifestInfo]: A list of UploadManifestInfo objects corresponding to the input manifests
+        containing manifest path, hash information, and source path
 
     Raises:
         NonValidInputError: raise when any of the input is not valid.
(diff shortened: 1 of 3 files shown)


src/deadline/job_attachments/api/attachment.py
```diff
@@ -98,7 +98,7 @@ def attachment_upload(
     path_mapping_rules: Optional[str] = None,
     upload_manifest_path: Optional[str] = None,
     logger: ClickLogger = ClickLogger(False),
-):
+) -> Dict[str, Dict[str, str]]:
```
leongdl: Can we define a structure similar to: 
https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/api/manifest.py#L96

Makes things nicer to serialize as a contract :)

src/deadline/job_attachments/api/attachment.py
```diff
@@ -179,6 +189,14 @@ def attachment_upload(
             f"Uploaded assets from {rule.destination_path}, to {s3_settings.to_s3_root_uri()}/Manifests/{key}, hashed data {data}"
         )
 
+        # Store manifest information in the dictionary
+        manifest_info_dict[rule.destination_path] = {
```
leongdl: This is prefect as a data structure!

One idea that Godot and I have been thinking is to include also the original manifest root into this structure so this data is not lost.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -179,6 +189,14 @@ def attachment_upload(
...
+            "outputManifestPath": f"{s3_settings.to_s3_root_uri()}/Manifests/{key}",
```
leongdl: nit, use a local variable to share with 189 above so we know when things mutates both are updated.

src/deadline/job_attachments/asset_sync.py
```diff
@@ -935,19 +945,31 @@ def sync_outputs(
...
+            return (SummaryStatistics(), {})
         if not attachments:
             self.logger.info(f"No attachments configured for Job {job_id}, no outputs to sync.")
-            return SummaryStatistics()
+            return (SummaryStatistics(), {})
 
         all_output_files: List[OutputFile] = []
 
+        # Initialize manifest info dictionary
+        manifest_info_dict: dict[str, dict[str, str]] = {}
```
leongdl: Same nit, anonymous dictionaries are scary :)

src/deadline/job_attachments/api/attachment.py
godobyte: +1

src/deadline/job_attachments/api/attachment.py
godobyte: Additionally, what do you think about returning a list of objects rather than a map? A list seems more natural as the API signature, it's up to the caller to decide what they want to use as a key.

src/deadline/job_attachments/api/attachment.py
godobyte: WDYT about a function in JobAttachmentS3Settings to construct output manifest path from the key?

mwiebe (approved):

leongdl (approved):

src/deadline/job_attachments/api/attachment.py
leongdl: My comment is resolved.

godobyte (approved):

PR #673, opened 2025-05-06
fix!: Implement improved job attachments containment warning

### LINKED ISSUE #407: Bug: `bundle submit` prompts for permission to upload files in a job bundle's directory

mwiebe:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/407

The intent of Deadline Cloud's job submission is to warn when submitting files from directories that the user hasn't explicitly specified. The current implementation doesn't follow this rule.

* Refactor create_job_from_job_bundle to support calls from both the CLI
  and from the GUI.
    * Modify the behavior for auto_accept as follows: If is_gui is true,
      it calls the confirmation function anyway whatn the default is to
      cancel, because GUI is always an interactive context. If is_gui is
      false, it automatically accepts the default. If there are unknown
      path warnings, the default is to cancel, and it cancels with a
      message describing what it did. If there are no unknown path
      warnings, the default is to proceed, and it does so.
    * Modify the return type from Union[str, None] to str. Before, it
      sometimes would raise an exception and other times return None.
      Now, if it returns a value, that value is a job id.
    * Add a new parameter known_asset_paths. This is a list of
      directories that are ok for uploading or downloading data to S3
      for running jobs. The values in this list are added to the known
      asset paths from the local configuration and any PATH parameter
      values that are passed in to the function's job_parameters
      parameter.
    * Modify the decide_cancel_submission_callback type from
      Callable[[AssetUploadGroup], bool] to Callable[[str, bool], bool],
      and move all the business logic about when to ask for user input
      about canceling into create_job_from_job_bundle so that it's
      shared between all usages (e.g. including the CLI and GUI code).
      Previously, the CLI and GUI implemented the same logic with
      duplicated code. The callback now should always present the
      requested prompt to the user, and the boolean parameter indicates
      whether the default should be to cancel or to proceed.
    * Add a printout summarizing the job attachments files to
      upload/collect outputs for download. Even in non-interactive
      contexts, it is useful to see this information.
* Add a new setting known_asset_paths to the config file. It's a list of
  paths that should not generate warnings when outside storage profile
  locations.
    * Update the configuration GUI to include editing of this list.
* Add descriptions to all the config file properties that didn't have
  descriptions. In particular, the 'auto_accept' setting is now defined
  to automatically accept the default.
* Fix the DEVELOPMENT.md sample showing cancelation to use
  CancelationFlag that decouples the value from the window lifetime.
* Change usage of deprecated log.warn into log.warning.
* Add missing `fresh_deadline_config` fixture to failing telemetry
  tests.

BREAKING CHANGE: The behavior for the auto_accept configuration option
    has changed to automatically accept the default option that would be
    presented interactively. One exception is in a GUI context, when
    automatically accepting would cancel the operation.

BREAKING CHANGE: The parameter decide_cancel_submission_callback is
    removed from api.create_job_from_job_bundle, and a new paramete
    interactive_confirmation_callback replaces it. The previous callback
    included business logic that is now in the create job function, and
    the callback's purpose is now solely to present an interactive
    confirmation prompt.

----

### DIFF
--- a/src/deadline/client/api/_job_attachment.py
+++ b/src/deadline/client/api/_job_attachment.py
@@ -1,14 +1,15 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-from deadline.client import api
-from deadline.client.config import config_file
-from deadline.job_attachments.models import AssetRootGroup, AssetRootManifest
-from deadline.job_attachments.upload import S3AssetManager, SummaryStatistics
+from .. import api
+from ..config import config_file
+from ...job_attachments.models import AssetRootGroup, AssetRootManifest
+from ...job_attachments.upload import S3AssetManager, SummaryStatistics
+from ...job_attachments.progress_tracker import ProgressReportMetadata, ProgressStatus
 
 
 import textwrap
 from configparser import ConfigParser
-from typing import Callable, Dict, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple
 
 
 def _hash_attachments(
@@ -25,7 +26,7 @@ def _hash_attachments(
     callback. Returns a list of the asset manifests of the hashed files.
     """
 
-    def _default_update_hash_progress(hashing_metadata: Dict[str, str]) -> bool:
+    def _default_update_hash_progress(hashing_metadata: ProgressReportMetadata) -> bool:
         return True
 
     if not hashing_progress_callback:
@@ -41,7 +42,18 @@ def _hash_attachments(
     api.get_deadline_cloud_library_telemetry_client(config=config).record_hashing_summary(
         hashing_summary
     )
-    print_function_callback("Hashing Summary:")
-    print_function_callback(textwrap.indent(str(hashing_summary), "    "))
+    if hashing_summary.total_files > 0:
+        print_function_callback("Hashing Summary:")
+        print_function_callback(textwrap.indent(str(hashing_summary), "    "))
+    else:
+        # Ensure to call the callback once if no files were processed
+        hashing_progress_callback(
+            ProgressReportMetadata(
+                status=ProgressStatus.PREPARING_IN_PROGRESS,
+                progress=100,
+                transferRate=0,
+                progressMessage="No files to hash",
+            )
+        )
 
     return hashing_summary, manifests
--- a/src/deadline/client/cli/_common.py
+++ b/src/deadline/client/cli/_common.py
@@ -17,6 +17,7 @@ import sys
 from configparser import ConfigParser
 from typing import Any, Callable, Optional, Set
 import logging
+import traceback
 
 import click
 from contextlib import ExitStack
@@ -65,13 +66,11 @@ def _handle_error(func: Callable) -> Callable:
         except click.ClickException:
             # Let click exceptions fall through
             raise
-        except Exception as e:
+        except Exception:
             # Log and print out unfamiliar exceptions with additional
             # messaging.
-            click.echo(f"The AWS Deadline Cloud CLI encountered the following exception:\n{e}")
-
-            if logging.DEBUG >= logger.getEffectiveLevel():
-                logger.exception("Exception details:")
+            click.echo("The AWS Deadline Cloud CLI encountered the following exception:")
+            click.echo(traceback.format_exc())
             _prompt_at_completion(ctx)
             sys.exit(1)
 
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -6,9 +6,9 @@ All the `deadline bundle` commands.
 
 from __future__ import annotations
 
-import textwrap
 import json
 import logging
+import sys
 import re
 from typing import Any, Optional
 
@@ -16,17 +16,20 @@ import click
 from botocore.exceptions import ClientError
 
 from ... import api
-from ...config import config_file, get_setting, set_setting
+from ...config import config_file
 from ....job_attachments.exceptions import (
     AssetSyncError,
     AssetSyncCancelledError,
     MisconfiguredInputsError,
 )
-from ....job_attachments.models import AssetUploadGroup, JobAttachmentsFileSystem
-from ....job_attachments._path_summarization import human_readable_file_size, summarize_path_list
+from ....job_attachments.models import JobAttachmentsFileSystem
 
 from ...exceptions import DeadlineOperationError, CreateJobWaiterCanceled
-from .._common import _apply_cli_options_to_config, _handle_error, _ProgressBarCallbackManager
+from .._common import (
+    _apply_cli_options_to_config,
+    _handle_error,
+    _ProgressBarCallbackManager,
+)
 from ._sigint_handler import SigIntHandler
 
 logger = logging.getLogger(__name__)
@@ -69,6 +72,20 @@ def validate_parameters(ctx, param, value):
         parameters_split.append({"name": regex_match[1], "value": regex_match[2]})
 
     return parameters_split
+
+
+def _interactive_confirmation_prompt(message: str, default_response: bool) -> bool:
+    """
+    Callback to decide if submission should continue or be canceled. Returns True to continue, False to cancel.
+
+    Args:
+        warning_message (str): The warning message to display.
+        default_response (bool): The default to present as the response (True to continue, False to cancel).
+    """
+    return click.confirm(
+        message,
+        default=default_response,
+    )
 
 
 @cli_bundle.command(name="submit")
@@ -128,13 +145,19 @@ def validate_parameters(ctx, param, value):
     type=click.STRING,
     help="Name of the application submitting the bundle.",
 )
+@click.option(
+    "--known-asset-path",
+    multiple=True,
+    help="Path that should not generate warnings when outside storage profile locations. "
+    "Can be specified multiple times for different paths.",
+)
 @click.argument("job_bundle_dir")
 @_handle_error
 def bundle_submit(
     job_bundle_dir,
     job_attachments_file_system,
     parameter,
-    yes,
+    known_asset_path,
     name,
     priority,
     max_failed_tasks_count,
@@ -156,60 +179,6 @@ def bundle_submit(
     def _check_create_job_wait_canceled() -> bool:
         return sigint_handler.continue_operation
 
-    def _decide_cancel_submission(upload_group: AssetUploadGroup) -> bool:
-        """
-        Callback to decide if submission should be cancelled or not. Return 'True' to cancel.
-        Prints a warning that requires confirmation if paths are found outside of configured storage profile locations.
-        """
-        warning_message = ""
-        warning_input_paths = set()
-        warning_output_paths = set()
-        for group in upload_group.asset_groups:
-            if not group.file_system_location_name:
-                warning_input_paths.update(group.inputs)
-                warning_output_paths.update(group.outputs)
-
-        if len(warning_input_paths) > 0:
-            warning_message += "Locations for upload:\n"
-            warning_message += textwrap.indent(summarize_path_list(warning_input_paths), "  ")
-        if len(warning_output_paths) > 0:
-            warning_message += "Locations to collect job outputs for download:\n"
-            # We expect the list of output paths to be small, but truncate it to an arbitrary limit just in case for the summary
-            warning_entry_count = 4
-            if len(warning_output_paths) == warning_entry_count + 1:
-                warning_entry_count += 1
-            for path in sorted(warning_output_paths)[:warning_entry_count]:
-                warning_message += f"  {path}\n"
-            if len(warning_output_paths) > warning_entry_count:
-                warning_message += (
-                    f"\n  ... and {len(warning_output_paths) - warning_entry_count} more\n"
-                )
-
-        # Exit early if there are no warnings and we've either set auto accept or there's no files to confirm
-        if not warning_message and (
-            yes
-            or config_file.str2bool(get_setting("settings.auto_accept", config=config))
-            or upload_group.total_input_files == 0
-        ):
-            return False
-
-        message_text = (
-            f"Job submission contains {upload_group.total_input_files} input files totaling {human_readable_file_size(upload_group.total_input_bytes)}. "
-            " All input files will be uploaded to S3 if they are not already present in the job attachments bucket."
-        )
-        if warning_message:
-            message_text += (
-                f"\n\nFiles were specified outside of the configured storage profile location(s). "
-                " Please confirm that you intend to submit a job that uses files from the following directories:\n"
-                f"{warning_message}\n"
-                "To permanently remove this warning you must only use files located within a storage profile location."
-            )
-        message_text += "\n\nDo you wish to proceed?"
-        return not click.confirm(
-            message_text,
-            default=not warning_message,
-        )
-
     try:
         job_id = api.create_job_from_job_bundle(
             job_bundle_dir=job_bundle_dir,
@@ -225,29 +194,29 @@ def bundle_submit(
             upload_progress_callback=upload_callback_manager.callback,
             create_job_result_callback=_check_create_job_wait_canceled,
             print_function_callback=click.echo,
-            decide_cancel_submission_callback=_decide_cancel_submission,
+            interactive_confirmation_callback=_interactive_confirmation_prompt,
             require_paths_exist=require_paths_exist,
-            submitter_name=submitter_name,
+            submitter_name=submitter_name or "CLI",
+            known_asset_paths=known_asset_path,
         )
 
         # Check Whether the CLI options are modifying any of the default settings that affect
         # the job id. If not, we'll save the job id submitted as the default job id.
         # If the submission is canceled by the user job_id will be None, so ignore this case as well.
         if (
-            job_id is not None
-            and args.get("profile") is None
+            args.get("profile") is None
             and args.get("farm_id") is None
             and args.get("queue_id") is None
             and args.get("storage_profile_id") is None
         ):
-            set_setting("defaults.job_id", job_id)
+            config_file.set_setting("defaults.job_id", job_id)
 
     except AssetSyncCancelledError as exc:
         if sigint_handler.continue_operation:
             raise DeadlineOperationError(f"Job submission unexpectedly canceled:\n{exc}") from exc
         else:
             click.echo("Job submission canceled.")
-            return
+            sys.exit(1)
     except AssetSyncError as exc:
         raise DeadlineOperationError(f"Failed to upload job attachments:\n{exc}") from exc
     except CreateJobWaiterCanceled as exc:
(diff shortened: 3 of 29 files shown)


src/deadline/client/api/_job_attachment.py
```diff
@@ -1,14 +1,15 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-from deadline.client import api
-from deadline.client.config import config_file
-from deadline.job_attachments.models import AssetRootGroup, AssetRootManifest
-from deadline.job_attachments.upload import S3AssetManager, SummaryStatistics
+from .. import api
```
leongdl, line 3: For consistency, why are we changing them all to relative imports? Any library preferences?

src/deadline/client/api/_job_attachment.py
```diff
@@ -25,7 +26,7 @@ def _hash_attachments(
     callback. Returns a list of the asset manifests of the hashed files.
     """
 
-    def _default_update_hash_progress(hashing_metadata: Dict[str, str]) -> bool:
+    def _default_update_hash_progress(hashing_metadata: ProgressReportMetadata) -> bool:
```
leongdl, line 29: Thank for making a data class - Dict[str, str] was not very meaningful!

src/deadline/client/api/_job_attachment.py
```diff
@@ -41,7 +42,18 @@ def _default_update_hash_progress(hashing_metadata: Dict[str, str]) -> bool:
...
         hashing_summary
     )
-    print_function_callback("Hashing Summary:")
-    print_function_callback(textwrap.indent(str(hashing_summary), "    "))
+    if hashing_summary.processed_files > 0:
+        print_function_callback("Hashing Summary:")
+        print_function_callback(textwrap.indent(str(hashing_summary), "    "))
+    else:
+        # Ensure to call the callback once if no files were processed
+        hashing_progress_callback(
```
leongdl, line 50: This was probably a UX bug?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -225,11 +205,15 @@ def _decide_cancel_submission(upload_group: AssetUploadGroup) -> bool:
...
             create_job_result_callback=_check_create_job_wait_canceled,
             print_function_callback=click.echo,
-            decide_cancel_submission_callback=_decide_cancel_submission,
+            decide_cancel_submission_callback=_decide_cancel_job_submission,
             require_paths_exist=require_paths_exist,
             submitter_name=submitter_name,
+            known_asset_paths=known_asset_paths,
         )
 
+        if job_id is None:
```
leongdl: Do we get any user feedback here?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -71,6 +75,22 @@ def validate_parameters(ctx, param, value):
...
 
+def _decide_cancel_job_submission(warning_message: str, cancel_by_default: bool) -> bool:
+    """
+    Callback to decide if submission should be cancelled or not. Returns 'True' to cancel.
+
+    Args:
+        warning_message (str): The warning message to display.
+        cancel_by_default (bool): If True, the default action should be to cancel, otherwise to submit the job
+    """
+    result = click.confirm(
```
leongdl: No more auto accept behavior like below?
```
# Exit early if there are no warnings and we've either set auto accept or there's no files to confirm
        if not warning_message and (
            yes
            or config_file.str2bool(get_setting("settings.auto_accept", config=config))
            or upload_group.total_input_files == 0
        ):
```

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -438,64 +440,26 @@ def on_submit(self):
...
-            self.create_job_response = job_progress_dialog.start_submission(
-                farm_id,
-                queue_id,
-                storage_profile,
-                self.job_history_bundle_dir,
-                queue_parameters,
-                asset_manager,
-                deadline,
-                auto_accept=str2bool(get_setting("settings.auto_accept")),
+            job_id = job_progress_dialog.start_job_submission(
```
leongdl, line 520: Are we losing telemetry?
```
            api.get_deadline_cloud_library_telemetry_client().record_event(
                event_type="com.amazon.rum.deadline.submission",
                event_details={
                    "submitter_name": self.submitter_name,
                },
                from_gui=True,
            )
```

src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -84,7 +84,7 @@ def on_create_job_bundle_callback(
         asset_references: AssetReferences,
         host_requirements: Optional[Dict[str, Any]] = None,
         purpose: JobBundlePurpose = JobBundlePurpose.SUBMISSION,
-    ) -> None:
+    ) -> dict[str, Any]:
```
leongdl, line 88: Can we have a data class :) ? Or What's `Any`

leongdl:
The PR is quite large and will take a few passes to grok all the nuances. 

Few things I spotted, eg missing auto accept, telemetry events were important.

src/deadline/client/api/_job_attachment.py
mwiebe: Yes, all the self-imports inside the library should be relative. This is necessary for example for vendoring to be possible.

src/deadline/client/api/_job_attachment.py
mwiebe: Yeah, this gives the GUI version an opportunity to know that there was no hashing and it's continuing to the next step.

src/deadline/client/cli/_groups/bundle_group.py
mwiebe: Before, the logic about whether to auto-accept was duplicated here and in the GUI code. I've moved it into the shared job bundle submission function.

src/deadline/client/cli/_groups/bundle_group.py
mwiebe: Yes, it calls the print function callback. On the other hand, we should probably change the function to only return success, and always raise on an error.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
mwiebe: I tried to make sure the job bundle submission function is sending the same telemetry as before, but it's worth double checking.

src/deadline/client/ui/job_bundle_submitter.py
mwiebe: The point of this is to have an open-ended interface that we can extend later. A data class and more strict typing on the values would make that much more complicated.

test/unit/deadline_client/cli/test_cli_bundle_known_paths.py
```diff
@@ -0,0 +1,269 @@
...
+                        "submit",
+                        temp_job_bundle_dir,
+                        "--known-asset-path",
+                        os.path.dirname(external_file_path),
+                    ],
+                )
+
+                # Verify warning was not shown
+                assert result.exit_code == 0, result.output
+                mock_confirm.assert_called_once_with(ANY, default=True)
```
jericht, line 357: nit: Could these three test cases be split out into separate test functions? If the first test case fails, the second and third won't run.

Maybe we can put these into a class and have the setup be a fixture or something like that?

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -303,7 +303,7 @@ def _build_ui(self):
         self.limited_max_worker_count.toggled.connect(
             self.limited_max_worker_count_radio_button_toggled
         )
-        self.max_worker_count_layout = QVBoxLayout(self)
+        self.max_worker_count_layout = QVBoxLayout()
```
jericht, line 306: Is there a benefit to not setting the parent widget here?

src/deadline/client/api/_job_attachment.py
leongdl: Got it - i'll keep an eye out and change them as I encounter them.

src/deadline/client/cli/_groups/bundle_group.py
leongdl: Got it!

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -64,11 +61,18 @@ class SubmitJobToDeadlineDialog(QDialog):
...
         on_create_job_bundle_callback: A function to call when the dialog needs to create a Job Bundle. It
-            is called with arguments (widget, job_bundle_dir, settings, queue_parameters, asset_references)
+            is called with arguments:
+            (widget, job_bundle_dir, settings, queue_parameters, asset_references, host_requirements, purpose).
+            It can return either None or a dict with parameters about the submission. Currently,
+            the additional parameters supported are:
+            {
+                # See documentation for deadline.client.api.create_job_from_job_bundle about this parameter
+                "known_asset_paths": ["/path/1", ...],
+            }
```
jericht: nit: Looks like the actual parameter doesn't have a typing annotation on it. Could we add one?

```py
    on_create_job_bundle_callback: typing.Callable[[<types of the args>], dict[str, typing.Any]],
```

We can also consider defining a [Protocol](https://mypy.readthedocs.io/en/stable/protocols.html#callback-protocols) for a more clear interface with the kwarg names. For a callback with 7 parameters, I think this could be helpful.

jericht (commented):
I need more time to finish my review, I'll do another pass later

src/deadline/client/ui/widgets/shared_job_settings_tab.py
mwiebe: It was giving an error message that there was already a layout, passing self here was a bug.

test/unit/deadline_client/cli/test_cli_bundle_known_paths.py
mwiebe: Yes, I could do those changes, but I kind of like the way it is now better. It's currently laid out in a linear fashion, so you can read the test from top to bottom and understand the setup, cases, etc. As a class with fixtures, that would be fragmented into pieces and then harder to see which parts have the same environment or not. I also think it's fine that the three cases don't run independently as separate failures - they're all confirming logic in the same parts of the code, and when debugging issues due to a failure, development iteration within the test worked naturally for me.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
mwiebe: Good idea, I think the Protocol should work well.

mwiebe: Hmm, seems that mypy doesn't do a structural check, even though I matched all the types it doesn't allow it. I will use the Callable[...] instead then. UPDATE: I think I have it working. It was kind of finicky.

mwiebe: Actually, Callable[] can't represent the keyword-only argument. I guess I'll put in the type annotation and use `#  type: ignore` to make things pass mypy.

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -71,6 +75,22 @@ def validate_parameters(ctx, param, value):
...
+        warning_message,
+        default=not cancel_by_default,
+    )
+    print("result:", result)
```
leongdl: This uses print, would it break any --json CLI usage flows?

src/deadline/client/ui/job_bundle_submitter.py
leongdl: I see, because in other places I saw various uses of the key and value so its harder to process which keys are there and how to interpret the value.

leongdl (commented):
Is this PR ready? its getting very large. Hard to track major changes across different revisions now.

src/deadline/client/cli/_groups/bundle_group.py
mwiebe: That's a debug print, will remove

mwiebe:
> Is this PR ready? its getting very large. Hard to track major changes across different revisions now.

Yes, with the config gui in place now I'm doing some final passes over it and then will remove draft status.

src/deadline/client/cli/_groups/bundle_group.py
mwiebe: I'm changing the return type of the function to just `str`, and making all the `return None` into raising exceptions. In most error cases it already raises, so this will make it more consistent.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -397,6 +431,36 @@ def on_export_bundle(self):
...
+                existing_job_parameters = []
+
+        # Overwrite any existing values, and add new values at the end
+        combined_job_parameters = []
+        for param in existing_job_parameters:
+            combined_job_parameters.append(job_parameters_dict.pop(param["name"], param))
+        combined_job_parameters.extend(job_parameters_dict.values())
+
+        with open(job_parameters_file, "w", encoding="utf8") as f:
+            json.dump({"parameterValues": existing_job_parameters}, f, indent=1)
```
leongdl: We are always converting back to JSON here. Is this something we want to the user? or do we want to allow the user to use their original YAML format?

leongdl (commented):
LGTM - I'm also going to try out your PR locally as I am debugging something to spot any side issues.

leongdl:
> LGTM - I'm also going to try out your PR locally as I am debugging something to spot any side issues.

Will also take a peek at the unit test failures on mac since I'm on mac.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
mwiebe: I did that to keep the logic simple here. JSON is a subset of YAML, and since the structure is fairly simple I found it readable that way.

leongdl:
> LGTM - I'm also going to try out your PR locally as I am debugging something to spot any side issues.

Great fixes spotted from our discussion today. Do we want to merge this now? Or do more exploratory testing prior to merge? We can merge before our next deadline release.

mwiebe:
> > LGTM - I'm also going to try out your PR locally as I am debugging something to spot any side issues.
> 
> Great fixes spotted from our discussion today. Do we want to merge this now? Or do more exploratory testing prior to merge? We can merge before our next deadline release.

I'm inclined to merge this now so anyone else testing on mainline gets the changes too.

leongdl (dismissed):

src/deadline/client/cli/_common.py
```diff
@@ -65,13 +66,11 @@ def wraps(ctx: click.Context, *args, **kwargs):
...
-        except Exception as e:
+        except Exception:
             # Log and print out unfamiliar exceptions with additional
             # messaging.
-            click.echo(f"The AWS Deadline Cloud CLI encountered the following exception:\n{e}")
-
-            if logging.DEBUG >= logger.getEffectiveLevel():
-                logger.exception("Exception details:")
+            click.echo("The AWS Deadline Cloud CLI encountered the following exception:")
+            click.echo(traceback.format_exc())
```
prestomation, line 73: Thank you for this one, I was trying to debug something last night and the limited messaging was driving me nuts.

src/deadline/client/config/config_file.py
```diff
@@ -33,7 +33,8 @@
 # The default AWS Deadline Cloud endpoint URL
 # Environment variable that, if set, overrides the value of DEFAULT_DEADLINE_ENDPOINT_URL
 DEFAULT_DEADLINE_ENDPOINT_URL = os.getenv(
```
prestomation, line 35: Not related to your change, but reading an AWS SDK env var like this smells to me, why would we do this? This doesn't look like it is used anyway

src/deadline/client/config/config_file.py
```diff
@@ -123,6 +135,10 @@
             "This multiplier is used to calculate the size threshold. (Small files are defined as those smaller than or equal to the chunk size multiplied by this factor.)"
         ),
     },
+    "settings.known_asset_paths": {
+        "default": "",  # OS-specific path separator delimited list
+        "description": "A list of paths that should not generate warnings when outside storage profile locations, separated by the OS path separator (semicolon on Windows, colon on Linux/macOS).",
```
prestomation:
```suggestion
        "description": "A list of paths that should not generate warnings when outside storage profile locations, separated by the OS path list separator (semicolon on Windows, colon on Linux/macOS).",
```

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -31,129 +29,178 @@
...
+
+def _submission_thread_runner(
+    cancelation_flag: CancelationFlag, dialog: SubmitJobProgressDialog, kwargs
+):
+    """Function to run api.create_job_from_job_bundle in a background thread."""
+    try:
+        job_id = api.create_job_from_job_bundle(**kwargs)
+        if not cancelation_flag.canceled:
+            dialog.job_id = job_id
+            dialog.submission_thread_succeeded.emit(job_id or "")
```
prestomation: is it possible for job_id to be None here?

prestomation (dismissed):

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
mwiebe: No, it used to be possible but not anymore.

src/deadline/client/config/config_file.py
mwiebe: Agreed, this would be a great refactor PR to apply.

prestomation (approved):

leongdl (approved):

PR #678, opened 2025-05-14
fix: Guarantee string input to line edit GUI fields

### LINKED ISSUE #663: Bug: Difference in gui-submit handling of parameter types crashes submission
### Describe the bug

When you submit a job template that has a string parameter, but pass in an integer, the job fails to submit with `deadline bundle gui-submit .` The same template and parameters work when submitted through `deadline bundle submit .`

### Expected Behaviour

Both submit commands should have identical behavior. I'd prefer if we kept the `deadline bundle submit .` behavior where it doesn't crash and just automatically converts ints to strings

### Current Behaviour

A popup with the following error appears:
```
Exception caught:
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/__init__.py", line 148, in gui_context_for_cli
    yield app
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/cli/_groups/bundle_group.py", line 317, in bundle_gui_submit
    submitter = show_job_bundle_submitter(
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/job_bundle_submitter.py", line 189, in show_job_bundle_submitter
    submitter_dialog = SubmitJobToDeadlineDialog(
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py", line 107, in __init__
    self._build_ui(
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py", line 161, in _build_ui
    self._build_job_settings_tab(job_setup_widget_type, initial_job_settings)
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py", line 262, in _build_job_settings_tab
    self.job_settings = job_setup_widget_type(
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/widgets/job_bundle_settings_tab.py", line 57, in __init__
    self._build_ui(initial_settings)
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/widgets/job_bundle_settings_tab.py", line 73, in _build_ui
    self.refresh_ui(initial_settings)
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/widgets/job_bundle_settings_tab.py", line 81, in refresh_ui
    self.parameters_widget = OpenJDParametersWidget(
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/widgets/openjd_parameters_widget.py", line 69, in __init__
    self.rebuild_ui(
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/widgets/openjd_parameters_widget.py", line 152, in rebuild_ui
    control = control_widget(self, parameter)
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/widgets/openjd_parameters_widget.py", line 279, in __init__
    self.set_value(value)
  File "/home/alex/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deadline/client/ui/widgets/openjd_parameters_widget.py", line 326, in set_value
    self.edit_control.setText(value)
TypeError: 'PySide2.QtWidgets.QLineEdit.setText' called with wrong argument types:
  PySide2.QtWidgets.QLineEdit.setText(int)
Supported signatures:
  PySide2.QtWidgets.QLineEdit.setText(str)
```

### Reproduction Steps

Try submitting the following job bundle with `gui-submit` and then the regular submit command:

[StringIntValueFail.zip](https://github.com/user-attachments/files/19763656/StringIntValueFail.zip)

### Environment

At minimum:
1. Operating system Happens on both Windows & Linux
2. Output of `python3 --version` Doesn't matter
3. Output of `deadline --version` 0.49.7
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? Happens across both

miabatta:
Fixes: #663 

We don't cast input into strings when loading the submitter GUI, so certain inputs into a LineEdit widget will crash the submitter GUI. 🧵

Cast to string when setting LineEdit and MultiLineEdit values. 🎣

Less crashes! 💥 More robust GUI! 💯

- Unit tests (ran existing, added one) 🧪
- Manual test with the job bundle provided in the issue 💼

nop

nop

nop

----

### DIFF
--- a/src/deadline/client/ui/widgets/openjd_parameters_widget.py
+++ b/src/deadline/client/ui/widgets/openjd_parameters_widget.py
@@ -323,7 +323,7 @@ class _JobTemplateLineEditWidget(_JobTemplateWidget):
         return self.edit_control.text()
 
     def set_value(self, value):
-        self.edit_control.setText(value)
+        self.edit_control.setText(str(value))
 
     def _handle_text_changed(self, text, callback):
         message = deepcopy(self.job_template_parameter)
@@ -377,7 +377,7 @@ class _JobTemplateMultiLineEditWidget(_JobTemplateWidget):
         return self.edit_control.toPlainText()
 
     def set_value(self, value):
-        self.edit_control.setPlainText(value)
+        self.edit_control.setPlainText(str(value))
 
     def _handle_text_changed(self, text, callback):
         message = deepcopy(self.job_template_parameter)
--- a/test/installer/test_installer.py
+++ b/test/installer/test_installer.py
@@ -165,9 +165,9 @@ def test_default_location(installer_path: Path):
         help_result = subprocess.run(
             [installer_path, *text_mode, "--help"], capture_output=True, text=True, timeout=5
         )
-        assert help_result.returncode == 0, (
-            f"Installer exited with non-zero code: {help_result.returncode}"
-        )
+        assert (
+            help_result.returncode == 0
+        ), f"Installer exited with non-zero code: {help_result.returncode}"
         assert help_result.stdout is not None, "No stdout from --help"
         help_output = iter(help_result.stdout.splitlines())
     except subprocess.TimeoutExpired as e:
@@ -182,9 +182,9 @@ def test_default_location(installer_path: Path):
             location = re.match(default_pattern, next(help_output, "").strip(), flags=re.IGNORECASE)
             break
 
-    assert location is not None, (
-        f"Could not find default install location in help output:\n{help_result.stdout}"
-    )
+    assert (
+        location is not None
+    ), f"Could not find default install location in help output:\n{help_result.stdout}"
     if platform.system() != "Windows":
         assert location.group(1) == default_install_location.as_posix()
     else:
@@ -245,9 +245,9 @@ def test_did_not_build_with_evaluation_mode(installer_path: Path, tmp_path: Path
     assert output
 
     for line in output:
-        assert eval_text not in line, (
-            "Installer was detected to have been built with Evaluation mode"
-        )
+        assert (
+            eval_text not in line
+        ), "Installer was detected to have been built with Evaluation mode"
 
 
 @pytest.mark.skipif(platform.system() == "Windows", reason="Only run on Linux and MacOS")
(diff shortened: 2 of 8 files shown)


mwiebe:
Do you have examples of what value types were being passed in? Are you certain those are correct code, or are the exceptions revealing bugs there?

moorec-aws (dismissed):

leongdl (dismissed):
Thanks for adding tests!

epmog (dismissed):

miabatta:
> Do you have examples of what value types were being passed in? Are you certain those are correct code, or are the exceptions revealing bugs there?

@mwiebe We don't have an exact use case, but the main concern is that this behaviour will crash the submitter GUI while working fine with `deadline bundle submit`. I prefer being more permissive than less in this case!

I'll need another round of approvals in a minute as this one is failing on formatting. Thanks all for taking a look!

moorec-aws (approved):

epmog:
> > Do you have examples of what value types were being passed in? Are you certain those are correct code, or are the exceptions revealing bugs there?
> 
> @mwiebe We don't have an exact use case, but the main concern is that this behaviour will crash the submitter GUI while working fine with `deadline bundle submit`. I prefer being more permissive than less in this case!
> 
> I'll need another round of approvals in a minute as this one is failing on formatting. Thanks all for taking a look!

I chatted with @miabatta offline, and what @mwiebe is saying makes sense. I don't think there's anything wrong with us ensuring that we're using strings when setting string values as it will prevent the crash, but it definitely seems like we're addressing the symptom (non-string value making its way to the GUI string field) as opposed to a maybe different root-cause (why do we have a non-string value at all - should we?).

I'm good with getting this change in the short-term (low-risk, quick) to stop the current known crashing, but we're going to want to look a bit further into how those errant values are being passed to through the system.

epmog (approved):

miabatta:
Quick response before I merge! After a bit more testing I found that YAML-formatted parameters will cause this error in a less obvious way since YAML strings don't need enclosing quotations. If this is indicative of a larger error in the submitter logic, it's an easy symptom to deal with immediately and could be a QoL improvement for YAML users. But yes, going to look more into the root cause (especially why it still works in the regular submitter 😄)

Thanks all for the comments and approvals!

PR #682, opened 2025-05-16
feat(experimental): fix atomicity for pid lock & add release lock support for incremental downloads cli

sakshie95:
Fixes: N/A

1. Pid locking mechanism added in PR had an issue when creating a ~tmp pid file & replacing the pid file with ~tmp file. The ~tmp file wasn't unique to each process & could pose problems when 2 concurrent threads try to obtain a lock at the same time. See https://github.com/aws-deadline/deadline-cloud/pull/665#discussion_r2055056092
2. We need a release pid locking mechanism for the feature

1. Fixed the pid locking mechanism by adding a unique attribute (pid) to the ~tmp file name before replacing the global pid file with it.
2. Added releasing pid locking mechanism
3. Also added pid as a parameter to all functions in the pid utility helper to make this utility usable across any class that wants to call it & obtain/release locks on any pid. We shouldn't need to call os.getpid() each time the functions are called.

Pid locking functionality complete for the feature.

Added unit tests for testing the entire functionality including concurrent threads trying to obtain the pid lock.

```
Required test coverage of 80.0% reached. Total coverage: 81.44%
19.04s call     test/unit/deadline_job_attachments/api/test_snapshot.py::TestSnapshotAPI::test_snapshot_includes_excludes[includes3-excludes3-results3]
16.89s call     test/unit/deadline_job_attachments/api/test_snapshot.py::TestSnapshotAPI::test_snapshot_time_diff
16.78s call     test/unit/deadline_job_attachments/api/test_snapshot.py::TestSnapshotAPI::test_snapshot_diff
15.96s call     test/unit/deadline_job_attachments/api/test_snapshot.py::TestSnapshotAPI::test_snapshot_size_diff
15.87s call     test/unit/deadline_job_attachments/api/test_snapshot.py::TestSnapshotAPI::test_diff_with_includes_excludes[exclude_filter-None-initial_exclude1-None-diff_exclude1-modified_files1-new_files1-expected_diff_files1]
============================================================== 1464 passed, 55 skipped, 1 xfailed in 139.25s (0:02:19) ==============================================================
(
```

E2E run of the cli, since no release lock is done currently:
```
deadline queue  incremental-output-download --farm-id farm-dce6609ed9844daca0c0d06cc4446d7a --queue-id queue-857988acee53489fafec961e61831a66 --saved-progress-checkpoint-location /Users/sakshie/automatic-downloadsexport ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD="True"
processing {'farm_id': 'farm-dce6609ed9844daca0c0d06cc4446d7a', 'queue_id': 'queue-857988acee53489fafec961e61831a66', 'conflict_resolution': 'OVERWRITE'}
Started incremental download....
Checking if another download is in progress at /Users/sakshie/automatic-downloads
Process with pid 52715 is not running. Deleting pid file.
Creating new pid file at /Users/sakshie/automatic-downloads/incremental_output_download.pid with pid 84391
```

No

No
----


src/deadline/client/_pid_utils.py
```diff
@@ -8,28 +8,28 @@
 
 
 def check_and_obtain_pid_lock_if_available(
-    saved_progress_checkpoint_location: str, logger: ClickLogger
+    pid_file_location: str, current_process_pid: str, logger: ClickLogger
```
mwiebe: I believe this would be better without any `current_process_id` parameters in these functions. In the PR description you describe the purpose to "make this utility usable across any class that wants to call it & obtain/release locks on any pid," but the whole point is to capture the lock for the current pid, there is no use case for doing that on behalf of a different process.

src/deadline/client/_pid_utils.py
```diff
@@ -8,28 +8,28 @@
...
 ) -> bool:
     """
-    Checks if the download progress file exists and if it does, it checks if the process is still running.
+    Checks if the pid lock file for the download exists and if it does, it checks if the process is still running.
```
mwiebe: Probably say "Checks if the specified pid lock file exists..." to be generic about the operation.

src/deadline/client/_pid_utils.py
```diff
@@ -8,28 +8,28 @@
```
mwiebe, line 9: This constant `PID_FILE_NAME` doesn't belong in `_pid_utils.py`. That's the specific filename for the incremental output download pid lock. Other operations would use different filenames.

src/deadline/client/_pid_utils.py
```diff
@@ -49,27 +49,67 @@ def check_and_obtain_pid_lock_if_available(
...
+    # Check if pid lock file does not exist.
+    if not os.path.exists(pid_file_full_path):
+        logger.echo(f"Pid lock file does not exist at {pid_file_full_path}")
+        return True
+
+    # Try to open pid file at download progress location in read mode
+    with open(pid_file_full_path, "r") as f:
+        # Read pid file and obtain the process id from file contents
+        pid = f.read()
+        # Process pid from file is same as current process pid - release pid lock
```
mwiebe: From here down should be de-dented a level.

test/unit/deadline_client/test_pid_utils.py
```diff
@@ -142,6 +148,103 @@ def test_check_pid_lock_with_errors(self, mock_logger, test_paths, error, expect
...
+        ), patch("os.remove") as mock_remove:
+            mock_exists.return_value = True
+
+            from deadline.client._pid_utils import release_pid_lock
+
+            result = release_pid_lock(test_paths["location"], "1234", mock_logger)
+
+            assert result is True
+            assert mock_logger.echo.called
+            assert "Deleting pid file" in mock_logger.echo.call_args_list[1][0][0]
```
mwiebe: These looked like temporary debug prints to me that we would delete later. The messages also aren't part of an important behavior of the code, so I suggest to remove all assertions throughout your tests about log messages.

test/unit/deadline_client/test_pid_utils.py
```diff
@@ -142,6 +148,103 @@ def test_check_pid_lock_with_errors(self, mock_logger, test_paths, error, expect
...
+
+            result = release_pid_lock(test_paths["location"], "1234", mock_logger)
+
+            assert result is False
+            assert mock_logger.echo.called
+            assert "Skipping pid file deletion" in mock_logger.echo.call_args_list[1][0][0]
+
+    def test_check_pid_lock_concurrent_access(self, mock_logger, test_paths):
+        """
+        Tests the scenario when two threads with different PIDs try to obtain a lock.
```
mwiebe: Small correction for this comment and within the tests - this should say "processes" instead of "threads" throughout. Thread locks are very different from process locks (of which the PID lock is).

src/deadline/client/_pid_utils.py
```diff
@@ -49,27 +49,67 @@ def check_and_obtain_pid_lock_if_available(
...
 
 
-def _obtain_pid_lock_atomically(pid_file_full_path: str, logger: ClickLogger) -> bool:
+def release_pid_lock(pid_file_location: str, current_process_pid: str, logger: ClickLogger) -> bool:
+    """
+    Releases the pid lock by deleting the pid file.
+    :param current_process_pid: current process's id
+    :param pid_file_location: location of the pid lock file
+    :param logger: Click logger component
+    :return:
```
leongdl: Return true if released with the right process ID?

src/deadline/client/_pid_utils.py
```diff
@@ -49,27 +49,67 @@ def check_and_obtain_pid_lock_if_available(
...
+    """
+    logger.echo(f"Releasing pid lock at {pid_file_location}")
+
+    # Get the full path of the pid file at download progress location
+    pid_file_full_path = os.path.join(pid_file_location, PID_FILE_NAME)
+
+    # Check if pid lock file does not exist.
+    if not os.path.exists(pid_file_full_path):
+        logger.echo(f"Pid lock file does not exist at {pid_file_full_path}")
+        return True
```
leongdl, line 125: is this true? Or should we raise? EG: if the file did not exist and we did not even lock it, how can it be true?

src/deadline/client/api/_queue_apis.py
```diff
@@ -41,9 +41,12 @@ def _incremental_output_download(
     """
 
     try:
-        # Check if a download is already ongoing with pid lock checking mechanism
+        # 1. First get the current process's pid
+        current_process_pid: str = str(os.getpid())
```
leongdl: Why not move `os.getpid()` into the function?

test/unit/deadline_client/api/test_queue_apis.py
```diff
@@ -21,7 +22,46 @@ def test_incremental_output_download_success(mock_pid_lock, tmp_path):
     queue_id = "queue-0123456789abcdef"
     boto3_session = MagicMock(spec=boto3.Session)
     saved_progress_checkpoint_location = str(tmp_path / "checkpoint")
-    logger = MagicMock(spec=ClickLogger)
+    logger: ClickLogger = ClickLogger(is_json=False)
+
+    # Test download progress
+    download_progress_json = {
```
leongdl: This blob doesn't seem related?

src/deadline/client/_pid_utils.py
```diff
@@ -8,28 +8,28 @@
...
+    Checks if the pid lock file for the download exists and if it does, it checks if the process is still running.
     If the process is still running, it raises an exception.
-    If the process is not running, it deletes the download progress file.
-    If the download progress file does not exist, it creates a new one.
-    :param saved_progress_checkpoint_location: location of the download progress file
+    If the process is not running, it deletes the pid lock file.
+    If the pid lock file does not exist, it creates a new one and acquires lock for this pid.
+    :param current_process_pid: current process's id
+    :param pid_file_location: location of the pid lock file
     :param logger: Click logger component
```
mwiebe: Because this file is part of the deadline client interface, it shouldn't be using anything from click. Click is specific to CLI functionality, which all lives under the `cli` module.

src/deadline/client/_pid_utils.py
```diff
@@ -49,27 +49,67 @@ def check_and_obtain_pid_lock_if_available(
...
+    :param pid_file_location: location of the pid lock file
+    :param logger: Click logger component
+    :return:
+    """
+    logger.echo(f"Releasing pid lock at {pid_file_location}")
+
+    # Get the full path of the pid file at download progress location
+    pid_file_full_path = os.path.join(pid_file_location, PID_FILE_NAME)
+
+    # Check if pid lock file does not exist.
```
mwiebe: This comment states exactly what the next line of code does. When writing comments like this, it's good to think about what information to communicate to a reader beyond what's immediately visible. In this case, it could be the reason why returning True in this case is the right thing to do.

mwiebe (changes requested):

src/deadline/client/_pid_utils.py
sakshie95: This is a good idea - done. I usually add comments for easy readability of the code irrespective but it's definitely more useful to add beyond what's immediately visible.

src/deadline/client/_pid_utils.py
sakshie95: The ClickLogger seems to be passed around from the ``cli`` to ``api`` modules and even helper classes in the manifest & attachment cli implementations where it was introduced. eg. https://github.com/aws-deadline/deadline-cloud/blob/6cc672d1990376bcbd2c73d72d717546ee58a9e1/src/deadline/job_attachments/_diff.py#L125

Entire code search: https://github.com/search?q=repo%3Aaws-deadline%2Fdeadline-cloud%20ClickLogger&type=code

Do we not intend to pass it around to echo what's needed from the ``api`` and helper classes and instead keep it scoped to just the contents under ``cli``?

test/unit/deadline_client/test_pid_utils.py
sakshie95: Done

test/unit/deadline_client/test_pid_utils.py
sakshie95: Yeah they were temporary debug prints as I wanted to keep the cli very verbose while it's in Experimental. Removed the unnecessary assertions from the tests.

src/deadline/client/_pid_utils.py
sakshie95: Good call, fixed it

src/deadline/client/_pid_utils.py
sakshie95: Done

src/deadline/client/_pid_utils.py
sakshie95: Done

src/deadline/client/_pid_utils.py
sakshie95: You're right there isn't a particular use-case I had in mind where we would want this class to do so on behalf of a different process. My thought was to keep ``_pid_utils`` like a static helper file which locks & releases given all information. I've updated it now

test/unit/deadline_client/api/test_queue_apis.py
sakshie95: Yes it was from next PR - removed it

src/deadline/client/api/_queue_apis.py
sakshie95: Similar as comment below, done.

src/deadline/client/_pid_utils.py
sakshie95: I would keep it as returning True as I want to keep it simple here - release the lock if it exists else True. The caller should handle the rest.

src/deadline/client/_pid_utils.py
sakshie95: Yes, added to docstring

src/deadline/client/_pid_utils.py
leongdl: Perhaps pooly named class, but we want a logger that prevents random prints that are not JSON.

Its more a wrapper Logger handler than specific to click.

src/deadline/client/_pid_utils.py
```diff
@@ -4,32 +4,28 @@
...
+    # Get the current process's id to obtain lock
+    current_process_pid: int = os.getpid()
+
     try:
         # Check if download progress file does not exist.
         if not os.path.exists(pid_file_full_path):
             logger.echo(f"Download progress file does not exist at {pid_file_full_path}")
             # Create a new pid file with the current process id
-            return _obtain_pid_lock_atomically(pid_file_full_path, logger)
+            return _obtain_pid_lock_atomically(pid_file_full_path, logger, current_process_pid)
```
leongdl: super nit on argument ordering, logger can be last :)

src/deadline/client/_pid_utils.py
```diff
@@ -40,7 +36,7 @@ def check_and_obtain_pid_lock_if_available(
                 psutil.Process(int(pid))
                 # Process with the pid exists, so we cannot obtain a lock
                 raise RuntimeError(
-                    f"Another download is in progress at {saved_progress_checkpoint_location}, use --force-bootstrap or wait for previous download to finish"
```
leongdl, line 43: This try - get process branch is re-dundent if the `pid` returned from `f.read() == current_process_pid` ?

src/deadline/client/_pid_utils.py
mwiebe: Got it, I would suggest to make it a print function callback the same as here: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/api/_submit_job_bundle.py#L65

src/deadline/client/_pid_utils.py
sakshie95: done

src/deadline/client/_pid_utils.py
sakshie95: I don't see the possibility of that happening because ``check_and_obtain_pid_lock_if_available`` is only called at the beginning to obtain pid lock for current process which isn't possible if the pid lock file already has the current process's id. If there was an interruption midway, the process id would always change.

src/deadline/client/_pid_utils.py
sakshie95: sure, I invoked ``ClickLogger.echo`` as a print function callback instead of passing it around as it. Hope it looks better now

src/deadline/client/_pid_utils.py
```diff
@@ -40,38 +43,88 @@ def check_and_obtain_pid_lock_if_available(
...
+                    f"Another download is in progress at {pid_file_full_path}, use --force-bootstrap or wait for previous download to finish"
```
mwiebe: This `{pid_file_full_path}` will print a file path, correct? Will users of the tool know what that means? I think this will make more sense as the process id. something like "Another process with ID {pid} is performing a download, ..." because they could look that up in task manager.

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,196 @@
...
+            os.remove(pid_file_full_path)  # Delete the pid file
+
+        # If the pid changed before we locked the file to release the inactive pid lock, we throw a runtime error and exit
+        # This could happen if another concurrent process overrode the inactive pid lock before we released it.
+        else:
+            raise RuntimeError(
+                f"Unable to acquire pid lock as process with pid {pid} exists on {pid_file_full_path}"
+            )
+    except Exception:
+        raise
```
mwiebe: You can delete these two lines. Using `finally` doesn't require there be an `except` clause.

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,196 @@
...
+    :param pid_file_full_path: full path of the pid file
+    :return: returns True if it was able to release the inactive pid dangling lock.
+        Throws a runtime exception if the pid was updated by a different concurrent process and does not try to delete file
+        Throws an unexpected exception if we get into an error while trying to obtain lock on file.
+    """
+    file_size: int = os.path.getsize(os.path.realpath(pid_file_full_path))
+    file_locked_for_delete = open(pid_file_full_path, "r+")
+    try:
+        # Acquire an exclusive lock
+        if sys.platform.startswith("linux") or sys.platform.startswith("darwin"):
```
mwiebe: It's better to test for Windows, and default to POSIX if it's not Windows, since the other OSs will generally be more like POSIX than Windows. Looking at https://docs.python.org/3/library/sys.html#sys.platform, there are two Windows values, so the check can be like `if sys.platform in ("win32", "cygwin"):`

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,196 @@
...
+    The primitive lock is obtained on the pid file so concurrent processes cant delete it
+    This would be caught in race conditions if primitive file locking is disabled on a customer's machine.
+    :param pid: inactive pid in the file
+    :param pid_file_full_path: full path of the pid file
+    :return: returns True if it was able to release the inactive pid dangling lock.
+        Throws a runtime exception if the pid was updated by a different concurrent process and does not try to delete file
+        Throws an unexpected exception if we get into an error while trying to obtain lock on file.
+    """
+    file_size: int = os.path.getsize(os.path.realpath(pid_file_full_path))
+    file_locked_for_delete = open(pid_file_full_path, "r+")
```
mwiebe: This should be `with open(pid_file_full_path, "r+") as file_locked_for_delete:`. Did you try this and run into any difficulties with the locking?

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,196 @@
...
+        return True
+    # Process pid from file is different from current process pid - do not release lock
+    else:
+        print_function_callback(
+            f"Process with pid {pid} is not the current process. Skipping pid file deletion."
+        )
+        return False
+
+
+def _lock_pid_file_and_release_dangling_pid_lock(pid: str, pid_file_full_path: str) -> bool:
```
sakshie95: Combined everything together into one function above for the locking for easy reasoning of the logic & concurrency - except extracting this method out for dealing with dangling pids from a different process that died midway during this cli operation.

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,196 @@
...
 import os
+import sys
 import psutil
-from deadline.client.cli._groups.click_logger import ClickLogger
-
-PID_FILE_NAME = "incremental_output_download.pid"
+from typing import Callable
 
 
 def check_and_obtain_pid_lock_if_available(
```
mwiebe: In locking APIs I've worked with, I've seen APIs that do this named like "tryLock" or "tryAcquire". See https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html and https://learn.microsoft.com/en-us/dotnet/api/java.util.concurrent.semaphore.tryacquire?view=net-android-35.0

I think it would be good to name this function similarly to these APIs so developers reading this will recognize or find related docs if they search about it. Maybe `def try_acquire_pid_lock`?

mwiebe (changes requested):

src/deadline/client/_pid_utils.py
leongdl: Just a thought for code organization, can the code just have 1 branch. Multiple OS branches in logic is hard to maintain overtime.

```
if sys.platform.startswith("linux") or sys.platform.startswith("darwin")
   # Posix logic, maybe a function?
else:
  # Windows logic, maybe a function?
```

src/deadline/client/_pid_utils.py
sakshie95: yeah makes sense, done. 
We cannot make the ``sys.platform in`` as ``hatch lint`` fails cross-platform with that. I followed what's been done in the repo in other places currently - https://github.com/search?q=repo%3Aaws-deadline%2Fdeadline-cloud+sys.platform+%3D%3D+%22win32%22&type=code but with checking both values for windows

src/deadline/client/_pid_utils.py
sakshie95: yes I had run into some issues with releasing the lock in this approach initially but ended up resolving them so I can move back to the ``with open`` implementation now.

src/deadline/client/_pid_utils.py
sakshie95: Yeah that definitely looks better, done

src/deadline/client/_pid_utils.py
sakshie95: good catch, done

src/deadline/client/_pid_utils.py
sakshie95: I added a print when we get the exception trying to acquire pid lock because a ``pid`` has it already for the customer: ``print_function_callback(f"Unable to acquire pid lock as process with pid {pid} exists on {pid_file_full_path}")
``

src/deadline/client/_pid_utils.py
sakshie95: This part has been removed now due to issues with windows msvcrt library. We will revisit it later since it's a concurrency prob in an edge case

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,148 @@
...
+            )
+
+            can_obtain_pid_lock = True
+
+        # 3. Try to read existing pid file to verify process with the pid lock is active & hasn't ended pre-maturely
+        else:
+            with open(pid_file_full_path, "r") as f:
+                pid = f.read()
+                try:
+                    psutil.Process(int(pid))
```
mwiebe: Instead of constructing a Process object and handling an exception, I see that there's a function in `psutil` specifically to check this: https://psutil.readthedocs.io/en/latest/#psutil.pid_exists

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,148 @@
...
+                    # TODO Try to minimize potential race conditions from concurrent processes trying to remove pid file
+                    os.remove(pid_file_full_path)
+
+                    can_obtain_pid_lock = True
 
     except Exception:
-        # We already checked the file exists before reading it.
-        # For any other unexpected exceptions, we should not delete the pid file and raise an error.
+        # For any other unexpected exceptions, we should raise an error.
         raise
```
mwiebe, line 71: What is the purpose of this `except Exception` here?

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,148 @@
...
+                    print_function_callback(
+                        f"Process with pid {pid} is not running. Deleting pid file."
+                    )
+
+                    f.close()  # Required for windows
+
+                    # Delete existing pid file as pid is not active
+                    # Once we've determined the process in pid file isn't running we do nothing except:
+                    # Closing file (required for windows) and deleting it
+                    # This will keep race conditions across concurrent processes trying this to be least possible without locking
```
mwiebe: The race starts at the `f.read()` call on line 46 and ends at the `os.remove` on line 69. I see that there's a call to `print_function_callback` in the line of execution, which makes it longer. It's a bit hard to see the execution paths because of the try/except pattern here, so I think switching it to use `psutil.pid_exists` will make it easier to read and see the race. Something like this:

```python
with open(pid_file_full_path, "r") as f:
    pid = f.read()
if not psutil.pid_exists(int(pid)):
    os.remove(pid_file_full_path)
    print_function_callback(...)
else:
    print_function_callback(...)
    raise ...
```

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,148 @@
...
-        # We already checked the file exists before reading it.
-        # For any other unexpected exceptions, we should not delete the pid file and raise an error.
+        # For any other unexpected exceptions, we should raise an error.
         raise
 
+    # 4. If we've determined we can obtain lock for this process, we do so now.
+    # We handle concurrency for processes racing for this pid lock by using atomic primitives
+    # The pid lock is obtained successfully for one process and fails for the other process
+    # We use os.rename() for windows and os.link() for linux/macOS as atomic primitives
+    if can_obtain_pid_lock:
```
mwiebe, line 77: Under what conditions is this value False? I think the function will have raised an exception in that case if I'm reading it correctly.

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,148 @@
...
+
+        print_function_callback(
+            f"Creating new pid file at {pid_file_full_path} with pid {current_process_pid}"
+        )
+
+        # Open tmp file in write mode and write the current process id to it
+        with open(tmp_file_name, "w+") as f:
+            f.write(str(current_process_pid))
+            f.flush()
+            os.fsync(f.fileno())
```
mwiebe, line 89: According to https://docs.python.org/3/library/io.html#io.IOBase.close, closing the file, which happens when the context handler exits, does a flush so this shouldn't be necessary.

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,148 @@
...
+
+        # Atomic write to pid file while handling concurrency for parallel processes trying to race for the pid lock
+        try:
+            if sys.platform.startswith("linux") or sys.platform.startswith("darwin"):
```
mwiebe: It would be better to reverse this logic - test for windows as the special case and use posix behavior for everything else. Per the documentation https://docs.python.org/3/library/sys.html#sys.platform, there are two values for Windows so it can look like:

```python
if sys.platform in ("win32", "cygwin"):
    ...
else:
    ...
```

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,148 @@
...
+    with open(pid_file_full_path, "r") as f:
+        # Read pid file and obtain the process id from file contents
+        pid = f.read()
+
+    # Process pid from file is same as current process pid - release pid lock
+    if int(pid) == current_process_pid:
+        print_function_callback(
+            f"Process with pid {pid} is the current process. Deleting pid file."
+        )
+        os.remove(pid_file_full_path)
```
mwiebe, line 137: I think it's better to remove the file first and then print that it was removed, to reduce the potential delay between the check and the `os.remove` call.

src/deadline/client/api/_queue_apis.py
```diff
@@ -36,20 +37,26 @@ def _incremental_output_download(
...
+        # 2. Check if a download is already ongoing with pid lock checking mechanism
+        _pid_utils.try_acquire_pid_lock(pid_file_full_path, print_function_callback)
+    except RuntimeError:
+        print_function_callback(
+            f"Another download is in progress at {saved_progress_checkpoint_location}, wait for previous download to finish"
         )
-    except RuntimeError as e:
-        logger.echo(f"Download failed because of error : {e}")
         return
     except Exception as e:
```
mwiebe, line 59: I see you're using `RuntimeError` to mean that the lock is already being held, and any other `Exception` to mean some other error occurred. `RuntimeError` is a built-in base class of many exceptions, so this feels like it would be hard to maintain as an invariant of the implementation. Creating a specific exception like `PidLockAlreadyHeld` would make this more clear to read the code and less error-prone.

mwiebe (changes requested):

src/deadline/client/api/_queue_apis.py
sakshie95: done

src/deadline/client/_pid_utils.py
sakshie95: done

src/deadline/client/_pid_utils.py
sakshie95: ah missed this, done

src/deadline/client/_pid_utils.py
sakshie95: I read It's a good practice to flush and fsync to make sure the files are synced from python's buffer to the OS buffer and to the disk so I followed that to make sure it's persisted. Even though it may seem redundant I added it as a good measure. Do you see concerns I may be missing?

src/deadline/client/_pid_utils.py
sakshie95: No you're reading it 100% right.. an exception is thrown from all cases this can be false but since the default value is false - I added this check so we only execute when explicitly set to ``True``

src/deadline/client/_pid_utils.py
sakshie95: Catching any unexpected exceptions from the try block above it and re-raising them. One example could be error in os.remove() because of concurrent processes trying to delete the file (the rare edge case)

src/deadline/client/_pid_utils.py
sakshie95: I had used ``psutil.Process(pid)`` with the intention that it can also provide metadata about the process if we need to supply that info to the customer to debug. But I think the pid should be sufficient for now so I can use this simpler wrapper instead

src/deadline/client/_pid_utils.py
```diff
@@ -1,85 +1,144 @@
...
+            if sys.platform.startswith("win32") or sys.platform.startswith("cygwin"):
+                os.rename(tmp_file_name, pid_file_full_path)
+            else:
+                os.link(tmp_file_name, pid_file_full_path)
+                os.remove(tmp_file_name)
+        except FileExistsError:
+            print_function_callback(
+                f"Concurrency issue when trying to obtain pid lock at {pid_file_full_path} for process id {current_process_pid}"
+            )
+            raise
```
leongdl, line 102: nit, does it make sense for this to raise a PidLock error like above? for error consistency sake. A File Exist exception here will not be descriptive.

Note, this can be done as a follow up PR, the logic is fine and a user experience edge case.

test/unit/deadline_client/test_pid_utils.py
```diff
@@ -142,6 +194,205 @@ def test_check_pid_lock_with_errors(self, mock_logger, test_paths, error, expect
...
+            test_paths["location"], "queue-12345_incremental_output_download.pid"
+        )
+
+        with patch("os.path.exists") as mock_exists, patch(
+            "builtins.open", mock_open(read_data="1234")
+        ), patch("os.remove") as mock_remove, patch("os.getpid") as mock_get_pid:
+            mock_exists.return_value = True
+            mock_get_pid.return_value = 1234
+
+            from deadline.client._pid_utils import release_pid_lock
```
leongdl, line 228: We are importing here so patch works?

leongdl: Super nit and not for revisions, I personally think we don't even need to patch path if we setup a folder as part of the setup. Mocking / patching without a debugger and lots of asserts introspecting internal states is hard to say the test is good :)

(Just my opinion)

leongdl (commented):
I think we should move forward - save for windows case which we can double click on with pywin32.

The real meat of the project we haven't even gotten to a PR yet.

I have some nits, but those are more personal opinions rather than functional ones.

leongdl (approved):

src/deadline/client/_pid_utils.py
mwiebe: Because the Python documentation says it flushes on close, I would trust it and not flush a second time. Do you have a reference to where you read about the fsync?

src/deadline/client/_pid_utils.py
mwiebe: In that case this check should not be here, and the `can_obtain_pid_lock` variable should be deleted. The code should follow either the pattern where it raises exceptions on errors and returns on success, or it should never raise exceptions and return the boolean. By mixing the patterns, it takes more effort for someone to read and understand the code.

test/unit/deadline_client/test_pid_utils.py
```diff
@@ -29,26 +27,78 @@ def test_paths(self):
...
+        self, mock_open_file, mock_file, pid_file, mock_getsize, mock_realpath
+    ):
+        """
+        Helper method to verify common assertions across tests.
+        """
+        mock_open_file.assert_called_once_with(pid_file, "a+")
+        mock_file.seek.assert_called_once_with(0)
+        mock_file.read.assert_called_once()
+        mock_getsize.assert_called_once_with(pid_file)
+        mock_realpath.assert_called_once_with(pid_file)
```
mwiebe, line 68: I haven't dug very much into these tests, but I see they're using a lot of mocks. Generally, using mocks this much means that the tests are testing "is the implementation doing what I wrote it to do" which is often not useful and slows down refactoring later. Tests that test "does the implementation satisfy the properties it should" with as few mocks as possible are much better, I recommend you think about how to reduce the amount of mocks your tests have to as close to zero as possible.

mwiebe (approved):
I agree with @leongdl . Let's merge this, and please make a follow-up PR for the comments that aren't yet addressed.

PR #685, opened 2025-05-20
feat!: Add root information manifest snapshot response

godobyte:
WA client need the root information from manifest snapshot response.

Added root information.

1. Always require both root and manifest path for `manifest snapshot`
2. Added root information to snapshot response

Yes, `manifest` and `root` are required for constructing `ManifestSnapshot`.

No

----

### DIFF
--- a/src/deadline/job_attachments/models.py
+++ b/src/deadline/job_attachments/models.py
@@ -390,7 +390,8 @@ class GlobConfig:
 class ManifestSnapshot:
     """Data structure to store the results of a manifest snapshot"""
 
-    manifest: str = field(default_factory=str)
+    root: str
+    manifest: str
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


jairaws (dismissed):

src/deadline/job_attachments/models.py
```diff
@@ -390,8 +391,65 @@ class GlobConfig:
...
+            data: Dictionary containing root and manifest fields
+
+        Returns:
+            ManifestSnapshot object
+
+        Raises:
+            ValueError: If the dictionary is missing required fields
+        """
+        if "root" not in data or "manifest" not in data:
+            raise ValueError("Invalid dictionary: missing required fields 'root' or 'manifest'")
```
jericht: Instead of having dedicated serialization functionality here (since we're really just wrapping built-in dataclasses/json functions), how about we use dataclasses `__post_init__` instead? It looks like we are mainly wanting to add this validation to the class.

https://docs.python.org/3/library/dataclasses.html#post-init-processing

This way the interface of the class is kept just to its fields defined at the class level, rather than having a separate interface for serialization/deserialization that we need to keep consistent

leongdl (approved):

jericht (approved):

PR #687, opened 2025-05-22
feat: Add function to construct partial s3 prefix for job attachments output manifests

godobyte:
We'd like to define where the output manifest should be uploaded too.

Create a function in JobAttachmentS3Settings.

WA caller can simply call this to construct the path to put output manifest, and pass that to attachment upload.

- [ ] Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended that you ensure that the docker-based unit tests pass.

Run WA E2E tests `pytest -- test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux` to make sure
1. job succeeded, which accounted for step-step dependencies
2. downloaded job output, and verified they're as expected
 
No

No

----

### DIFF
--- a/src/deadline/job_attachments/models.py
+++ b/src/deadline/job_attachments/models.py
@@ -27,6 +27,7 @@ from deadline.job_attachments.exceptions import (
 from ._utils import (
     _generate_random_guid,
     _join_s3_paths,
+    _float_to_iso_datetime_string,
 )
 
 S3_DATA_FOLDER_NAME = "Data"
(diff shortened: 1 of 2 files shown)


src/deadline/job_attachments/models.py
```diff
@@ -290,6 +291,32 @@ def full_output_prefix(
             session_action_id,
         )
 
+    @staticmethod
+    def partial_output_manifest_prefix(
```
leongdl: Naming is hard :)

leongdl: I don't have a better name right now, but how can we name this?

godobyte: Yea open to any name recommendation. I picked partial_output_manifest_prefix simply because there is a similar partial_manifest_prefix for input manifest below.

leongdl (approved):

jericht (approved):

PR #691, opened 2025-05-28
feat(experimental): command bootstrap logic and reading from, saving to state file for incremental downloads cli

sakshie95:
Fixes: N/A

We need the bootstrapping logic for the first run of the incremental download cli command or if the customer chooses to force bootstrap in a later run and continue from current progress if none of the above are true. We also need to save the progress at the end of the command.

1. Functionality to bootstrap command if it is being run for the first time (i.e the download progress a.k.a the state file is missing) or if the option ``--force-bootstrap`` is provided.
2. Bootstrap the command with the ``--bootstrap-lookback-in-minutes`` if provided or ``0`` default
3. State file needs to be read from if the command is already bootstrapped
4. State file needs to be written to at the end of the orchestration once all downloads are done.
5. Downloads are not implemented in this PR, so we simply set the updated download progress to current download progress.
6. Releasing pid lock at the end of the command

Bootstrapping and reading from/writing to state file complete.

Ran the tests on local with all changes from this PR and dependent PR https://github.com/aws-deadline/deadline-cloud/pull/682

```
Required test coverage of 80.0% reached. Total coverage: 81.93%
20.07s call     test/unit/deadline_job_attachments/api/test_manifest_diff.py::TestDiffAPI::test_diff_no_change
12.07s call     test/unit/deadline_client/api/test_api_telemetry.py::test_record_decorator_success
11.28s call     test/unit/deadline_job_attachments/api/test_manifest_upload.py::TestManifestUpload::test_upload
8.93s call     test/unit/deadline_job_attachments/api/test_manifest_diff.py::TestDiffAPI::test_diff_modified_file_size
8.20s call     test/unit/deadline_job_attachments/api/test_manifest_diff.py::TestDiffAPI::test_diff_deleted_file
======================================================== 1525 passed, 55 skipped, 1 xfailed, 2 warnings in 62.06s (0:01:02) =========================================================
```

No

No

No

----


src/deadline/client/api/_queue_apis.py
```diff
@@ -41,10 +46,11 @@ def _incremental_output_download(
     """
 
     try:
-        # Check if a download is already ongoing with pid lock checking mechanism
-        _pid_utils.check_and_obtain_pid_lock_if_available(
-            saved_progress_checkpoint_location, logger
-        )
+        # 1. Construct pid file full path
+        pid_file_full_path = os.path.join(saved_progress_checkpoint_location, PID_FILE_NAME)
```
leongdl: Isn't this change part of https://github.com/aws-deadline/deadline-cloud/pull/682. It is appearing here too?

leongdl: Or you need to rebase after merging 682? Just seemed odd the first line of change gave me de-ja-vu

src/deadline/job_attachments/incremental_downloads/models/state_file.py
```diff
@@ -0,0 +1,72 @@
...
+class StateFileModel:
+    """
+    Model representing the download progress state file structure.
+    State file structure:
+    {
+        "lastLookbackTime": "2025-04-04T05:30:00",
+        "jobs":
+        [
+            {
+                "jobId": "Job-1234353453443",
```
leongdl: nit, all of our IDs are lower case. `job-00000000000000000000000000000000`

src/deadline/job_attachments/incremental_downloads/models/state_file.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __init__(self, last_lookback_time=None, jobs=None):
+        """
+        Initialize a StateFileModel instance.
+        Args:
+            last_lookback_time (str): ISO format timestamp of the last lookback time
+            jobs (list): List of job dictionaries containing job_id and sessions information
+        """
+        self.last_lookback_time = last_lookback_time
+        self.jobs = jobs or []
```
leongdl: Typing please, python without types is very scary. Especially since this is a data structure.

src/deadline/job_attachments/incremental_downloads/orchestrator.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+            current_download_progress.last_lookback_time = (
+                datetime.datetime.utcnow()
+                - datetime.timedelta(minutes=float(bootstrap_lookback_in_minutes or 0))
+            )
+
+        # 2. Load progress from current download progress state file
+        else:
+            try:
+                with open(saved_progress_checkpoint_full_path, "r") as file:
```
leongdl: More of a discussion question - since StateFileModel already uses a static method to load, can we just make lines 64 to 69 the static method that does the loading and return a true false.

Reduces the complexity of this function since it is already a multi-step function

leongdl: I like linear code with as little branches as possible :)

EG:

``` 
if not (state_data := StateFileModel.load(file): 
   raise .....
```

src/deadline/job_attachments/incremental_downloads/orchestrator.py
```diff
@@ -0,0 +1,136 @@
...
+
+            except Exception as e:
+                logger.echo(
+                    f"Failed to load download progress from {saved_progress_checkpoint_full_path}: {str(e)}"
+                )
+                return False
+
+        # 3. Download and update download progress.
+        # Right now it is set to no change in progress except setting the last lookback time to now
+        updated_download_progress: StateFileModel = current_download_progress
```
leongdl: This line means it is a todo right? 

I read the comment above and was like, this isn't finished?

src/deadline/job_attachments/incremental_downloads/orchestrator.py
```diff
@@ -0,0 +1,136 @@
...
+                os.fsync(file.fileno())
+
+            # 5. Atomically replace the target file with the temporary file
+            os.replace(temp_file_path, current_saved_progress_checkpoint_full_path)
+
+            logger.echo(
+                f"Successfully saved download progress to {current_saved_progress_checkpoint_full_path}"
+            )
+        except Exception as e:
+            logger.echo(
```
leongdl: We don't re-raise here? At least make the app exit with error!

src/deadline/job_attachments/incremental_downloads/models/state_file.py
leongdl: Can we fully fledge out this data structure as this PR? 

What i see now is a class with a time and a wild card array. The PR calls it to load / save the file, but nothing really uses it. its a shell that I will not know if it will serve the download purpose.

src/deadline/client/api/_queue_apis.py
sakshie95: yes it is.. it will need to be rebased once the other PR is merged. This had to be done for 2 common classes - you can find details in the description

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: My thought was to not make the app exit on this error to prevent manual action from customers at the cost of downloading more than EXACTLY once for one-off save errors. I've updated this to hard fail for now & added a comment - we can revisit this behaviour later if we need to

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: yes but i didn't add it as a todo cz we discussed it looked off in the previous PR given these PRs are incremental changes for the feature. This PR is scoped to just the info in the description.

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: makes sense, I extracted this bit as a static function. Had to modify what it returns to be the current progress though.

src/deadline/job_attachments/incremental_downloads/models/state_file.py
sakshie95: Discussed this offline. This PR is scoped just to info in the description - loads progress from an empty state file with just lookback time & saves it back with updated lookback time.

src/deadline/job_attachments/incremental_downloads/models/state_file.py
sakshie95: done

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -311,5 +311,5 @@ def incremental_output_download(
             bootstrap_lookback_in_minutes=bootstrap_lookback_in_minutes,
             force_bootstrap=force_bootstrap,
             path_mapping_rules=path_mapping_rules,
-            logger=logger,
+            print_function_callback=logger.echo,
```
leongdl: :) nice, I like this usage.

src/deadline/job_attachments/incremental_downloads/models/state_file.py
leongdl: OK - but I will still look for proper typing in the future.

src/deadline/job_attachments/incremental_downloads/orchestrator.py
```diff
@@ -0,0 +1,162 @@
...
+
+            # 4. Write the state data to the temporary file
+            with open(temp_file_path, "w") as file:
+                json.dump(state_data, file, indent=2)
+                # Ensure data is written to disk
+                file.flush()
+                os.fsync(file.fileno())
+
+            # 5. Atomically replace the target file with the temporary file
+            os.replace(temp_file_path, current_saved_progress_checkpoint_full_path)
```
leongdl: From offline conversation, is this line changing? If yes, another revision or?

src/deadline/job_attachments/incremental_downloads/orchestrator.py
```diff
@@ -0,0 +1,162 @@
...
+
+            # 3. Get a unique id for atomic file update and create a temporary file in the given directory
+            unique_process_identifier = str(os.getpid())
+            temp_file_path = f"{saved_progress_checkpoint_location}/{DOWNLOAD_PROGRESS_FILE_NAME}_{unique_process_identifier}.tmp"
+
+            # 4. Write the state data to the temporary file
+            with open(temp_file_path, "w") as file:
+                json.dump(state_data, file, indent=2)
+                # Ensure data is written to disk
+                file.flush()
```
leongdl: Since this is already in a python context, why is flush and fsync needed?

src/deadline/job_attachments/incremental_downloads/orchestrator.py
```diff
@@ -0,0 +1,162 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mwiebe: I expect the overall structure of this command to be better if there is no concept like "orchestrator" in the design, so I recommend you remove this file.

As a better alternative, think about the lower level operations the command needs to run, and structure them into a callable library form. Those operations should be self-contained and isolated from the overall command, they should look like useful library calls that other commands could use as well.

src/deadline/job_attachments/incremental_downloads/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+from __future__ import annotations
+
+from .state_file import StateFileModel
+
+__all__ = ["StateFileModel"]
```
mwiebe: The number of models probably won't grow for this particular command, right? Because of that, it will be nicer to not have this `models` directory and instead move the object that holds state one level up. That will reduce the number of indirections/boilerplate necessary to interact with this code.

src/deadline/job_attachments/incremental_downloads/models/state_file.py
```diff
@@ -0,0 +1,72 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mwiebe: The essential thing about this object is that it holds download state, not that it's for a file. Because of that, something like "incremental_download_state.py" with a class `IncrementalDownloadState` is a name that will be more directly understandable.

src/deadline/job_attachments/incremental_downloads/orchestrator.py
```diff
@@ -0,0 +1,162 @@
...
+        saved_progress_checkpoint_full_path: str = (
+            f"{saved_progress_checkpoint_location}/{DOWNLOAD_PROGRESS_FILE_NAME}"
+        )
+        current_download_progress: StateFileModel = StateFileModel()
+
+        # 1. Bootstrap to ignore current download progress & set lookback based on bootstrap_lookback_in_minutes if:
+        # this is the first run of the command & save progress checkpoint file does not exist
+        # OR
+        # if force bootstrap option is provided by customer
+        if force_bootstrap or not os.path.exists(saved_progress_checkpoint_full_path):
```
mwiebe: Since the bootstrap logic is related to handling the CLI options, it's better to put that code inside the CLI command handler. That will reduce coupling between the parts of the code.

In the library code as part of incremental_download_state.py, you could have two functions, `load_from_file` and `bootstrap_from_queue`.

mwiebe (changes requested):

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: makes sense.. will do

src/deadline/job_attachments/incremental_downloads/models/__init__.py
sakshie95: The number of models is going to grow with the incremental changes. Right now I added a StateFileModel, I need enums for Statuses of Sessions/SessionActions for eg. and the expectation is to have a model class to hold the data for the job list as well.

src/deadline/job_attachments/incremental_downloads/models/state_file.py
sakshie95: Got it, will make the update

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: Makes sense. I wanted to keep the cli very thin but this would make more sense there you're right.

src/deadline/job_attachments/incremental_downloads/models/state_file.py
```diff
@@ -0,0 +1,72 @@
...
+        Create a StateFileModel instance from a dictionary.
+        Args:
+            data (dict): Dictionary containing state file data
+        Returns:
+            StateFileModel: A new instance populated with the data
+        """
+        if not data:
+            return cls()
+
+        return cls(last_lookback_time=data.get("lastLookbackTime"), jobs=data.get("jobs", []))
```
leongdl: By the way, .get here will pass and return an empty object even with `{}` as input. 

Do we care about file format checking?

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: No the offline conversation was regarding : https://github.com/aws-deadline/deadline-cloud/pull/682. Concurrency handling to be specific

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: I read it's needed to properly persist the data on disk in python as the data passes through various buffers: Without flush(), data may remain in Python's buffer and not be written to the OS. Without fsync(), data might remain in the OS buffer and be lost in a system crash or power failure before reaching the disk. So I've done this as a good measure

src/deadline/job_attachments/incremental_downloads/orchestrator.py
sakshie95: I've implemented ``incremental_download_state.py`` with a Class ``IncrementalDownloadState`` to save the state and kept the library functions as methods on the file instead of class functions to keep them re-usable and isolated from the state itself. Let me know if you had a different idea in mind, happy to incorporate.

src/deadline/job_attachments/incremental_downloads/models/state_file.py
sakshie95: No I would think that's expected behaviour for this class. If the input is {} we return an empty object with the keys

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -297,19 +298,29 @@ def incremental_output_download(
...
         queue_id = config_file.get_setting("defaults.queue_id", config=config)
-
         boto3_session: boto3.Session = api.get_boto3_session(config=config)
 
-        # Call the incremental output download api
+        # Get download progress file name appended by the queue id - a unique progress file exists per queue
+        download_progress_file_name: str = f"{queue_id}_{DOWNLOAD_PROGRESS_FILE_NAME}"
+
+        # Get saved progress file full path now that we've validated all file inputs are valid
+        saved_progress_checkpoint_full_path: str = os.path.join(
```
leongdl, line 317: This logic is now duplicated in _queue_apis.py, either make a utility, or find a way for it to be passed. Otherwise one code path may change without the other!

src/deadline/client/api/_queue_apis.py
```diff
@@ -7,11 +7,19 @@
 from typing import Optional, Callable
 import boto3
 from deadline.client import _pid_utils
+from deadline.job_attachments.incremental_downloads.incremental_download_state import (
```
mwiebe: Imports within the same library should be relative, not absolute. Would be good to fix the line just above this too.

src/deadline/client/api/_queue_apis.py
```diff
@@ -42,15 +50,47 @@ def _incremental_output_download(
...
         _pid_utils.try_acquire_pid_lock(pid_file_full_path, print_function_callback)
+
+        # Construct the saved progress checkpoint full path
+        saved_progress_checkpoint_full_path: str = os.path.join(
+            saved_progress_checkpoint_location, f"{queue_id}_{DOWNLOAD_PROGRESS_FILE_NAME}"
+        )
+        current_download_progress: IncrementalDownloadState = IncrementalDownloadState()
+
+        # 3. If bootstrap is required, then bootstrap using helper function
+        if force_bootstrap:
```
mwiebe: Should this include `or not os.path.exists(saved_progress_checkpoint_full_path)`?

src/deadline/client/cli/_groups/queue_group.py
mwiebe: Rather than a utility function, better to pass this as a parameter to the function so it doesn't have to re-calculate it.

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -297,19 +298,29 @@ def incremental_output_download(
...
         _queue_apis._incremental_output_download(
             boto3_session=boto3_session,
             farm_id=farm_id,
             queue_id=queue_id,
             saved_progress_checkpoint_location=saved_progress_checkpoint_location,
             bootstrap_lookback_in_minutes=bootstrap_lookback_in_minutes,
-            force_bootstrap=force_bootstrap,
+            # True if either the option is provided by customer or saved progress checkpoint file does not exist
+            force_bootstrap=force_bootstrap
+            or not os.path.exists(saved_progress_checkpoint_full_path),
```
mwiebe: Seems better to include this test inside the function logic.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,180 @@
...
+    def to_dict(self):
+        """
+        Convert the IncrementalDownloadState to a dictionary.
+        Returns:
+            dict: Dictionary representation of the state file model
+        """
+        return {"lastLookbackTime": self.last_lookback_time, "jobs": self.jobs}
+
+
+def bootstrap_fresh_state(
```
mwiebe: A nice name for this might be `IncrementalDownloadState.from_bootstrap`. That would match the existing `from_dict` constructor it has.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,180 @@
...
+    print_function_callback(
+        "Bootstrapping command. Ignoring download progress location and creating new"
+    )
+    current_download_progress.last_lookback_time = datetime.datetime.utcnow() - datetime.timedelta(
+        minutes=float(bootstrap_lookback_in_minutes or 0)
+    )
+    return current_download_progress
+
+
+def load_progress_from_state_file(
```
mwiebe: A nice name for this might be `IncrementalDownloadState.from_file` or `IncrementalDownloadState.load_file`.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,180 @@
...
+
+    except Exception as e:
+        print_function_callback(
+            f"Failed to load existing state file from download progress checkpoint location {saved_progress_checkpoint_full_path}: {str(e)}"
+        )
+        # Raise as this is an unexpected exception in reading state file, we already checked its existence earlier
+        raise
+
+
+def save_progress_to_state_file(
```
mwiebe: A nice name for this might be `IncrementalDownloadState.to_file` or `IncrementalDownloadState.save_file`.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,180 @@
...
+    saved_progress_checkpoint_location: str,
+    current_saved_progress_checkpoint_full_path: str,
+    current_download_progress: IncrementalDownloadState,
```
mwiebe: Because of the context this is in (saving a state file), shorter names for all these variables will be more accurate and readable. E.g. `current` is incorrect, this function is independent of whether what it is saving is current or not.

test/unit/deadline_client/api/test_queue_apis.py
```diff
@@ -11,234 +11,327 @@
...
+    """Create a checkpoint directory for all tests to use."""
+    checkpoint_dir = tmp_path_factory.mktemp("checkpoint")
+    yield str(checkpoint_dir)
+    # No cleanup needed here as tmp_path_factory handles it automatically
+
+
+@pytest.fixture(scope="module")
+def queue_id():
+    """Return a consistent queue ID for all tests."""
+    return "queue-0123456789abcdef"
```
mwiebe: Using a fixture for a constant looks like overkill to me. Other tests are already using a constant for this, can you match what they do?

test/unit/deadline_client/api/test_queue_apis.py
```diff
@@ -11,234 +11,327 @@
...
+    progress_file_path = os.path.join(checkpoint_dir, f"{queue_id}_download_progress.json")
+    # File will be created by the test that needs it
+    yield progress_file_path
+    # Clean up after each test
+    if os.path.exists(progress_file_path):
+        os.remove(progress_file_path)
+
+
+@pytest.fixture
+def pid_file(checkpoint_dir, queue_id):
```
mwiebe: nit: Renaming to `pid_lock_file` will make reading the tests more clear.

test/unit/deadline_client/api/test_queue_apis.py
```diff
@@ -11,234 +11,327 @@
...
+    """Create a PID file path for tests that need it."""
+    pid_file_path = os.path.join(checkpoint_dir, f"{queue_id}_incremental_output_download.pid")
+    yield pid_file_path
+    # Clean up
+    if os.path.exists(pid_file_path):
+        os.remove(pid_file_path)
+
+
+@pytest.fixture
+def rules_file(tmp_path_factory):
```
mwiebe: What is this? The name and comments are just repeating "rules file" that doesn't explain what these rules are.

mwiebe (changes requested):

src/deadline/client/cli/_groups/queue_group.py
sakshie95: I thought about this - for the api available under ``_queue_apis.py`` as ``_incremental_output_download`` it would probably not make sense for the caller to construct the full file path for the saved progress file? Also, had to move the bootstrap logic to the cli as per previous comment - https://github.com/aws-deadline/deadline-cloud/pull/691#discussion_r2127415284 which is why the cli needs the full path too. Thoughts?

src/deadline/client/api/_queue_apis.py
sakshie95: No I moved the entire bootstrap logic to the cli as per previous comment: https://github.com/aws-deadline/deadline-cloud/pull/691#discussion_r2127415284. Maybe I should call this parameter ``bootstrap`` (being passed to the api from the cli) to make it clear?

test/unit/deadline_client/api/test_queue_apis.py
sakshie95: This is the path mapping rules.. will make it explicit

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: ah I took inspiration from previous comment for these names https://github.com/aws-deadline/deadline-cloud/pull/691#discussion_r2127415284 :))

mwiebe: Yep! Honing in on a great design is an iterative process, where each iteration usually isn't a leap to the result we want.

test/unit/deadline_client/api/test_queue_apis.py
sakshie95: done

test/unit/deadline_client/api/test_queue_apis.py
sakshie95: done

test/unit/deadline_client/api/test_queue_apis.py
sakshie95: done

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: done

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: done :))

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: done, also moved as a classmethod as discussed.

src/deadline/client/cli/_groups/queue_group.py
sakshie95: done

src/deadline/client/api/_queue_apis.py
sakshie95: done. We discussed moving this to the cli and keep the api independent of the cli params completely.

src/deadline/client/api/_queue_apis.py
sakshie95: done

src/deadline/client/cli/_groups/queue_group.py
sakshie95: discussed this offline. We decided moving all of this to `queue_group.py` at the cli level. The api will be independent of cli params and will be such that it can be called by a different caller in the future for a different use-case

src/deadline/client/api/_queue_apis.py
```diff
@@ -6,61 +6,44 @@
...
     :param boto3_session: boto3 session
     :param print_function_callback: Callback to print messages produced in this function.
                 Used in the CLI to print to stdout using click.echo. By default, ignores messages.
-    :return: None
+    :return: updated downloaded state
     """
+    # Download outputs for ongoing jobs using current download progress
+    # Right now it is set to no change in progress except setting the last lookback time to now
+    updated_download_progress: IncrementalDownloadState = download_state
+    updated_download_progress.last_lookback_time = datetime.datetime.utcnow().isoformat()
```
mwiebe: It would be logical to keep this as a datetime object. Note that `utcnow` is deprecated so we shouldn't be using it anymore (see e.g. https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated). Here's the preferred pattern:

```python
>>> import datetime
>>> datetime.datetime.now(datetime.timezone.utc)
datetime.datetime(2025, 6, 16, 16, 47, 50, 152426, tzinfo=datetime.timezone.utc)
```

src/deadline/client/api/_queue_apis.py
```diff
@@ -6,61 +6,44 @@
...
     :param path_mapping_rules: path mapping rules for cross OS path mapping
     :param boto3_session: boto3 session
     :param print_function_callback: Callback to print messages produced in this function.
                 Used in the CLI to print to stdout using click.echo. By default, ignores messages.
-    :return: None
+    :return: updated downloaded state
     """
+    # Download outputs for ongoing jobs using current download progress
+    # Right now it is set to no change in progress except setting the last lookback time to now
+    updated_download_progress: IncrementalDownloadState = download_state
```
mwiebe: nit: It's helpful to maintain consistency through renames, here would rename `updated_download_progress` `updated_download_state`. This makes it easier for everyone reading the code later.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        current_download_progress: IncrementalDownloadState = IncrementalDownloadState()
+
+        print_function_callback(
+            "Bootstrapping command. Ignoring download progress location and creating new"
+        )
+        current_download_progress.last_lookback_time = (
+            datetime.datetime.utcnow()
+            - datetime.timedelta(minutes=float(bootstrap_lookback_in_minutes or 0))
+        )
```
mwiebe, line 101: Here `last_lookback_time` is being set to a datetime value, not a string as elsewhere.

A good way to protect against errors like this is adding type information to the class, something like:

```python
class IncrementalDownloadState:
    ...
    last_lookback_time: datetime
    jobs: list[dict]

    ...
```

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,185 @@
...
+        :param saved_progress_checkpoint_full_path: full path of the saved progress checkpoint file
+        :param print_function_callback: Callback to print messages produced in this function.
+                Used in the CLI to print to stdout using click.echo. By default, ignores messages.
+        :return: Returns the loaded state file,
+        or throws an exception if we're unable to read it as we already validated its existence
+        """
+        try:
+            with open(saved_progress_checkpoint_full_path, "r") as file:
+                state_data = json.load(file)
+                current_download_progress: IncrementalDownloadState = (
```
mwiebe: nit: Good idea to de-dent from here down, the file doesn't need to stay open longer.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,185 @@
...
+            print_function_callback(
+                f"Failed to load existing state file from download progress checkpoint location {saved_progress_checkpoint_full_path}: {str(e)}"
+            )
+            # Raise as this is an unexpected exception in reading state file, we already checked its existence earlier
+            raise
+
+    @classmethod
+    def save_file(
+        cls,
+        saved_progress_checkpoint_location: str,
```
mwiebe: Should remove this parameter, because it's included as part of the filename path.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,185 @@
...
+            return current_download_progress
+
+        except Exception as e:
+            print_function_callback(
+                f"Failed to load existing state file from download progress checkpoint location {saved_progress_checkpoint_full_path}: {str(e)}"
+            )
+            # Raise as this is an unexpected exception in reading state file, we already checked its existence earlier
+            raise
+
+    @classmethod
```
mwiebe: This shouldn't be a class method.

mwiebe (changes requested):

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: Done!
I had it typed as a "str" in the final version of the file in a future PR-  https://github.com/aws-deadline/deadline-cloud/pull/709/files#diff-592fb7f096549deb33a6bf9ddc37390d1ee7f2e95f2709b5cfcfbe921b927570R139. However, it'll be better to type it as ``datetime`` & serialize as iso format str in state file.

src/deadline/client/api/_queue_apis.py
sakshie95: oh good call.. done

src/deadline/client/api/_queue_apis.py
sakshie95: Makes sense, done!

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: makes sense done.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: I moved it to be a class level function from the module method it was earlier. I made it a ``@classmethod`` so it's accessible as a stateless library function without instantiating an object of ``IncrementalDownloadState``. Did you have a different suggestion for this? happy to incorporate

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: removed

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
mwiebe: The function accepts a non-optional parameter `download_state`, which means you have to have an instantiated `IncrementalDownloadState` to be able to call it. That parameter should be `self` instead.

sakshie95: Got it, done!

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+
+        try:
+            # Get checkpoint location directory from full path
+            saved_progress_checkpoint_location = os.path.dirname(
+                saved_progress_checkpoint_full_path
+            )
+
+            # 1. Create directory if it doesn't exist
+            os.makedirs(os.path.dirname(saved_progress_checkpoint_location), exist_ok=True)
```
mwiebe: I think this one doesn't need the `os.path.dirname` since that was already applied at line 156

test/unit/deadline_client/cli/test_cli_queue_incremental_download.py
```diff
@@ -0,0 +1,419 @@
...
+)
+@patch("os.access")
+@patch.dict(os.environ, {"ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD": "True"})
+def test_validate_file_inputs_mapping_rules_not_readable(
+    mock_access, checkpoint_dir, path_mapping_rules_file
+):
+    """Test validation when mapping rules file is not readable"""
+    # Create the rules file
+    with open(path_mapping_rules_file, "w") as f:
+        f.write('{"rules": []}')
```
mwiebe, line 389: We'll want to look at what's happening here later, because I'm pretty sure the path mapping rules need to follow what's specified here: https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run#path-mapping

sakshie95: we'll remove path mapping rules from this cli and rely on customer's storage profile instead. I'll make sure this is done as a follow up

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: done

mwiebe (dismissed):

mwiebe (approved):

leongdl (approved):

PR #697, opened 2025-06-04
job 'wait' and 'logs API/CLI commands

prestomation:
The Deadline library/client does have facilities to wait for a job to complete and go from failures to logs.

This is  common workflow during the building and debugging of new conda packages or job templates

#### Waiting for Job Completion

After submitting a job, you can wait for it to complete using the `wait` command:

```sh
# Wait for a job to complete with default settings
$ deadline job wait --job-id job-12345

# Customize the polling interval (default is 10 seconds)
$ deadline job wait --job-id job-12345 --poll-interval 30

# Set a timeout (default is 0, meaning no timeout)
$ deadline job wait --job-id job-12345 --timeout 3600

# Get the result in JSON format
$ deadline job wait --job-id job-12345 --output json
```

The command blocks until the job reaches a terminal state (SUCCEEDED, FAILED, CANCELED, or NOT_COMPATIBLE), then returns information about the job's status and any failed tasks. 

### Retrieving Job Logs

You can monitor job status and retrieve logs using the CLI:

```sh
# Get logs for a specific session
$ deadline job logs --session-id session-12345

# Get logs for a job (automatically uses the latest session based on endedAt time if multiple sessions exist)
$ deadline job logs --job-id job-12345

# Limit the number of log lines returned
$ deadline job logs --session-id session-12345 --limit 50

# Filter logs by time range
$ deadline job logs --session-id session-12345 --start-time 2023-01-01T12:00:00Z --end-time 2023-01-01T13:00:00Z

# Get logs in JSON format
$ deadline job logs --session-id session-12345 --output json

# Paginate through logs
$ deadline job logs --session-id session-12345 --next-token next-token-value
```

When using a Deadline Cloud monitor profile, the `job logs` command will use the Queue role credentials to read logs. Otherwise, the chosen profile credentials are used for all API invocations. This allows you to access logs with the appropriate permissions based on your authentication method.

Debugging tools and workflows do not have to reinvent the wheel for these operations

New unit tests
I'm using this in a new local workflow

- Have you run the integration tests? No

- Are relevant docstrings in the code base updated? No
- Has the README.md been updated? If you modified CLI arguments, for instance. Not yet

### Is this a breaking change? No

### Does this change impact security? No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,64 @@ and removing them by logging out:
 $ deadline auth logout
 ```
 
+## Job Monitoring and Logs
+
+### Waiting for Job Completion
+
+After submitting a job, you can wait for it to complete using the `wait` command:
+
+```sh
+# Wait for a job to complete with default settings
+$ deadline job wait --job-id job-12345
+
+# Customize the maximum polling interval (default is 120 seconds)
+# The polling interval starts at 0.5 seconds and doubles until reaching this maximum
+$ deadline job wait --job-id job-12345 --max-poll-interval 30
+
+# Set a timeout (default is 0, meaning no timeout)
+$ deadline job wait --job-id job-12345 --timeout 3600
+
+# Get the result in JSON format
+$ deadline job wait --job-id job-12345 --output json
+```
+
+The command blocks until the job reaches a terminal state (SUCCEEDED, FAILED, CANCELED, SUSPENDED, NOT_COMPATIBLE), then returns information about the job's status and any failed tasks. It uses exponential backoff for polling, starting at 0.5 seconds and doubling the interval after each check until it reaches the maximum polling interval.
+
+**Exit Codes:**
+- `0` - Job succeeded
+- `1` - Timeout waiting for job completion
+- `2` - Job failed or has failed tasks
+- `3` - Job was canceled
+- `4` - Job was archived
+- `5` - Job is not compatible 
+
+### Retrieving Job Logs
+
+You can monitor job status and retrieve logs using the CLI. The logs lines are returned starting from the most recent log event.:
+
+```sh
+# Get logs for a specific session
+$ deadline job logs --session-id session-12345
+
+# Get logs for a job (automatically selects session: ongoing sessions preferred, then most recently started/ended)
+$ deadline job logs --job-id job-12345
+
+# Limit the number of log lines returned to the 50 most recent.
+$ deadline job logs --session-id session-12345 --limit 50
+
+# Filter logs by time range
+$ deadline job logs --session-id session-12345 --start-time 2023-01-01T12:00:00Z --end-time 2023-01-01T13:00:00Z
+
+# Get logs in JSON format
+$ deadline job logs --session-id session-12345 --output json
+
+# Paginate through logs
+$ deadline job logs --session-id session-12345 --next-token next-token-value
+```
+
+When using a Deadline Cloud monitor profile, the `job logs` command will use the Queue role credentials to read logs. Otherwise, the chosen profile credentials are used for all API invocations. This allows you to access logs with the appropriate permissions based on your authentication method.
+
+
 ## AWS Credentials Integration
 
 You can use the Deadline Cloud client to obtain temporary AWS credentials for a queue and use them with the AWS CLI or SDK. This enables you to create AWS profiles that have queue-specific permissions for use in programmatic workflows.
--- a/src/deadline/client/api/_job_monitoring.py
+++ b/src/deadline/client/api/_job_monitoring.py
@@ -1,0 +1,320 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Functions for monitoring job status and retrieving job information.
+"""
+
+from __future__ import annotations
+import datetime
+import time
+from typing import List, Dict, Any, Optional, Callable
+from configparser import ConfigParser
+from dataclasses import dataclass
+
+from botocore.exceptions import ClientError
+
+from deadline.client.exceptions import DeadlineOperationError, DeadlineOperationTimedOut
+from deadline.client.api._session import (
+    get_boto3_client,
+    get_queue_user_boto3_session,
+    get_user_and_identity_store_id,
+)
+
+
+@dataclass
+class FailedTask:
+    """
+    Represents a failed task in a job.
+    """
+
+    step_id: str
+    task_id: str
+    step_name: str
+    parameters: Dict[str, Any]
+    session_id: Optional[str] = None
+
+
+@dataclass
+class JobCompletionResult:
+    """
+    Result of waiting for a job to complete.
+    """
+
+    status: str
+    failed_tasks: List[FailedTask]
+    elapsed_time: float
+
+
+@dataclass
+class LogEvent:
+    """
+    Represents a single log event from CloudWatch Logs.
+    """
+
+    timestamp: datetime.datetime
+    message: str
+    ingestion_time: Optional[datetime.datetime] = None
+    event_id: Optional[str] = None
+
+
+@dataclass
+class SessionLogResult:
+    """
+    Result of retrieving logs for a session.
+    """
+
+    events: List[LogEvent]
+    next_token: Optional[str]
+    log_group: str
+    log_stream: str
+    count: int
+
+
+def wait_for_job_completion(
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    max_poll_interval: int = 120,
+    timeout: int = 0,
+    config: Optional[ConfigParser] = None,
+    status_callback: Optional[Callable] = None,
+) -> JobCompletionResult:
+    """
+    Wait for a job to complete and return information about its status and any failed tasks.
+
+    This function blocks until the job's taskRunStatus reaches a terminal state
+    (SUCCEEDED, FAILED, CANCELED, SUSPENDED, or NOT_COMPATIBLE), then returns a JobCompletionResult
+    object containing the final status and any failed tasks.
+
+    The function uses exponential backoff for polling, starting at 0.5 seconds and doubling
+    the interval after each check until it reaches the maximum polling interval.
+
+    Args:
+        farm_id: The ID of the farm containing the job.
+        queue_id: The ID of the queue containing the job.
+        job_id: The ID of the job to wait for.
+        max_poll_interval: Maximum time in seconds between status checks (default: 120).
+        timeout: Maximum time in seconds to wait (0 for no timeout).
+        config: Optional configuration object.
+        status_callback: Optional callback function that receives the current status during polling.
+
+    Returns:
+        A JobCompletionResult object containing the job's final status and any failed tasks.
+
+    Raises:
+        DeadlineOperationError: If the timeout is reached or there's an error retrieving job information.
+    """
+    deadline = get_boto3_client("deadline", config=config)
+
+    start_time = datetime.datetime.now()
+    terminal_states = ["SUCCEEDED", "FAILED", "CANCELED", "SUSPENDED", "NOT_COMPATIBLE"]
+    status = ""
+
+    # Initial polling interval of 0.5 seconds
+    current_interval = 0.5
+
+    while True:
+        # Check timeout
+        if timeout > 0:
+            elapsed = (datetime.datetime.now() - start_time).total_seconds()
+            if elapsed > timeout:
+                raise DeadlineOperationTimedOut(
+                    f"Timeout waiting for job {job_id} to complete after {elapsed:.1f} seconds"
+                )
+
+        # Get job status
+        try:
+            job = deadline.get_job(farmId=farm_id, queueId=queue_id, jobId=job_id)
+            status = job.get("taskRunStatus", "")
+
+            # Call the status callback if provided with elapsed time and timeout info
+            if status_callback:
+                elapsed = (datetime.datetime.now() - start_time).total_seconds()
+                status_callback(status, elapsed, timeout)
+
+        except ClientError as exc:
+            raise DeadlineOperationError(f"Failed to get job status: {exc}") from exc
+
+        if status in terminal_states:
+            break
+
+        # Sleep using current interval
+        time.sleep(current_interval)
+
+        # Exponential backoff with a maximum interval
+        current_interval = min(current_interval * 2, max_poll_interval)
+
+    elapsed_time = (datetime.datetime.now() - start_time).total_seconds()
+
+    # If job failed, collect failed tasks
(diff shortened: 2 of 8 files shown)


src/deadline/client/api/_job_monitoring.py
```diff
@@ -0,0 +1,311 @@
...
+
+def wait_for_job_completion(
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    poll_interval: int = 10,
+    timeout: int = 0,
+    config: Optional[ConfigParser] = None,
+    status_callback: Optional[callable] = None,
+) -> JobCompletionResult:
```
mwiebe, line 81: This command reminds me of the `job cancel` command we already have:

```
$ deadline job cancel -h
Usage: deadline job cancel [OPTIONS]

  Cancel job from running.

Options:
  --profile TEXT                  The AWS profile to use.
  --farm-id TEXT                  The farm to use.
  --queue-id TEXT                 The queue to use.
  --job-id TEXT                   The job to cancel.
  --mark-as [canceled|failed|succeeded]
                                  The status to apply to all active tasks in
                                  the job.
  --yes                           Skip any confirmation prompts
```

In particular, that defaults to marking a job as canceled, but if you pass `deadline job cancel --mark-as failed` you can get a different status.

Would this command be comfortable as `deadline job wait-for` where by default it waits for completion, and it has an option to give it different statuses it could wait for?

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -707,6 +725,207 @@ def job_download_output(step_id, task_id, output, **args):
...
+@click.option("--end-time", help="End time for logs in ISO format (e.g., 2023-01-01T13:00:00Z).")
+@click.option("--next-token", help="Token for pagination of results.")
+@click.option(
+    "--output",
+    type=click.Choice(["verbose", "json"], case_sensitive=False),
+    default="verbose",
+    help="Output format (verbose or json).",
+)
+@_handle_error
+def job_get_logs(session_id, limit, start_time, end_time, next_token, output, **args):
```
mwiebe: I wonder if a `deadline job tail-logs` would be more helpful as the first command to add in this vein? It could print the last 20-30 lines of the log, and then continue tailing it.

I think there's another missing command that we need alongside this one, that prints all the active workers/sessions of the job. `deadline job show-running` or similar.

prestomation: Agree tail-logs is likely more useful to an operator. My unsaid requirement is I want something to present to an LLM, thus the classic pagination controls.

prestomation: another option for my particular usecase is to not expose any new functionality in the CLI as my tool won't use it, but these seemed like commands a CLI user would also use

src/deadline/client/api/_job_monitoring.py
prestomation: Does this make sense for this wait-for command? The job is going to end up in one particular terminal state. If the user asked to wait for it to get to status A but it ends up in status B it's going to wait forever right? maybe I'm being dense and missing something here.

One nuance is this is waiting on the job-level property 'taskRunStatus'.  That means that all steps(and tasks) need to transition to a terminal state until this returns. 
An alternative tool may have a 'report-status' list, that prints out any tasks/sessions that do reach that state, but it would still 'wait' until the job reaches a terminal state.  This would let you see the failed(or succeed) IDs as they finish. something like this?

mwiebe: Good point, probably the original name is the right idea. Is there something shorter to type than "completion" though?

One thought is that the exit code should reflect the status - 0 for success, 1 for failure, maybe 2 for something else.

src/deadline/client/cli/_groups/job_group.py
mwiebe: Maybe there's a way to have it make sense for both use cases while still being a simple and succinct command to type. `deadline job logs` that prints logs, with an `--tail` option?

mwiebe: A simple common case is to have a job with one task in one step. In that case I want to just type `deadline job logs --tail` or similar to see what's happening.

src/deadline/client/api/_job_monitoring.py
epmog: I guess `complete` doesn't imply success or failure. We opposed on going with `deadline job wait` and/or adding a `--wait` on the bundle submit command?

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -707,6 +725,207 @@ def job_download_output(step_id, task_id, output, **args):
...
+
+    is_json_output = output.lower() == "json"
+
+    # Define a status callback for verbose output
+    def status_callback(status):
+        if not is_json_output:
+            click.echo(f"Current status: {status}. Waiting {poll_interval} seconds...")
+
+    if not is_json_output:
+        click.echo(f"Waiting for job {job_id} to complete...")
```
epmog: Would it be possible to show a progress bar once the job is picked up instead of constantly dumping output that it's waiting (maybe with the verbose option as opposed to JSON)? If I'm using it as a blocking command in a terminal I'd personally want to visually see how far it's progressing ie. like job attachment/s3 upload/download (otherwise I'd just go to the Monitor).

If I'm scripting, I probably don't want any output like you've done here.

If we want to use the command to chain together some commands, it may help the implementation to show those commands to know what should/should not be put in stdout.

```
deadline bundle submit . && deadline job wait | deadline logs
```
or
```
deadline bundle submit . --wait | deadline logs --first --failed
```

src/deadline/client/cli/_groups/job_group.py
epmog: There's probably a few options we could pass that allow us to choose which logs to grab when running a `deadline job logs` command

Common use-cases for myself are getting the first failure, last failure, tailing, and comparing a successful vs. failed task run. So maybe something like:

```bash
# Get the first session action that failed
deadline job logs --first --failed
# Get the last task logs that succeeded
deadline job logs --last --succeeded
# Get all of them, could either save to separate files on disk or pipe to jq?
deadline job logs --all
# grabs a currently running task, we can be opinionated about which one we choose to display if they don't specify the id
deadline job logs --tail
# takes 2 task actions, default is last success and first failure?
deadline job logs --compare
```
Just spitballing

src/deadline/client/cli/_groups/job_group.py
prestomation: for scripting cases there is the `--output json` or using the underlying API from a python script. agreed a progress bar would be better

src/deadline/client/cli/_groups/job_group.py
prestomation: agreed there are lots of options here! I rarely look at logs that aren't from a failed job or the most recent so I'm making that case the default. showing multiple has UX implications, how exactly would this be displayed? for that reason I'd prefer to punt on those variations but am all ears!

src/deadline/client/api/_job_monitoring.py
prestomation: I've changed it to `wait`. making it a switch on the submit command is an interesting choice. That would work in my case. I can see cases where the submission and the thing waiting on it being done are  different so a separate command makes sense too.

mwiebe: A switch on the submit command would need a bit more subtle thought. I've thought a `--no-wait` would be nice, to return immediately and not wait for the asynchronous success/failure of job submission itself (currently it always waits for that). That means there are two possible different waits that could be controlled during submission.

README.md
```diff
@@ -186,6 +186,62 @@ and removing them by logging out:
...
+After submitting a job, you can wait for it to complete using the `wait` command:
+
+```sh
+# Wait for a job to complete with default settings
+$ deadline job wait --job-id job-12345
+
+# Customize the polling interval (default is 10 seconds)
+$ deadline job wait --job-id job-12345 --poll-interval 30
+
+# Set a timeout (default is 0, meaning no timeout)
```
mwiebe, line 203: Worth mentioning how you tell if it timed out here. Exit code 2 or similar?

README.md
```diff
@@ -186,6 +186,62 @@ and removing them by logging out:
...
+
+### Waiting for Job Completion
+
+After submitting a job, you can wait for it to complete using the `wait` command:
+
+```sh
+# Wait for a job to complete with default settings
+$ deadline job wait --job-id job-12345
+
+# Customize the polling interval (default is 10 seconds)
```
mwiebe: Feels like we can do better than a constant polling interval. Intuitively, I'd like something that avoids excessive polling but has low percentage overhead:

* If the job finishes quickly, the `wait` command should detect it quickly. E.g. a 1 second job should be seen as finished in less than a second, not have to wait 10 or 30 seconds because it just barely missed the completion.
* If the job finishes slowly, the `wait` command can detect it slower. A 1 hour job where `wait` detects it after 2 minutes would be 3% overhead of waiting time.

Maybe a strategy to start polling with a 1 second interval, and do an exponential backoff until this polling interval? Then make the default longer, maybe 2 minutes?

prestomation: Great feedback. I've incorporated it

crowecawcaw (dismissed):

README.md
```diff
@@ -186,6 +186,64 @@ and removing them by logging out:
...
+$ deadline job wait --job-id job-12345 --max-poll-interval 30
+
+# Set a timeout (default is 0, meaning no timeout)
+$ deadline job wait --job-id job-12345 --timeout 3600
+
+# Get the result in JSON format
+$ deadline job wait --job-id job-12345 --output json
+```
+
+The command blocks until the job reaches a terminal state (SUCCEEDED, FAILED, CANCELED, or NOT_COMPATIBLE), then returns information about the job's status and any failed tasks. It uses exponential backoff for polling, starting at 0.5 seconds and doubling the interval after each check until it reaches the maximum polling interval.
```
mwiebe: Looks like `SUSPENDED` should be in the terminal state list as well.

README.md
```diff
@@ -186,6 +186,64 @@ and removing them by logging out:
...
+
+### Retrieving Job Logs
+
+You can monitor job status and retrieve logs using the CLI:
+
+```sh
+# Get logs for a specific session
+$ deadline job logs --session-id session-12345
+
+# Get logs for a job (automatically uses the latest session based on endedAt time if multiple sessions exist)
```
mwiebe: > the latest session based on endedAt time

Does this mean that it won't get logs for in-progress sessions, because they don't have an endedAt yet?

README.md
```diff
@@ -186,6 +186,64 @@ and removing them by logging out:
...
+$ deadline job logs --job-id job-12345
+
+# Limit the number of log lines returned
```
mwiebe: Does this give the first 50 or the last 50 lines?

src/deadline/client/api/_job_monitoring.py
```diff
@@ -0,0 +1,322 @@
...
+    Returns:
+        A JobCompletionResult object containing the job's final status and any failed tasks.
+
+    Raises:
+        DeadlineOperationError: If the timeout is reached or there's an error retrieving job information.
+    """
+    deadline = get_boto3_client("deadline", config=config)
+
+    start_time = datetime.datetime.now()
+    terminal_states = ["SUCCEEDED", "FAILED", "CANCELED", "ARCHIVED", "NOT_COMPATIBLE"]
```
mwiebe: I think this needs to include `SUSPENDED`. Also, `ARCHIVED` shouldn't be here, that's for `lifecycleStatus` but the code is looking at `taskRunStatus`. (Values are listed in https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_GetJob.html#deadlinecloud-GetJob-response-taskRunStatus)

src/deadline/client/api/_job_monitoring.py
```diff
@@ -0,0 +1,322 @@
...
+            raise DeadlineOperationError(f"Failed to get queue credentials: {e}")
+    else:
+        # Use the same boto session as for deadline
+        logs_client = get_boto3_client("logs", config=config)
+
+    # Construct the log group name
+    log_group_name = f"/aws/deadline/{farm_id}/{queue_id}"
+
+    # Ensure session ID has the correct format
+    if not session_id.startswith("session-"):
```
mwiebe: In what circumstances was this happening? This seems to mean that it also accepts values that are not session ids, and tries to fix them up.

src/deadline/client/api/_job_monitoring.py
```diff
@@ -0,0 +1,322 @@
...
+
+    try:
+        response = logs_client.get_log_events(**params)
+
+        # Convert to strongly typed objects
+        events = []
+        for event in response.get("events", []):
+            events.append(
+                LogEvent(
+                    timestamp=datetime.datetime.fromtimestamp(event["timestamp"] / 1000),
```
mwiebe: From https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp, it looks like this creates the timestamp in the local timezone, not UTC. should this be:

```
datetime.datetime.fromtimestamp(event["timestamp"] / 1000, tz=datetime.timezone.utc)
```

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -706,6 +724,307 @@ def job_download_output(step_id, task_id, output, **args):
...
+            exit_code = 5
+        else:
+            # Unknown status, treat as failure
+            exit_code = 2
+
+        sys.exit(exit_code)
+
+    except DeadlineOperationError as e:
+        # Check if this is a timeout error
+        if "Timeout waiting for job" in str(e):
```
mwiebe: In https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/exceptions.py we have `DeadlineOperationError` and `DeadlineOperationCanceled`. Maybe there could be another `DeadlineOperationTimedOut` to model this case explicitly instead of depending on the contents of the message?

README.md
prestomation: Great point, this is ambiguous. Today this always starts with the most recent, so 50 most recent log lines.

src/deadline/client/api/_job_monitoring.py
prestomation: yes the model added this and I kind of glossed over it. I agree I don't like this 'magical' behavior, will remove

README.md
prestomation: Great callout. if there are any sessions that have not ended yet, it will use the session with the latest startedAt instead. i'll update the docs.

mwiebe (approved):

crowecawcaw (approved):

PR #701, opened 2025-06-09
feat(experimental): populate list of ongoing jobs on a queue from search for incrementally downloading output on a queue

sakshie95:
Fixes: N/A

We need to maintain list of active/ongoing jobs on a queue to download outputs incrementally for that queue. 

1. Created a job_processor.py to give the list of active/ongoing jobs on a queue.
2. It gives the list of jobs updated since given ``timestamp`` time.
3. For fetching the list of jobs we make a SearchJobs call to deadline. Since the api is paginated and we might face issues across pages for currently updating jobs we tweak the search call a bit to adjust for this inconsistency - given that the Search max page size is 100, we call SearchJobs with an overlap like 0-99, 90-189, 180-279, ... Then, we verify that at every 10 job overlap, we have matching job ids. If they do not match, it means something changed about the state and we restart

Maintaining list of active/ongoing jobs complete for feature with this PR

Case1. E2E testing of the cli with force bootstrap & a lookback:
```
deadline queue  incremental-output-download --farm-id farm-dce6609ed9844daca0c0d06cc4446d7a --queue-id queue-857988acee53489fafec961e61831a66 --saved-progress-checkpoint-location /Users/sakshie/automatic-downloads --force-bootstrap --bootstrap-lookback-in-minutes 1440000
processing {'farm_id': 'farm-dce6609ed9844daca0c0d06cc4446d7a', 'queue_id': 'queue-857988acee53489fafec961e61831a66', 'conflict_resolution': 'OVERWRITE'}
Started incremental download....
Checking if another download is in progress at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Download pid lock file does not exist at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Creating new pid file at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid with pid 67029
Bootstrapping command. Ignoring download progress location and creating new
Querying for jobs in queue queue-857988acee53489fafec961e61831a66 since 2022-09-13T00:58:31.837983
Searching jobs from page 1, offset 0
Page 1: Added 4 new job IDs
Current total job IDs: 4
Found 4 jobs updated since 2022-09-13T00:58:31.837983
Total ongoing jobs: 4
Got the set of ongoing jobs: {'job-458d5882933a4d3c8aff4eefe93bec74', 'job-a83b7ef5abc1415f9f6f7077e796e5fa', 'job-316e5171f00b4072aa3045b2559b8d53', 'job-7b7fff8216074430b4baf84736713d8f'} on queue queue-857988acee53489fafec961e61831a66
Successfully saved state file to /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_download_progress.json
Releasing pid lock at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Process with pid 67029 is the current process. Deleting pid file.
```

Case 2: E2E testing of the cli with force bootstrap and no lookback (starts from now):
```
deadline queue  incremental-output-download --farm-id farm-dce6609ed9844daca0c0d06cc4446d7a --queue-id queue-857988acee53489fafec961e61831a66 --saved-progress-checkpoint-location /Users/sakshie/automatic-downloads --force-bootstrap                                             
processing {'farm_id': 'farm-dce6609ed9844daca0c0d06cc4446d7a', 'queue_id': 'queue-857988acee53489fafec961e61831a66', 'conflict_resolution': 'OVERWRITE'}
Started incremental download....
Checking if another download is in progress at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Download pid lock file does not exist at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Creating new pid file at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid with pid 70219
Bootstrapping command. Ignoring download progress location and creating new
Querying for jobs in queue queue-857988acee53489fafec961e61831a66 since 2025-06-09T01:14:00.812615
Searching jobs from page 1, offset 0
Found 0 jobs updated since 2025-06-09T01:14:00.812615
Total ongoing jobs: 0
Got the set of ongoing jobs: set() on queue queue-857988acee53489fafec961e61831a66
Successfully saved state file to /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_download_progress.json
Releasing pid lock at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Process with pid 70219 is the current process. Deleting pid file.
```

Case 3: E2E testing of the cli with continuing downloading from current progress:
```
deadline queue  incremental-output-download --farm-id farm-dce6609ed9844daca0c0d06cc4446d7a --queue-id queue-857988acee53489fafec961e61831a66 --saved-progress-checkpoint-location /Users/sakshie/automatic-downloads                       
processing {'farm_id': 'farm-dce6609ed9844daca0c0d06cc4446d7a', 'queue_id': 'queue-857988acee53489fafec961e61831a66', 'conflict_resolution': 'OVERWRITE'}
Started incremental download....
Checking if another download is in progress at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Download pid lock file does not exist at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Creating new pid file at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid with pid 70880
Loaded existing state file from download progress checkpoint location /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_download_progress.json
Querying for jobs in queue queue-857988acee53489fafec961e61831a66 since 2025-06-09T01:14:01.512438
Searching jobs from page 1, offset 0
Found 0 jobs updated since 2025-06-09T01:14:01.512438
Total ongoing jobs: 0
Got the set of ongoing jobs: set() on queue queue-857988acee53489fafec961e61831a66
Successfully saved state file to /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_download_progress.json
Releasing pid lock at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Process with pid 70880 is the current process. Deleting pid file.
```

No

No

----


src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -0,0 +1,287 @@
...
+        Returns:
+            Set[str]: Set of job IDs
+        """
+        return {job.job_id for job in self.jobs}
+
+    def get_last_lookback_time(self) -> str:
+        return self.last_lookback_time
+
+
+def bootstrap_fresh_state(
```
sakshie95: code from hereon in this file is already in review in https://github.com/aws-deadline/deadline-cloud/pull/691/files#diff-592fb7f096549deb33a6bf9ddc37390d1ee7f2e95f2709b5cfcfbe921b927570. This will be fixed with rebase after it is merged

src/deadline/client/api/_queue_apis.py
```diff
@@ -36,23 +49,76 @@ def _incremental_output_download(
...
+    # 1. Construct pid file full path
+    pid_file_full_path = os.path.join(
+        saved_progress_checkpoint_location, f"{queue_id}_{PID_FILE_NAME}"
+    )
 
     try:
-        # Check if a download is already ongoing with pid lock checking mechanism
-        _pid_utils.check_and_obtain_pid_lock_if_available(
-            saved_progress_checkpoint_location, logger
+        # 2. Check if a download is already ongoing with pid lock checking mechanism
```
leongdl: These changes looks like they come from another PR, I guess it is due to all the forks of code you have.

src/deadline/job_attachments/incremental_downloads/job_processor.py
```diff
@@ -0,0 +1,150 @@
...
+        if item_offset > 0 and not restart_needed:
+            # Get job IDs from the overlapping section of the previous page
+            previous_overlap_ids = set(all_job_ids[-OVERLAP_SIZE:])
+
+            # Get job IDs from the overlapping section of the current page
+            current_overlap_ids = set(current_page_job_ids[:OVERLAP_SIZE])
+
+            # If overlap doesn't match, we need to restart from the beginning
+            if previous_overlap_ids != current_overlap_ids:
+                print_function_callback(
```
leongdl: nit; This basically forces the loop to infinitely spin until the full set matches in this page. 

If it is a super busy far, we probably want an escape hatch where > X minutes of retrying we have to give up.

src/deadline/client/api/_queue_apis.py
sakshie95: yes it was intentional for keeping the usage of the library function introduced in this PR. But now that we're aligned on the interaction I've removed the usage of the library functions and just kept them in the scope of this review

src/deadline/job_attachments/incremental_downloads/updated_jobs_fetcher.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Check for overlap consistency if not the first page and not after a restart
+        if item_offset > 0 and not restart_needed:
+            # Get job IDs from the overlapping section of the previous page
+            previous_overlap_ids = set(all_job_ids[-OVERLAP_SIZE:])
+
+            # Get job IDs from the overlapping section of the current page
+            current_overlap_ids = set(current_page_job_ids[:OVERLAP_SIZE])
+
+            # If overlap doesn't match, we need to restart from the beginning
```
mwiebe: This strategy for treating the overlaps looks like it can be forced to restart indefinitely, and I think we need to avoid that. I see a way to define a simpler single pass algorithm for this.

The key is to sort the jobs by increasing updatedAt timestamp, so that each search_jobs query gets the oldest 100 jobs that are newer than the timestamp used in the filter expression. We use a property that the updatedAt timestamp will obey: Any time a job is created or updated, it becomes the newest job in the queue, i.e. it becomes the last item in the list.

Here's some pseudo-code based on this reasoning. There are details to clean up about it to make it robust, like the overlap, but this approach should be solid:

```python
def get_jobs_updated_since_timestamp(timestamp):
    threshold_timestamp = timestamp
    result = set()
    jobs = []
    total_results = -1
    while len(jobs) != total_results:
        response = search_jobs(filter=job.timestamp >= threshold_timestamp, sort=increasing by job.timestamp)
        jobs = response.jobs
        total_results = response.totalResults
        if jobs:
            result.update(job.jobId for job in jobs)
            threshold_timestamp = jobs[-1].updatedAt
    return result
```

mwiebe (changes requested):

src/deadline/job_attachments/incremental_downloads/updated_jobs_fetcher.py
sakshie95: This is a good idea. While I was implementing this I saw a use-case where this may fail - if the ``updatedAt`` is the exact same to the last millisecond for all 100 jobs in a page in a very busy farm in which case this algorithm may be stuck in an infinite loop. Is an edge case but is definitely possible.. what do you think?

mwiebe: I can see two ways we could deal with it: 1. Try to come up with a way to implement correct behavior in this case. 2. Detect this case and fail the operation if it happens.

Based on your reasoning about the probability, which do you think we should do?

src/deadline/job_attachments/incremental_downloads/updated_jobs_fetcher.py
```diff
@@ -0,0 +1,148 @@
...
+    item_offset = 0
+    page_number = 1
+
+    # Flag to track if we need to restart due to inconsistency
+    restart_needed = False
+
+    while True:
+        print_function_callback(f"Searching jobs from page {page_number}, offset {item_offset}")
+
+        deadline = boto3_session.client("deadline", config=get_default_client_config())
```
leongdl, line 69: I don't think we need to create a new client every time here, we can and should use a shared one.

EG: a frequent PR from Brian showed how much cost it is to create this.

src/deadline/job_attachments/incremental_downloads/updated_jobs_fetcher.py
sakshie95: Since the probability of this happening is low I've chosen the second option for now and thrown an exception.

src/deadline/job_attachments/incremental_downloads/updated_jobs_fetcher.py
sakshie95: ack makes sense.. I'll fix this

PR #706, opened 2025-06-10
feat(experimental): downloading session action outputs with merged manifest attachment download for incremental download cli

sakshie95:
Fixes: N/A

We need a helper class to provide the function to create merged manifest and download the outputs for session action ids for taskRuns that have been identified as completed.

Implemented ``aggregate_manifest_and_download_outputs`` in a ``manifest_download_handler.py`` that aggregates output manifests using a set of SessionActionMappings (we need session_action_id, step_id, task_id and job_id to create a merged output manifest) and then use the existing attachment download module to download attachments using the merged manifest. 

Download implemented for session action ids that have been identified as completed for incremental download cli.

This change is tested e2e using the CLI - outputs get downloaded for the session actions identified as completed:
```
Found output paths for job job-7b7fff8216074430b4baf84736713d8f, session action sessionaction-8a0a10050d444ee7a5fbf1b8f0564d61-3: {'/Users/sakshie/Documents/maya/projects/default': AssetManifest(hashAlg=<HashAlgorithm.XXH128: 'xxh128'>, paths=[ManifestPath(path='scenes/images/cube.exr', hash='ba951a6fa7d804bed6c7aa3aa0163920', size=18510, mtime=1749456409958651)], manifestVersion=<ManifestVersion.v2023_03_03: '2023-03-03'>, totalSize=18510)}
Processed 1 file totaling 18.51 KB.
Skipped re-processing 0 files totaling 0 B.
Total processing time of 0.22269 seconds at 83.12 KB/s.

Processing session action sessionaction-412ab6a091014446ae1b5a6eb5fb4f40-3 (Step: step-1e285402d4e544b4bcd8bc94e161d7d3, Task: task-1e285402d4e544b4bcd8bc94e161d7d3-0)
Found output paths for job job-7b7fff8216074430b4baf84736713d8f, session action sessionaction-412ab6a091014446ae1b5a6eb5fb4f40-3: {'/Users/sakshie/Documents/maya/projects/default': AssetManifest(hashAlg=<HashAlgorithm.XXH128: 'xxh128'>, paths=[ManifestPath(path='scenes/images/cube.exr', hash='ba951a6fa7d804bed6c7aa3aa0163920', size=18510, mtime=1749456409958651)], manifestVersion=<ManifestVersion.v2023_03_03: '2023-03-03'>, totalSize=18510)}
Processed 1 file totaling 18.51 KB.
Skipped re-processing 0 files totaling 0 B.
Total processing time of 0.19147 seconds at 96.67 KB/s.

Processing session action sessionaction-efaf5a8cbfd94426afeeb6470f8afcce-3 (Step: step-1e285402d4e544b4bcd8bc94e161d7d3, Task: task-1e285402d4e544b4bcd8bc94e161d7d3-0)
Found output paths for job job-7b7fff8216074430b4baf84736713d8f, session action sessionaction-efaf5a8cbfd94426afeeb6470f8afcce-3: {'/Users/sakshie/Documents/maya/projects/default': AssetManifest(hashAlg=<HashAlgorithm.XXH128: 'xxh128'>, paths=[ManifestPath(path='scenes/images/cube.exr', hash='ba951a6fa7d804bed6c7aa3aa0163920', size=18510, mtime=1749456409958651)], manifestVersion=<ManifestVersion.v2023_03_03: '2023-03-03'>, totalSize=18510)}
Processed 1 file totaling 18.51 KB.
Skipped re-processing 0 files totaling 0 B.
Total processing time of 0.18868 seconds at 98.11 KB/s.

Processing session action sessionaction-3925869816c24bec9a5846461a362372-3 (Step: step-1e285402d4e544b4bcd8bc94e161d7d3, Task: task-1e285402d4e544b4bcd8bc94e161d7d3-0)
Found output paths for job job-7b7fff8216074430b4baf84736713d8f, session action sessionaction-3925869816c24bec9a5846461a362372-3: {'/Users/sakshie/Documents/maya/projects/default': AssetManifest(hashAlg=<HashAlgorithm.XXH128: 'xxh128'>, paths=[ManifestPath(path='scenes/images/cube.exr', hash='ba951a6fa7d804bed6c7aa3aa0163920', size=18510, mtime=1749456409958651)], manifestVersion=<ManifestVersion.v2023_03_03: '2023-03-03'>, totalSize=18510)}
Processed 1 file totaling 18.51 KB.
Skipped re-processing 0 files totaling 0 B.
Total processing time of 0.20176 seconds at 91.74 KB/s.

Processing session action sessionaction-3996be85d3ca4c23a2c8926101e0cc6b-3 (Step: step-1e285402d4e544b4bcd8bc94e161d7d3, Task: task-1e285402d4e544b4bcd8bc94e161d7d3-0)
Found output paths for job job-7b7fff8216074430b4baf84736713d8f, session action sessionaction-3996be85d3ca4c23a2c8926101e0cc6b-3: {'/Users/sakshie/Documents/maya/projects/default': AssetManifest(hashAlg=<HashAlgorithm.XXH128: 'xxh128'>, paths=[ManifestPath(path='scenes/images/cube.exr', hash='ba951a6fa7d804bed6c7aa3aa0163920', size=18510, mtime=1749456409958651)], manifestVersion=<ManifestVersion.v2023_03_03: '2023-03-03'>, totalSize=18510)}
Processed 1 file totaling 18.51 KB.
Skipped re-processing 0 files totaling 0 B.
Total processing time of 0.19472 seconds at 95.06 KB/s.

Processing session action sessionaction-607850c7e7e7464aba297527f47b2dc7-3 (Step: step-1e285402d4e544b4bcd8bc94e161d7d3, Task: task-1e285402d4e544b4bcd8bc94e161d7d3-0)
Found output paths for job job-7b7fff8216074430b4baf84736713d8f, session action sessionaction-607850c7e7e7464aba297527f47b2dc7-3: {'/Users/sakshie/Documents/maya/projects/default': AssetManifest(hashAlg=<HashAlgorithm.XXH128: 'xxh128'>, paths=[ManifestPath(path='scenes/images/cube.exr', hash='ba951a6fa7d804bed6c7aa3aa0163920', size=18510, mtime=1749456409958651)], manifestVersion=<ManifestVersion.v2023_03_03: '2023-03-03'>, totalSize=18510)}
Processed 1 file totaling 18.51 KB.
Skipped re-processing 0 files totaling 0 B.
Total processing time of 0.20273 seconds at 91.31 KB/s.

Processing session action sessionaction-813f81abc3b745629abe9b72fc2ea2aa-3 (Step: step-1e285402d4e544b4bcd8bc94e161d7d3, Task: task-1e285402d4e544b4bcd8bc94e161d7d3-0)
Found output paths for job job-7b7fff8216074430b4baf84736713d8f, session action sessionaction-813f81abc3b745629abe9b72fc2ea2aa-3: {'/Users/sakshie/Documents/maya/projects/default': AssetManifest(hashAlg=<HashAlgorithm.XXH128: 'xxh128'>, paths=[ManifestPath(path='scenes/images/cube.exr', hash='ba951a6fa7d804bed6c7aa3aa0163920', size=18510, mtime=1749456409958651)], manifestVersion=<ManifestVersion.v2023_03_03: '2023-03-03'>, totalSize=18510)}
Processed 1 file totaling 18.51 KB.
Skipped re-processing 0 files totaling 0 B.
Total processing time of 0.2801 seconds at 66.08 KB/s.

Downloaded outputs for session action ids: ['sessionaction-8a0a10050d444ee7a5fbf1b8f0564d61-3', 'sessionaction-412ab6a091014446ae1b5a6eb5fb4f40-3', 'sessionaction-efaf5a8cbfd94426afeeb6470f8afcce-3', 'sessionaction-3925869816c24bec9a5846461a362372-3', 'sessionaction-3996be85d3ca4c23a2c8926101e0cc6b-3', 'sessionaction-607850c7e7e7464aba297527f47b2dc7-3', 'sessionaction-813f81abc3b745629abe9b72fc2ea2aa-3']
Successfully saved state file to /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_download_progress.json
Releasing pid lock at /Users/sakshie/automatic-downloads/queue-857988acee53489fafec961e61831a66_incremental_output_download.pid
Process with pid 51041 is the current process. Deleting pid file.
```

- Has the README.md been updated? If you modified CLI arguments, for instance. Not right now, this will be done later

No

No
----


src/deadline/job_attachments/api/attachment.py
```diff
@@ -52,13 +52,31 @@ def attachment_download(
...
+        boto3_session,
+        file_name_manifest_dict,
+        s3_root_uri,
+        conflict_resolution,
+        path_mapping_rules,
+        logger,
+    )
+
+
+def _attachment_download_with_root_manifests(
```
mwiebe, line 70: Could we switch `attachment_download` itself to have this interface? I see it's marked as follows:

> BETA API - This API is still evolving.

Maybe even better is to rename it to `_attachment_download` to more clearly advertise that this isn't part of the public interface yet.

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,145 @@
...
+        boto3_session: boto3.Session,
+        session_actions: List[SessionAction],
+        farm_id: str,
+        queue_id: str,
+        file_conflict_resolution: FileConflictResolution,
+        path_mapping_rules: Optional[str] = None,
+        print_function_callback: Callable[[str], None] = lambda msg: None,
+) -> None:
+    """
+    Aggregate manifests and download outputs for the given session actions.
```
mwiebe: This is only for taskRun session actions, correct? Would be good to specify that.

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,145 @@
...
+    )
+
+    downloaded_session_action_ids = []
+    failed_session_action_ids: List[str] = []
+
+    # Get job attachment settings once
+    ja_settings = JobAttachmentS3Settings(**queue["jobAttachmentSettings"])
+    s3_root_uri = ja_settings.to_s3_root_uri()
+
+    # Function to process all session actions serially
```
mwiebe: I see the logic here is:

1. for each session action:
   1. download and aggregate its manifests
   2. download all its files

I believe the operation will be better overall if we do it as:

1. download and aggregate the manifests of all session actions
2. download all the files of the fully aggregated manifests

mwiebe (changes requested):

src/deadline/job_attachments/api/attachment.py
sakshie95: Yeah I thought about it but I think this CLI is being tested & released as we speak in the current state so I don't want to switch this right now as a part of these changes to cause any conflicts.

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: yes, I can specify it

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: I plan to download and aggregate the manifests of all sessions to make this faster & parallelizable per-session in a future iteration (with concurrent futures). See https://github.com/sakshie95/deadline-cloud/blob/feature/early-access/src/deadline/job_attachments/incremental_downloads/session_action_processor.py#L314 for a rough reference to what I'm thinking.

But the error handling is trickier with the above approach and it not currently implemented in the code shared above. We will need to return all session action ids associated with an aggregated manifest if there was any failure in processing even if some session action id outputs would have downloaded successfully - that might lead to more than once download for the customer.

This is because of the way the existing libraries are designed for output aggregation & attachment download - all we get as a response is a list of local directory root (which may be modified in windows long paths) & the number of files downloaded in that directory.

mwiebe: If we switch this operation to no longer return partial success, the error handling stays reasonable. An example benefit to aggregating all of the manifests first is if a later task run overwrites a file that a previous task run wrote, it will no longer double-download that file.

sakshie95: Agree with the benefit and I've switched this to an all or nothing download - aggregating manifests for all session action ids and downloading once. Here's some performance data I gathered for manifest aggregation and download which is a bit concerning:
```
Manifest aggregation completed in 167.59 seconds for Total session actions (taskrun) to download: 1273
Download completed in 0.34 seconds
Total aggregation and download completed in 176.98 seconds
```

While download is very fast with this approach for the aggregated manifest - I think manifest aggregation seems to be a bottleneck for downloads at scale.

mwiebe: How many manifests are being aggregated in that test? What does the fully aggregated manifest look like? We can't draw any conclusions about the design from that data without looking carefully at the details.

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,123 @@
...
+
+        except Exception as e:
+            print_function_callback(
+                f"Error getting manifests for session action {session_action.session_action_id}: {e}"
+            )
+            raise AssetDownloadFailedException("Aggregation & Download Failed")
+
+    # Created a merged manifest per root
+    for root, manifests in all_output_manifests_by_root.items():
+        merged_manifest = merge_asset_manifests(manifests)
```
mwiebe: The merge_asset_manifests operation has later manifests effectively overwrite the files of earlier ones, because merging two manifests is not commutative. How does the code here ensure that manifests are being merged in the correct order?

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: I'll gather this data and we can address this as a part of the optimizations PR

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: Good call, looks like this is missing even for the manifests download api/library too. I can create a bug for it too.
For our handling, I've added some logic to sort session actions so that the merged manifests are created from that ordered list. I've also added some handling to the merging logic to make sure later manifests overwrite the earlier ones.

src/deadline/job_attachments/api/attachment.py
leongdl: We can't change it without breaking worker agent via a major bump - https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/774ef56a5ecb25beb0d1aa293439311717e28f1c/src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py#L24

Can we use a union or optional parameter for now so we can migrate worker agent? We need some sort of 3 phase change. 

Given the API is "beta", I don't think we should or need version bumps.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -92,13 +110,13 @@ def attachment_download(
...
-    boto3_session: boto3.Session,
-    root_dirs: List[str] = [],
-    path_mapping_rules: Optional[str] = None,
-    upload_manifest_path: Optional[str] = None,
-    logger: ClickLogger = ClickLogger(False),
+        manifests: List[str],
+        s3_root_uri: str,
+        boto3_session: boto3.Session,
+        root_dirs: List[str] = [],
+        path_mapping_rules: Optional[str] = None,
```
leongdl: IMO, this API should mirror the download as a path mapping object, but that is out of scope. 

We need a way to define the interface in/out at each level. CLI, API, modules, and then keep them consistent.

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,129 @@
...
+
+            # If no output manifests, continue to the next session action
+            if not output_manifests_by_root:
+                continue
+
+            print_function_callback(
+                f"Found output paths for session action {session_action.session_action_id}"
+            )
+
+            # Merge with existing manifests, ensuring later session actions override earlier ones
```
leongdl: Lines 93 to 105 I've seen in a few places, this is a good function / utility to have separately. (if not already)

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,129 @@
...
+                    # The order matters here - the new manifest should override the existing one
+                    combined_manifest = merge_asset_manifests([merged_output_manifests[root]] + manifests)
+                    if combined_manifest:
+                        merged_output_manifests[root] = combined_manifest
+
+        except Exception as e:
+            print_function_callback(
+                f"Error getting manifests for session action {session_action.session_action_id}: {e}"
+            )
+            raise AssetDownloadFailedException("Aggregation & Download Failed")
```
leongdl: Where did we download? I dont see it in the current code up to here.

leongdl: Maybe for the error message to be clear we could not merge the manifest? But I can't forsee any merge manifest errors, more likely cannot get output manifest by root since it scans S3. Can we scope the exception block better?

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # If we have merged manifests, download them
+    if merged_output_manifests:
```
leongdl: else:
 -> Nothing to download or cannot download?

src/deadline/job_attachments/api/attachment.py
leongdl: function docs please.

src/deadline/job_attachments/download.py
```diff
@@ -256,6 +256,83 @@ def _get_tasks_manifests_keys_from_s3(
     return manifests_keys
 
 
+def _get_tasks_manifests_keys_from_s3_sorted_by_last_modified(
+    manifest_prefix: str, s3_bucket: str, session: Optional[boto3.Session] = None
+) -> List[Tuple[str, datetime]]:
+    """
```
leongdl, line 283: nit: Function arguments docs please.

src/deadline/job_attachments/download.py
```diff
@@ -256,6 +256,83 @@ def _get_tasks_manifests_keys_from_s3(
...
+            action="listing bucket contents",
+            error_details=str(bce),
+        ) from bce
+    except JobAttachmentsError:
+        raise  # pass along JobAttachmentsErrors if we get them
+    except Exception as e:
+        raise AssetSyncError(e) from e
+
+    # Sort all manifest keys across all session actions by their LastModified timestamp (oldest first)
+    manifests_keys_with_metadata.sort(key=lambda x: x[1])
```
leongdl: `x[1]` ? What are we sorting here.

leongdl: Oh I see, its a pair.... manifests_keys_with_metadata.append((content["Key"], content["LastModified"]))

src/deadline/job_attachments/download.py
```diff
@@ -741,6 +818,54 @@ def get_output_manifests_by_asset_root(
...
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    step_id: Optional[str] = None,
+    task_id: Optional[str] = None,
+    session_action_id: Optional[str] = None,
+    session: Optional[boto3.Session] = None,
+) -> dict[str, list[Tuple[BaseAssetManifest, datetime]]]:
+    """
+    For a given job/step/task, gets a map from each root path to a corresponding list of
```
leongdl: function docs please.

src/deadline/job_attachments/download.py
```diff
@@ -741,6 +818,54 @@ def get_output_manifests_by_asset_root(
     return outputs
 
 
+def get_output_manifests_by_asset_root_sorted_by_last_modified_asc(
```
leongdl: Curious - why is a manifest function in download.py.

Can we move it to a manifest utility file?

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,166 @@
...
+            except Exception as e:
+                print_function_callback(
+                    f"Error getting manifests for session action {session_action.session_action_id}: {e}"
+                )
+                raise AssetDownloadFailedException("Aggregation & Download Failed")
+
+        # Process each root directory separately
+        merged_output_manifests = {}
+
+        for root, manifests_with_timestamps in all_manifests_by_root.items():
```
leongdl: We should refactor this logic into the standard merge manifest function. This will also make the manifest handling consistent for all other use cases. 

Otherwise we may have different behvior between this tool and other usages. (regular job download, worker agent etc)

src/deadline/job_attachments/api/attachment.py
```diff
@@ -52,19 +52,51 @@ def attachment_download(
...
+    path_mapping_rule_list: Optional[List[PathMappingRule]] = None,
+    logger: ClickLogger = ClickLogger(False),
+):
+    """
+    Function to use for attachment download when the caller has manifests and path mapping rule list,
+    instead of reading these from input files.
+    We should make this the default API Interface eventually to make it flexible
+
+    :param boto3_session: boto3 session
+    :param file_name_manifest_dict: manifests classified by file name
```
mwiebe: What is the file name referred to here, is it a local path on the file system? The description is ambiguous and could be misinterpreted.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -52,19 +52,51 @@ def attachment_download(
...
         # File name is supposed to be prefixed by a hash of source path in path mapping, use that to determine destination
         # If it doesn't appear in path mapping or mapping doesn't exist, download to current directory instead
         destination = next(
             (
                 rule.destination_path
-                for rule in path_mapping_rule_list
+                for rule in (path_mapping_rule_list or [])
                 if rule.get_hashed_source_path(manifest.get_default_hash_alg()) in file_name
             ),
             # Write to current directory partitioned by manifest name when no path mapping defined
```
mwiebe, line 103: This isn't from your PR, but this behavior looks like something we should re-evaluate. We may want to do something else in this case.

src/deadline/job_attachments/download.py
```diff
@@ -256,6 +273,62 @@ def _get_tasks_manifests_keys_from_s3(
...
+    :param session: Optional boto3 session
+    :return: None
+    """
+
+    # Update manifests for each session action
+    for session_action in session_actions:
+        # Check if any session action already has output manifests
+        if session_action.get("manifests"):
+            # At least one session action already has output manifests, no need to update
+            return
```
mwiebe: This should be a `continue`, because a job can run on a mix of older and newer worker agents.

src/deadline/job_attachments/download.py
```diff
@@ -256,6 +273,62 @@ def _get_tasks_manifests_keys_from_s3(
     return manifests_keys
 
 
+def update_manifest_output_paths_for_session_actions(
```
mwiebe: Let's keep this function prefixed with `_` until we've used it as an internal function of the library for a while. That way we can make changes, and then remove the `_` once we're confident in the interface.

src/deadline/job_attachments/download.py
```diff
@@ -256,6 +273,62 @@ def _get_tasks_manifests_keys_from_s3(
...
+            # Convert paths to the expected format
+            manifests = []
+            for path in manifest_paths:
+                manifests.append({"outputManifestPath": path})
+
+            # Set the session action manifests in the incoming object
+            session_action["manifests"] = manifests
+        except JobAttachmentsError:
+            # Rethrow an exception here because we were unable to set session action output manifest paths successfully
+            raise
```
mwiebe: Better to fully remove this try/except. The exception behavior will be the same without it.

src/deadline/job_attachments/download.py
```diff
@@ -256,6 +273,62 @@ def _get_tasks_manifests_keys_from_s3(
...
+
+            # Get manifest paths from S3 using prefix
+            manifest_paths = _get_tasks_manifests_keys_from_s3(
+                manifest_prefix, s3_settings.s3BucketName, session=session
+            )
+
+            # Convert paths to the expected format
+            manifests = []
+            for path in manifest_paths:
+                manifests.append({"outputManifestPath": path})
```
mwiebe: I suspect this will not put the manifest paths in the correct order. Maybe put that as a TODO comment to investigate and fix, and then we can merge this before verifying/resolving that.

src/deadline/job_attachments/download.py
```diff
@@ -741,6 +814,42 @@ def get_output_manifests_by_asset_root(
     return outputs
 
 
+def get_output_manifest_files_by_asset_root_with_last_modified(
+    s3_settings: JobAttachmentS3Settings,
+    output_manifest_paths: list[str],
+    session: Optional[boto3.Session] = None,
+) -> dict[str, list[Tuple[datetime, BaseAssetManifest]]]:
```
mwiebe: This interface would be more flexible if it returns a list that exactly mirrors the provided `output_manifest_paths`, like list[tuple[str, datetime, BaseAssetManifest]]`. If the caller doesn't actually want them grouped by asset root, they have to undo this grouping with some tricky code. On the other hand, if a caller does want the grouping, and is given the list of 3-tuples, it's a simple transformation to perform that grouping.

src/deadline/job_attachments/download.py
```diff
@@ -741,6 +814,42 @@ def get_output_manifests_by_asset_root(
...
+    """
+    For a given output manifest path, gets a map from each root path to a corresponding list of
+    output manifests with their last modified timestamps.
+
+    Returns:
+        A dictionary mapping asset root paths to lists of tuples containing (last_modified, manifest)
+    """
+    outputs: DefaultDict[str, list[Tuple[datetime, BaseAssetManifest]]] = DefaultDict(list)
```
mwiebe: To get the mirroring behavior I've suggested, you can initialize with:

```
outputs: list[tuple[str, datetime, BaseAssetManifest]] = [None] * len(output_manifest_paths)
```

Then make the parallel loop be `for index in range(len(output_manifest_paths))` so it can grab the correct item from `output_manifest_paths` and place the result into `outputs`.

src/deadline/job_attachments/download.py
```diff
@@ -908,6 +1017,38 @@ def merge_asset_manifests(manifests: list[BaseAssetManifest]) -> BaseAssetManife
     return output_manifest
 
 
+def merge_asset_manifests_sorted_asc_by_last_modified(
```
mwiebe: Do we need this function? What this does will be simple to read and understand in any client code calling it. Maybe prefix it with `_` to keep it internal for now, so that we can decide that later.

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,144 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mwiebe: It would be good to rename the `incremental_downloads` directory to `_incremental_downloads` and keep these functions internal-only while we're working out details.

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,144 @@
...
+        queue_display_name=queue["displayName"],
+    )
+
+    # Get job attachment settings once
+    ja_settings = JobAttachmentS3Settings(**queue["jobAttachmentSettings"])
+    s3_root_uri = ja_settings.to_s3_root_uri()
+
+    try:
+        # Get all manifests with their last modified timestamps, grouped by asset root
+        output_manifest_with_last_modified_per_asset_root = (
```
mwiebe: I think there's some complications we need to work out because we're merging manifests from multiple different jobs together. The paths themselves can overlap without the asset roots being identical.

What seems best to me is to prefix `incremental_downloads` with `_` so that we can treat it as an internal-only detail and iterate on the implementation that way.

mwiebe (changes requested):

src/deadline/job_attachments/api/attachment.py
leongdl: Please have a todo to resolve how to get this function working with Worker Agent. Probably needs some 3 phase change or magic.

src/deadline/job_attachments/download.py
```diff
@@ -89,9 +89,16 @@ def get_manifest_from_s3(
     return manifest
 
 
-def get_asset_root_and_manifest_from_s3(
+def get_asset_root_and_manifest_from_s3_with_last_modified(
```
leongdl: At this rate the function return value will be a structure, can / should we do that?  A pair of 2 is ok, but a tuple of 3 starts to get weird and unweildy. Especially when this function is not internal.

src/deadline/job_attachments/download.py
```diff
@@ -141,6 +149,15 @@ def get_asset_root_and_manifest_from_s3(
         raise AssetSyncError(e) from e
 
 
+def get_asset_root_and_manifest_from_s3(
```
leongdl: Is this because of backwards compatibility? Who is using this function right now?

leongdl: https://github.com/search?q=repo%3Aaws-deadline%2Fdeadline-cloud%20get_asset_root_and_manifest_from_s3&type=code

I see the usages are all in this file, can we just refactor it instead of making a bridge?

src/deadline/job_attachments/download.py
```diff
@@ -256,6 +273,62 @@ def _get_tasks_manifests_keys_from_s3(
     return manifests_keys
 
 
+def update_manifest_output_paths_for_session_actions(
+    s3_settings: JobAttachmentS3Settings,
+    farm_id: str,
+    queue_id: str,
+    session_actions: list[dict[str, Any]],
+    session: Optional[boto3.Session],
```
leongdl, line 281: nit, the word Session is very overloaded. boto_session?

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+
+class AssetDownloadFailedException(Exception):
+    """
+    Asset Download Failed Exception to throw to caller if any of the aggregations & download failed
+    """
+
+
+def aggregate_manifest_and_download_outputs(
```
leongdl: I'm not convinced this is a good interface function to have as a building block.

From the code flow of any client, they want to download a job or jobs from a queue.

How would they know what `output manifest paths` are ?

I tend to think this we should have functions such as:
Downloader.py
   - get all output manifest. (encapsulates both pre-chunking and post chunking behavior)
   - Partition output manifests by root (output is a dictionary of a list)
   - aggregate all manifest. (Input is a list of manifest for 1 root)
   - generate the equivalent path mapping by root (given Storage profile as well)
   - Download an aggregated manifest of 1 root given the path mapping.

Somehow lumping points 3-5 is very odd. I want to see a nice pyramid like set of functions that builds up to the end functionality :)

src/deadline/job_attachments/download.py
mwiebe: This is quite confusing, yeah. It's also in the interface of existing functions for quite some time. Maybe it would be worth doing a PR to change it throughout to `boto_session` and flagging it as a breaking change in the release notes?

src/deadline/job_attachments/api/attachment.py
sakshie95: Yes I had the same thought and we'll fix it atleast for our use-case with path mapping rules, I'll create an item to re-evaluate bcz we checked with the owner that this was intended behaviour for attachment clis

src/deadline/job_attachments/download.py
sakshie95: Yes that's a good call and we can re-evaluate and create an object for this to carry the manifest with last modified - I'll treat this as non-blocking for this PR though unless you have concerns :)

src/deadline/job_attachments/download.py
sakshie95: Agree with both of you.. I've been confused reading the code whether this is a deadline session or boto session - I'll create a SIM to track this throughout our repo

src/deadline/job_attachments/download.py
sakshie95: Ack, discussed this offline. I would expect homogeneous workers behaviour but we should handle this for CMF

src/deadline/job_attachments/download.py
sakshie95: This doesn't need to be in the correct order bcz we sort all manifests later according to their last modified. I can add a TODO to re-evaluate sure

src/deadline/job_attachments/download.py
sakshie95: There was a different comment in an earlier revision with exporting the sorting logic to merge function so it can be re-used across multiple use-cases in the service. I'll add a ``_`` to keep it internal for now sure

src/deadline/job_attachments/api/attachment.py
sakshie95: I've re-used what's there in the current api interface for attachment api (we should eventually move the interface to be this), I can update the description to make it understandable

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: I think it would work based on what I understand but it's a good idea to park this for now marking it as internal & we can discuss :))

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: I like your suggestion. Want to give you some context for the current approach though - for the current iteration, I've based it off of an offline discussion for building a pyramid like structure as well but with a slightly different model. We have an interface for updating the output manifests for session action (``update_manifest_output_paths_for_session_actions``) as a fallback until we have manifest data from the list session actions api. Next up the pyramid is this function which the caller calls with the ``output_manifests``. It's much like how the other manifest snapshotting, download etc. work.

This is worth re-visiting I think, I've made the functions as internal as per the other suggestion so we can iterate through this process in future PRs

src/deadline/job_attachments/download.py
sakshie95: Sure I can do that

src/deadline/job_attachments/download.py
sakshie95: All of the manifest utilities are in this file currently, eg. the non sorted version of this function that exists: https://github.com/aws-deadline/deadline-cloud/blob/5fdc5bf7ce03599286d48cbbfdf6f4701724dec7/src/deadline/job_attachments/download.py#L701. I think we need an overall cleanup of this huge ``download.py`` file :))

I'll track a task with all of these refactors on github

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+    :param print_function_callback: Function for logging messages
+    :param farm_id: The farm ID
+    :param queue_id: The queue ID
+    :return: Returns None
+    Throws: AssetDownloadFailedException, when there is an exception during aggregation & download for any of the input
+    session action ids
+    """
+
+    # Get queue for assuming queue role session to access JA bucket
+    deadline = boto3_session.client("deadline")
```
leongdl, line 85: That's odd, I think my prior comment is gone.

This function seems awkward as a standalone function.

leongdl: Why not just aggregate and download as 2 separate functions? The caller can just call them back to back.

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+
+    try:
+        # Get all manifests with their last modified timestamps, grouped by asset root
+        output_manifests_list = get_output_manifest_files_by_asset_root_with_last_modified(
+            s3_settings=ja_settings,
+            output_manifest_paths=output_manifest_paths,
+            session=queue_role_session,
+        )
+
+        # Group manifests by asset root
```
leongdl, line 107: lines 107 to 134 is a very good independent function to have across worker agent, attachment download and this tool. Can it be its own utility?

src/deadline/job_attachments/incremental_downloads/asset_downloader_with_manifest_aggregation.py
leongdl: I guess so, given this file and submodule are now marked as internal.

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
leongdl: OK saw the other thread did not move. I guess its ok given its an internal module now. Just keep in mind how I was thinking about modules and building up the code in a pyramid like function. We all like building blocks lego like code patterns.

leongdl: Before we remove the `_`, good to have a walk through and then see which bits makes sense to cut out as re-usable functions.

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+from deadline.client.api._session import _get_queue_user_boto3_session
+from typing import Optional, Callable, Tuple, DefaultDict
+import boto3
+from ...job_attachments.download import (
+    get_output_manifest_files_by_asset_root_with_last_modified,
+    _merge_asset_manifests_sorted_asc_by_last_modified,
+)
+from ...job_attachments.api.attachment import _attachment_download_with_root_manifests
+from ...job_attachments.models import JobAttachmentS3Settings, FileConflictResolution
+from ..asset_manifests.base_manifest import BaseAssetManifest
```
benl-2023: should we use absolute imports instead of relative?

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class SessionAction:
```
benl-2023, line 16: I think this could be more concisely implemented with python dataclasses, and dictionary conversion would be provided out of the box with `asdict`

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+
+class AssetDownloadFailedException(Exception):
+    """
+    Asset Download Failed Exception to throw to caller if any of the aggregations & download failed
+    """
+
+
+def aggregate_manifest_and_download_outputs(
```
benl-2023, line 60: Do we have unit tests for this function? I don't see one. Given this function's name it sounds like its doing 2 distinct things. Can we split it into functions with single responsibilities? That would probs make unit testing easier as well.

src/deadline/job_attachments/download.py
```diff
@@ -85,13 +85,20 @@
...
 ) -> BaseAssetManifest:
-    _, manifest = get_asset_root_and_manifest_from_s3(manifest_key, s3_bucket, session)
+    _, _, manifest = get_asset_root_and_manifest_from_s3(manifest_key, s3_bucket, session)
     return manifest
 
 
 def get_asset_root_and_manifest_from_s3(
     manifest_key: str, s3_bucket: str, session: Optional[boto3.Session] = None
-) -> Tuple[Optional[str], BaseAssetManifest]:
+) -> Tuple[Optional[str], datetime, BaseAssetManifest]:
```
mwiebe, line 103: Since `get_asset_root_and_manifest_from_s3` is a public interface of the library, changing it is a breaking change. Probably better to keep this one the same, and introduce an internal-only `_*` function we can iterate on. You can read https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#public-contracts to see a fairly detailed description of what's in the public contract and what isn't.

src/deadline/job_attachments/download.py
```diff
@@ -741,6 +802,45 @@ def get_output_manifests_by_asset_root(
     return outputs
 
 
+def get_output_manifest_files_by_asset_root_with_last_modified(
```
mwiebe: Can you make this an internal-only function for now?

mwiebe (changes requested):

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: I'm following this from a previous comment to use relative imports instead of absolute :))

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: There are no unit tests as of now, we wanted to get the functionality out for e2e integration with the other pieces and write unit tests later. There is a different utility function already to aggregate manifests under ``download.py``. This one is a level built on top of it which is why it is combined

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: I've created this redundant class only for development purposes. It'll be a part of the other piece under development which'll be continued from https://github.com/aws-deadline/deadline-cloud/pull/726/files#diff-592fb7f096549deb33a6bf9ddc37390d1ee7f2e95f2709b5cfcfbe921b927570 so I'll make sure this comment gets applied there

benl-2023: Thanks!

benl-2023 (dismissed):

src/deadline/job_attachments/download.py
sakshie95: Discussed offline since there was conflict with another comment: https://github.com/aws-deadline/deadline-cloud/pull/706#discussion_r2194056860. We discussed it is a breaking change to change this in the public modules, so we'll stick with the earlier version

src/deadline/job_attachments/_incremental_downloads/asset_downloader_with_manifest_aggregation.py
sakshie95: Sure.. I'll incorporate this in my follow-up PR for unit tests. Will have to think on where to put it

mwiebe (approved):

leongdl (approved):

PR #708, opened 2025-06-12
feat: Adding a helper to initialize the lru_cached get_s3_client meth…

baxeaz:
Adding a helper to initialize the lru_cached get_s3_client method.  

Adding a cached get_session_client method to allow get_boto3_client calls which end up using the same session and same service name to use cached clients.

While investigating slowness at job submission time in the unreal integration it was discovered that in create_job_from_job_bundle we were creating a new deadline client each time so our subsequent call to get_queue_user_boto3_session would then create a new session, causing our S3AssetUploader to create a new s3 client.  

Additionally from the unreal side we wanted to "pre cache" the s3 client which we'd later reuse with the above changes.

Modify the get_boto3_client method which can't directly use lru_caching due to the ConfigParser parameter to internally retrieve the client from a new get_session_client option which does use lru_caching.  In the case where the same session is used to retrieve a client from the same service, the same client should be returned.

Much faster job submission in the unreal integration pre hashing (~30 seconds -> 1-2 seconds) and likely improved performance from other integrations.

All unit tests pass, new unit tests added which pass, manual testing in the unreal integration (Follow on PR to use this feature coming to the unreal integration once a new release is out)

Yes

No

No

No

----

### DIFF
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -12,7 +12,7 @@ from configparser import ConfigParser
 from contextlib import contextmanager
 from enum import Enum
 from functools import lru_cache
-from typing import Optional
+from typing import Optional, Tuple
 
 import boto3  # type: ignore[import]
 import botocore
@@ -27,6 +27,7 @@ from botocore.session import get_session as get_botocore_session
 from .. import version
 from ..config import get_setting
 from ..exceptions import DeadlineOperationError
+from ...job_attachments._aws.aws_clients import get_s3_client
 
 
 class AwsCredentialsSource(Enum):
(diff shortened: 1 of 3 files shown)


src/deadline/client/api/_session.py
```diff
@@ -280,6 +300,65 @@ def check_authentication_status(config: Optional[ConfigParser] = None) -> AwsAut
             return AwsAuthenticationStatus.CONFIGURATION_ERROR
 
 
+def initialize_queue_user_s3_client(
```
crowecawcaw: Could we make this init function work for all clients and not be specific to S3? The slow part of the process is getting the queue credentials right? It might be more generally useful and easier to understand if it were just a "preload_credentials" function. I might be off though since I don't know all the layers of caching and abstractions involved.

baxeaz: It was actually the service discovery process in botocore in this case that was eating up the majority of the time.  I could add additional client initializations here if that'd be useful in other cases, sure!  Which other clients/methods are you thinking?  This one will create the deadline and s3 client

crowecawcaw: Interesting... Do you know why the service discovery is so slow?

As an idea, what do you think about naming this `precache_clients`? `initialize` sounds like it's a required step when it's actually an optional optimization. Also it will also cache the Deadline client.

baxeaz: Haven't solved that piece of the puzzle yet but wanted to get this mitigation out as it's useful on its own and we had a customer request we investigate and solve it as they're about to have additional internal teams look at the integration.  Will rename!

leongdl (approved):

crowecawcaw (approved):

PR #711, opened 2025-06-16
feat: Remember last used priority, max retries per task, max failed t…

gianelli-lagos-aws:
Fixes: *< https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/95 >*

The fields priority, initial_state , max_failed_tasks_count, max_retries_per_task, and max_worker_count do not have sticky settings which can be inconvenient for the customer. 

* Update the refresh_ui() method to use getattr() with fallback values for backward compatibility.
* Modify the method to load saved values for all four fields when they exist in the settings object.
* Update the update_settings() method to save all four field values when they exist in the settings object.

Customers will have settings that are saved and reused between submissions.

* Since there were also changes made to the deadline-cloud-for-cinema-4d repo, both packages were built with the changes and tested together to ensure proper integration.
* Verified that all settings persist after closing and reopening Cinema 4D.
* Tested with both new scenes and existing scenes to ensure backward compatibility.
* Validated that the JSON settings file contains all the expected fields with correct values.

Yes. 
0.31s call     test/unit/deadline_adaptor_for_cinema4d/test_copyright_headers.py::test_copyright_headers
0.06s call     test/unit/deadline_submitter_for_cinema4d/test_scene.py::TestAnimation::test_get_frame_list_data_input[RDATA_FRAMESEQUENCE_CUSTOM-3,6-10:2,15-17]
0.06s setup    test/unit/deadline_submitter_for_cinema4d/test_scene.py::TestFrameRange::test_frame_range_repr[start=1, stop=None, step=7]
0.03s call     test/unit/deadline_submitter_for_cinema4d/test_submitter.py::test_get_job_template
0.03s call     test/unit/deadline_submitter_for_cinema4d/test_scene.py::test_get_output_directores
===================================================================================================================== 54 passed in 4.67s ============================================

Yes. 
417.31s call     test/integ/test_cinema4d.py::test_integ[physical_multi_takes]
91.88s call     test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
75.28s call     test/integ/test_cinema4d.py::test_integ[redshift_textured]
74.50s call     test/integ/test_cinema4d.py::test_integ[redshift]
74.04s call     test/integ/test_cinema4d.py::test_integ[redshift_takes]
=============================================================================================================== 7 passed in 837.70s (0:13:57) ==========================

----


src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -316,12 +316,12 @@ def limited_max_worker_count_radio_button_toggled(self, state):
...
 
     def refresh_ui(self, settings: Any):
-        self.sub_name_edit.setText(settings.name)
-        self.desc_edit.setText(settings.description)
-        self.initial_status_box.setCurrentText("READY")
-        self.max_failed_tasks_count_box.setValue(20)
-        self.max_retries_per_task_box.setValue(5)
-        self.priority_box.setValue(50)
+        self.sub_name_edit.setText(getattr(settings, "name", ""))
+        self.desc_edit.setText(getattr(settings, "description", ""))
```
karthikbekalp: Hmm, I think settings will always contain "name" and "description". https://github.com/aws-deadline/deadline-cloud/blob/947133f84b06ff18efb9a955b7061f745249d049/src/deadline/client/ui/widgets/shared_job_settings_tab.py#L247-L249 

So I think its best to limit to other fields only.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -316,12 +316,12 @@ def limited_max_worker_count_radio_button_toggled(self, state):
...
+        self.initial_status_box.setCurrentText(getattr(settings, "initial_state", "READY"))
+        self.max_failed_tasks_count_box.setValue(getattr(settings, "max_failed_tasks_count", 20))
+        self.max_retries_per_task_box.setValue(getattr(settings, "max_retries_per_task", 5))
+        self.priority_box.setValue(getattr(settings, "priority", 50))
```
karthikbekalp: Hmm, I wonder if we should check if the values that are sent from other integrations are in correct types? 

For example, what happens if priority that is sent is float and not int? 
Should we check if the field exists and if its the correct type before setting the values? I'll leave this question for the SMEs of deadline-cloud.

karthikbekalp (commented):
Few comments and clarifying questions. Feel free to respond to them whenever you get time.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
karthikbekalp: The fields "name" and "description" do not test for types so I believe this is Ok. But let's see what deadline-cloud SMEs say.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
gianelli-lagos-aws: Ohh I see, that's true. Thank you!

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -318,10 +318,10 @@ def limited_max_worker_count_radio_button_toggled(self, state):
...
         self.desc_edit.setText(settings.description)
-        self.initial_status_box.setCurrentText("READY")
-        self.max_failed_tasks_count_box.setValue(20)
-        self.max_retries_per_task_box.setValue(5)
-        self.priority_box.setValue(50)
+        self.initial_status_box.setCurrentText(getattr(settings, "initial_state", "READY"))
+        self.max_failed_tasks_count_box.setValue(getattr(settings, "max_failed_tasks_count", 20))
+        self.max_retries_per_task_box.setValue(getattr(settings, "max_retries_per_task", 5))
+        self.priority_box.setValue(getattr(settings, "priority", 50))
         self.unlimited_max_worker_count.setChecked(True)
```
karthikbekalp: Maybe we can add sticky settings for the other fields too?

gianelli-lagos-aws: Yes!

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -318,13 +318,15 @@ def limited_max_worker_count_radio_button_toggled(self, state):
...
-        self.max_worker_count_box.setHidden(True)
+        self.initial_status_box.setCurrentText(getattr(settings, "initial_state", "READY"))
+        self.max_failed_tasks_count_box.setValue(getattr(settings, "max_failed_tasks_count", 20))
+        self.max_retries_per_task_box.setValue(getattr(settings, "max_retries_per_task", 5))
+        self.priority_box.setValue(getattr(settings, "priority", 50))
+        is_limited = getattr(settings, "limited_max_worker_count", False)
+        self.unlimited_max_worker_count.setChecked(not is_limited)
+        self.limited_max_worker_count.setChecked(is_limited)
+        self.max_worker_count_box.setHidden(not is_limited)
+        self.max_worker_count_box.setValue(getattr(settings, "max_worker_count", 1))
```
karthikbekalp: Hmm, shouldn't we only set this value if the `max_worker_count_box` is not hidden? 

Also, can we simplify this to something like this? What do you think? 

```
has_max_worker_count = hasattr(settings, "max_worker_count")
self.unlimited_max_worker_count.setChecked(not has_max_worker_count)
self.limited_max_worker_count.setChecked(has_max_worker_count)
self.max_worker_count_box.setHidden(not has_max_worker_count)
if has_max_worker_count:
    self.max_worker_count_box.setValue(settings.max_worker_count)
```

karthikbekalp: Discussed this offline. Instead of checking if the attribute exists or not, we can make it even simpler by just checking if the value is greater than 0. If it is, then we set the value and update the settings and the boxes. 

```
has_max_worker_count = getattr(settings, "max_worker_count", -1) > 0
self.unlimited_max_worker_count.setChecked(not has_max_worker_count)
self.limited_max_worker_count.setChecked(has_max_worker_count)
self.max_worker_count_box.setHidden(not has_max_worker_count)
if has_max_worker_count:
    self.max_worker_count_box.setValue(settings.max_worker_count)
```

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -411,6 +413,12 @@ def update_settings(self, settings):
         """
         settings.name = self.sub_name_edit.text()
         settings.description = self.desc_edit.text()
+        settings.initial_state = self.initial_status_box.currentText()
+        settings.max_failed_tasks_count = self.max_failed_tasks_count_box.value()
+        settings.max_retries_per_task = self.max_retries_per_task_box.value()
+        settings.priority = self.priority_box.value()
+        settings.limited_max_worker_count = self.limited_max_worker_count.isChecked()
+        settings.max_worker_count = self.max_worker_count_box.value()
```
karthikbekalp: Hmm, what if the settings doesn't have this field? Shouldn't we check if these fields exist before we set the values? 

For example, other DCCs may not have implemented a dataclass with `initial_state` and other fields. Wouldn't this break them?

karthikbekalp: Hmm, should we be setting the value of `max_worker_count` if the box is hidden? 
Just curious, but what is the value when the box is hidden, is it also -1?

gianelli-lagos-aws: I agree, I can check if the attributes exist beforehand to avoid any issues. Ty!

gianelli-lagos-aws: I checked and when the box was hidden, it maintained the previously used number. The json file however did have -1. I think the max_worker_count can be set to -1 when the box is hidden when updating the settings so that it can be more consistent. I will add that to the code and since this has been addressed in the file, I can remove the handling of max_worker_count in the cinema4d repo.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -412,6 +415,25 @@ def update_settings(self, settings):
...
+        for attr_name, getter_func in SETTINGS_FIELDS:
+            if hasattr(settings, attr_name):
+                if attr_name == "max_worker_count":
+                    if self.unlimited_max_worker_count.isChecked():
+                        setattr(settings, attr_name, -1)
+                    else:
+                        setattr(settings, attr_name, getter_func())
+                else:
+                    setattr(settings, attr_name, getter_func())
+
```
karthikbekalp: nit: Hmm, do you think we can keep it simpler with just the if statements? 

For example, with code like this: 

```
        # Set other fields only if they exist in the settings object
        if hasattr(settings, "initial_state"):
            settings.initial_state = self.initial_status_box.currentText()
            
        if hasattr(settings, "max_failed_tasks_count"):
            settings.max_failed_tasks_count = self.max_failed_tasks_count_box.value()
            
        if hasattr(settings, "max_retries_per_task"):
            settings.max_retries_per_task = self.max_retries_per_task_box.value()
            
        if hasattr(settings, "priority"):
            settings.priority = self.priority_box.value()
            
        if hasattr(settings, "limited_max_worker_count"):
            settings.limited_max_worker_count = self.limited_max_worker_count.isChecked()
            
        if hasattr(settings, "max_worker_count"):
            if self.unlimited_max_worker_count.isChecked():
                settings.max_worker_count = -1 # -1 denotes no max worker count limits. 
            else:
                settings.max_worker_count = self.max_worker_count_box.value()
```
There are pros and cons to this approach.  For example, this code is more explicit as its more straightforward. 
But the cons if that If new fields are added, then this will become longer instead of just adding a line to the dict.  (Interestingly, even being explicit the line count is less than the dict approach)

What do you think?

gianelli-lagos-aws: Yes, I noticed that with the if statements the code can become quite lengthy especially if more fields are added. This is why i decided to use the dict since the for loop can reduce the length. I am not sure if having many if statements would be better.

karthikbekalp (dismissed):
Looks good to me. Let's wait for deadline-cloud SMEs.

prestomation:
The use of getattr/setattr seems quite fragile to me. Should this be full new settings in config_file.py? We lose all our type safety and introspection and useful defaults in non-UI cases among other things by doing this in the UI layer like this

src/deadline/client/ui/widgets/shared_job_settings_tab.py
miabatta: Talked a bit on Slack about this:
> My iniiiitial thought was that not every integrated submitter may respect those type definitions- they should, it's just something we'd have to smoke test a bit. Or ask individual DCC owners if it'll break their submitters :)

It's without a doubt a good idea to make sure all integrations are handling parameter types correctly, so what matters most here is probably the order of any necessary changes s.t. nothing breaking is introduced by this change.

miabatta:
> The use of getattr/setattr seems quite fragile to me. Should this be full new settings in config_file.py? We lose all our type safety and introspection and useful defaults in non-UI cases among other things by doing this in the UI layer like this

Is `config_file.py` the best place? A while ago we discussed extracting sticky settings from per-DCC implementation to centralized in this repo... Might not be the scope for this change/PR, but to me there's value in keeping config separate from sticky values.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -318,13 +318,16 @@ def limited_max_worker_count_radio_button_toggled(self, state):
...
         self.sub_name_edit.setText(settings.name)
         self.desc_edit.setText(settings.description)
-        self.initial_status_box.setCurrentText("READY")
-        self.max_failed_tasks_count_box.setValue(20)
-        self.max_retries_per_task_box.setValue(5)
-        self.priority_box.setValue(50)
-        self.unlimited_max_worker_count.setChecked(True)
-        self.limited_max_worker_count.setChecked(False)
-        self.max_worker_count_box.setHidden(True)
+        self.initial_status_box.setCurrentText(getattr(settings, "initial_state", "READY"))
```
epmog: What's the underlying API call for this? The label for the parameter is `initial state`, but we have `initial_status` as the property.

My guess is that the property and label should probably be `initial_status`: https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_CreateJob.html#deadlinecloud-CreateJob-request-targetTaskRunStatus

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -412,6 +415,25 @@ def update_settings(self, settings):
...
+
+        for attr_name, getter_func in SETTINGS_FIELDS:
+            if hasattr(settings, attr_name):
+                if attr_name == "max_worker_count":
+                    if self.unlimited_max_worker_count.isChecked():
+                        setattr(settings, attr_name, -1)  # -1 denotes no max worker count limits.
+                    else:
+                        setattr(settings, attr_name, getter_func())
+                else:
+                    setattr(settings, attr_name, getter_func())
```
epmog: TL;DR - To merge this:
1. I think if we change the `initial_state` to `initial_status` and 
2. update the comment for this code block to better describe why we even have this check in the first place, since it is indeed fragile.

------

I don't disagree with @prestomation that this `hasattr`/`getattr` is more fragile than we'd perhaps like. But there's some interesting bits about the history of the sticky setting code that don't make it easy to actually address/fix that issue here. Additionally sticky settings themselves are currently just a UI construct, "populate the gui with these settings next time based on some anchor value (ie. scene file)", they do not exist as a non-ui component to a workflow.

Deadline-cloud itself has a settings dataclass where name/description are defined (and where these new attributes are not defined), however that doesn't even talk about sticky settings. And in fact, each DCC ends up defining their own settings dataclass which does NOT inherit from this deadline-cloud one. This makes it hard to actually offer any reasonable safety measures on what is and is not available on the object passed in, and what deadline-cloud actually does with it.

Sooner rather than later we should probably plan out how we can actually define the right behaviour here, since I know there's a desire for sticky settings in other use-cases. Like gui-submit anchored on specific parameters, or even something like queue settings for the default values.

epmog (changes requested):

karthikbekalp (dismissed):

src/deadline/client/ui/widgets/shared_job_settings_tab.py
epmog: At the very least, we should ensure that this code can't break an already existing submitter. So if it is the wrong type, or rather if the set fails, then we should fall back to previous default

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -315,16 +315,51 @@ def limited_max_worker_count_radio_button_toggled(self, state):
         """
         self.max_worker_count_box.setHidden(not state)
 
+    def _has_compatible_attr(self, obj, attr_name, expected_type):
+        """
+        Determine if attribute exists and if the type is correct.
+        """
+        # Users can have anything in the settings object since they define their own dataclass to pass in.
```
karthikbekalp: I think it should say DCCs instead of Users. The settings object that is passed along is defined by the teams that are owned by specific DCCs.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -315,16 +315,51 @@ def limited_max_worker_count_radio_button_toggled(self, state):
...
+        # Changing what we look for below may cause breaking changes in usage of this library.
+        if hasattr(obj, attr_name) and isinstance(getattr(obj, attr_name), expected_type):
```
karthikbekalp: nit: I believe we can simplify this function to
```
return isinstance(getattr(obj, attr_name, None), expected_type):
```

What do you think?

karthikbekalp (changes requested):
Just few nits and change to the comment. But on the whole looks good! I think we are close on getting this in.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
gianelli-lagos-aws: Yep! I agree, I will change that.

src/deadline/client/ui/widgets/shared_job_settings_tab.py
gianelli-lagos-aws: I agree! This is much better.

karthikbekalp (dismissed):

src/deadline/client/ui/widgets/shared_job_settings_tab.py
gianelli-lagos-aws: Yes, I agree that it should be initial_status. I have changed it!

src/deadline/client/ui/widgets/shared_job_settings_tab.py
```diff
@@ -409,9 +441,46 @@ def update_settings(self, settings):
...
+        # TODO: Extract sticky settings from per-DCC implementation to centralized.
         settings.name = self.sub_name_edit.text()
         settings.description = self.desc_edit.text()
 
+        # Set all fields with type checking
+        settings.initial_status = (
+            self.initial_status_box.currentText()
+            if self._has_compatible_attr(settings, "initial_status", str)
+            else "READY"
+        )
```
epmog: Do we know if this is right? We're type checking the value on the settings object, and if it doesn't exist we set to "READY". But also if it exists with a different type, we overwrite it?

When does this function actually get called by the DCC?

gianelli-lagos-aws: In the submitter, initial_status has 2 options "READY" and "SUSPENDED" , customers can chose which status to set their job to. The default should be "READY". If the customer chose "SUSPENDED" then it get overwritten to preserve the sticky setting.

karthikbekalp: Hmm, I think @epmog was referring to the concern of overwriting the field when the type does not match. 

I agree on this, I think its better to keep it simple like this so that we don't overwrite: 

```
if self._has_compatible_attr(settings, "initial_status", str):
    settings.intial_status = self.initial_status_box.currentText()
```

gianelli-lagos-aws: Yep! I agree as well. It is best not to overwrite the settings a DCC might have with those defaults. The defaults should only be used at the beginning in refresh_ui when the settings object has no values yet.

karthikbekalp (approved):

epmog (approved):

PR #717, opened 2025-06-23
feat(experimental): Add client library API to list recently updated jobs

mwiebe:

For the incremental downloads CLI command in development, we need a robust way to get
all jobs that are new, or have had tasks requeued, since a specified timestamp.

* Added experimental candidate API function in deadline.client.api._list_jobs_recent_by_timestamp_field
  * It returns all jobs in a queue whose selected timestamp field is >= a provided timestamp, or since a provided timedelta ago.
  * The name starts with list_jobs_, so that someone using tab-complete or browsing the list of functions sees it alphabetically adjacent to the straight list_jobs function.
* The purpose of this function is to support the new incremental downloads CLI command. We expect it to be a useful primitive, but are marking it internal-only for now.

We can build the rest of the CLI command using this function.

Added unit tests that cover the function by generating random fake job datasets, and a mocked SearchJobs implementation that implements the subset of that API used by the function.

Tested the function on a Deadline Cloud queue.

No.

No.

No.

No.

----

### DIFF
--- a/src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
+++ b/src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
@@ -1,0 +1,212 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from __future__ import annotations
+
+__all__ = ["_list_jobs_recent_by_timestamp_field"]
+
+from typing import Any, Optional
+import boto3
+
+from deadline.client.api._session import get_default_client_config
+from botocore.exceptions import ClientError
+from datetime import datetime, timedelta, timezone
+from deadline.client.exceptions import DeadlineOperationError
+
+
+class JobFetchFailure(RuntimeError):
+    """
+    Failure fetching jobs updated since timestamp
+    """
+
+
+def _list_jobs_recent_by_timestamp_field(
+    boto3_session: boto3.Session,
+    farm_id: str,
+    queue_id: str,
+    timestamp_field_name: str,
+    *,
+    timestamp: Optional[datetime] = None,
+    look_back_duration: Optional[timedelta] = None,
+) -> list[dict[str, Any]]:
+    """
+    This function retrieves jobs that have a value for `timestamp_field_name` timestamp equal or newer than
+    the provided timestamp. It returns a list of jobs in the form returned by the deadline:SearchJobs API.
+
+    CAUTION:
+        Eventual consistency in the deadline:SearchJobs API means that the result set can be missing jobs
+        with a timestamp close to the current time. If you are repeatedly calling this function to collect
+        all jobs over time, use a sufficiently large overlap window to account for this eventual consistency.
+
+    TODO: This is an experimental function under development, and is exposed under the internal-only name
+          deadline.client.api._list_jobs_recent_by_timestamp_field._list_jobs_recent_by_timestamp_field.
+          If it proves useful, deadline.client.api.list_jobs_recent_by_timestamp_field is the likely public function.
+
+    NOTE:
+        There's an edge case where 100 jobs with identical timestamps will cause the function to raise
+        a JobFetchFailure exception. Because this would require all 100 jobs be created/updated with
+        identical timestamp recorded at millisecond precision, we do not expect this to occur in practice.
+
+    Example:
+        boto3_session = boto3.Session()
+        farm_id = ...
+        queue_id = ...
+        saved_timestamp = ...
+
+        # Get the jobs created in the last 5 minutes
+        since_5minutes = _list_jobs_recent_by_timestamp_field(
+            boto3_session,
+            farm_id,
+            queue_id,
+            "createdAt",
+            look_back_duration=timedelta(minutes=5)
+        )
+
+        # Continuously get jobs created, querying every 5 minutes
+        job_ids = set()
+        timestamp = <starting_timestamp>
+        while True:
+            # Determine the next threshold timestamp before calling the function
+            # using a 90 second overlap to account for eventual consistency.
+            next_timestamp = datetime.now(timezone.utc) - timedelta(seconds=90)
+
+            jobs = _list_jobs_recent_by_timestamp_field(
+                boto3_session,
+                farm_id,
+                queue_id,
+                "endedAt",
+                timestamp=timestamp
+            )
+            # Jobs can be seen more than once, so add them to a set
+            job_ids.update(job["jobId"] for job in jobs)
+
+            # Sleep for 5 minutes
+            time.sleep(5*60)
+
+
+    Args:
+      boto3_session (boto3.Session): The boto3 Session for AWS API access.
+      farm_id (str): The Farm ID.
+      queue_id (str): The Queue ID.
+      timestamp_field_name (str): The name of the job timestamp field to use. Can be
+            "createdAt", "endedAt", "startedAt", or "updatedAt". See
+            https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_GetJob.html#API_GetJob_ResponseSyntax
+      timestamp (Optional[datetime]): The timestamp from which to retrieve jobs. Either timestamp
+            or look_back_duration must be provided.
+      look_back_duration (Optional[timedelta]): A timedelta subtracted from the current time to
+            get the timestamp from which to retrieve jobs. Either timestamp or look_back_duration
+            must be provided.
+
+    Returns:
+      The list of all jobs in the queue whose timestamp field is equal or greater than the provided
+      timestamp. Each job is as returned by the deadline:SearchJobs API.
+    """
+    # This function gets the full set of jobs whose timestamp field is newer than the threshold timestamp,
+    # except potentially jobs whose timestamp is near datetime.now(), due to eventual consistency.
+    # It uses deadline:SearchJobs as a primitive, unioning subsets of up to 100 jobs at a time,
+    # designed to guarantee we end up with the full correct set.
+    #
+    # Let J = {j ∈ queue | j["timestamp_field_name"] >= threshold_timestamp} be the set we want
+    #     J' = union {J_i | i ∈ range(search_count)} be the set the algorithm produces,
+    #            where J_i = {j ∈ queue | j["timestamp_field_name"] >= timestamp_i} [limit 100, ordered by timestamp_field_name asc]
+    #                  timestamp_0 = input_timestamp
+    #                  timestamp_i = max {j["timestamp_field_name"]: j in J_{i-1}}
+    #
+    # To prove these sets are equal, we show they are each a subset of the other.
+    #
+    # J' is a subset of J, because each J_i is from a deadline:SearchJobs that filters to a subset of J.
+    #
+    # Consider a job j ∈ J. A property of the system is that if the timestamp field is updated, it will be updated to the current time.
+    #    1. If the timestamp field of j is not updated during the algorithm, and is visible to deadline:SearchJobs, it will be
+    #       in one of the sets J_i because by construction the union of the query time intervals
+    #       cover the time interval for J. If it is not visible to deadline:SearchJobs, its timestamp is in the eventual
+    #       consistency window close to datetime.now().
+    #    2. If j has the timestamp field created or modified during the algorithm, its timestamp will be set to the current time.
+    #       and depending on how quickly the deadline:SearchJobs API updates, it may appear in a later set J_n. Eventual
+    #       consistency of deadline:SearchJobs means it may not end up in J.
+    #
+    # These are all the possibilities, so therefore j is in J' and J is a subset of J' except for some jobs whose timestamps
+    # are within the eventual consistency window close to datetime.now().
+
+    # This holds {job_id: job_from_search_jobs_call, ...}
+    result_jobs = {}
+
+    # Initialize threshold timestamp with the input timestamp/timedelta
+    if timestamp is not None and look_back_duration is not None:
+        raise ValueError("Only one of timestamp and look_back_duration may be provided.")
+    elif timestamp is not None:
+        if timestamp.tzinfo is None:
+            raise ValueError(f"The input timestamp {timestamp} is missing a required time zone.")
+        threshold_timestamp = timestamp
+    elif look_back_duration is not None:
+        threshold_timestamp = datetime.now(timezone.utc) - look_back_duration
+    else:
+        raise ValueError("One of timestamp or look_back_duration must be provided.")
+
+    permitted_field_names = ["createdAt", "endedAt", "startedAt", "updatedAt"]
+    if timestamp_field_name not in permitted_field_names:
+        raise ValueError(
+            f"The provided timestamp field name {timestamp_field_name!r} is not one of {permitted_field_names}"
+        )
+
+    deadline = boto3_session.client("deadline", config=get_default_client_config())
+
+    # The endedAt field is called ENDED_AT in sort and filter expressions
+    expression_field_name = timestamp_field_name.replace("At", "_at").upper()
+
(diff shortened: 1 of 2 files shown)


src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,177 @@
...
+            or look_back_duration must be provided.
+      look_back_duration (Optional[timedelta]): A timedelta subtracted from the current time to
+            get the timestamp from which to retrieve jobs. Either timestamp or look_back_duration
+            must be provided.
+
+    Returns:
+      The list of all jobs in the queue whose timestamp field is equal or greater than the provided
+      timestamp. Each job is as returned by the deadline:SearchJobs API.
+    """
+    # This function gets the full set of jobs whose timestamp field is newer than the thresho9ld timestamp.
```
leongdl: `thresho9ld` :)

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,177 @@
...
+            "createdAt",
+            look_back_duration=timedelta(minutes=5)
+      )
+
+    Args:
+      boto3_session (boto3.Session): The boto3 Session for AWS API access.
+      farm_id (str): The Farm ID.
+      queue_id (str): The Queue ID.
+      timestamp_field_name (str): The name of the job timestamp field to use. Can be
+            "createdAt", "endedAt", "startedAt", or "updatedAt". See
```
leongdl, line 91: Are we still considering `updatedAt` considering the corner cases we found?

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,177 @@
...
+            queue_id,
+            "createdAt",
+            look_back_duration=timedelta(minutes=5)
+      )
+
+    Args:
+      boto3_session (boto3.Session): The boto3 Session for AWS API access.
+      farm_id (str): The Farm ID.
+      queue_id (str): The Queue ID.
+      timestamp_field_name (str): The name of the job timestamp field to use. Can be
```
leongdl, line 90: We assume `timestamp_field_name` to be correct, do we want to guard rail this function with the enum set of `createdAt`, `endedAt`, `startedAt`, `updatedAt`, especially if we are going to make this a proper API?

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,177 @@
...
+            # Rare edge case where the timestamp field is the same for all 100 jobs in the page, that
+            # we expect to never see in practice. The timestamp value is stored with
+            # millisecond precision, and jobs are scheduled independently from each other,
+            # with updates of running jobs generally being multiple seconds apart.
+            raise JobFetchFailure(
+                f"Failure fetching recent jobs based on the {timestamp_field_name} field as more then 100 jobs have the exact same timestamp value."
+            )
+        else:
+            # Continue processing from the largest timestamp value we saw so far
+            threshold_timestamp = jobs[-1][timestamp_field_name]
```
leongdl, line 210: I think there's a case to considering here I was thinking over.

Consider a farm with 99 jobs already existing. We can call them [ j1 .... j99 ]. Now consider the user submits 2 jobs, j100, j101 in this order with sequential created at. And pretend the job number is also T=n

If this tool runs when the following sets are eventually consistent we may have a corner case if `j101` is indexed or returns earlier than `j100`.  `[ j1.... j99,  j101 ]`, leading to 1 page of data, meeting the `len(jobs) == total_results` exit condition. 

The next time this tool runs, and all 101 jobs are now indexed, we have a condition to consider. Would `threshold_timestamp` be set to the prior value of `j101` ? If yes, then the `filter_expression` will no longer include `j100`

Will we need to do an overlap algorithm? sort of like a time based ladder?

More so thinking of correctness if we lost 1 job that we did not download.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: From the perspective of this primitive function, yes. The corner cases were about how we tried to apply the updatedAt timestamp.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: To confirm what you're describing, do you mean the following?

What do we do if the assumption about system behavior in this part of the proof is wrong:

2. If j has the timestamp field created or modified during the algorithm, its value will be set to the current time
      and it will appear in a later set J_n, which is the last one at the time of create/update, but may not be
      the last time when the algorithm completes.

In particular, the trouble is when jobs get added/updated in the SearchJobs materialized view in a different order compared to the timestamp.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,183 @@
...
+    Failure fetching jobs updated since timestamp
+    """
+
+
+def _list_jobs_recent_by_timestamp_field(
+    boto3_session: boto3.Session,
+    farm_id: str,
+    queue_id: str,
+    timestamp_field_name: str,
+    *,
```
sakshie95, line 27: Why do we need this?

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,183 @@
...
+    timestamp: Optional[datetime] = None,
+    look_back_duration: Optional[timedelta] = None,
```
sakshie95, line 29: Thanks for incorporating this over the union :) It's crystal clear to use as a library now

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,183 @@
...
+) -> list[dict[str, Any]]:
```
sakshie95, line 30: Should we model this into a ``Job`` class instead to make this strongly typed?

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,183 @@
...
+    #
+    # To prove these sets are equal, we show they are each a subset of the other.
+    #
+    # J' is a subset of J, because each J_i is from a deadline:SearchJobs that filters to a subset of J.
+    #
+    # Consider a job j ∈ J. A property of the system is that if the timestamp field is updated, it will be updated to the current time.
+    #    1. If the timestamp field of j is not updated during the algorithm, it will be
+    #       in one of the sets J_i because by construction the union of the query time intervals
+    #       cover the time interval for J.
+    #    2. If j has the timestamp field created or modified during the algorithm, its value will be set to the current time
```
sakshie95: This actually depends on the field they choose in the ``timestamp_field_name``. For eg. a created timestamp is never modified.. an updated timestamp is currently never modified except when the customer calls ``update_job``. Should we limit this algorithm to permit the timestamp fields that make a 100% sense for this algorithm to work as expected instead of the entire set of supported timestamps on deadline? i.e the variable ``permitted_field_names`` below

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,183 @@
...
+        # Some jobs may not have the requested time stamp field.
+        jobs = [job for job in jobs if timestamp_field_name in job]
+
+        result_jobs.update({job["jobId"]: job for job in jobs})
+
+        if len(jobs) == total_results:
+            # If the jobs we got are the total results, result_jobs is now the full set
+            break
+        elif jobs[0][timestamp_field_name] == jobs[-1][timestamp_field_name]:
+            # Rare edge case where the timestamp field is the same for all 100 jobs in the page, that
```
sakshie95, line 201: Maybe this should be in the library's docstring instead given that it completely fails the algorithm - so it's a callout when someone tries to use this library for their use-case

test/unit/deadline_client/api/test_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,372 @@
...
+    def _fake_search_jobs(
+        farmId, queueIds, itemOffset, pageSize=100, filterExpressions={}, sortExpressions=[]
+    ):
+        # Make deep copies of these parameters so we can destructively assert about them
+        filterExpressions = deepcopy(filterExpressions)
+        sortExpressions = deepcopy(sortExpressions)
+
+        assert farmId == farmIdForJobs
+        assert queueIdForJobs in queueIds
+        # We're only implementing ascending order by updatedAt
```
sakshie95: Is this accurate? We've learnt updatedAt isn't updated for operations other than customer actions on resources so I'm not sure how this would help?

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
```diff
@@ -0,0 +1,183 @@
...
+            # we expect to never see in practice. The timestamp value is stored with
+            # millisecond precision, and jobs are scheduled independently from each other,
+            # with updates of running jobs generally being multiple seconds apart.
+            raise JobFetchFailure(
+                f"Failure fetching recent jobs based on the {timestamp_field_name} field as more then 100 jobs have the exact same timestamp value."
```
benl-2023, line 206: Could we shrink the probability we hit this edge case by increasing the pageSize used in the search call from the default to say 1000?

sakshie95:
SonarCloud analysis is pretty extra strict about duplicated lines even in tests :) You might wanna make some param lists common in the test functions looking at the report https://sonarcloud.io/component_measures?id=aws-deadline_deadline-cloud&pullRequest=717&metric=new_duplicated_lines_density&view=list

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
leongdl: Sure - more of a consumer question if it gets raised. Maybe a site note to add of the behaviour where it works, or where it does not.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
leongdl: > In particular, the trouble is when jobs get added/updated in the SearchJobs materialized view in a different order compared to the timestamp.

Effectively yes. Given the nature of eventual consistency and of search (and DDB), it is likely in a sequence of jobs items will be returned out of order, and therefore exiting the algorithm. The timestamp of each job is not controlled by this algorithm, and regardless of whether it is updated, or created during the algorithm the assumption is the result set can be unstable.

test/unit/deadline_client/api/test_list_jobs_recent_by_timestamp_field.py
mwiebe: Yep, it's a stale comment. Updating it to reflect that this fake function does ascending order by any of the timestamp fields.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: This makes `timestamp` and `look_back_duration` into keyword-only arguments. Because you can't specify both, it's awkward to try using them as positional. See https://peps.python.org/pep-3102/ about keyword-only arguments in general (this PEP is nice and short).

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: Thanks for the input!

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: From my analysis in the comment, the probability of this is already vanishingly small. It would be great to do some stress tests to double check that theory, since there could be factors I didn't account for in that.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: Yeah, that makes sense. I don't expect this to ever happen in practice, so I would put it as a footnote there.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: I don't believe we need to change anything because the logic still applies to the fields that are never modified. In those cases we could simplify the statement to:

> If j has the timestamp field created during the algorithm, ...

By keeping the statement more generally, the reasoning applies to all the timestamps.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: The reason I avoided that is because the functions in this api namespace are structured to be like boto3 but with a little bit more convenience. We don't have a full wrapper of the Deadline Cloud APIs. If we modeled it as a `Job` class, then code that uses these functions and boto3 directly won't be as consistent.

mwiebe:
> SonarCloud analysis is pretty extra strict about duplicated lines even in tests :) You might wanna make some param lists common in the test functions looking at the report https://sonarcloud.io/component_measures?id=aws-deadline_deadline-cloud&pullRequest=717&metric=new_duplicated_lines_density&view=list

Yeah SonarCloud is picky! We haven't marked it as required in github, so we can apply our judgment about how much to satisfy what it suggests. I try to think through "is this or that more readable" rather than worry specifically about duplication. Often too much duplicated code is unreadable, so it does reduce that, but not always. Sometimes duplicated code is better, because deduplication would move the lines of code you need to see far apart in the file. I'll see if I find a way to remove that duplication but not harm readability.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: We need to update https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_GetJob.html#deadlinecloud-GetJob-response-updatedAt to be more clear.

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: Following some higher bandwidth conversation, I've updated the docstring and the reasoning about correctness to include eventual consistency. What I've tried to do is document clearly what this function provides, and provide a generous overlap window in an example of 90 seconds, so that someone using it can reason appropriately.

leongdl (approved):
Thank you for the deep dive into the correctness of this function!

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
sakshie95: interesting.. I would think we should have a full wrapper of the deadline cloud apis in that case for consistency but that's probably out of the scope of this PR

sakshie95 (approved):

src/deadline/client/api/_list_jobs_recent_by_timestamp_field.py
mwiebe: Definitely something to consider! When I set this up, I decided against the idea because boto3 is the standard way to access AWS APIs from Python, and I tried to find the least-effort way to provide some patterns around using the APIs when logged in from Deadline Cloud Monitor vs with other AWS credentials.

PR #718, opened 2025-06-25
feat(experimental): Refine the IncrementalDownloadState class

mwiebe:

Want to document the logic in the IncrementalDownloadState object, including things like invariants, pre-conditions, post-conditions. This is a partial step towards that.

This commit takes some parts from several in-progress PRs, and includes some further docstring refinement and code polishing.

* Modify the properties of the IncrementalDownloadState, and expand the docstring with information about the state and how it evolves.
  * Add downloads_started_timestamp to hold the bootstrap time of the incremental download operation.
  * Rename last_lookback_time to downloads_completed_timestamp and have it represent the end of the interval over which downloads are completed.
  * Add eventual_consistency_max_duration property, so that the checkpoint file is more self-documenting.
* Add an IncrementalDownloadJob class to represent the job within the incremental download state.
* Move _pid_utils.py into the cli module and rename it to
      _pid_file_lock.py. PID file locking is functionality for CLI commands
      to use.
* Refactor the PID locking tests to use lock files in a temporary
      directory.

Progress on the incremental downloads command.

Updated and ran the unit tests.

Includes docstring updates.

No

No, changes are to internal details prefixed with `_`.

No

----


src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -329,36 +333,33 @@ def incremental_output_download(
             # 2. Check if a download is already ongoing with pid lock checking mechanism
             _pid_utils.try_acquire_pid_lock(pid_file_full_path, logger.echo)
 
-            current_download_progress: IncrementalDownloadState = IncrementalDownloadState()
+            current_download_state: IncrementalDownloadState
 
-            # 3. If bootstrap is required, then bootstrap using helper function
```
leongdl, line 334: If we are getting rid of these comments we can remove 1 and 2 too ?

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -329,36 +333,33 @@ def incremental_output_download(
...
-            updated_download_progress: IncrementalDownloadState = (
-                _queue_apis._incremental_output_download(
-                    boto3_session=boto3_session,
-                    farm_id=farm_id,
-                    queue_id=queue_id,
-                    download_state=current_download_progress,
-                    path_mapping_rules=path_mapping_rules,
-                    print_function_callback=logger.echo,
-                )
+            updated_download_state = _incremental_output_download(
```
leongdl: Question, no mypy for?

Yeah VS code knows but just something we all do for readability.

leongdl (commented):
Easy PR, questions only.

src/deadline/client/cli/_groups/queue_group.py
mwiebe: Sure! I left those alone because we had planned some polishing of the code in _pid_utils.py. Might as well do some of that in this PR then.

src/deadline/client/cli/_groups/queue_group.py
mwiebe: Good call, I'll fix that up.

src/deadline/client/cli/_groups/queue_group.py
mwiebe: I've refactored that into a context manager now.

leongdl (dismissed):
LGTM! Looks like mostly refactoring and moving code around.

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -266,7 +267,7 @@ def queue_get(**args):
     def incremental_output_download(
         path_mapping_rules: str,
         json: bool,
-        bootstrap_lookback_in_minutes: Optional[int],
+        bootstrap_lookback_in_minutes: float,
```
sakshie95, line 270: Is there a use-case for this to be float not int?

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -290,15 +291,17 @@ def incremental_output_download(
 
         logger.echo("Started incremental download....")
 
-        try:
-            # Validate file path inputs for downloading outputs incrementally.
-            _queue_apis._validate_file_inputs_for_incremental_output_download(
-                saved_progress_checkpoint_location=saved_progress_checkpoint_location,
-                path_mapping_rules=path_mapping_rules,
+        # Check if download progress location is a valid directory on the os
```
sakshie95, line 294: Interesting.. I wanted to keep the validations separate from the cli code and available as a library to whichever caller wants to use the incremental download api (under ``_queue_apis.py``) to make the validations first. Is there a reason this needs to be here?

src/deadline/client/cli/_incremental_download.py
```diff
@@ -41,41 +39,9 @@ def _incremental_output_download(
     # Download outputs for ongoing jobs using current download progress
     # Right now it is set to no change in progress except setting the last lookback time to now
```
sakshie95: I think this comment needs to be updated as per the implementation change to subtract the ``eventual_consistency_max_duration`` from now?

src/deadline/client/cli/_pid_file_lock.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+    else:
+        # Process pid from file is different from current process pid.
+        logger.warning(
+            f"Another process with pid {lock_holder_pid} claimed the pid lock {pid_file_path} while {current_process_id} was holding it. Skipping pid file deletion."
+        )
+
+
+@contextmanager
+def PidFileLock(
```
sakshie95, line 177: I think we should have a docstring here cz now we've delegated the whole responsibility to this and all the caller needs to do is call it

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -314,64 +317,46 @@ def incremental_output_download(
...
-            _pid_utils.try_acquire_pid_lock(pid_file_full_path, logger.echo)
-
-            current_download_progress: IncrementalDownloadState = IncrementalDownloadState()
-
-            # 3. If bootstrap is required, then bootstrap using helper function
-            if force_bootstrap or not os.path.exists(saved_progress_checkpoint_full_path):
-                current_download_progress = IncrementalDownloadState.from_bootstrap(
-                    bootstrap_lookback_in_minutes,
-                    logger.echo,
+        with PidFileLock(
```
sakshie95, line 330: Interesting implementation :)

src/deadline/client/cli/_pid_file_lock.py
```diff
@@ -0,0 +1,149 @@
...
+        os.remove(tmp_file_name)
+        # In this case, another process jumped in and got the lock between the earlier check
+        # and when we tried to claim it.
+        raise PidLockAlreadyHeld(
+            f"Unable to perform {operation_name} as process with pid {lock_holder_pid} already holds the lock {pid_file_path}"
+        ) from None
+
+
+def _release_pid_lock(pid_file_path: str, print_function_callback: Callable[[str], None] = print):
+    """
```
sakshie95, line 131: I think we should also make sure we delete the tmp file for the process id when we release the pid lock if it exists - since this is executed in the ``finally`` block and in case of an issue in that flow (before we make the atomic write/replace to shared pid file lock) -  we wanna make sure cleanup is done for any files created.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -1,13 +1,82 @@
...
+    A new download becomes available whenever a TASK_RUN session action completes. The state
+    includes some informational fields that are not strictly necessary, to help make the data
+    on disk easier to understand on inspection.
+
+    * https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_GetSessionAction.html#API_GetSessionAction_ResponseSyntax
+    * https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_SessionActionDefinition.html
+
+    We track state at three levels, and use the resource state at one level to prune queries at lower levels when we can:
+
+    1. Job - The jobs list contains every job that entered an active status within the time interval [downloads_started_timestamp, downloads_completed_timestamp],
```
sakshie95, line 71: Maybe for a future PR but I think we need to define exit criteria for the ``session`` and ``sessionactions`` and then ``jobs`` too from the ``IncrementalDownloadState`` to make this crystal clear. I would expect exit criteria to be linked heir-archly so jobs are evicted when we've downloaded all session outputs on it and it has ENDED for example. And a job gets added back to this state if its ``ended_at`` changes (re-queued tasks, re-queued jobs)

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -43,17 +112,44 @@ class IncrementalDownloadState:
...
+        "eventualConsistencyMaxSeconds",
+        "jobs",
+    ]
 
-    def __init__(self, last_lookback_time=None, jobs=None):
+    downloads_started_timestamp: datetime
+    """The timestamp of when the download state was bootstrapped."""
+    downloads_completed_timestamp: Optional[datetime]
+    """The timestamp up to which we are confident downloads are complete."""
+    eventual_consistency_max_duration: timedelta = timedelta(seconds=120)
```
sakshie95, line 126: are these 2 minutes of eventual consistency somewhere from our SLA?

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -117,68 +210,52 @@ def from_file(
...
         Save the current download progress to a state file atomically.
 
-        :param saved_progress_checkpoint_full_path: Absolute path of file with saved progress
+        :param file_path: Where to save the file.
         :param print_function_callback: Callback to print messages produced in this function.
                 Used in the CLI to print to stdout using click.echo. By default, ignores messages
-        :return: None if save was successful,
-        or throws an exception if we're unable to save progress file to download location.
-            Hard fail for customer's assets to be downloaded EXACTLY once.
+        :return: None if save was successful, throws an exception if we're unable to save progress file to download location.
```
sakshie95, line 232: Is this accurate still? I don't see the exception being thrown anymore in the code below?

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -117,68 +210,52 @@ def from_file(
...
-                f"Successfully saved state file to {saved_progress_checkpoint_full_path}"
-            )
-        except Exception as e:
-            print_function_callback(
-                f"Failed to save state file to {saved_progress_checkpoint_full_path}: {str(e)}"
-            )
-
-            # Raise as this is an unexpected exception in saving state file,
-            # Hard fail for customer's files to be downloaded EXACTLY once.
-            raise
```
sakshie95, line 184: No more hard fail if there is an exception in saving the state file?

test/unit/deadline_client/test_pid_file_lock.py
```diff
@@ -0,0 +1,185 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sakshie95, line 1: Maybe I'm blind, but I don't see a test case for the race b/w two processes to claim the lock file after this refactoring. Could you add it if that's the case?

src/deadline/client/cli/_groups/queue_group.py
mwiebe: Not a significant use case, but as float you can specify 3.5 minutes and it will work. The code using the value is `timedelta(minutes=bootstrap_lookback_in_minutes)`, and I looked up the documentation https://docs.python.org/3/library/datetime.html#datetime.timedelta to see that the arguments accept floats.

src/deadline/client/cli/_groups/queue_group.py
mwiebe: The function these were in had four checks, and after removing the two path mapping checks it just had two left. I thought about how easy it was to understand what was happening in the CLI command function here, and putting these two checks here instead of in a separate function of a different file made it easier without growing the size of the function a lot.

I like your motivation to provide those validations for callers using the Python interface. Maybe moving them into the incremental_output_download function is the right call? I'll make that change, then we can look at it again and see if that's good.

mwiebe: As I go to do that, realizing that incremental_output_download takes the object, not the paths, so the validation can't go in there.

src/deadline/client/cli/_incremental_download.py
mwiebe: Yeah, I'll clean that up.

src/deadline/client/cli/_pid_file_lock.py
mwiebe: Yes, great point since the code change makes it the primary interface to use.

src/deadline/client/cli/_pid_file_lock.py
mwiebe: Good idea, will add that.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
mwiebe: Yes, 100%. Doing that as we step through implementing the job/session/sessionaction hierarchy makes the best sense to me, as that's when we'll be double checking the precise API definitions to make sure the state tracking is robust.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
mwiebe: There's not a specific SLA about this, but the number is based on taking what we saw during some heavy load testing and making it bigger by a generous amount.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
mwiebe: Yes, still accurate because the file open or write operations would raise an exception, and the code here lets it fall through.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
mwiebe: Instead of catching and re-throwing the exception, it falls through untouched. I see I missed doing the equivalent in the from_file function, I'll change it there too to simplify that one.

test/unit/deadline_client/test_pid_file_lock.py
mwiebe: You're correct, there is no test for that. This is a very tricky thing to test: a test that confirms mocked calls would only validate that the code is what we wrote it to be, and not exercise the race condition. I'll push the rest of the changes and think about this one some more.

mwiebe: I've changed the try_acquire function to first optimistically claim the lock, instead of doing a check followed by a claim. This makes the happy path more efficient, and exposes the main race to claim the lock to the tests better. I like the improvement, and I think this moves the coverage in the direction you were looking for.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: I see.. can you make this duration a constant somewhere then instead of hard-coding it here. We might want to play around with this number when we have more data. We can do this in a later PR ofcourse it's very nit :))

src/deadline/client/cli/_groups/queue_group.py
sakshie95: Ah i see.. is there value in moving it ``_queue_apis.py`` instead then? 
Also, thanks for removing the path mapping rule validation, we'll get rid of that param soon from the cli!

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
sakshie95: I see that makes sense, thanks

test/unit/deadline_client/test_pid_file_lock.py
sakshie95: I see you changed the flow so not sure if its fully applicable but I wrote some tests for this validation in the code earlier https://github.com/aws-deadline/deadline-cloud/blob/mainline/test/unit/deadline_client/test_pid_utils.py#L333 that uses mocks but kind of helps to validate the race condition as well. But maybe you find a better way for testing this.. in a later revision

sakshie95 (approved):

leongdl (approved):

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
mwiebe: Good thought, I'll do that.

PR #719, opened 2025-06-26
feat(experimental): Add client library API to list jobs by filter expr

mwiebe:

Incremental downloads needs to track the correct list of jobs that might produce data for download by completing session actions. What we've added so far is a part of that, but need some more primitives. This one is for all the active jobs in a queue, and we are planning to add one that returns all active but requeued jobs.

* Added experimental candidate API function in
  deadline.client.api._list_jobs_by_filter_expression
  * It returns all jobs in a queue that satisfy a provided
    filterExpression in the same form that the deadline:SearchJobs API
    accepts. Because it nests the provided filterExpression in a
    groupFilter, the provided filter cannot itself nest groupFilters.
  * The name starts with list_jobs_, so that someone using tab-complete
    or browsing the list of functions sees it alphabetically adjacent
    to the straight list_jobs function.
* This new experimental function generalizes _list_jobs_recent_by_timestamp_field,
  so removing that function. Modified all the test cases to use the new
  interface and supply the timestamp thresholding as filter expressions.
* Move the code that implements a mock local deadline:SearchJobs API
  into its own file, and extend it to dynamically construct more general
  filter expressions.
* The purpose of this function is to support the new incremental
  downloads CLI command. We expect it to be a useful primitive,
  but are marking it internal-only for now.

Progress on incremental downloads feature.

Generalized and updated the unit tests.

No

No

No

No

----


src/deadline/client/api/_list_jobs_active.py
```diff
@@ -0,0 +1,182 @@
...
+    deadline = boto3_session.client("deadline", config=get_default_client_config())
+
+    # Sort jobs in ascending order of the timestamp field
+    sort_expressions = [{"fieldSort": {"name": "CREATED_AT", "sortOrder": "ASCENDING"}}]
+
+    # Filter for any of the active statuses READY, ASSIGNED, STARTING, SCHEDULED, or RUNNING
+    active_jobs_filter = {
+        "groupFilter": {
+            "filters": [
+                {
```
leongdl: I wish we could do the small project to do the group filter, this input is so not obvious. (obviously not a comment to change in this CR)

src/deadline/client/api/_list_jobs_active.py
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    filter_expressions = {
+        "filters": [active_jobs_filter],
+        "operator": "AND",
+    }
+
+    # Continue until we've processed all jobs
+    while True:
```
leongdl: This loop is nearly the same code as the prior primitive. Do we care about code duplication here? Or make this loop a function of its own?

src/deadline/client/api/_list_jobs_active.py
```diff
@@ -0,0 +1,182 @@
...
+            # with updates of running jobs generally being multiple seconds apart.
+            raise JobFetchFailure(
+                "Failure fetching jobs based on the createdAt field as more then 100 jobs have the exact same timestamp value."
+            )
+        else:
+            # Continue processing from the largest timestamp value we saw so far
+            threshold_timestamp = jobs[-1]["createdAt"]
+
+            # Update jobs to have job["createdAt"] >= threshold_timestamp
+            filter_expressions = {
```
sakshie95: This is not a 100 percent accurate for active jobs because ``created_at`` doesn't get updated for re-queued jobs/steps/tasks even though the job becomes "active". We might want to rename the library function accordingly and add this to the docstring

test/unit/deadline_client/api/mock_search_jobs.py
```diff
@@ -0,0 +1,205 @@
...
+    "RUNNING",
+    "SUSPENDED",
+    "CANCELED",
+    "FAILED",
+    "SUCCEEDED",
+    "NOT_COMPATIBLE",
+]
+
+
+def _snake_to_camel(value: str) -> str:
```
sakshie95: nit: might be useful to move to a test util.. I had thought of making this function too at some point

src/deadline/client/api/_list_jobs_active.py
mwiebe: Yeah, part of my motivation to make the code nearly the same is to make the similarities more obvious and lead to some kind of merging of them in an update. These are internal-only at the moment, so we should be able to do that and clean up all the traces.

src/deadline/client/api/_list_jobs_active.py
mwiebe: The key that makes it work is that the first time it runs SearchJobs, there is no filtering by threshold timestamp, and the sorting is by create_at, so it starts from the earliest active job by created_at. That means it will get all the active jobs in the queue.

src/deadline/client/api/_list_jobs_active.py
mwiebe: Generalizing to a single function ended up looking similar in difficulty to other options, so I went ahead with that and now there's just one.

src/deadline/client/api/_list_jobs_by_filter_expression.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def _list_jobs_by_filter_expression(
+    boto3_session: boto3.Session,
+    farm_id: str,
+    queue_id: str,
+    filter_expression: dict[str, Any],
+) -> list[dict[str, Any]]:
+    """
+    This function retrieves all jobs in the queue that satisfy a provided filter expression, except potentially
```
sakshie95, line 28: Want to understand how this generalization is working. Do we want to order and query next batches by ``created_at`` even when a ``dateTimeFilter`` is provided in the filter expression? How would that work reliably?

mwiebe: When we were looking at listing recent jobs, we saw that we could order and filter by the same timestamp field to get a reliable way of listing every job except for jobs updated within an eventual consistency window. The key to this generalization is that the ordering and filtering for the entire query don't need to use that same timestamp. We can take an arbitrary filter expression as input, and use any linear ordering + filtering to walk through the full dataset. The important property is that the ordering/filtering is based on a value in the job instead of being an integer index in the query result list.

After writing that down, I think I didn't generalize it enough. Probably we could use Job ID for that purpose, and the resulting code will be even simpler. I'm going to try that.

mwiebe: After trying it, found that we can't filter with inequalities on Job ID right now. We'll have to update that in the SearchJobs API before we can make this change. The current use of `createdAt` will work fine to provide that ordering for now.

sakshie95: Makes sense, thanks for the explanation. We could still get into the situation where we miss newer jobs (close to the current timestamp) because we're using ``createdAt`` for the sorting after we filter and query the results but that will fall under the "eventual consistency window" & handled in the next run of the cli as per our discussion

sakshie95 (approved):

src/deadline/client/api/_list_jobs_by_filter_expression.py
```diff
@@ -0,0 +1,172 @@
...
+            boto3_session,
+            farm_id,
+            queue_id,
+        )
+
+    Args:
+      boto3_session (boto3.Session): The boto3 Session for AWS API access.
+      farm_id (str): The Farm ID.
+      queue_id (str): The Queue ID.
+      filter_expressions (dict[str, Any]): The filter expression to apply to jobs.
```
leongdl, line 63: for next PR, not blocking, given the union query at line 158, the nested depth of filter expression and the number of clauses is limited. (We can add it later if we make this a public API, the caller would know the constraint anyways)

leongdl (approved):

PR #721, opened 2025-07-01
feat(experimental): Track jobs list for incremental download over time

mwiebe:

For incremental output download for a queue, want to track all the jobs that have downloads started, and compare it with all the jobs that might start new downloads.

Update the incremental output command to record the full state of the job as returned by SearchJobs, and show a diff with previous state. Running this command to view the output as a queue evolves over time will help refine how this should work.

* Move the eventual consistency max time to a constant
* Add the full job state from the SearchJobs responses to the
  IncrementalDownloadState object. This is to help explore what options
  we have for tracking/optimizing job state changes.
* Use the _list_jobs_by_filter_expression function to collect
  all the candidate jobs for downloading tasks.
* In the incremental download function, perform a diff between the
  in-progress and new candidate job lists, printing info about changes.
* Print a summary of the job updates at the end.
* Remove debug printing from the pid file lock implementation and the
  incremental download state load/save.
* Adjust the CLI help output to use the docstring instead of from the
  decorator. Shorten some of the option names where the context makes it
  clear. Added a default checkpoint directory for the operation.
* Modified how the opt-in environment variable works. Now, the command
  is always available and showin in the help output, but running it only
  succeeds if the environment variable is set.
* Update the CLI tests to use the mocked search_jobs API, and
  implemented a mocked get_job API that works the same way. The golden
  path test now calls the CLI twice, modifying the state of the mocked
  job between calls, and validates that the output of the CLI command
  reflects that job's state appropriately.

Progress implementing the incremental download feature.

```
$ deadline queue -h
Usage: deadline queue [OPTIONS] COMMAND [ARGS]...

  Commands for queues.

Options:
  -h, --help  Show this message and exit.

Commands:
  export-credentials           Export queue credentials in a format...
  get                          Get the details of a queue.
  incremental-output-download  BETA - Downloads job attachments output...
  list                         Lists the available queues.
  paramdefs                    Lists a Queue's Parameters Definitions.

$ deadline queue incremental-output-download -h
Usage: deadline queue incremental-output-download [OPTIONS]

  BETA - Downloads job attachments output incrementally for all jobs in a
  queue. When run for the first time or with the --force-bootstrap option, it
  starts downloading from --bootstrap-lookback-minutes in the past. When run
  each subsequent time, it loads  the previous checkpoint and continues where
  it left off.

  To try this command, set the ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD environment
  variable to 1 to acknowledge its incomplete beta status.

  [NOTE] This command is still WIP and partially implemented right now.

Options:
  --bootstrap-lookback-minutes FLOAT
                                  Downloads outputs for job-session-actions
                                  that have been completed since these many
                                  minutes at bootstrap. Default value is 0
                                  minutes.
  --checkpoint-dir TEXT           Proceed downloading from the previous
                                  progress file stored in this directory, if
                                  it exists. If the file does not exist, the
                                  download will initialize using the bootstrap
                                  lookback in minutes.
  --force-bootstrap               Forces command to start from the bootstrap
                                  lookback period and overwrite any previous
                                  checkpoint. Default value is False.
  --conflict-resolution [skip|overwrite|create_copy]
                                  How to handle downloads if an output file
                                  already exists: CREATE_COPY: Download the
                                  file with a new name, appending '(1)' to the
                                  end SKIP: Do not download the file OVERWRITE
                                  (default): Download and replace the existing
                                  file. Default behaviour is to OVERWRITE.
```

Wrote a separate script that continually submits random jobs, and viewed the output over time. Refined the output to include the summary at the end, clean up the debug prints, etc.

No.

No.

No.

No.

----


src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -211,38 +211,29 @@ def queue_get(**args):
...
 if os.environ.get("ENABLE_INCREMENTAL_OUTPUT_DOWNLOAD") is not None:
 
-    @cli_queue.command(
-        name="incremental-output-download",
-        help="BETA - Download Job Output data incrementally for all jobs running on a queue as session actions finish.\n"
-        "The command bootstraps once using a bootstrap lookback specified in minutes and\n"
-        "continues downloading from the last saved progress thereafter until bootstrap is forced.\n"
-        "[NOTE] This command is still WIP and partially implemented right now",
-    )
+    @cli_queue.command(name="incremental-output-download")
```
sakshie95: why did we remove the help text?

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -314,49 +293,65 @@ def incremental_output_download(
...
-                    checkpoint_file_path, logger.echo
+                current_download_state = IncrementalDownloadState.from_file(checkpoint_file_path)
+
+                # Print the previous download completed time in local time
+                logger.echo("Checkpoint found")
+                logger.echo(
+                    f"Continuing from: {current_download_state.downloads_completed_timestamp.astimezone().isoformat()}"
                 )
 
+            logger.echo()
```
sakshie95: typo?

src/deadline/client/cli/_incremental_download.py
```diff
@@ -3,13 +3,166 @@
...
+    #                     }
+    #                 ],
+    #                 "operator": "AND",
+    #             },
+    #         )
+    #     }
+    # )
+    # WORKAROUND: Get all jobs with a SUCCEEDED or SUSPENDED task run status, and filter by endedAt client-side.
+    #             We want to download everything that is succeeded or suspended, but not
+    #             FAILED, CANCELED, or NOT_COMPATIBLE.
```
sakshie95, line 125: can a failed job not have partial outputs for the steps/tasks that were successfully finished?

src/deadline/client/cli/_incremental_download.py
```diff
@@ -3,13 +3,166 @@
...
+        farm_id: The farm ID.
+        queue_id: The queue ID in the farm.
+        starting_timestamp: The point in time from which to look for new download outputs.
+
+    Returns:
+        A dictionary mapping job id to the job as returned by the deadline.search_jobs API.
+    """
+    # Construct the full set of jobs that may have new available downloads.
+    # - Any active job (job with taskRunStatus in READY, ASSIGNED,
+    #   STARTING, SCHEDULED, or RUNNING), that has at least one SUCCEEDED task.
```
sakshie95, line 41: Wouldn't this slow down the detection because we're waiting for atleast one task to succeed for picking a job and adding it to this set? i.e in the case that the task is very long running

src/deadline/client/cli/_pid_file_lock.py
```diff
@@ -46,7 +45,6 @@ def _claim_pid_lock_with_rename(tmp_file_name: str, pid_file_path: str) -> bool:
 def _try_acquire_pid_lock(
     pid_file_path: str,
     operation_name: str = "the operation",
-    print_function_callback: Callable[[str], None] = print,
```
sakshie95, line 49: No prints for the caller from this library? Since it's expected to be used elsewhere my thought was to keep the prints here and pass it back to the caller with a callback if provided

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -4,29 +4,48 @@
...
-from datetime import datetime, timedelta
-from typing import Any, Optional, Callable
+from datetime import datetime
+from typing import Any, Optional
 import tempfile
 
+# This as an upper bound to allow for eventual consistency into the materialized view that
+# the deadline:SearchJobs API is based on. It's taken from numbers seen in heavy load testing,
+# increased by a generous amount.
+EVENTUAL_CONSISTENCY_MAX_SECONDS = 120
```
sakshie95, line 14: thanks for making this configurable!

src/deadline/client/cli/_groups/queue_group.py
mwiebe: Click uses the docstring for help if it's not supplied here. The other commands do it that way, so this change brings it in line with them

src/deadline/client/cli/_groups/queue_group.py
mwiebe: No, this provides vertical space to separate different parts of the output.

src/deadline/client/cli/_incremental_download.py
mwiebe: Yes, that's something to think about as we refine the command. What should the default behavior be for failed jobs, and should we add an option to customize that?

src/deadline/client/cli/_incremental_download.py
mwiebe: It should pick up the job with a long-running task right when it finishes, because then the count will increase. That's also the first time any output will be available so this reduces the size of the checkpoint file.

src/deadline/client/cli/_pid_file_lock.py
mwiebe: These were debugging prints to help with implementing the primitive operations correctly. As lower level library components, it's better to be silent here and let the caller choose what to print about the operation.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -3,13 +3,166 @@
...
+    download_candidate_jobs_dict = {
+        job["jobId"]: job
+        for job in _list_jobs_by_filter_expression(
+            boto3_session,
+            farm_id,
+            queue_id,
+            filter_expression={
+                "filters": [
```
leongdl, line 49: I thought we could to 3 groups of 3,

EG:
group1 = SearchGroupFilter( READY, ASSIGNED, STARTING, OR )
group1 = SearchGroupFilter( SCHEDULED, RUNNING, OR )
aggregate = SearchGroupFilter( group1, group2, OR)

src/deadline/client/cli/_incremental_download.py
```diff
@@ -3,13 +3,166 @@
...
+                            },
+                        }
+                        for status_value in ["SCHEDULED", "RUNNING"]
+                    ],
+                    "operator": "OR",
+                },
+            )
+        }
+    )
+    print(f"DEBUG: Got {len(download_candidate_jobs_dict)} active jobs")
```
leongdl, line 87: nit for debug, I guess we are removing these before shipping.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -3,13 +3,166 @@
...
+    download_candidate_jobs_dict = {
+        job_id: _datetimes_to_str(job)
+        for job_id, job in download_candidate_jobs_dict.items()
+        if job["taskRunStatusCounts"]["SUCCEEDED"] > 0
+    }
+    print(
```
leongdl, line 93: nit print.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -3,13 +3,166 @@
...
+        filter_expression={
+            "filters": [
+                {
+                    "stringFilter": {
+                        "name": "TASK_RUN_STATUS",
+                        "operator": "EQUAL",
+                        "value": status_value,
+                    },
+                }
+                # Maximum of 3 filters are permitted, so the 5 statuses are split
```
leongdl: Copy paste :)?

src/deadline/client/cli/_incremental_download.py
leongdl: I see why, https://github.com/aws-deadline/deadline-cloud/blob/93c71a6ed266197317cdc18f9c3d0f550963e4e0/src/deadline/client/api/_list_jobs_by_filter_expression.py#L110, if the input was a list of filters then we maybe able to optimize.

src/deadline/client/cli/_incremental_download.py
sakshie95: I would think the default behaviour should be what the default behavior is for job download command. If it downloads partial outputs automatic downloads should do the same

src/deadline/client/cli/_incremental_download.py
```diff
@@ -38,14 +189,159 @@ def _incremental_output_download(
...
+        else:
+            # Carry over the attachments manifest metadata
+            dc_job["attachments"] = ip_job["attachments"]
+    updated_job_ids.difference_update(attachments_free_job_ids)
+
+    # Prune jobs that we are certain have no changes by looking at its task status counts. A job is unchanged if both of these are true:
+    # 1. job["taskRunStatusCounts"]["SUCCEEDED"] stayed the same. Except for when a task is requeued, this count will always increase
+    #    when new output is available to download. If a task is requeued, this value could drop and then return to the same value
+    #    when new output is generated.
+    # 2. job["updatedAt"] stayed the same. If a task is requeued, this timestamp will be updated, so this catches anything missed
```
sakshie95, line 243: This isn't a true assumption right? If a task is re-queued the job updated_at doesn't change. If a job is re-queued the job updated_at changes

src/deadline/client/cli/_incremental_download.py
```diff
@@ -38,14 +189,159 @@ def _incremental_output_download(
...
+        ):
+            print_function_callback(f"  {line}")
+
+        updated_in_progress_jobs.append(dc_job)
+
+    # Process all the jobs that are new
+    for job_id in new_job_ids:
+        dc_job = download_candidate_jobs[job_id]
+
+        # Call deadline:GetJob to retrieve attachments manifest information
```
sakshie95, line 302: Do we need this or this is just for this iteration? Given that we can get the manifest info directly from s3 using the session actions at the end.. why do we need this?

src/deadline/client/cli/_incremental_download.py
mwiebe: The outer level needs the AND operator in _list_jobs_by_filter_expression, so combining would need a third nesting level.

src/deadline/client/cli/_incremental_download.py
mwiebe: Yes, it's useful to see this for now to understand how much client-side filtering is happening.

src/deadline/client/cli/_incremental_download.py
mwiebe: Yep!

src/deadline/client/cli/_incremental_download.py
mwiebe: This is what I observed in testing. It's worth making double-sure, yeah.

src/deadline/client/cli/_incremental_download.py
mwiebe: The job itself is the source of truth, it's better to directly use the data from it. I believe the metadata attached to the S3 object won't be enough for everything we want either, we can look closer when we're doing that part.

sakshie95 (approved):

src/deadline/client/cli/_incremental_download.py
mwiebe: I've double-checked, and found that you are correct. Looks like I hadn't properly re-checked that after I removed https://github.com/aws-deadline/deadline-cloud-samples/pull/81 from my test farm (since its updates were setting updatedAt). We'll have to think through the requeue case again more carefully.

src/deadline/client/cli/_incremental_download.py
leongdl: I see - ok definitely need to work on the grouped filter instead.

leongdl (approved):

PR #725, opened 2025-07-03
feat: Add job name/id to job waitoutput and improve timestamp output for 'logs' command

prestomation:
I received some offline feedback for the previous change adding these commands.

I have added job name/id to the output of the 'wait' command, as if you use the default job there is no indication of which job you are waiting on.

I've improved the timestamp handling in the logs command, it now returns ISO8601 and there is a '--timezone local' switch which will print out in local timezone:

`deadline job wait` had no output indicting which job was being waited on
`deadline job logs` output no precision timestamps with ambiguous timezones

`deadline job wait` now prints the job id and name
`deadline job logs` now prints iso8601 timestamps, and has a `--timezone local` switch to use the system timezone.

More precise information for users of these commands
 

- Have you run the integration tests? no

- Has the README.md been updated? If you modified CLI arguments, for instance. yes

Technically I guess? but the base change was merged just a few hours ago

No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ The command blocks until the job reaches a terminal state (SUCCEEDED, FAILED, CA
 
 ### Retrieving Job Logs
 
-You can monitor job status and retrieve logs using the CLI. The logs lines are returned starting from the most recent log event.:
+You can monitor job status and retrieve logs using the CLI. The logs lines are returned starting from the most recent log event with timestamps in ISO 8601 format:
 
 ```sh
 # Get logs for a specific session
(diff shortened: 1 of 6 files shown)


leongdl (dismissed):

leongdl:
DCO is failing, please use `git commit --amend -s`

leongdl:
```
FAILED test/unit/deadline_client/cli/test_cli_job_logs.py::test_cli_job_logs_json_error - assert 'error' in 'The AWS Deadline Cloud CLI encountered the following exception:\nTraceback (most recent call last):\n  File "/Users/r... _endpoint_for_partition\n    raise NoRegionError()\nbotocore.exceptions.NoRegionError: You must specify a region.\n\n'
 +  where 'The AWS Deadline Cloud CLI encountered the following exception:\nTraceback (most recent call last):\n  File "/Users/r... _endpoint_for_partition\n    raise NoRegionError()\nbotocore.exceptions.NoRegionError: You must specify a region.\n\n' = <Result SystemExit(1)>.output
```

Does this mean it needs to be integ tests?

evanspearman-a:
I don't think this should be considered a breaking change. Adding a field to json _output_ shouldn't be considered breaking in this sort of context.

evanspearman-a (dismissed):

prestomation:
I've fixed up the tests. it's not an integ test but the way the SDK was being used it tried to load creds even though they aren't used. I've added an autouse fixture to override with an empty aws/config file so this doesn't pollute tests in the future. 

The breaking change is not due to adding the new field, but because the timestamp format is changing which someone could be relying on.

prestomation:
The SonarQube failure is about some test duplication which I'm not inclined to refactor but happy to do so if there are existing patterns I should follow

leongdl:
> I've fixed up the tests. it's not an integ test but the way the SDK was being used it tried to load creds even though they aren't used. I've added an autouse fixture to override with an empty aws/config file so this doesn't pollute tests in the future.
> 
> The breaking change is not due to adding the new field, but because the timestamp format is changing which someone could be relying on.

If this is a breaking changes, please use `feat!` in the commit message for our version bump workflows.

leongdl (approved):

prestomation:
Hm actually I will leave it as it is because the change it 'breaks' has not yet been released.

evanspearman-a (approved):

PR #726, opened 2025-07-04
feat(experimental): Retrieve sessions/session actions for auto download

mwiebe:

The auto download feature needs to determine all the sessions and session actions that change across an incremental update. The session actions will later be used to determine which downloads to perform.

* Move the code to categorize and update the jobs from the checkpoint into a _categorize_jobs_in_checkpoint function
* Update the storage about sessions in IncrementalDownloadJob to be a dict from session ID to the latest session action index that is finished downloading. When a session's endedAt is before the downloadsCompletedTimestamp, we no longer have to track the session, and remove it from the checkpoint file.
* Add a field to IncrementalDownloadJob that contains the maximum endedAt timestamp of a session for the job that has been downloaded.
* Implement a function _get_job_sessions that uses deadline:ListSessions and deadline:ListSessionActions to retrieve all the new session actions to download for the job. It uses the maximum endedAt timestamp stored in the IncrementalDownloadJob as a threshold to filter out old sessions, and populates the sessionActionIndex in the session action objects by extracting it from the sessionActionId.
* Implement a function _update_checkpoint_jobs_list to take the categorized job ids and other data collected, and produce an updated list of IncrementalDownloadJob objects to put in the checkpoint.

Progress towards an auto downloads feature.

Some unit tests, but will add more later. Mostly testing in a farm with random job submissions being done periodically.

No

No

No

No

----

### DIFF
--- a/src/deadline/client/cli/_incremental_download.py
+++ b/src/deadline/client/cli/_incremental_download.py
@@ -6,6 +6,7 @@ __all__ = ["_incremental_output_download"]
 from datetime import datetime, timedelta, timezone
 import difflib
 import textwrap
+from typing import Optional
 
 from .. import api
 from typing import Any, Callable
@@ -20,7 +21,11 @@ from ._common import _cli_object_repr
 
 
 def _get_download_candidate_jobs(
-    boto3_session: boto3.Session, farm_id: str, queue_id: str, starting_timestamp: datetime
+    boto3_session: boto3.Session,
+    farm_id: str,
+    queue_id: str,
+    starting_timestamp: datetime,
+    print_function_callback: Callable[[str], None] = lambda msg: None,
 ) -> dict[str, dict[str, Any]]:
     """
     Uses deadline:SearchJobs queries to get a dict {job_id: job} of download candidates for the queue.
@@ -36,10 +41,13 @@ def _get_download_candidate_jobs(
     Returns:
         A dictionary mapping job id to the job as returned by the deadline.search_jobs API.
     """
+    print_function_callback("Retrieving updated data from Deadline Cloud...")
+    start_time = datetime.now(tz=timezone.utc)
+
     # Construct the full set of jobs that may have new available downloads.
     # - Any active job (job with taskRunStatus in READY, ASSIGNED,
     #   STARTING, SCHEDULED, or RUNNING), that has at least one SUCCEEDED task.
-    download_candidate_jobs_dict = {
+    download_candidate_jobs = {
         job["jobId"]: job
         for job in _list_jobs_by_filter_expression(
             boto3_session,
@@ -61,7 +69,7 @@ def _get_download_candidate_jobs(
             },
         )
     }
-    download_candidate_jobs_dict.update(
+    download_candidate_jobs.update(
         {
             job["jobId"]: job
             for job in _list_jobs_by_filter_expression(
@@ -84,21 +92,21 @@ def _get_download_candidate_jobs(
             )
         }
     )
-    print(f"DEBUG: Got {len(download_candidate_jobs_dict)} active jobs")
-    download_candidate_jobs_dict = {
+    print(f"DEBUG: Got {len(download_candidate_jobs)} active jobs")
+    download_candidate_jobs = {
         job_id: _datetimes_to_str(job)
-        for job_id, job in download_candidate_jobs_dict.items()
+        for job_id, job in download_candidate_jobs.items()
         if job["taskRunStatusCounts"]["SUCCEEDED"] > 0
     }
     print(
-        f"DEBUG: Filtered down to {len(download_candidate_jobs_dict)} active jobs based on SUCCEEDED task filter"
+        f"DEBUG: Filtered down to {len(download_candidate_jobs)} active jobs based on SUCCEEDED task filter"
     )
 
     # - Any recently ended job (job went from active to terminal with a taskRunStatus
     #   in SUSPENDED, CANCELED, FAILED, SUCCEEDED, NOT_COMPATIBLE), that has at least
     #   one SUCCEEDED task. The endedAt timestamp field gets updated when that occurs.
     # TODO: Enable this when filtering by ENDED_AT works.
-    # download_candidate_jobs_dict.update(
+    # download_candidate_jobs.update(
     #     {
     #         job["jobId"]: job
     #         for job in _list_jobs_by_filter_expression(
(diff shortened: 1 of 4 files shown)


src/deadline/client/cli/_incremental_download.py
```diff
@@ -157,127 +165,138 @@ def _get_download_candidate_jobs(
...
+
+    became_inactive: set[str] = set()
+    updated: set[str] = set()
+    added: set[str] = set()
+    unchanged: set[str] = set()
+    attachments_free: set[str] = set()
+    completed: set[str] = set()
+
+
+def _categorize_jobs_in_checkpoint(
```
leongdl, line 193: I think the logic looks right, but I think we need many test cases to test out everything. This function has quite a bit of business logic.

mwiebe: I've expanded the test cases that cover the various state transitions of jobs for this part. The tests create scenarios with a job evolving in various ways, and assert that expected output about its status is printed in subsequent calls of the operation.

What doesn't have coverage yet is the processing of sessions and session actions.

mwiebe: I'm planning to handle coverage of sessions/session actions in a future PR, first complete things end-to-end with tests running against the service.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -36,10 +41,13 @@ def _get_download_candidate_jobs(
     Returns:
         A dictionary mapping job id to the job as returned by the deadline.search_jobs API.
     """
+    print_function_callback("Retrieving updated data from Deadline Cloud...")
+    start_time = datetime.now(tz=timezone.utc)
+
     # Construct the full set of jobs that may have new available downloads.
     # - Any active job (job with taskRunStatus in READY, ASSIGNED,
     #   STARTING, SCHEDULED, or RUNNING), that has at least one SUCCEEDED task.
-    download_candidate_jobs_dict = {
```
sakshie95, line 42: I know we've been through some iterations for this so this might be a repetitive question but are we not capturing any jobs that don't have atleast one succeeded task now? I thought we were gonna have this delay of atleast one task only for re-queued jobs but this looks like there's a delay of 1 task for every new job as well?

src/deadline/client/cli/_incremental_download.py
```diff
@@ -157,127 +166,142 @@ def _get_download_candidate_jobs(
...
+    # that a session ending before the downloads completed timestamp was already processed.
     for job_id in updated_job_ids:
-        ip_job = in_progress_jobs[job_id]
+        ip_job = checkpoint_jobs[job_id]
         dc_job = download_candidate_jobs[job_id]
 
         if ip_job["taskRunStatusCounts"]["SUCCEEDED"] == dc_job["taskRunStatusCounts"][
             "SUCCEEDED"
-        ] and ip_job.get("updatedAt") == dc_job.get("updatedAt"):
+        ] and ip_job.get("endedAt") == dc_job.get("endedAt"):
```
sakshie95, line 263: thanks for updating this :)

src/deadline/client/cli/_incremental_download.py
```diff
@@ -320,27 +353,333 @@ def _incremental_output_download(
...
+    return result
+
+
+def _get_job_sessions(
+    boto3_session: boto3.Session,
+    farm_id: str,
+    queue_id: str,
+    checkpoint_job_session_completed_indexes: dict[str, dict[str, int]],
+    categorized_job_ids: CategorizedJobIds,
+    checkpoint: IncrementalDownloadState,
```
sakshie95, line 385: should we call this the same everywhere for consistency. I like ``checkpoint`` but we called it ``download_state`` in other places including the api interface

src/deadline/client/cli/_incremental_download.py
```diff
@@ -320,27 +353,333 @@ def _incremental_output_download(
...
+        job.job_id: job.session_ended_timestamp
+        for job in checkpoint.jobs
+        if job.session_ended_timestamp is not None
+    }
+
+    deadline = boto3_session.client("deadline")
+    job_sessions: dict[str, list] = {}
+    sessions_paginator = deadline.get_paginator("list_sessions")
+    for job_id in job_ids:
+        # Use the greater of the bootstrap command timestamp and the session ended timestamps
```
sakshie95, line 410: Is this accurate? Maybe I'm blind but I think the code below tries to fetch the session ended timestamp from the job or from the checkpoint file. I don't see a check to determine which is greater here? Also would like to understand the thought behind this

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
```diff
@@ -90,46 +118,12 @@ class IncrementalDownloadState:
     a stream by tracking state at three levels. Where possible, we use the resource state at one level to prune queries at lower levels:
 
     1. Job - The jobs list contains every job that is active and that we have downloaded output from in a previous incremental download command.
```
sakshie95, line 120: Good time to add the eviction strategies to this file. This or next PR

src/deadline/client/cli/_incremental_download.py
leongdl: If its not ended, then it takes the checkpoint time, isn't that right?

leongdl (approved):
I'm ok with this change, only `_categorize_jobs_in_checkpoint` worries me in terms of code complexity and ongoing code maintainability / coverage.

src/deadline/client/cli/_incremental_download.py
mwiebe: The way I thought that through is that any new job with 0 succeeded tasks can never have a download. The first time it would have a download is when succeeded tasks is >= 1. If a job has 0 succeeded tasks, and has downloads, it means that the previously succeeded tasks were requeued.

src/deadline/client/cli/_incremental_download.py
mwiebe: Yes, here was my thought process to change it in a few places: It felt like a good change, but I wasn't sure. By changing it in a few places, can get a feel for the name while developing, and then later choose whether to finish the rest of the rename or revert to the previous name, to be consistent again. We still have free reign over these names before making them public interface, so it's a good time to try different ones on and see how they fit.

src/deadline/client/cli/_incremental_download.py
mwiebe: The session_ended_timestamp values are all bigger than downloads_started_timestamp, that's why it doesn't bother doing the comparison and only falls back when there is no session_ended_timestamp to use.

src/deadline/job_attachments/incremental_downloads/incremental_download_state.py
mwiebe: Great call out, I'll do that in the next PR.

sakshie95 (approved):

PR #730, opened 2025-07-11
feat(experimental): Download outputs for the auto download feature

mwiebe:

For the auto/incremental download, it needs to download the output files after it determines which session actions are in scope for processing. This involves downloading all the manifest files, determining the correct order to process them, merging them, then downloading all the files.

* Added code that fills in the manifest keys when they're not already available on the session action. The output manifests of session actions can be stored in session actions, see https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_ListSessionActions.html, but this was added recently so is not guaranteed to be present.
* Added a --dry-run option to the incremental download command, that prints out information about what would be downloaded but does not perform the download and does not save the checkpoint file.
* Add a new file _incremental_download_s3_access.py to contain all the S3 functionality needed by the incremental operation. Some of this code is copied from the job attachments download.py, to be able to change the interfaces and work on optimization independently of the existing functionality.
* Use thread-based concurrency for the manifest file download, the session and session action retrieval, and the output file download.
* Print a summary of all the download filenames before proceeding to do the download.
* Add statistics about the number of files and total bytes downloaded.
* Modified the multi-threaded download function to download files
      smaller than 1MB with get_object instead of transfer manager. This
      improved performance on a test dataset from 4m50s to 1m30s.
 
The feature works in some manual testing end-to-end. This means we can expand testing and evaluation of it.

Performed many downloads with different look-back periods from a test farm. Filling in more of the unit tests is for a later follow-up.

No

No

No

No

----

### DIFF
--- a/src/deadline/client/cli/_incremental_download.py
+++ b/src/deadline/client/cli/_incremental_download.py
@@ -5,19 +5,46 @@ __all__ = ["_incremental_output_download"]
 
 from datetime import datetime, timedelta, timezone
 import difflib
-import textwrap
 from typing import Optional
+from configparser import ConfigParser
+from typing import Any, Callable
+import time
+import concurrent.futures
 
 from .. import api
-from typing import Any, Callable
 import boto3
+from botocore.client import BaseClient  # type: ignore[import]
 from ..api._list_jobs_by_filter_expression import _list_jobs_by_filter_expression
+from ...job_attachments.api import summarize_path_list
 from ...job_attachments._incremental_downloads.incremental_download_state import (
     IncrementalDownloadState,
     IncrementalDownloadJob,
     _datetimes_to_str,
 )
-from ._common import _cli_object_repr
+from ...job_attachments._incremental_downloads._manifest_s3_downloads import (
+    _add_output_manifests_from_s3,
+    _download_all_manifests_with_absolute_paths,
+    _merge_absolute_path_manifest_list,
+    _download_manifest_paths,
+)
+from ...job_attachments.asset_manifests import (
+    BaseAssetManifest,
+    BaseManifestPath,
+)
+from ...job_attachments.asset_manifests import (
+    HashAlgorithm,
+)
+from ...job_attachments.models import (
+    FileConflictResolution,
+)
+from ...job_attachments.progress_tracker import (
+    ProgressReportMetadata,
+)
+from ._common import _cli_object_repr, sigint_handler
+from ...job_attachments._path_summarization import human_readable_file_size
+
+
+SESSIONS_API_MAX_CONCURRENCY = 3
 
 
 def _get_download_candidate_jobs(
@@ -33,10 +60,11 @@ def _get_download_candidate_jobs(
     the provided starting_timestamp.
 
     Args:
-        boto3_session: The boto3 session for calling AWS APIs.
-        farm_id: The farm ID.
-        queue_id: The queue ID in the farm.
+        boto3_session: The boto3.Session for accessing AWS.
+        farm_id: The farm id for the operation.
+        queue_id: The queue id for the operation.
         starting_timestamp: The point in time from which to look for new download outputs.
+        print_function_callback: Callback for printing output to the terminal or log.
 
     Returns:
         A dictionary mapping job id to the job as returned by the deadline.search_jobs API.
@@ -180,14 +208,24 @@ class CategorizedJobIds:
     """
     Takes jobs loaded from a loaded checkpoint and a query to get download candidate jobs,
     analyzes all the jobs by looking at fields like task run status counds to categorize them.
+
+    Job categories:
+        added: The job was created or requeued so it now can produce new downloads.
+        updated: The job changed since the previous incremental download operation.
+        unchanged: The job did not change since the previous incremental download operation.
+        completed: The job finished running so all output is available for download.
+        inactive: The job can no longer have any new downloads unless it is requeued. Minimal
+            metadata is tracked to detect if it is requeued.
+        attachments_free: The job has no job attachments associated that can produce
+            outputs for download.
     """
 
-    inactive: set[str] = set()
-    updated: set[str] = set()
     added: set[str] = set()
+    updated: set[str] = set()
     unchanged: set[str] = set()
-    attachments_free: set[str] = set()
     completed: set[str] = set()
+    inactive: set[str] = set()
+    attachments_free: set[str] = set()
 
 
 def _categorize_jobs_in_checkpoint(
@@ -203,7 +241,18 @@ def _categorize_jobs_in_checkpoint(
     Categorizes the provided download candidate jobs by id into a CategorizedJobIds object,
     updating the jobs within download_candidate_jobs where necessary.
 
-    * Calls deadline:GetJob to get job attachments manifest information if it is not stored yet.
+    * Calls boto3 deadline.get_job() to get job attachments manifest information if it is not stored yet.
+
+    Args:
+        boto3_session: The boto3.Session for accessing AWS.
+        farm_id: The farm id for the operation.
+        queue_id: The queue id for the operation.
+        checkpoint: The checkpoint for the incremental download.
+        download_candidate_jobs: The result of a _get_download_candidate_jobs call, {job_id: job} where
+            job is a result from a deadline.search_jobs() or deadline.get_job() call.
+        new_completed_timestamp: This is the timestamp value that will be placed in
+            checkpoint.downloads_completed_timestamp when saving the checkpoint.
+        print_function_callback: Callback for printing output to the terminal or log.
     """
     deadline = boto3_session.client("deadline")
     checkpoint_jobs = {job.job_id: job.job for job in checkpoint.jobs}
@@ -351,7 +400,11 @@ def _categorize_jobs_in_checkpoint(
             attachments_free_job_ids.add(job_id)
             print_function_callback("  Job does not use job attachments.")
         else:
-            print_function_callback(textwrap.indent(_cli_object_repr(dc_job["attachments"]), "  "))
+            print_function_callback("  Manifest file system paths:")
+            for manifest in dc_job["attachments"]["manifests"]:
+                print_function_callback(
+                    f"    - {manifest['rootPath']} ({manifest['rootPathFormat']})"
+                )
 
         if (
             dc_succeeded_task_count == dc_total_task_count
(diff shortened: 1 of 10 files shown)


src/deadline/client/cli/_incremental_download_s3_access.py
```diff
@@ -0,0 +1,488 @@
...
+    session_action_list = [
+        session_action
+        for session_action in session_action_list
+        if "manifests" not in session_action
+    ]
+    if not session_action_list:
+        return
+
+    # Process the job's attachments list to generate the hashes
+    job_manifests_length = len(job["attachments"]["manifests"])
```
leongdl: `attachments` isn't required parameter of the job response, can this fail?

src/deadline/client/cli/_incremental_download.py
```diff
@@ -429,51 +462,125 @@ def _get_job_sessions(
...
+            queueId=queue["queueId"],
+            jobId=job_id,
+            sessionId=session["sessionId"],
+        ):
+            # Include only succeeded taskRun actions, and filter out any actions that ended before the
+            # start of the update time interval
+            session_action_list.extend(
+                session_action
+                for session_action in session_actions_page.get("sessionActions", [])
+                if session_action.get("status") == "SUCCEEDED"
```
leongdl: super nit for testability, can we refactor this type of logic into a function so we can write a unit test. While the v0 here is correct by the comment, follow changes or fixes by others makes this too hard to not have regressions.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -429,51 +462,125 @@ def _get_job_sessions(
...
+            for session_action in session.get("sessionActions", [])
+        ]
+        _add_output_manifests_from_s3(farm_id, queue, job, queue_role_session, session_action_list)
+
+    duration = datetime.now(tz=timezone.utc) - start_time
+    print_function_callback(f"...populated in {duration}")
+
+    # Filter out session actions without any output manifests, and print a warning message for them.
+    # These are all jobs that use job attachments, so if there's no output it may be a problem.
+    for job_id, session_list in job_sessions.items():
```
leongdl, line 725: nit, something like the block 559-582 can be a separate utility? Keeps the method much shorter for reviews.

leongdl (commented):
Did a few passes of the code. The logic reads correctly. Given the PR is in draft, I'll make another pass as the code is updated.

We need someone to add unit tests, the logic is complex and easy to break since it is all many large linear functions.

src/deadline/client/cli/_incremental_download.py
benl-2023: +1

src/deadline/client/cli/_incremental_download.py
```diff
@@ -429,51 +462,125 @@ def _get_job_sessions(
...
+                    "endedAt" in session_action
+                    and session_action["endedAt"] < checkpoint.downloads_completed_timestamp
                 )
+            )
+        if session_action_list:
+            # Extract the session action indexes from the ids
+            for session_action in session_action_list:
+                # Session action IDs look like "sessionaction-abc123-12" for index 12
+                session_action_index = int(session_action["sessionActionId"].rsplit("-", 1)[-1])
+                session_action["sessionActionIndex"] = session_action_index
```
benl-2023: We should add unit tests cases for this kind of string splitting logic and error handling in case `session_action["sessionActionId"]` is not of the shape we expect.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -429,51 +462,125 @@ def _get_job_sessions(
...
-                    for session_action in session_actions_page.get("sessionActions", [])
-                    if session_action.get("status") == "SUCCEEDED"
-                    and "taskRun" in session_action.get("definition", {})
-                    and not (
-                        "endedAt" in session_action
-                        and session_action["endedAt"] < checkpoint.downloads_completed_timestamp
-                    )
+
+    def retrieve_session_actions_for_session(job_id: str, session: dict[str, Any]):
+        session_action_list: list[dict[str, Any]] = []
```
sakshie95: I think we should move this outside of the scope of ``_get_job_sessions`` to make this more readable

src/deadline/client/cli/_incremental_download.py
```diff
@@ -378,11 +408,13 @@ def _categorize_jobs_in_checkpoint(
 
 def _get_job_sessions(
```
benl-2023, line 529: The new additions to this function make it quite long. Can we modularize some of the code chunks here to shorten the function?

It should make it easier to write unit tests against those smaller functions.

src/deadline/client/cli/_incremental_download.py
sakshie95: +1

src/deadline/client/cli/_incremental_download.py
```diff
@@ -429,51 +462,125 @@ def _get_job_sessions(
...
+            future.result()
 
     duration = datetime.now(tz=timezone.utc) - start_time
     print_function_callback(f"...retrieval completed in {duration}")
 
+    print_function_callback("")
+    print_function_callback("Populating missing manifest S3 keys...")
+    start_time = datetime.now(tz=timezone.utc)
+
+    # Run through job_sessions and populate any missing "manifests" fields in the session actions
```
sakshie95: I think this should be a separate utility function too, which we can deprecate once the sessions have manifests field always. Hard for someone without context to understand why we needed it

src/deadline/client/cli/_incremental_download.py
```diff
@@ -429,51 +462,125 @@ def _get_job_sessions(
...
+
+    # Filter out session actions without any output manifests, and print a warning message for them.
+    # These are all jobs that use job attachments, so if there's no output it may be a problem.
+    for job_id, session_list in job_sessions.items():
+        job = download_candidate_jobs[job_id]
+        total_count = 0
+        filtered_count = 0
+        for session in session_list:
+            total_count += len(session.get("sessionActions", []))
+            # Filter out session actions
```
sakshie95: Filter out session actions  that don't have any manifests? Didn't we filter them out already above?

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -0,0 +1,497 @@
...
+        job = download_candidate_jobs[job_id]
+        for session in session_list:
+            for session_action in session.get("sessionActions", []):
+                # The manifests lists from the job and session action correspond, so we can zip them
+                # together to attach the root path with the S3 manifest key
+                for job_manifest, session_action_manifest in zip(
+                    job["attachments"]["manifests"], session_action["manifests"]
+                ):
+                    if "outputManifestPath" in session_action_manifest:
+                        manifests_to_download.append(
```
benl-2023, line 250: Could we refactor this into separate private functions? To reduce the level of nesting in this function?

src/deadline/job_attachments/download.py
```diff
@@ -262,12 +266,16 @@ def _get_tasks_manifests_keys_from_s3(
...
-            set(f.split("/")[len(task_folder.split("/"))] for f in files), reverse=True
-        )[0]
-        manifests_keys += [f for f in files if f.startswith(f"{task_folder}/{last_subfolder}/")]
+    if select_latest_per_task:
+        # 2. Select all files in the last subfolder (alphabetically) under each "task-{any}" folder.
+        for task_folder, files in task_prefixes.items():
+            last_subfolder = sorted(
+                set(f.split("/")[len(task_folder.split("/"))] for f in files), reverse=True
+            )[0]
+            manifests_keys += [f for f in files if f.startswith(f"{task_folder}/{last_subfolder}/")]
```
benl-2023, line 275: Do we need some error handling here in case the files aren't conforming to this path structure?

src/deadline/job_attachments/download.py
```diff
@@ -262,12 +266,16 @@ def _get_tasks_manifests_keys_from_s3(
...
+    else:
+        # Include all the keys, not just the latest per task
```
sakshie95, line 277: Wouldn't this break the existing use-cases? I see the function is internal and marked with ``_`` so technically backwards compatibility isn't a concern, but for where-ever it gets used in the existing https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/download.py#L790 ``get_output_manifests_by_asset_root`` still needs to be backwards compatible?

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -0,0 +1,497 @@
...
+
+It consists fully of internal-only functionality. We would like to iteratively refine these
+interfaces over time, with the goal that we deprecate the current interfaces
+in deadline.job_attachments.download and replace them with more general and flexible interfaces.
+"""
+
+SESSION_ACTION_ID_FROM_KEY_RE = re.compile(r"(sessionaction-[^/-]+-[^/-]+)/")
+
+
+def _add_output_manifests_from_s3(
```
benl-2023, line 77: While this is marked a non public, I think we need to have some more unit test coverage on these methods to check happy path and corner cases. In the original module some of this functionality is forked from, are there also additional unit tests we can copy over?

src/deadline/client/cli/_incremental_download.py
sakshie95, line 529: Overhauling comment from all of the below review comments : Would it be possible to break this function down so it's easier to follow? It has a bunch of logic right now from fetching sessions, session actions and populating missing stuff etc.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -5,19 +5,45 @@
 
 from datetime import datetime, timedelta, timezone
```
sakshie95, line 6: I may be blind but where's the logic to evict the jobs, sessions, session actions out of the download checkpoint once they are fully downloaded?

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -0,0 +1,497 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sakshie95, line 1: Still have to read this more thoroughly. Quick question, was there an issue you encountered in re-using https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/download.py#L276 and https://github.com/aws-deadline/deadline-cloud/blob/e5176598cc859584217eaea00d5e6a14a40070d5/src/deadline/job_attachments/download.py#L814 for our use-case?

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -0,0 +1,497 @@
...
+
+def _add_output_manifests_from_s3(
+    farm_id: str,
+    queue: dict[str, Any],
+    job: dict[str, Any],
+    boto3_session: boto3.Session,
+    session_action_list: list[dict[str, Any]],
+):
+    """
+    This function takes a list of session actions (as returned by boto3 deadline.list_session_actions),
```
sakshie95, line 85: I think this function does a little bit more than this doc string? Not sure why but there's portions that group session actions by task Id "Organize the session actions by task id".

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -0,0 +1,497 @@
...
+            if output_manifest_path.startswith(manifest_prefix):
+                output_manifest_path = output_manifest_path[len(manifest_prefix) :]
+            session_action_for_key["manifests"][manifests_index] = {
+                "outputManifestPath": output_manifest_path,
+            }
+
+
+def _download_all_manifests_with_absolute_paths(
+    queue: dict[str, Any],
+    download_candidate_jobs: dict[str, dict[str, Any]],
```
benl-2023, line 223: We could build more structure into this function signature by opting to use dataclasses instead of `dict[str, Any]`. Its not immediately obvious what `download_candidate_jobs` may have as fields without looking at the caller code.

src/deadline/client/cli/_incremental_download_s3_access.py
mwiebe: True, in that case it should leave the session actions untouched. I'll add a check to the beginning.

src/deadline/client/cli/_incremental_download.py
mwiebe: I've split this out into a separate function `_filter_session_actions_without_manifests_from_job_sessions`

src/deadline/client/cli/_incremental_download.py
mwiebe: Fixed the comment. What we filtered out earlier is jobs that have no job attachments. Here, the job does have a list of manifests, but the particular session action produced no outputs so its manifests list consists of empty `{}`.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
mwiebe: I didn't try to use it yet, because it doesn't have the logic to put the manifests in the correct order yet. Once that is fixed up, that will likely be a more efficient replacement for the function here, though it will be good to measure the difference in a few scenarios.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
mwiebe: Yeah, I've added new unit tests for this particular function that wasn't forked from a different implementation, but many of the others were and do have unit tests to source from. I'll do that before marking this PR as ready.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
mwiebe: That "Organize the session actions by task id" was stale, I've edited the comment to be accurate and describe the purpose of that dict. It should just be populating the missing manifest paths, what else do you see it doing?

src/deadline/client/cli/_incremental_download.py
mwiebe: When a job is fully downloaded, it will no longer appear in the `_get_download_candidate_jobs` query output. In `_categorize_jobs_in_checkpoint` that job will be in the `became_inactive_job_ids` set, When `_update_checkpoint_jobs_list` constructs the list of jobs to put in the updated checkpoint, it looks at the inactive jobs, and keeps a minimal record if `session_ended_timestamp` has a value. That's necessary in order to track requeues. I'll add some comments to make sure these pieces are described in the code.

src/deadline/client/cli/_incremental_download.py
mwiebe: Yes, I've split things up now. Even if we can unit test the implementation pieces, it is still always nicer when we can write our unit tests at a higher level. That lets us refactor the code more quickly with confidence, because then we don't have to change the unit tests as much alongside the code.

mwiebe: Yes, I've split it up into more functions now.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
mwiebe: I'd prefer to keep these functions matching the boto3 return structures where possible. When clients of the library use them together with other boto3 functionality, it results in a more cohesive coding style. It's also more efficient, because it means the code isn't re-formatting the data into different structures as often.

I think another way to help with that understanding is expanding the docstrings to describe the arguments. I'll go ahead and do that.

mwiebe: I've updated all the docstrings here and in _incremental_download.py.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
mwiebe: I've moved this into a separate function. It is fairly nested, but when I look at the result of splitting it up it looks less cohesive and harder to follow, so I think keeping this loop together is nicer.

src/deadline/job_attachments/download.py
mwiebe: It is backwards compatible because the new `select_latest_per_task` parameter defaults to the previous behavior.

src/deadline/job_attachments/download.py
mwiebe: This particular code wasn't changed by this PR, so I'd rather not grow the scope to include it. I agree with you that its error handling behavior could be better, it would be worth looking at this in a separate task.

src/deadline/client/cli/_incremental_download.py
mwiebe: I've moved that out, thanks.

src/deadline/client/cli/_incremental_download.py
mwiebe: I've moved it into a separate function. For backwards compatibility, we'll want to keep this around for a long time even once the manifests fields are being consistently populated by the worker agent.

src/deadline/client/cli/_incremental_download.py
mwiebe: I've refactored this in place instead of splitting out a function, including removing the filter by timestamp that it was doing. Take a fresh look and let me know what you think.

leongdl (dismissed):

src/deadline/client/cli/_incremental_download.py
mwiebe: I haven't added this specific unit test, but it is an example of error case testing that would be great to expand. I'd prefer to merge this PR, and then follow up with a more comprehensive sweep of tests for these error cases.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
mwiebe: There are more tests to expand, but I'd like to merge this PR and do that in follow-ups.

leongdl (approved):

src/deadline/client/cli/_incremental_download.py
benl-2023: Sounds good to me!

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
benl-2023: Thanks!

sakshie95 (approved):

benl-2023 (approved):

PR #742, opened 2025-07-21
feat: Add account IDs to telemetry events

miabatta:
We can provide more detailed support and messaging to users if we know their account IDs.

- Added an `sts:GetCallerIdentity` call when creating telemetry events to retrieve account IDs
- Edited unit tests to include account IDs in event details

When users are authenticated, their AWS account IDs are included in telemetry events.

- Unit tests edited & run
- Manual testing to verify that account IDs appear in telemetry events

n/a

no

no

----


src/deadline/client/api/_telemetry.py
```diff
@@ -313,6 +314,16 @@ def record_error(
     def record_event(
         self, event_type: str, event_details: Dict[str, Any], *, from_gui: bool = False
     ):
+        # Include the account ID if we are authenticated
+        # If not (error), omit the ID and provid the error message
+        try:
+            self.update_common_details(
+                {"accountId": get_boto3_client("sts").get_caller_identity()["Account"]}
+            )
```
erico-aws: What was the reason for not caching the account id after the first call?

test/unit/deadline_client/cli/test_cli_job.py
```diff
@@ -1540,6 +1540,9 @@ def test_cli_job_download_output_with_different_asset_root_path_format_than_job(
 """
             in result.output
         )
+
+        print(result.output)
```
erico-aws: Was this intentional?

miabatta: Testing artifact, thanks- will delete

src/deadline/client/api/_telemetry.py
miabatta: Thought it might be redundant since the `get_boto3_client` result is cached, but I'll edit to cache the ID itself

erico-aws (approved):

AWS-Samuel (approved):

PR #743, opened 2025-07-22
feat: Support path mapping of auto download assets via Storage Profiles

benl-2023:
* We want to allow auto downloads to work across a variety of submission and download machines.
* For that to happen, we need to support storage profile and source the file system locations to map output files as specified by the manifests.

* Fetch the locally configured storage profile the user's deadline config.
* When file system location names present in the user's storage profile match those in the job's manifest, perform path re-mapping of the output files.
* Otherwise fall back to the root paths as specified in the job's manifest metadata.

* We can support cross machine and cross operating system downloading of manifests

Yes
```
============================================================== 1707 passed, 62 skipped, 1 xfailed, 1708 warnings in 36.14s ==============================================================
```

Yes

```
======================================================================= 51 passed, 4 skipped in 123.71s (0:02:03) =======================================================================

```

No

N/A
N/A

No

No

----


src/deadline/client/cli/_incremental_download.py
```diff
@@ -901,6 +905,20 @@ def _incremental_output_download(
         queue_display_name=queue["displayName"],
     )
 
+    # Fetch the storage profile data that's currently selected in the config
+    local_storage_profile_id = get_setting("settings.storage_profile_id", config)
+    local_storage_profile: Optional[Dict[str, Any]] = None
```
mwiebe, line 910: nit: Because of `from __future__ import annotations` at the top, you can simplify this to `dict[str, Any]` instead of importing `Dict`. We've got a mix of this convention across our codebase, but it's nice to tidy things like this up as we're working through new code.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -236,7 +257,7 @@ def _get_manifests_to_download(
     Returns:
         A list of (rootPath, manifest_s3_key) tuples for the manifest objects that need to be downloaded.
```
mwiebe, line 258: Since the return type changed, this docstring needs updating.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -901,6 +905,20 @@ def _incremental_output_download(
...
+    if local_storage_profile_id:
+        try:
+            local_storage_profile = deadline.get_storage_profile(
+                farmId=farm_id,
+                storageProfileId=local_storage_profile_id,
+            )
+        except ClientError as exc:
+            raise DeadlineOperationError(
```
sakshie95, line 918: I think we might need to be more verbose with this error msg. Especially because job download output doesn't have any of this handling right now so a customer might see different behaviour for downloading in the 2 clis so this is their chance of resolving the error without reaching out to us

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -213,16 +215,35 @@ def _download_manifest_and_make_paths_absolute(
...
 
 
 def _get_manifests_to_download(
     job_attachments_root_prefix: str,
     download_candidate_jobs: dict[str, dict[str, Any]],
     job_sessions: dict[str, list],
-) -> list[tuple[str, str]]:
+) -> list[tuple[str, str, str]]:
     """
     Collect a list of (rootPath, manifest_s3_key) tuples for all the job attachments that need to be downloaded.
```
sakshie95, line 248: This needs to be updated to what it is after your changes. What's the last ``str`` going to be for these tuples? I think the second ``str`` has been updated to be ``file_system_location_name`` as per code below but it needs to go in the doc string

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -213,16 +215,35 @@ def _download_manifest_and_make_paths_absolute(
...
+            ]
+            absolute_path = os.path.join(local_file_system_root_path, original_path)
+
+        else:
+            absolute_path = os.path.join(root_path, original_path)
+
+        # Validate path compatibility
+        is_compatible, error_message = _validate_path_compatibility(absolute_path)
+        if not is_compatible:
+            raise PathCompatibilityError(
```
sakshie95, line 234: Does this get caught/handled somewhere or is this what the customer sees in their cli logs?

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -250,6 +271,7 @@ def _get_manifests_to_download(
                         manifests_to_download.append(
                             (
                                 job_manifest["rootPath"],
+                                job_manifest.get("fileSystemLocationName", None),
```
sakshie95, line 274: Does the default ``None`` need to be handled here?

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -285,10 +308,18 @@ def _download_all_manifests_with_absolute_paths(
         A list of BaseAssetManifest objects containing local absolute file paths sorted by the last_modified timestamp.
     """
     # Get the list of (rootPath, manifest_s3_key) tuples to download from S3.
-    manifests_to_download: list[tuple[str, str]] = _get_manifests_to_download(
+    manifests_to_download: list[tuple[str, str, str]] = _get_manifests_to_download(
         queue["jobAttachmentSettings"]["rootPrefix"], download_candidate_jobs, job_sessions
     )
 
+    # Build a mapping between the file system location names and their path from the local storage profile.
```
sakshie95, line 315: Don't we also need the ``os`` in this mapping?

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -85,7 +85,7 @@ def attachment_download(
     config = _apply_cli_options_to_config(**args)
```
sakshie95, line 85: I see you didn't need to change the attachment_cli in this PR. Have you validated the case where there's no storage profile and none is needed works (eg. downloading from same machine as the one used to submit the job)? And it downloads to the full path instead of current directory/full path? Bcz I don't see us passing path mapping rules to it in this PR

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -602,3 +637,67 @@ def _download_manifest_paths(
...
+        # Check for Windows-specific path issues
+        if path.startswith("/"):
+            return (
+                False,
+                "Path starts with '/' which is not compatible with Windows. Windows paths should use drive letters (e.g., 'C:\\').",
+            )
+
+    else:  # Linux/macOS
+        # Check for Windows drive letters in Linux/macOS paths
+        if re.match(r"^[a-zA-Z]:\\", path):
```
mwiebe, line 697: Paths on Windows can also have a UNC form that looks like `\\server\mount\path\to\file.png`. I remember us running into something similar with path validation before, we should double check the best thing to do, which is hopefully using something from the Python standard library.

test/unit/deadline_job_attachments/incremental_downloads/test_manifest_s3_downloads.py
```diff
@@ -513,3 +519,397 @@ def test_add_output_manifests_from_s3_edge_cases(fresh_deadline_config):
...
+                output_manifests,
+            )
+
+        # Verify paths were processed correctly
+        result_manifest = output_manifests[0][1]
+        for i, path in enumerate(linux_paths):
+            expected_path = os.path.normpath(os.path.join(local_storage_root, path))
+            assert result_manifest.paths[i].path == expected_path
+
+    @patch("os.name", "nt")  # Mock Linux/macOS environment
```
sakshie95, line 867: Isn't that windows? :)

sakshie95:
There is a security hotspot in the tests and some duplication in the new code that Sonar has detected

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -213,16 +215,35 @@ def _download_manifest_and_make_paths_absolute(
     )
     # Convert all the manifest paths to have absolute normalized local paths
     for manifest_path in manifest.paths:
-        manifest_path.path = os.path.normpath(os.path.join(root_path, manifest_path.path))
-        # TODO: Apply path mapping rules to manifest_path.path right here
+        original_path = manifest_path.path
+        if (
+            manifest_file_system_location_name
+            and manifest_file_system_location_name in local_file_system_locations
+        ):
```
mwiebe, line 222: It looks like this transformation will only work when the file system location names from storage profiles and the manifest root paths line up precisely. I believe the current implementation works this way, but that isn't a requirement of the system, so we should make the path mapping code more general. The way to generalize it is to produce something to perform path mappings, such as a list of path mapping rules, in the per-job code since each job has a storage profile associated with the submitting workstation. Then the parameter that goes in here is the path mapping object.

mwiebe (changes requested):

src/deadline/client/cli/_groups/attachment_group.py
benl-2023: Yeah, I tested exactly that case and it works.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
leongdl: I think you mean what the worker agent is doing. Given a set of asset roots, job storage profiles, compute the path mapping of 
```
[
  {
     input: `/mnt/work`
     output: `/home/david/project`
  }
.....
]
```

Then this code uses the path mapping to unpack to the local path by a dictonary lookup.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -236,7 +257,7 @@ def _get_manifests_to_download(
     Returns:
         A list of (rootPath, manifest_s3_key) tuples for the manifest objects that need to be downloaded.
     """
-    manifests_to_download: list[tuple[str, str]] = []
+    manifests_to_download: list[tuple[str, str, str]] = []
     for job_id, session_list in job_sessions.items():
         job = download_candidate_jobs[job_id]
         for session in session_list:
```
leongdl, line 263: This maybe a good place to create a job level manifest and compute the path mapping first.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -285,10 +308,18 @@ def _download_all_manifests_with_absolute_paths(
         A list of BaseAssetManifest objects containing local absolute file paths sorted by the last_modified timestamp.
     """
     # Get the list of (rootPath, manifest_s3_key) tuples to download from S3.
-    manifests_to_download: list[tuple[str, str]] = _get_manifests_to_download(
+    manifests_to_download: list[tuple[str, str, str]] = _get_manifests_to_download(
         queue["jobAttachmentSettings"]["rootPrefix"], download_candidate_jobs, job_sessions
     )
 
+    # Build a mapping between the file system location names and their path from the local storage profile.
+    local_file_system_locations = {}
```
leongdl, line 316: This bit of logic would be good to refactor into computing a general path mapping first.

PR #745, opened 2025-07-23
feat!: add asset type differentiation and name manifest accordingly for attachment upload api

godobyte:
- `attachment upload` currently takes manifest files, and use the file names as s3 object names for upload. However input and output manifests follows certain naming convention, specifically input manifest file follows `{root hash}_input` and output manifest follows `{root hash}_output`.
- There is still discussion about the API shape, we'll keep the API internal. 

- Give the option to specify the asset type and name accordingly.
- Rename `attachment_upload` to `_attachment_upload` to better reflect its beta status mentioned in the docstring.

The final manifest file name would follow the convention for input and output

- [N/A] Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended

  - We have a follow-up task to update job attachments related CLI together

Yes. We renamed `attachment_upload` to `_attachment_upload` to better reflect its beta status mentioned in the docstring. Callers need to update and call the `_attachment_upload` now.

No

----


test/unit/deadline_client/cli/test_cli_attachment.py
```diff
@@ -4,21 +4,68 @@
...
 """
 
-import os
-
 from click.testing import CliRunner
 
 from deadline.client.cli import main
 
 
-def test_cli_attachment_existence(fresh_deadline_config):
```
kenyrish-amazon, line 14: Just wondering why this test was deleted?

src/deadline/job_attachments/api/attachment.py
```diff
@@ -184,7 +187,7 @@ def attachment_upload(
             (
                 rule
                 for rule in path_mapping_rule_list
-                if rule.get_hashed_source_path(manifest.get_default_hash_alg()) in file_name
+                if rule.get_hashed_source_path(manifest.hashAlg) in file_name
```
kenyrish-amazon, line 190: Is there a reason why this is changed? If so it is used in the same file [here](https://github.com/aws-deadline/deadline-cloud/blob/a07f02d5215b2fff9510e6151ab00b3aa10fc47b/src/deadline/job_attachments/api/attachment.py#L101)

kenyrish-amazon (dismissed):
Left a few questions but looks good!

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -154,6 +158,14 @@ def attachment_download(
 @click.option(
     "--upload-manifest-path", default=None, help="File path for uploading the manifests to CAS."
 )
+@click.option(
```
leongdl: I don't think this is useful because how will customers know what is the type? 


I think we need to revisit the customer interface here as a CLI argument.

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -167,6 +179,7 @@ def attachment_upload(
     path_mapping_rules: str,
     s3_root_uri: str,
     upload_manifest_path: str,
+    asset_type: str,
```
leongdl: I think `upload_manifest_type` maybe a better name, if we are exposing this. 

Still, I don't think it makes sense to be in the CLI. It maybe more suitable in the API arg only.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -208,12 +211,22 @@ def attachment_upload(
         if rule.source_path_format:
             metadata["Metadata"]["file-system-location-name"] = rule.source_path_format
 
+        # Determine manifest file name based on asset type
+        if asset_type == AssetType.INPUT:
+            hashed_source_path = rule.get_hashed_source_path(manifest.get_default_hash_alg())
+            manifest_file_name = f"{hashed_source_path}_input"
```
leongdl: This `f` string should be its own utility function with a test.

`Thou shall never change file names`

src/deadline/job_attachments/api/attachment.py
```diff
@@ -208,12 +211,22 @@ def attachment_upload(
...
+        elif asset_type == AssetType.OUTPUT:
+            hashed_source_path = rule.get_hashed_source_path(manifest.get_default_hash_alg())
+            manifest_file_name = f"{hashed_source_path}_output"
```
leongdl: This `f` string should be its own utility function with a test.

`Thou shall never change file names`

src/deadline/client/cli/_groups/attachment_group.py
leongdl: i had an idea for a name for this - `upload_manifest_s3_file_name_format`, the options are INPUT, OUTPUT, OTHER. Then it can nicely fit with `upload_manifest_path`

godobyte: I used asset type simply because it's what we use for `manifest download`, where we have asset type being input, output, or all. I figured it's applicable and can make our naming consistent.

Maybe we can loop in more people to come up with a name we all align?

src/deadline/job_attachments/api/attachment.py
godobyte: Good catch, I'll update the other place.

test/unit/deadline_client/cli/test_cli_attachment.py
godobyte: We were using env var to control the visibility to user when we just started developing. The logic for the env var is already removed, and the test is no longer applicable.

src/deadline/job_attachments/api/attachment.py
godobyte: Sure I was thinking where would be a good place for this, and putting here just for the time being. I can move to a function for now.

src/deadline/client/cli/_groups/attachment_group.py
leongdl: Yeah, and we need to think carefully if we really did remove the beta flag.

src/deadline/job_attachments/api/attachment.py
leongdl: Thanks :)

src/deadline/client/cli/_groups/attachment_group.py
godobyte: This is the same reasoning as the one I followed for the CLI. I was trying to make the term consistent. Maybe we should spend a bit more time to polish it and think about other suggestions.

src/deadline/client/cli/_groups/attachment_group.py
godobyte: Keeping it here for visibility, open to name suggestions. Will integrate the name we align on in the final revision.

godobyte: Discussed offline, decided to remove this from the command and only keep it in api.

src/deadline/job_attachments/models.py
```diff
@@ -469,6 +474,32 @@ class PathMappingRule:
     def get_hashed_source_path(self, hash_alg: HashAlgorithm) -> str:
         return hash_data(self.source_path.encode("utf-8"), hash_alg)
```
mwiebe, line 475: This needs to include the location name if it's available too, not just the root path. It needs to produce the same value as https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/asset_sync.py#L536

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -161,7 +164,7 @@ def attachment_download(
 )
 @click.option("--json", default=None, is_flag=True, help="Output is printed as JSON for scripting")
 @_handle_error
-def attachment_upload(
+def _attachment_upload(
```
leongdl, line 167: Does this function `_` matter?

src/deadline/job_attachments/api/__init__.py
```diff
@@ -2,15 +2,15 @@
 
 __all__ = [
     "attachment_download",
-    "attachment_upload",
+    "_attachment_upload",
```
leongdl, line 5: IMO all of these should be `_`

mwiebe (changes requested):

src/deadline/job_attachments/api/__init__.py
godobyte: I was planning to update that in a separate PR, not to couple with the additional changes here.

src/deadline/client/cli/_groups/attachment_group.py
godobyte: Since we want to indicate they're internal and beta, I'd add the leading underscores to both functions to be more explicit.

godobyte:
This PR is no longer applicable, there are other changes to replace the logic here.

PR #746, opened 2025-07-23
fix: UI Elements in Deadline Config Dialog stretch and squish undesirably

ttblanchard:

The AWS Deadline Cloud configuration dialog had UI/UX issues that impacted user experience:

- **Poor window sizing**: Small default Window size compressed visually elements 
- **Unnatural group box appearance**: Group boxes stretched vertically to fill space instead of sizing to content
- **UI element truncation**: "Remove Selected" button was cut off at default window width

- Implemented screen-aware `sizeHint()` that calculates optimal height based on available screen space
- Modified `DeadlineWorkstationConfigWidget` to use natural group box sizing instead of stretched layout and added consistent spacing.
- `DeadlineWorkstationConfigWidget._build_ui()`: Removed stretch factors from group box widgets, added vertical spacer
- Enhanced form layouts with `ExpandingFieldsGrow` for consistent horizontal expansion

- Dialog opens at optimal size for any screen resolution without unnecessary scrolling and UI element squishing
- Adaptive sizing works across different display configurations
- Maintains full resizing functionality for user preference

- Manual testing across different screen resolutions and window sizes
- Verified adaptive sizing behavior and scroll bar functionality
- Confirmed all UI elements display properly without truncation

**BEFORE**

Squishy

**AFTER**

On a small monitor:
SmallMonitorDefault

On a higher res/larger monitor:
LargeMonitorDefault

Stretched/compressed:
VerticalCompress-HorizontalStretch

- Added inline comments explaining adaptive sizing logic

No - UI improvements only, no API changes.

No - limited to Qt widget layout and window sizing.

----


src/deadline/client/ui/dialogs/deadline_config_dialog.py
```diff
@@ -85,14 +86,39 @@ def __init__(self, parent=None) -> None:
...
+
+        # Calculate optimal dialog height based on screen size
+        content_height = 850  # Height needed to show all content without scroll bars
+        max_dialog_height = int(available_height * 0.9)  # 90% of screen height
+
+        # Use smaller of content height or max screen percentage
+        optimal_height = min(content_height, max_dialog_height)
+
+        # Set minimum height to ensure usability
+        optimal_height = max(optimal_height, 600)
```
mwiebe: This could only increase the height to be larger than the maximum dialog height, since the content height starts at 850. Should remove the line?

ttblanchard: Yea, I see what you are saying. On a really small display this would enforce a min size of 600 - which would potentially extend the dialog beyond the display. 

I think on displays that small, usability is going to be pretty heavily impacted regardless. That said I'll remove if you think it's the right thing.

AWS-Samuel: I think it's worth removing. Yes, the current code only really impacts displays with a height of < 600, but might happen more often in cases where the display is a scalable window such as when connecting to a remote desktop.

AWS-Samuel (dismissed):

AWS-Samuel (approved):

lucaseck (approved):

PR #748, opened 2025-07-23
chore: Replace jsonschema with handwritten validation

evanspearman-a:

A recent version of jsonschema has rpds-py (https://github.com/crate-py/rpds) as a transitive dependency. rpds-py is Python bindings for a rust package. While there's no inherent issue with including a rust-based dependency in deadline-cloud, rpds-py does not include a NOTICES file or equivalent attributing rust dependencies. This is a problem for us because we have to redistribute our dependencies and it would be difficult for us to attribute the rust dependency tree.

Our usage of jsonschema is pretty trivial, so I opted to just write code to do the validations manually.

OSS license compliance.

- Updated tests
- Ran `hatch run test`
- Submitted a job with attachments
- Submitted a job with attachments using a client with these changes applied
- The job was https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/test/e2e/job_attachment_bundle/dep_data_flow/linux_bundle/template.yaml and I ran it on a worker agent that had deadline cloud with this change applied and the `ASSET_SYNC_JOB_USER_FEATURE` flag set.

N/A

No, it removes them

No

I don't think so. The new validation mirrors what jsonschema was doing.

----


leongdl (dismissed):

leongdl (dismissed):

src/deadline/job_attachments/asset_manifests/decode.py
```diff
@@ -26,20 +24,97 @@ def _get_schema(version) -> dict[str, Any]:
         return json.load(schema_file)
 
 
+_REQUIRED_FIELDS_2023_03_03: list[str] = [
```
godobyte: Does this have to be in the same file with decoding? Could we formalize it in its own file with versioning to avoid accidentally touching it?

src/deadline/job_attachments/asset_manifests/decode.py
```diff
@@ -26,20 +24,97 @@ def _get_schema(version) -> dict[str, Any]:
...
+        return False, "paths must be a list"
+    elif len(paths) < 1:
+        return False, "paths must have a least one item"
+    else:
+        for path_object in paths:
+            ok, message = _validate_path_2023_03_03(path_object)
+            if not ok:
+                return False, message
+
+    return True, None
```
godobyte: What do you think about making the validation function more generic? We could probably re-use existing definitions in `_REQUIRED_FIELDS_2023_03_03` and `_PATH_REQUIRED_FIELDS_2023_03_03` and avoid hard coding the value.

src/deadline/job_attachments/asset_manifests/decode.py
evanspearman-a: Seems reasonable.

src/deadline/job_attachments/asset_manifests/decode.py
evanspearman-a: I think that's premature. We don't know what a future version will look like. If we release a future specification, we can refactor the code once we know what would actually be shared.

src/deadline/job_attachments/asset_manifests/decode.py
evanspearman-a: Updated. I re-ran the tests, but I didn't do the full test with the worker agent since this is just a refactor and no logic has changed.

src/deadline/job_attachments/asset_manifests/decode.py
godobyte: Keeping the validation logic isolated in a versioned file is a good first step. This encapsulation prevents accidental modifications to the hard-coded validation rules and maintains clear version boundaries. 

With this versioned validation in place, I agree we can evaluate potential abstraction patterns when needed.

godobyte (approved):

leongdl (approved):

PR #751, opened 2025-07-25
feat!: refactor the autorization UI dialog

### LINKED ISSUE #727: Maintenance: Improve submitter messaging when not logged in
### Description

The submitter window presents a fairly non-user friendly error message when the monitor is not logged in.

Image

The all-caps red text that says "UNAVAILABLE" makes it seem like something has broken and may convince some non-technical users that something was improperly installed.

### Solution

We should change the messages to look less like critical error messages and find a more gentle way to inform the user of what they need to do. We should also take out the underscore. Something like "Login Required" would work.

lucaseck:
Fixes: #727 

We have received feedback multiple times that the authentication status widgets at the bottom of the submitters that use deadline-cloud are unclear and ugly(see #727)

Reworked the authentication status widget section to more nicely display the authentication status and provide more helpful hints to users when there are issues with their current profile or configuration.

Additionally fixed a bug where changing profiles in the config menu and then selecting cancel would result in the profile change persisting, but the default farm/queue still being for the previous profile

The authentication status is displayed in a nicer manner and there are now help dialogs to point users in the right direction when there are problems with their credentials

##### User logged in using a DCM or named profile
image

##### User logged out using a DCM profile
image

##### User attempting to use a profile without the right permissions
image

More Info Text
image

##### User attempting to use a profile with a misconfigured credential process
image

More Info Text
image

##### Refreshing state when switching profiles
image

##### Dropdown menu to change profiles or log out
* The log out option only shows if using a DCM profile
* The switch profile option is only available if not already in the configuration dialog
image

Example in the configuration menu where switch profile and logout don't show when using a named profile
image

* Validated all button functionality and UI state changes as well as did a job submission when using Pyside6 and doing a `deadline bundle gui-submit`
* Validated all button functionality and UI state changes as well as did a job submission when using Pyside2 and doing a `deadline bundle gui-submit`
* Validated that the UI appeared correctly in a few of the DCC submitters on Windows to catch any glaring issues as a smoke test. As this is a breaking change each DCC repo will need to test with these changes as they choose to upgrade.
  * No issues in Nuke 15 on Windows
  * No issues in Maya 2024 on Windows
  * No issues in Houdini 19.5/20.5 on Windows
  * No issues in Blender 3.6 on Windows
* Had a team member run manual tests on Mac OS with Pyside6 and Pyside2 running `deadline bundle gui-submit` and verifying the behaviour and look
* Ran my own tests on RHEL 9 and verified the UI looked and behaved correctly with both Pyside2 and Pyside6

  - Yes(not that they test any UI functionality as far as I know) 
  - No 
  - N/A

In the commit alone

Yes, and the commit has a description with the breaking details.

```
BREAKING_CHANGE: The DeadlineAuthenticationStatusWidget class has been
rewritten and the AuthenticationStatusGroup widget class has been removed.
The login/logout buttons have been removed from the SubmitJobToDeadlineDialog
and the DeadlineConfigDialog.
```

No

----

### DIFF
--- a/src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
+++ b/src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
@@ -2,112 +2,417 @@
 
 """
 Provides a widget to place in AWS Deadline Cloud submitter dialogs, that shows
-the current status of AWS Deadline Cloud authentication and API.
+the current status of AWS Deadline Cloud authentication.
 The current status is handled by DeadlineAuthenticationStatus.
 """
 
-from logging import getLogger
-from typing import Optional
+import enum
 
-from qtpy.QtCore import Qt
+from dataclasses import dataclass
+from logging import getLogger
+from typing import Callable, Union, Dict
+
+from qtpy.QtCore import Signal
 from qtpy.QtWidgets import (  # pylint: disable=import-error; type: ignore
-    QFormLayout,
-    QGroupBox,
     QHBoxLayout,
     QLabel,
     QWidget,
+    QApplication,
+    QStyle,
+    QFrame,
+    QMenu,
+    QPushButton,
+    QMessageBox,
 )
 
 from ... import api
 from ..deadline_authentication_status import DeadlineAuthenticationStatus
+from ...config import config_file
 
 logger = getLogger(__name__)
 
 
+class RightAlignedQMenu(QMenu):
+    """
+    A modified QMenu that positions itself to the right of its parent widget
+    instead of the default left positioning.
+    """
+
+    def showEvent(self, event):
+        """
+        Override showEvent to position the menu to the right of the parent button.
+        """
+        if self.parent():
+            parent_top_right = self.parent().mapToGlobal(self.parent().rect().topRight())
+            self.move(parent_top_right.x(), parent_top_right.y())
+
+        super().showEvent(event)
+        self.adjustSize()  # The menu was displaying incorrectly on Mac until a resize event
+
+
+@dataclass
+class AuthenticationStateConfig:
+    """
+    Configuration dataclass that encapsulates all the visual and behavioral
+    settings needed to display the authentication widget in different states,
+    including which icon to show, what text to display, and which buttons should
+    be visible.
+
+    Attributes:
+        icon (QStyle.StandardPixmap): The standard Qt icon to display for this state.
+        text (Union[str, Callable[[], str]]): The text to display, either as a string
+            or a callable that returns a string for dynamic text.
+        switch_profile_button_visible (bool): Whether the switch profile button should
+            be visible. If the button isn't visible that switch profile menu item will be.
+            Defaults to False.
+        login_visible (bool): Whether the login button should be visible. Defaults to False.
+        logout_visible (Union[bool, Callable[[], bool]]): Whether the logout option should
+            be visible, either as a boolean or callable returning boolean. Defaults to False.
+        more_info_visible (bool): Whether the more info button should be visible.
+            Defaults to False.
+    """
+
+    icon: QStyle.StandardPixmap
+    text: Union[str, Callable[[], str]]
+    switch_profile_button_visible: bool = False
+    login_visible: bool = False
+    logout_visible: Union[bool, Callable[[], bool]] = False
+    more_info_visible: bool = False
+
+
+class AuthenticationState(enum.Enum):
+    """
+    Enum defining all the possible states that the authentication status widget can display,
+    each corresponding to a different combination of authentication status and
+    API availability.
+
+    Values:
+        REFRESHING: Authentication status is being loaded or refreshed.
+        AUTHENTICATED_READY: User is authenticated and has API access to list-farms
+        AUTHENTICATED_NO_API: User is authenticated but lacks API access permissions to list-farms
+        NEEDS_LOGIN: User needs to log in to authenticate with a DCM profile
+        CONFIGURATION_ERROR: There is a configuration issue with the AWS profile.
+        UNEXPECTED_ERROR: An unknown or unexpected error occurred during authentication.
+    """
+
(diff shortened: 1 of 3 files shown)


justinsaws (dismissed):

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
```diff
@@ -2,112 +2,411 @@
...
+        be visible.
+
+        Returns:
+            Dict[AuthenticationState, AuthenticationStateConfig]: A dictionary mapping
+                each authentication state to its corresponding UI configuration.
+        """
+        return {
+            AuthenticationState.REFRESHING: AuthenticationStateConfig(
+                icon=QStyle.StandardPixmap.SP_BrowserReload,
+                text=lambda: self._get_profile_name(),
```
edwards-aws: nit: feel free to not address this, I'm pretty sure the lambdas that only call a function can just be the function. So I think the following would work. 

`text=self._get_profile_name`

Unless type hinting gets weird because it expects no arguments, but `self._get_profile_name` has the implicit `self` argument

The being said I don't think this is a huge deal, this isn't on a hot path or anything I think.

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
```diff
@@ -2,112 +2,411 @@
...
+                "  • deadline:CreateJob\n"
+                "  • deadline:GetJob\n"
+                "  • deadline:GetQueue\n"
+                "  • deadline:GetQueueEnvironment\n"
+                "  • deadline:GetStorageProfileForQueue\n"
+                "  • deadline:GetStorageProfile\n"
+                "  • deadline:ListFarms\n"
+                "  • deadline:ListQueues\n"
+                "  • deadline:ListQueueEnvironments\n"
+                "  • deadline:ListStorageProfilesForQueue"
```
edwards-aws, line 332: Are there docs we could link to instead? I'm worried about this getting out of date, although there's always the worry of having dead links in code. What do you think?

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
```diff
@@ -2,112 +2,411 @@
...
+            )
+        elif self._status.auth_status == api.AwsAuthenticationStatus.CONFIGURATION_ERROR:
+            title = "Issue With Profile Configuration"
+            message = (
+                f"There is a configuration issue with the profile '{self._get_profile_name()}'.\n\n"
+                "To resolve this issue:\n"
+                "• Verify your AWS config and credentials files are correct\n"
+                "• Verify that any credential process being used is able to retrieve the credentials\n"
+                "• Verify that the correct AWS region is set\n"
+                "• If you are not using a AWS Deadline Cloud Monitor profile, check if your credentials have expired"
```
edwards-aws: Is there any more info or documentation that we could link to that tell customers how to do these things? I always get a little frustrated when I get an error message tells me to do something that I don't know how to do and isn't obvious.

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
```diff
@@ -2,112 +2,411 @@
...
+            )
         else:
-            color = "red"
-            text = "UNAVAILABLE"
-        self.deadline_authorized_group.label.setText(f"<b style='color:{color};'>{text}</b>")
+            title = "Unknown Issue With Configured Profile"
+            message = (
+                f"There was an unknown issue when trying to authenticate with the profile '{self._get_profile_name()}'.\n\n"
+                "Check any available console logs for errors to try and diagnose the problem."
```
edwards-aws: Is there anyway that we can guide customers where to find these logs?

edwards-aws (commented):
Great work on this Lucas! Just a couple questions about how we could make the error messages even better.

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
lucaseck: I had asked around and there doesn't appear to be a docs page anywhere with this 😞 . Good callout though and can check if we should add to the docs maybe in https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html. The only thing is I'm not sure we really document anywhere right now that you don't have to use a DCM profile even in those docs everything is assuming DCM is being used.

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
lucaseck: For sure, can try to add some more details. I had figured keep it a little open ended to not scare people who have no idea what an aws config file is and probably shouldn't be poking around in it. Some more details I don't think are the end of the world though.

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
lucaseck: All just depends where the DCC logs to. Like for Maya it'll be in their console, for Nuke I think it's either in the terminal you launched Nuke from or a popup. Blender will be in the blender console etc. I can add a note that it will in most cases be either in the terminal that the DCC or standalone GUI were launched from or in the built-in console in the DCC, but they'd still have to know where that is.

lucaseck: For context as well, for that message to print right now there would have to be a state that doesn't exist. Everything should be being caught as configuration errors unless someone adds a new enum value etc or something goes kaboom. Most users will see the configuration error always for issues.

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
lucaseck: Good catch. Fixed.

edwards-aws (approved):

justinsaws (approved):

PR #758, opened 2025-07-30
refactor(experimental): Refinements in incremental download

mwiebe:

We updated the deadline:SearchJobs API to enable filtering by ENDED_AT. This can reduce the number of candidate jobs to filter client-side. The log output "DROPPED Job" is unclear that it's dropped from tracking, not from downloading, should change that wording for clarity.

* Add service-side filtering of jobs by ENDED_AT. Update boto3 dependency range that supports the parameter.
* Rename "DROPPED Job:" to "FINISHED TRACKING Job:" to make the log output of the incremental download more straightforward to understand.
* Require Python >= 3.9 for the incremental download command. It relies
on filtering by endedAt in the deadline:SearchJobs call available with
boto3 1.39.10 and later, released after the Python 3.8 April 2025 date
documented for boto3 in
https://aws.amazon.com/blogs/developer/python-support-policy-updates-for-aws-sdks-and-tools/

Slightly improved performance, and better logging.

Ran unit tests, tested the operation on a queue with randomly submitted jobs.

No

No

No

No

----

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -18,9 +18,23 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
     uses: aws-deadline/.github/.github/workflows/reusable_python_build.yml@mainline
     with:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
       ref: ${{inputs.tag}}
+
+  TestPy38:
+    name: Python
+    strategy:
+      matrix:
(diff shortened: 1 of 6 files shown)


sakshie95:
Looks good to me. Curious how the PR improves performance for the CLI as said in the description.

sakshie95 (approved):

baxeaz (approved):

mwiebe:
> Looks good to me. Curious how the PR improves performance for the CLI as said in the description.

Thanks! Before this change, the SearchJobs calls were getting all the jobs in the listed terminal states, and then filtering that list client-side to just the ones with a recent endedAt timestamp. If you've been submitting jobs to the queue for a while, and have 10,000 jobs that are completed, that means every incremental operation gets all of those jobs and filters them down to the few that ended recently. After this change, the SearchJobs call performs that filtering, so the client only gets those few jobs. That means fewer SearchJobs calls and less client-side processing.

leongdl (approved):

PR #764, opened 2025-08-04
feat: Add functions to support path mapping

mwiebe:

For the incremental download work, we need functionality to generate path mapping rules from a pair of storage profiles and then to apply those rules to a large number of paths.

Write functions to support path mapping between two storage profiles, with a structure for applying them to many paths.

* Functions are marked internal-only for now.
* _generate_path_mapping_rules takes two storage profiles and generates a list of path mapping rules.
* _PathMappingRuleApplier builds an acceleration structure from a list of path mapping rules for mapping a large number of paths.
* Unit tests for new funcitonality.

Wrote unit tests.

No

No

No 

No 

----

### DIFF
--- a/src/deadline/job_attachments/_path_mapping.py
+++ b/src/deadline/job_attachments/_path_mapping.py
@@ -1,0 +1,186 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from __future__ import annotations
+
+from pathlib import Path, PurePosixPath, PureWindowsPath
+from typing import Any, Callable, Optional, Union
+
+from .models import PathFormat, PathMappingRule, StorageProfileOperatingSystemFamily
+
+__all__ = ["_generate_path_mapping_rules", "_PathMappingRuleApplier"]
+
+"""
+This file contains functionality related to path mapping rules. This functionality is internal-only for now,
+to be marked as public after developing some experience with it.
+"""
+
+
+def _generate_path_mapping_rules(
+    source_storage_profile: dict[str, Any],
+    destination_storage_profile: dict[str, Any],
+) -> list[PathMappingRule]:
+    """
+    Given a pair of storage profiles, generate all the path mapping rules to transform paths
+    from the source to the destination.
+
+    A mapping rule is generated for every file system location name that's shared between
+    the storage profile regardless of the type (SHARED vs LOCAL), to account for the broadest
+    possible storage profile configurations.
+
+    Args:
+        source_storage_profile: A storage profile as returned by boto3 deadline.get_storage_profile or
+            deadline.get_storage_profile_for_queue.
+        destination_storage_profile: A storage profile as returned by boto3 deadline.get_storage_profile or
+            deadline.get_storage_profile_for_queue.
+    Returns:
+        A list of path mapping rules to transform paths.
+    """
+    # If the source and destination are identical, no transformation is needed
+    if (
+        source_storage_profile["storageProfileId"]
+        == destination_storage_profile["storageProfileId"]
+    ):
+        return []
+
+    # Put the locations into dictionaries to match up the names
+    source_locations = {
+        location["name"]: location for location in source_storage_profile["fileSystemLocations"]
+    }
+    destination_locations = {
+        location["name"]: location
+        for location in destination_storage_profile["fileSystemLocations"]
+    }
+
+    if source_storage_profile["osFamily"] == StorageProfileOperatingSystemFamily.WINDOWS.value:
+        source_path_format = PathFormat.WINDOWS.value
+    else:
+        source_path_format = PathFormat.POSIX.value
+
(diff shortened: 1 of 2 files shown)


mwiebe:
SonarCloud warnings are all paths used in tests to validate mapping, there is no issue with them.

test/unit/deadline_job_attachments/test_path_mapping.py
```diff
@@ -0,0 +1,337 @@
...
+        applier.transform("/mnt/shared/file with spaces.txt")
+        == dest_paths.path1 / "file with spaces.txt"
+    )
+
+    # The second rule never applies, because the first rule is shorter and is used instead
+    assert applier.transform("/mnt/shared/projects") == dest_paths.path1 / "projects"
+    assert (
+        applier.transform("/mnt/shared/projects/file.txt")
+        == dest_paths.path1 / "projects" / "file.txt"
+    )
```
crowecawcaw: Seems like the more specific / longer rule should take precedence?

mwiebe: Yes, I've implemented the wrong behavior here. We did some review of where we mention this case in other docs and code, and we found two places:

* https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run#applying-path-mapping-rules-within-a-job-template
* https://github.com/OpenJobDescription/openjd-sessions-for-python/blob/ac475ba39847522d65d177b7aedafa716ff588bd/src/openjd/sessions/_session.py#L389-L392

In addition to fixing things here to behave that way, I think it's worth updating that wiki to use more clear language and to give an example of why it's important.

mwiebe: I've updated the implementation to do this, and added a link to the openjd-specifications wiki in the docstring.

crowecawcaw (approved):

src/deadline/job_attachments/_path_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+    This class provides an accelerated implementation for transforming paths according to a list of
+    path mapping rules. For details about how rules are applied, see the documentation
+    https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run#applying-path-mapping-rules-within-a-job-template
+
+    When mapping a path, the most specific rule is the one that applies. For example, if there are two rules
+        * '/mnt/Projects -> X:\\Projects'
+        * '/mnt/Projects/Special -> Y:\\'
+    then '/mnt/Projects/Special/data.txt' maps to 'Y:\\data.txt', not to 'X:\\Projects\\Special\\data.txt'.
+
+    The implementation uses a trie data structure for acceleration, as follows:
```
sakshie95, line 84: Interesting implementation :))

src/deadline/job_attachments/_path_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+            self.source_path_format = path_mapping_rules[0].source_path_format
+            if not all(
+                rule.source_path_format == self.source_path_format for rule in path_mapping_rules
+            ):
+                formats = list({rule.source_path_format for rule in path_mapping_rules})
+                raise ValueError(
+                    f"The path mapping rules included multiple source path formats {', '.join(formats)}, only one is permitted."
+                )
+
+            if self.source_path_format == PathFormat.POSIX.value:
```
sakshie95, line 124: Not blocking but would've been nice to have a base interface for this and windows/posix specific implementations so this logic can be simpler to read. Given that the trie works differently for the two implementation and has a base functionality that stays the same.

sakshie95 (approved):

src/deadline/job_attachments/_path_mapping.py
mwiebe: Cool, thanks! There are lots of different ways you can slice up an abstraction like this. :)

What you're suggesting would more closely match how Python's pathlib has its PurePath/PureWindowsPath/PurePosixPath/Path/... inheritance diagram shown in https://docs.python.org/3/library/pathlib.html

src/deadline/job_attachments/_path_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    Given a pair of storage profiles, generate all the path mapping rules to transform paths
+    from the source to the destination.
+
+    A mapping rule is generated for every file system location name that's shared between
+    the storage profile regardless of the type (SHARED vs LOCAL), to account for the broadest
+    possible storage profile configurations.
+
+    Args:
+        source_storage_profile: A storage profile as returned by boto3 deadline.get_storage_profile or
```
leongdl, line 31: Super nit - I had to go read the docs to check the strucutre - https://boto3.amazonaws.com/v1/documentation/api/1.35.43/reference/services/deadline/client/get_storage_profile.html

Would be nice to link it some how.

PR #773, opened 2025-08-07
feat(experimental): Add storage profile support for incremental download

mwiebe:

Ben, working on https://github.com/aws-deadline/deadline-cloud/pull/743, moved to other priorities before being able to address code review comments on that PR. We need to support storage profile and source the file system locations from the submitting workstation and the downloading workstation to map output files as specified by the manifests.

We considered two CLI interface options, whether to try to do the best thing possible, or to require storage profiles by default with an explicit override. We chose the latter so that when someone starts using the command but haven't set up storage profiles, they learn this immediately and what benefit storage profiles can give them, but then have an immediate way to ignore them if they see it doesn't apply.

Take elements from the previous PR, along with consideration of the comments there, into this new PR.

* Change the default behavior of 'deadline queue
  incremental-output-download' to require storage profiles.
   * Add an option --ignore-storage-profiles that lets you override this behavior
     and ignore them. This option should only be used if all jobs in the
     queue are submitted and downloaded from the same machine.
   * Add an option --storage-profile-id that lets you set the storage
     profile to download for independently of the configuration file.
* Add the localStorageProfileId to IncrementalDownloadState so that subsequent runs can validate that the value remains identical across runs.
* Generalize the path summarization code to accept a PathFormat (WINDOWS or POSIX) so that the incremental download can summarize paths that had no path mapping rule appropriately for the source operating system.
* Modify the path summarization code to de-duplicate repeated paths within the input.
  * Add a test case for this.
* Modify the path summarization code to normalize all Windows paths with "\\" separators instead of "/". Update the tests to reflect this.
* Change the unit tests in test_cli_queue_incremental_download.py to be based on moto instead of mocking boto3.Session. Used the documented https://docs.getmoto.org/en/latest/docs/services/patching_other_services.html for mocking deadline service calls. This allows features like paginators to work in the tests, that were not working with the boto3.Session mock approach.

The incremental download command will download job outputs to the correct locations when they were run on different operating systems. It uses storage profile configurations that customers can use to input corresponding paths, e.g. projects are on '/Volumes/Projects' on Mac but 'X:\Projects' on Windows.

Added unit tests for the new functionality. Created a farm configuration with various storage profiles on Linux and Windows, and manually confirmed expected behavior in a variety of cases.

New options are in the CLI help output.

No

No

No

----

### DIFF
--- a/src/deadline/client/cli/_groups/queue_group.py
+++ b/src/deadline/client/cli/_groups/queue_group.py
@@ -29,7 +29,6 @@ from ....job_attachments._incremental_downloads.incremental_download_state impor
     IncrementalDownloadState,
 )
 
-PID_FILE_NAME = "incremental_output_download.pid"
 DOWNLOAD_CHECKPOINT_FILE_NAME = "download_checkpoint.json"
 
 
@@ -213,6 +212,10 @@ def queue_get(**args):
 @cli_queue.command(name="incremental-output-download")
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
+@click.option(
+    "--storage-profile-id",
+    help="The storage profile to use for mapping paths to local. Cannot be used together with --ignore-storage-profiles",
+)
 @click.option("--json", default=None, is_flag=True, help="Output is printed as JSON for scripting.")
 @click.option(
     "--bootstrap-lookback-minutes",
@@ -231,6 +234,13 @@ def queue_get(**args):
     "--force-bootstrap",
     is_flag=True,
     help="Forces command to start from the bootstrap lookback period and overwrite any previous checkpoint.\n"
+    "Default value is False.",
+    default=False,
+)
+@click.option(
+    "--ignore-storage-profiles",
+    is_flag=True,
+    help="Ignores the storage profile configuration. Only use if all jobs in the queue are submitted and downloaded from the same machine. Downloads all jobs to unmapped paths regardless of operating system.\n"
     "Default value is False.",
     default=False,
 )
@@ -263,6 +273,7 @@ def incremental_output_download(
     bootstrap_lookback_minutes: float,
     checkpoint_dir: str,
     force_bootstrap: bool,
+    ignore_storage_profiles: bool,
     dry_run: bool,
     **args,
 ):
@@ -287,6 +298,11 @@ def incremental_output_download(
             "The incremental-output-download command requires Python version 3.9 or later"
         )
 
+    if ignore_storage_profiles and args.get("storage_profile_id") is not None:
+        raise click.UsageError(
+            "Options '--storage-profile-id' and '--ignore-storage-profiles' cannot be provided together"
+        )
+
     logger: ClickLogger = ClickLogger(is_json=json)
 
     # Expand '~' to home directory and create the checkpoint directory if necessary
@@ -309,13 +325,49 @@ def incremental_output_download(
     queue_id = config_file.get_setting("defaults.queue_id", config=config)
     boto3_session: boto3.Session = api.get_boto3_session(config=config)
 
-    # Get download progress file name appended by the queue id - a unique progress file exists per queue
-    download_checkpoint_file_name: str = f"{queue_id}_{DOWNLOAD_CHECKPOINT_FILE_NAME}"
+    deadline = boto3_session.client("deadline")
+
+    if ignore_storage_profiles:
+        local_storage_profile_id = None
+        logger.echo("Ignoring all storage profiles.")
+    else:
+        local_storage_profile_id = config_file.get_setting(
+            "settings.storage_profile_id", config=config
+        )
+        if not local_storage_profile_id:
+            raise DeadlineOperationError(
+                "The incremental-output-download operation requires a storage profile configured locally\n"
+                "or provided with the --storage-profile-id option in order to determine file system paths\n"
+                "for download. Storage profiles are used to generate path mappings when a job was submitted\n"
+                "from a machine with a different operating system or file system mount locations than the download machine. \n\n"
+                "See https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/modeling-your-shared-filesystem-locations-with-storage-profiles.html\n\n"
+                "If you only submit and download jobs from one machine, you can use the --ignore-storage-profiles option\n"
+                "to ignore the storage profiles, and download all job outputs to whereever the paths are configured.\n"
+                "This is not recommended if more than one machine will submit or download jobs."
+            )
+
+        try:
+            local_storage_profile = deadline.get_storage_profile_for_queue(
+                farmId=farm_id,
+                queueId=queue_id,
+                storageProfileId=local_storage_profile_id,
+            )
+        except ClientError as e:
+            id_source = (
+                "configured locally"
+                if args.get("storage_profile_id") is None
+                else "provided with --storage-profile-id"
+            )
+            raise DeadlineOperationError(
+                f"Could not retrieve the storage profile {local_storage_profile_id!r}, {id_source}, from Deadline Cloud:\n{e}"
+            )
+
+    # Get download progress file name appended by the queue id and storage profile id - a unique progress file exists per queue/storage profile
+    download_checkpoint_file_name: str = f"{queue_id}_{local_storage_profile_id or 'ignore-storage-profiles'}_{DOWNLOAD_CHECKPOINT_FILE_NAME}"
 
     # Get saved progress file full path now that we've validated all file inputs are valid
     checkpoint_file_path: str = os.path.join(checkpoint_dir, download_checkpoint_file_name)
 
-    deadline = boto3_session.client("deadline")
     queue = deadline.get_queue(farmId=farm_id, queueId=queue_id)
     if "jobAttachmentSettings" not in queue:
         raise DeadlineOperationError(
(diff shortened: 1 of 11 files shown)


src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -213,6 +212,10 @@ def queue_get(**args):
 @cli_queue.command(name="incremental-output-download")
 @click.option("--farm-id", help="The AWS Deadline Cloud Farm to use.")
 @click.option("--queue-id", help="The AWS Deadline Cloud Queue to use.")
+@click.option(
```
leongdl, line 215: Can we add the usual reg-ex checks for the storage profile ID? Its too easy to type in the wrong ID and then getting confused with some call stack.

mwiebe: I looked at the other place with a `--storage-profile-id` option, in the job bundle submission, but didn't find code matching what you described. Do you have more details?

leongdl: I mean apply a general `sp-{uuid}` input validation generally so customers get early notice of a failure. Instead of getting a backend error message mid way through the code execution.

Fail early on start, with a simple and easier to understand message.

mwiebe: I don't think adding a regex here is a good idea, but calling get_storage_profile_for_queue at the start of the command to fail early on start for malformed ids and other issues sounds nice. Here's the error message the service returns for a malformed id:

```
$ aws deadline get-storage-profile-for-queue --farm-id `deadline config get defaults.farm_id` --queue-id `deadline config get defau
lts.queue_id` --storage-profile-id xyz --profile `deadline config get defaults.aws_profile_name`

An error occurred (ValidationException) when calling the GetStorageProfileForQueue operation: Total 1 validation error found
 storageProfileId
 string does not match regex "^sp-[0-9a-f]{32}$"
```

mwiebe: I've updated the PR to get the storage profile early and write a better message:

```
$ deadline queue incremental-output-download --force-bootstrap --dry-run --bootstrap-lookback-minutes 3000 --storage-profile-id sp-12345
Could not retrieve the storage profile 'sp-12345', provided with --storage-profile-id, from Deadline Cloud:
An error occurred (ValidationException) when calling the GetStorageProfileForQueue operation: Total 1 validation error found
 storageProfileId
 string does not match regex "^sp-[0-9a-f]{32}$"
```

leongdl (approved):

src/deadline/client/cli/_groups/queue_group.py
leongdl: Thanks! This would work just as well.

leongdl (dismissed):

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -29,7 +29,6 @@
     IncrementalDownloadState,
 )
 
-PID_FILE_NAME = "incremental_output_download.pid"
 DOWNLOAD_CHECKPOINT_FILE_NAME = "download_checkpoint.json"
```
sakshie95, line 32: I wouldn't conflate the pid file with the checkpoint file name cz IMO it's harder to understand when it's ``queue123_download_checkpoint.json.pid``

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -309,13 +325,49 @@ def incremental_output_download(
...
-    # Get download progress file name appended by the queue id - a unique progress file exists per queue
-    download_checkpoint_file_name: str = f"{queue_id}_{DOWNLOAD_CHECKPOINT_FILE_NAME}"
+    deadline = boto3_session.client("deadline")
+
+    if ignore_storage_profiles:
+        local_storage_profile_id = None
+        logger.echo("Ignoring all storage profiles.")
+    else:
+        local_storage_profile_id = config_file.get_setting(
+            "settings.storage_profile_id", config=config
```
sakshie95, line 335: may not be local storage profile anymore when we pick it from the settings?

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -309,13 +325,49 @@ def incremental_output_download(
...
+        )
+        if not local_storage_profile_id:
+            raise DeadlineOperationError(
+                "The incremental-output-download operation requires a storage profile configured locally\n"
+                "or provided with the --storage-profile-id option in order to determine file system paths\n"
+                "for download. Storage profiles are used to generate path mappings when a job was submitted\n"
+                "from a machine with a different operating systemor file system mount locations than the download machine. \n\n"
```
sakshie95: nit: typo "system or"

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -309,13 +325,49 @@ def incremental_output_download(
...
+                "See https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/modeling-your-shared-filesystem-locations-with-storage-profiles.html\n\n"
+                "If you only submit and download jobs from one machine, you can use the --ignore-storage-profiles option\n"
```
sakshie95, line 344: I like this message, very verbose!

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -309,13 +325,49 @@ def incremental_output_download(
...
+                "configured locally"
+                if args.get("storage_profile_id") is None
+                else "provided with --storage-profile-id"
+            )
+            raise DeadlineOperationError(
+                f"Could not retrieve the storage profile {local_storage_profile_id!r}, {id_source}, from Deadline Cloud:\n{e}"
+            )
+
+    # Get download progress file name appended by the queue id and storage profile id - a unique progress file exists per queue/storage profile
+    download_checkpoint_file_name: str = f"{queue_id}_{local_storage_profile_id or 'ignore-storage-profiles'}_{DOWNLOAD_CHECKPOINT_FILE_NAME}"
```
sakshie95, line 366: interesting. Why is there a unique progress file per storage profile on the queue? That would mean the customer can never update their storage profile mid downloads as they'll lose all progress? I would rather add a "user" to this?

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -355,12 +417,27 @@ def incremental_output_download(
...
+                if checkpoint.local_storage_profile_id is None:
+                    raise DeadlineOperationError(
+                        "The checkpoint was created with the --ignore-storage-profiles, you must use the same option to continue from it."
+                    )
+                if local_storage_profile_id is None:
+                    raise DeadlineOperationError(
+                        "The checkpoint was created without the --ignore-storage-profiles, you must leave out the option to continue from it."
+                    )
+                raise DeadlineOperationError(
+                    f"The checkpoint was created with local storage profile {checkpoint.local_storage_profile_id}, but the configured storage profile is {local_storage_profile_id}"
```
sakshie95, line 436: Same as the above comment. Not sure why we're tightly coupling the storage profile to download progress. Maybe there's a reason I can't understand based on the diff

src/deadline/client/cli/_incremental_download.py
```diff
@@ -930,21 +964,109 @@ def _incremental_output_download(
         print_function_callback,
     )
 
+    # If storage profiles are being used, get them and construct all the path mapping rules
```
sakshie95, line 1081: I feel like we should move it out as a utility, I know you created ``_generate_path_mapping_rules`` but maybe a wrapper on top of that - but this can be out of the scope of this PR. For now can you move this to a separate function as the core function ``_incremental_output_download`` got too long with this

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
```diff
@@ -15,6 +15,8 @@
 import concurrent.futures
 from threading import Lock
 from pathlib import Path
```
sakshie95, line 17: Are we planning to extend all this to the attachment clis ever? We could re-use some of the path mapping stuff there

src/deadline/client/cli/_groups/queue_group.py
mwiebe: I think it's more intuitive to make the name match so closely, because the lock is effectively protecting atomic access to that checkpoint file. Do you have a suggestion for how to name it to communicate that relationship and be easier to understand?

src/deadline/client/cli/_groups/queue_group.py
mwiebe: This is always the local storage profile, yes. By definition this is the storage profile being applied to the local machine.

src/deadline/client/cli/_groups/queue_group.py
mwiebe: I added it here because the state tracking over time can depend on the storage profile - especially checkpoints ignoring storage profiles are incompatible with ones using one. The user isn't in the name because nothing the command does depends on the user. It would make sense to add an option to filter by user, after which that user id should also be in this filename to provide similar protection with it.

The customer can edit the storage profile in the AWS console, that doesn't change the ID.

src/deadline/client/cli/_groups/queue_group.py
mwiebe: Yes, same answer about protecting data consistency.

src/deadline/client/cli/_incremental_download.py
mwiebe: Good thought, I'll go ahead and do this.

src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py
mwiebe: Yes, I agree we should do that. Keeping it isolated here for now has allowed us to be more experimental without affecting the existing functionality.

src/deadline/client/cli/_incremental_download.py
mwiebe: I've implemented this, done some basic testing, and pushed the change.

src/deadline/client/cli/_groups/queue_group.py
sakshie95: Personal opinion but I think the current pid locking nomenclature ``queue123_incremental_output_download.pid`` was good since it communicated having a lock on the incremental download operation for that queue which is closely linked with the CLI name & easy to understand. The checkpoint file is an internal detail which customers may not even want to know in most cases - especially non-builders.

Approving PR as this is a non-blocking two way door until we have the CLI in experimental state.

sakshie95 (approved):

src/deadline/client/cli/_groups/queue_group.py
mwiebe: Cool, thanks! I like your reasoning, I would apply it to both files and rename them to include the cli command name.

leongdl (approved):

Issue #768, opened 2025-08-05
Bug: 0.51.0 submitter window stays open after successful submissions


lucaseck:
### Describe the bug

When using the 0.51.0 release of deadline the job submission dialog UI doesn't close after a successful submission when it should.

### Expected Behaviour

Running `deadline bundle gui-submit --output json --submitter-name Testing <BUNDLE_PATH>` should result in the dialog opening and after submitting a job successfully the submission dialog should __close__ and print a message like:
```
{"status": "SUBMITTED", "jobId": "job-asdasdasdasdasdasdasdasdas", "jobHistoryBundleDirectory": "<HISTORY_DIRECTORY>"}
```
This is what happens with deadline 0.50.1

### Current Behaviour

With deadline 0.51.0, running `deadline bundle gui-submit --output json --submitter-name Testing <BUNDLE_PATH>`, and submitting results in the dialog staying open. Manually closing the dialog results in:
```
{"status": "CANCELED"}
```

### Reproduction Steps

With deadline 0.51.0 installed run `deadline bundle gui-submit --output json --submitter-name Testing <BUNDLE_PATH>` . The dialog will not close like it should after a successful submission.

The value of create_job_response is not being set anymore in [SubmitJobToDeadlineDialog](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py#L539). This is causing the submission dialog to not close after a submission like it should be. 

### Environment

All environments using deadline 0.51.0 and the SubmitJobToDeadlineDialog.

### Workaround
At this time the only workaround is to downgrade to deadline 0.50.1

AWS-Samuel:
This is fixed in release 0.51.1

PR #775, opened 2025-08-07
fix: Submission window doesn't close after a successful submission

AWS-Samuel:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/768

The Submission window doesn't close after a successful submission.

The root cause is that a function had its return value changed from the create job response, to `None`. 

The solution is to hook up passing the jobId back into the submission window, as well as adding logic to close the window if the job submission is successful. This is a bit more ugly that I'd like it to be as a result of needing to close the dialog from the main thread.

The job submission window closes after a successful submit if running inside an integrated submitter. 

I ran 

```
deadline bundle gui-submit --output json --submitter-name Testing <path to job bundle>
```

And confirmed that

1. The job window closed after success and I clicked either `ok` or closed the submission progress window
2. The output in the terminal was `{"status": "SUBMITTED", "jobId": "job-0f452c2c73b942f5b05e2937c15ebedc", "jobHistoryBundleDirectory": "<job history bundle>"}`

Also ran 

```
deadline bundle gui-submit --output json <path to job bundle>
```

And confirmed that

1. The job window stayed open after success and I clicked either `ok` or closed the submission progress window
2. The output in the terminal was `{"status": "SUBMITTED", "jobId": "job-0f452c2c73b942f5b05e2937c15ebedc", "jobHistoryBundleDirectory": "<job history bundle>"}`

yes

yes

No

No, and it is not required

No

No
----

### DIFF
--- a/src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
+++ b/src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
@@ -112,7 +112,6 @@ class SubmitJobProgressDialog(QDialog):
     to AWS Deadline Cloud.
     """
 
-    job_id: Optional[str] = None
     cancelation_flag: CancelationFlag
 
     # This signal is sent when the background thread raises an exception.
@@ -126,6 +125,9 @@ class SubmitJobProgressDialog(QDialog):
 
     # This signal is sent when the background thread succeeds.
     submission_thread_succeeded = Signal(str)
+    progress_window_closed = Signal(None)
+
+    job_id: Optional[str] = None
 
     def __init__(self, parent: QWidget) -> None:
         super().__init__(parent=parent)
@@ -152,7 +154,7 @@ class SubmitJobProgressDialog(QDialog):
         job_bundle_dir: str,
         job_parameters: list[dict[str, Any]] = [],
         **kwargs,
-    ):
+    ) -> None:
         """
         Starts a job submission background thread and returns immediately. It wires up
         appropriate callbacks and then forwards all arguments.
@@ -291,6 +293,9 @@ class SubmitJobProgressDialog(QDialog):
             self.status_label.setText("Submission complete")
             self.button_box.setStandardButtons(QDialogButtonBox.Ok)
             self.button_box.button(QDialogButtonBox.Ok).setDefault(True)
+            self.button_box.button(QDialogButtonBox.Ok).clicked.connect(
+                self.progress_window_closed.emit
+            )
         else:
             if self.cancelation_flag or self._warning_dialog_canceled:
                 self.status_label.setText("Submission canceled")
@@ -314,6 +319,7 @@ class SubmitJobProgressDialog(QDialog):
         closing the dialog. If the submission is complete then any button, even
         'X', should result in the dialog being accepted.
         """
+        self.progress_window_closed.emit()
         if self._submission_complete:
             self.accept()
         else:
--- a/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
+++ b/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
@@ -124,11 +124,12 @@ class SubmitJobToDeadlineDialog(QDialog):
         self.job_settings_type = type(initial_job_settings)
         self.submitter_name = submitter_name or self.job_settings_type().submitter_name
         self.on_create_job_bundle_callback = on_create_job_bundle_callback
-        self.create_job_response: Optional[Dict[str, Any]] = None
+        self.job_id = None
         self.job_history_bundle_dir: Optional[str] = None
         self.deadline_authentication_status = DeadlineAuthenticationStatus.getInstance()
         self.show_host_requirements_tab = show_host_requirements_tab
         self.known_asset_paths = known_asset_paths or []
+        self.should_close = False
 
         self._build_ui(
             job_setup_widget_type,
@@ -141,6 +142,15 @@ class SubmitJobToDeadlineDialog(QDialog):
 
         self.gui_update_counter: Any = None
         self.refresh_deadline_settings()
+
+    def _submission_succeeded_signal_receiver(self, job_id: str):
+        self.job_id = job_id
+
+        set_setting("defaults.job_id", job_id)
+
+    def _close_event_receiver(self):
+        if self.submitter_name != "JobBundle" and self.job_id:
+            self.close()
 
     def sizeHint(self):
         return QSize(540, 700)
@@ -465,9 +475,6 @@ class SubmitJobToDeadlineDialog(QDialog):
         """
         Perform a submission when the submit button is pressed
         """
-        # Unset any cached response
-        self.create_job_response = None
-
         # Retrieve all the settings into the dataclass
         settings = self.job_settings_type()
         self.shared_job_settings.update_settings(settings)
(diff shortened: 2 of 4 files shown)


src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -334,6 +337,23 @@ def exec_(self) -> Optional[str]:  # type: ignore[override]
...
+
+    @job_id.setter
+    def job_id(self, value: str) -> None:
+        """
+        Sets the job ID if the submission was successful.
+        """
+        self.__job_id = value
+
+        # The submit job to deadline dialogn depends on having this value set.
+        self.parentWidget().job_id = value
```
epmog: I'm not gonna claim to know good UI code, but this seems off to me. Feels like we're gonna have another case of "here's this odd redirection that makes it hard to see what sets or uses it".

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -124,11 +124,21 @@ def __init__(
         self.job_settings_type = type(initial_job_settings)
         self.submitter_name = submitter_name or self.job_settings_type().submitter_name
         self.on_create_job_bundle_callback = on_create_job_bundle_callback
-        self.create_job_response: Optional[Dict[str, Any]] = None
```
epmog, line 127: So previously this was public, is this an additional breaking change we want to make as part of fixing it? We did effectively half remove it. I guess the option is to fully remove since we're not doing anything or sort of revert the prior change

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -517,7 +534,7 @@ def on_submit(self):
             if job_parameters:
                 self.save_job_parameters_to_job_bundle(self.job_history_bundle_dir, job_parameters)
 
-            job_id = job_progress_dialog.start_job_submission(
```
epmog, line 520: I guess linting didn't catch that start_job_submission doesn't explicitly return anything so this was always `None`? Seems odd to not flag that.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -142,6 +152,16 @@ def __init__(
         self.gui_update_counter: Any = None
         self.refresh_deadline_settings()
 
+    def _check_should_close(self):
+        """Check if we should close the dialog"""
+        if (
+            self.submitter_name != "JobBundle"
+            and self.job_progress_dialog
+            and self.job_progress_dialog.succeeded.is_set()
```
epmog: Do we have to introduce this new event? Can't we rely on the dialog being accepted? ie. in the dialog's closeEvent, we override so that it is only accepted if it was a successful job submission

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -142,6 +152,16 @@ def __init__(
...
+        ):
+            self.close()
+            self.close_check_timer.stop()
```
epmog: There's a few other close calls for this dialog, should we be overriding to ensure we clean this up on all of them?

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: Correct, when this was added in https://github.com/aws-deadline/deadline-cloud/commit/a8bf21bc1af85b6f67694defdb93fee49f3516a6#diff-f2c39b9d80f73b3766a8781dc1bb5579d8f769cd875c9c736a39265a40da8478, no return type was annotated, but there was never a job ID returned from this function.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: I can update the changelog to call out that we've removed `create_job_response` in favour of `job_id`

Basically this parameter was only ever used (internally, at least) for getting the job id. Since create_job_response no longer being set was part of the change released in 0.51, i think we can still bump this as a patch release.

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
AWS-Samuel: An alternative would be to update `_check_should_close` to set the job ID based on the job progress dialog, that would at least prevent the child widget from setting the attribute on the parent. If that seems better, I can update this

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: I can implement `closeEvent()` so that the timer always stops when the dialog closes

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: The alternative would be for the parent widget to check if the child widget is no longer visible

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: I've changed this so that the job ID is received from the job succeeded Qt Signal

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: I've kept this event but removed the timer, I modified the event so that it uses Qt SIgnals which I think makes a bit more sense in the GUI context

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: Removed the timer completely

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -142,6 +143,15 @@ def __init__(
         self.gui_update_counter: Any = None
         self.refresh_deadline_settings()
 
+    def _submission_succeeded_signal_receiver(self, job_id: str):
+        self.job_id = job_id
+
+        set_setting("defaults.job_id", job_id)
+
+    def _close_event_receiver(self):
+        if self.submitter_name != "JobBundle" and self.job_id:
```
mwiebe, line 152: You've maintained the existing behavior intent here, but closing or not depending on the value of submitter_name feels like a design choice asking for improvement. Maybe in a future change we could change this to a parameter to the dialog like "close_on_complete" instead, and then have each caller make the choice explicitly.

mwiebe (approved):

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
AWS-Samuel: Agreed. I'll cut an issue to capture this

AWS-Samuel: https://github.com/aws-deadline/deadline-cloud/issues/777

moorec-aws (approved):

PR #784, opened 2025-08-08
feat(experimental): Add mcp to deadline client tools

kavmur:
AWS Deadline Cloud users need a way to interact with their render farms through AI assistants and natural language interfaces. Currently, users must use CLI commands or Python APIs directly, which creates friction for AI-assisted workflows and automation. The Model Context Protocol (MCP) provides a standardized way for AI assistants to interact with external services, but AWS Deadline Cloud lacked MCP integration.

Added MCP server implementation to the AWS Deadline Cloud client tools. 
The MCP server provides access to all allowlisted/configured Deadline Cloud API functions through automatic registration, we currently have the following 8 tools:

- `deadline_list_farms()`: List available farms
- `deadline_list_queues()`: List queues in a farm  
- `deadline_list_jobs()`: List jobs in a queue
- `deadline_list_fleets()`: List fleets in a farm
- `deadline_list_storage_profiles_for_queue()`: List storage profiles for a queue

- `deadline_check_authentication_status()`: Check current authentication status

- `deadline_submit_job()`: Submit an Open Job Description job bundle to AWS Deadline Cloud
- `deadline_download_job_output()`: Download job output files from AWS Deadline Cloud

- **AI Integration**: Enables AI assistants (Claude, Cline, etc.) to manage AWS Deadline Cloud resources through natural language
- **Workflow Automation**: Enables AI-driven automation of common Deadline Cloud tasks like job submission and output downloading
- **No Breaking Changes**: Purely additive feature that doesn't modify existing functionality

- [x] Have you run the unit tests? - *Unit tests implemented and passing for MCP components*
- [x] Have you run the integration tests? - *Integration tests implemented for server functionality*
- [ ] Have you made changes to the `download` or `asset_sync` modules? - *No changes to these modules*

- [x] **MCP Guide**: Added detailed `docs/mcp_guide.md` with:
  - Installation and setup instructions
  - Architecture overview and project structure
  - Complete tool reference with usage examples
- [x] **README Updates**: Updated main README.md with MCP server section including installation and usage
- [x] **Code Documentation**: All tools have comprehensive docstrings with parameter descriptions

No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Notable features include:
 * A library of functions that implement AWS Deadline Cloud's Job Attachments functionality.
 * A library of functions for creating a job submission UI within any content creation tool that supports Python 3.8+ based plugins and
   the Qt GUI framework.
+* A Model Context Protocol (MCP) server for AI assistant integration, enabling natural language interaction with AWS Deadline Cloud resources.
 
 [cas]: https://en.wikipedia.org/wiki/Content-addressable_storage
 [deadline-cloud]: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/what-is-deadline-cloud.html
@@ -60,6 +61,11 @@ $ pip install deadline
 or if you want the optional gui dependencies:
 ```sh
 $ pip install "deadline[gui]"
+```
+
+if you want the optional mcp dependencies:
+```sh
+$ pip install "deadline[mcp]"
 ```
 
 ## Usage
@@ -115,8 +121,6 @@ Or with the configuration GUI:
 deadline config gui
 ```
 
-
-Shared storage is possible with customer-managed fleets (CMF) but not service-managed fleets (SMF). See [shared storage][shared-storage] for more information.
 
 ## Job Bundles
 
@@ -307,6 +311,12 @@ $ aws s3 ls --profile deadline-queue
 Available modes:
 - `USER`: Credentials with full queue-role permissions.
 - `READ`: Credentials with read-only permissions for queue logs
+
+## Model Context Protocol (MCP) Server
+
+The AWS Deadline Cloud client includes an MCP server that enables AI assistants to interact with AWS Deadline Cloud resources through natural language. The MCP server uses the [MCP Python SDK](https://github.com/modelcontextprotocol/python-sdk) for simplified tool registration while maintaining full protocol compliance.
+
+See [MCP Guide](https://github.com/aws-deadline/deadline-cloud/blob/release/docs/mcp_guide.md) for more information.
 
 
 ## Code of Conduct
--- a/docs/mcp_guide.md
+++ b/docs/mcp_guide.md
@@ -1,0 +1,131 @@
+# AWS Deadline Cloud MCP Server Guide
+
+**MCP ([Model Context Protocol](https://modelcontextprotocol.io/docs/getting-started/intro))** is an open standard that enables AI assistants to securely connect to external data sources and tools. It acts a bridge that allows Large Language Models (LLMs) like Claude, GPT, or other AI assistants to interact with applications and services through natural language.
+
+With the AWS Deadline Cloud MCP Server, you can use natural language for various Deadline Cloud workflows such as submitting a job, reading job/farm/queue information etc.
+
+## User Guide
+
+1. **Install the server** 
+  ```bash
+   pip install 'deadline[mcp]'
+   ```
+2. **Verify the MCP server command is available:**
+   ```bash
+   deadline mcp-server --help
+   ```
+   You should see:
+   ```
+   Usage: deadline mcp-server [OPTIONS]
+
+   Start the AWS Deadline Cloud MCP (Model Context Protocol) server.
+
+   The MCP server provides LLM tools with access to AWS Deadline Cloud
+   operations through the Model Context Protocol. This allows AI assistants to
+   interact with Deadline Cloud services on your behalf.
+
+   The server will run until interrupted with Ctrl+C or Ctrl+D.
+   ```
+3. **Configure AWS credentials:**
+   - Standard AWS credentials (AWS Profiles, environment variables)
+   - Or Deadline Cloud monitor credentials:
+     ```bash
+     deadline auth login
+     ```
+   - Verify authentication status:
+     ```bash
+     deadline auth status
+     ```
+4. **Configure your MCP client** to connect to the server 
+    Add to your MCP configuration (e.g., `./settings/mcp.json`):
+    ```json
+    {
+      "mcpServers": {
+        "deadline-cloud": {
+          "command": "deadline",
+          "args": ["mcp-server"],
+          "disabled": false,
+          "autoApprove": []
+        }
+      }
+    }
+    ```
+5. **Start having conversations** with your AI assistant about your rendering workflows
+
+## Example prompts
+
+```
+- "List all my AWS Deadline Cloud farms"
+- "Show me the queues in my farm"
+- "List the jobs in my queue"
+- "Submit the render job in /path/to/my-job-bundle"
+- "Submit a job with priority 80 to my render queue"
+- "Show me the status of job job-3a907bac684841f69fc344867ee166de"
+- "Download output from job job-3a907bac684841f69fc344867ee166de"
+- "Download output from step step-render in job job-3a907bac684841f69fc344867ee166de"
+```
+
+## Available Tools
+
+The MCP server provides access to all allowlisted/configured Deadline Cloud API functions through automatic registration:
+
+- `deadline_list_farms()`: List available farms
+- `deadline_list_queues()`: List queues in a farm  
+- `deadline_list_jobs()`: List jobs in a queue
+- `deadline_list_fleets()`: List fleets in a farm
+- `deadline_list_storage_profiles_for_queue()`: List storage profiles for a queue
+
+- `deadline_check_authentication_status()`: Check current authentication status
+
+- `deadline_submit_job()`: Submit an Open Job Description job bundle to AWS Deadline Cloud
+- `deadline_download_job_output()`: Download job output files from AWS Deadline Cloud
+
+
+## Developer Guide
+
+The MCP server exposes public Deadline Cloud operations as tools that AI assistants can call directly. Tools are defined in `config.py` where each entry maps a tool name to its corresponding function and parameters. 
+
+## Project Structure
+
+```
+src/deadline/mcp/
+├── server.py                # Main server with FastMCP setup and auto-registration
+├── config.py                # Tool configuration definitions
+├── utils.py                 # Auto-registration utilities
+└── tools/                   # Tool modules
+    └── job.py               # Job management tools (submit, download)
+```
+
+### Adding New MCP Tools
+
+#### Step 1: Use public api operations
+
+Prefer consistency with existing CLI patterns by using functions from the public API module `deadline.client.api.*` when available. Use direct boto3 calls when no wrapper exists or when the wrapper doesn't provide the needed functionality. Tools like `submit_job` and `download_job_output` are exceptions because they require job attachment functionality that are not available in the public API.
+
+```python
+# ✅ Good: Use existing public API
+from deadline.client.api import list_farms, list_queues
+
+# ❌ Bad: Don't use internal modules
+from deadline.client._internal.some_module import internal_function
+```
+
+#### Step 2: Configure the MCP Tool
+
+Add your new tool to the `API_TOOLS_CONFIG` in `src/deadline/mcp/config.py`:
+
+```python
+API_TOOLS_CONFIG: dict[str, ToolConfig] = {
+    # ... existing tools ...
+    "your_new_tool": {
+        "func": api.your_new_function,  # Must be from deadline.client.api
+        "params": ["param1", "param2", "optional_param"],  # List all parameters, or None if no params
+    },
+}
+```
+
+#### Step 3: Test the Integration
+
+1. **Unit Tests**: Add tests to `test/unit/deadline_mcp/test_mcp.py`
+2. **Integration Tests**: Add tests to `test/integ/deadline_mcp/test_mcp_server_integration.py`
+3. **Manual Testing**: Test with real MCP clients
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,9 @@ gui = [
   # If the version changes, update the version in deadline/client/ui/__init__.py
   "PySide6-essentials >= 6.6,< 6.10",
 ]
+mcp = [
+  "mcp >= 1.13.0",
+]
 
 [project.scripts]
 deadline = "deadline.client.cli:main"
@@ -132,6 +135,7 @@ known-first-party = ["deadline"]
 # XML, HTML, and terminal reports.
 [tool.pytest.ini_options]
 xfail_strict = true
+
 addopts = [
   "--durations=5",
   "--cov=src/deadline",
@@ -147,6 +151,7 @@ markers = [
   "integ: tests that run against AWS resources",
   "docker: marks tests to be run only in a Docker environment",
   "cross_account: tests that run against other aws accounts",
+  "asyncio: mark test as async",
 ]
 # looponfailroots is deprecated, this removes the deprecation from the test output
 filterwarnings = ["ignore::DeprecationWarning"]
--- a/src/deadline/_mcp/server.py
+++ b/src/deadline/_mcp/server.py
@@ -1,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from mcp.server.fastmcp import FastMCP
+
+from .utils import register_api_tools
+
+app = FastMCP("deadline-cloud")
+
+register_api_tools(app, prefix="deadline_")
+
+
+def main():
+    app.run()
--- a/src/deadline/client/cli/_deadline_cli.py
+++ b/src/deadline/client/cli/_deadline_cli.py
@@ -20,3 +20,6 @@ from ._groups import (  # noqa: F401
     attachment_group,
     manifest_group,
 )
+
+# MCP server command
+from ._mcp_server import cli_mcp_server  # noqa: F401
--- a/src/deadline/client/cli/_mcp_server.py
+++ b/src/deadline/client/cli/_mcp_server.py
@@ -1,0 +1,51 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+The `deadline mcp-server` command.
+"""
+
+import sys
+import click
+
+from ._common import _handle_error
+from ._main import main
+from deadline.client.api._telemetry import get_deadline_cloud_library_telemetry_client
+
+
+@main.command(name="mcp-server")
+@_handle_error
+def cli_mcp_server():
+    """
+    Start the AWS Deadline Cloud MCP (Model Context Protocol) server.
+
+    The MCP server provides LLM tools with access to AWS Deadline Cloud operations
+    through the Model Context Protocol. This allows AI assistants to interact with
+    Deadline Cloud services on your behalf.
+
+    The server will run until interrupted with Ctrl+C or Ctrl+D.
+
+    Note: This command requires MCP dependencies. Install them with:
+    pip install 'deadline[mcp]'
+    """
+    try:
+        from ..._mcp.server import main as mcp_main
+    except ImportError:
+        click.echo(
+            "Error: MCP dependencies not installed.\n"
+            "Please install them with: pip install 'deadline[mcp]'",
+            err=True,
+        )
+        sys.exit(1)
+
+    # Record server startup telemetry
+    try:
+        telemetry_client = get_deadline_cloud_library_telemetry_client()
+        telemetry_client.record_event(
+            event_type="com.amazon.rum.deadline.mcp.server_startup",
+            event_details={"usage_mode": "MCP", "startup_method": "cli"},
+        )
+    except Exception:
+        # Don't let telemetry errors affect server startup
+        pass
+
+    mcp_main()
--- a/test/integ/deadline_mcp/test_mcp_server_integration.py
+++ b/test/integ/deadline_mcp/test_mcp_server_integration.py
@@ -1,0 +1,456 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""Integration test for the Deadline Cloud MCP Server."""
+
+import json
+import pytest
+import boto3
+import asyncio
+import sys
+from unittest.mock import patch, MagicMock
+
+# Skip all tests in this module if MCP dependencies are not available
+pytest.importorskip("mcp", reason="MCP dependencies not available")
+
+from deadline._mcp import server
+from deadline._mcp.registry import TOOL_REGISTRY, get_tool_definition
+
+from deadline.client.api import list_farms
+from deadline.client.cli._mcp_server import cli_mcp_server
+from click.testing import CliRunner
+
+
+@pytest.fixture(scope="session")
+def get_boto_session():
+    """
+    Fixture to establish authenticated session for MCP integration tests.
+    """
+    session = boto3.Session()
+    sts_client = session.client("sts")
+    _ = sts_client.get_caller_identity()
+    return session
+
+
+@pytest.mark.asyncio
+@pytest.mark.integ
+async def test_mcp_server_integration(get_boto_session):
+    """Test that MCP server starts, registers tools, and handles tool calls."""
+    with patch("deadline.client.api._session.get_boto3_session") as mock_session:
+        mock_session.return_value = get_boto_session
+
+        app = server.app
+
+        # 1. Check if all tools in TOOL_REGISTRY are available
+        tools_result = await app.list_tools()
+        tools = tools_result if isinstance(tools_result, list) else tools_result.tools
+        available_tool_names = [tool.name for tool in tools]
+        expected_tools = [f"deadline_{tool_name}" for tool_name in TOOL_REGISTRY.keys()]
+        missing_tools = [tool for tool in expected_tools if tool not in available_tool_names]
+        assert not missing_tools, f"Missing expected tools from TOOL_REGISTRY: {missing_tools}"
+
+        # 2. Call deadline_check_authentication_status tool and verify response
+        auth_result = await app.call_tool("deadline_check_authentication_status", {})
+        assert auth_result is not None, "Authentication tool call should return a result"
+
+        if auth_result and hasattr(auth_result[0], "text"):  # type: ignore[index]
+            try:
+                auth_data = json.loads(auth_result[0].text)  # type: ignore[index]
+                assert "error" not in auth_data, (
+                    f"Authentication check failed with error: {auth_data.get('error')}"
+                )
+            except json.JSONDecodeError as e:
+                pytest.fail(
+                    f"Check authentication tool returned non-JSON response: {auth_result[0].text[:200]}... (JSONDecodeError: {e})"  # type: ignore[index]
+                )
+
+        # 3. Call deadline_list_farms tool and verify response
+        farms_result = await app.call_tool("deadline_list_farms", {})
+        assert farms_result is not None, "List farms tool call should return a result"
+        if farms_result and hasattr(farms_result[0], "text"):  # type: ignore[index]
+            try:
+                farms_data = json.loads(farms_result[0].text)  # type: ignore[index]
+                assert "error" not in farms_data, (
+                    f"List farms failed with error: {farms_data.get('error')}"
+                )
+            except json.JSONDecodeError as e:
+                pytest.fail(
+                    f"List farms returned non-JSON response: {farms_result[0].text[:200]}... (JSONDecodeError: {e})"  # type: ignore[index]
+                )
+
+        # 4. Check metadata on deadline_list_queues tool
+        list_queues_tool = None
+        for tool in tools:
+            if tool.name == "deadline_list_queues":
+                list_queues_tool = tool
+                break
+
+        assert list_queues_tool is not None, "deadline_list_queues tool not found"
+
+        expected_params = get_tool_definition("list_queues")["param_names"]
+        assert expected_params is not None, (
+            "list_queues should have parameters defined in TOOL_REGISTRY"
+        )
+
+        assert hasattr(list_queues_tool, "inputSchema"), (
+            "deadline_list_queues tool should have an inputSchema"
+        )
+
+        input_schema = list_queues_tool.inputSchema
+        assert input_schema is not None, "inputSchema should not be None"
+
+        available_params = []
+        if isinstance(input_schema, dict) and "properties" in input_schema:
+            available_params = list(input_schema["properties"].keys())
+        elif hasattr(input_schema, "properties") and input_schema.properties:
+            available_params = list(input_schema.properties.keys())
+        elif hasattr(input_schema, "model_fields"):
+            available_params = list(input_schema.model_fields.keys())
+
+        assert "farmId" in available_params, (
+            f"farmId parameter should be available in list_queues tool. "
+            f"Available params: {available_params}, Expected params: {expected_params}"
+        )
+
+        missing_params = [param for param in expected_params if param not in available_params]
+        assert not missing_params, (
+            f"Missing expected parameters in list_queues tool: {missing_params}. "
+            f"Available: {available_params}, Expected: {expected_params}"
+        )
+
+        return True
+
+
+@pytest.mark.asyncio
+@pytest.mark.integ
+async def test_tool_description_extraction(get_boto_session):
+    """Test that tool descriptions are properly extracted from original function docstrings."""
+    app = server.app
+
+    tools_result = await app.list_tools()
+    tools = tools_result if isinstance(tools_result, list) else tools_result.tools
+
+    list_farms_tool = None
+    for tool in tools:
+        if tool.name == "deadline_list_farms":
+            list_farms_tool = tool
+            break
+
+    assert list_farms_tool is not None, "deadline_list_farms tool not found"
+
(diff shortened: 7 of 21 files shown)


DEVELOPMENT.md
```diff
@@ -160,6 +160,11 @@ export JA_TEST_ROOT_PREFIX=$(
     --query 'jobAttachmentSettings.rootPrefix' | tr -d '"'
 )
 ```
+For MCP integration tests set-up env to use [AWS Deadline Cloud monitor][deadline-cloud-monitor] credentials, you can authenticate by logging in:
```
prestomation: Are these coupled to DCM creds? They probably shouldn't be. Is this different than the other integration tests?

pyproject.toml
```diff
@@ -145,13 +149,14 @@ markers = [
...
   "cross_account: tests that run against other aws accounts",
+  "asyncio: mark test as async",
 ]
 # looponfailroots is deprecated, this removes the deprecation from the test output
 filterwarnings = ["ignore::DeprecationWarning"]
 
 
 [tool.coverage.run]
-disable_warnings = ["module-not-measured"]
+disable_warnings = ["module-not-measured", "no-data-collected"]
```
prestomation: Why are we adding this supression?

src/deadline/client/api/_get_apis.py
```diff
@@ -0,0 +1,37 @@
...
+        farm_id (str, optional): The farm ID to get details for. If not provided,
+                                uses the default farm ID from configuration.
+        config (ConfigParser, optional): If provided, the AWS Deadline Cloud config to use.
+        **kwargs: Additional parameters to pass to the API call.
+
+    Returns:
+        dict: The farm details from the API response.
+    """
+    if farm_id is None:
+        farm_id = kwargs.get("farmId")
```
prestomation: I don't think it's a good interface for farmId to be an arg OR a kwarg. Also, get_farm has no other possible parameters, so why have kwargs? If it is non-empty then the call will fail right? or maybe it will succeed and the args do nothing, confusing the user more?

src/deadline/client/api/_get_apis.py
```diff
@@ -0,0 +1,37 @@
...
+        if farm_id is None:
+            farm_id = config_file.get_setting("defaults.farm_id", config=config)
+
+    if farm_id is None:
+        raise ValueError("farm_id must be provided either as parameter or in configuration")
+
+    deadline = get_boto3_client("deadline", config=config)
+    response = deadline.get_farm(farmId=farm_id, **kwargs)
+
+    response.pop("ResponseMetadata", None)
```
prestomation: This might actually be useful data for the LLM? should we pop this in the CLI layer?

src/deadline/mcp/__main__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Entry point for running the Deadline Cloud MCP Server as a module.
```
prestomation: what's the usecase for this?

src/deadline/mcp/server.py
```diff
@@ -0,0 +1,19 @@
...
+tools = [config["func"] for config in API_TOOLS_CONFIG.values()]
+register_api_tools(app, tools, prefix="deadline_")
+
+
+def main():
+    app.run()
+
+
+if __name__ == "__main__":
+    main()
```
prestomation: what's the usecase for this? Looks like there are three ways to run the server which don't really differ?

src/deadline/mcp/tools/utils.py
```diff
@@ -0,0 +1,157 @@
...
+        "params": None,
+        "description": "Check the authentication status of the current session",
+    },
+    "get_credentials_source": {
+        "func": api.get_credentials_source,
+        "params": None,
+        "description": "Get the source of the current credentials",
+    },
+    "submit_job": {
+        "func": job.submit_job,
```
prestomation: The docs say to use stuff right from API, but we don't for these. why? can/should we factor them better?

src/deadline/mcp/tools/utils.py
```diff
@@ -0,0 +1,157 @@
...
+import json
+import logging
+from typing import Any, Callable, Dict, List, Optional
+
+from ...client import api
+from . import job
+
+logger = logging.getLogger(__name__)
+
+API_TOOLS_CONFIG: Dict[str, Dict[str, Any]] = {
```
prestomation: can we type this better?

src/deadline/mcp/tools/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    },
+    "list_queues": {
+        "func": api.list_queues,
+        "params": ["farmId", "dryRun", "principalId", "status", "nextToken", "maxResults"],
+        "description": "List queues in a farm",
+    },
+    "list_jobs": {
+        "func": api.list_jobs,
+        "params": ["farmId", "queueId", "dryRun", "principalId", "nextToken", "maxResults"],
+        "description": "List jobs in a queue",
```
prestomation: I suspect this interface won't be good enough. does the model get the existing docstring from the func?  if so, this description seems unnecessary. if not, these descriptions need to be a lot better. seems like just the one should be good, to help both models and humans?

src/deadline/mcp/tools/utils.py
prestomation: Also, for style. this is a core data structure, doesn't make a lot of sense to be in a file calledd `utils.py`. can we put this somewhere else?

src/deadline/mcp/tools/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    signature = inspect.Signature(mcp_params, return_annotation=dict)
+
+    def wrapper(**kwargs) -> dict:
+        try:
+            filtered_kwargs = {
+                k: v for k, v in kwargs.items() if v is not None and v != "" and v != "null"
+            }
+
+            if "config" in inspect.signature(func).parameters:
+                filtered_kwargs["config"] = None
```
prestomation: This seems to be some conventional magic that needs to be documented for MCP devs

src/deadline/mcp/tools/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    serializer = serializer or _default_serializer
+    func_to_name = {config["func"]: name for name, config in API_TOOLS_CONFIG.items()}
+
+    for func in tools:
+        if not callable(func) or hasattr(func, "_mcp_tool_registered"):
+            continue
+
+        tool_name = func_to_name.get(func)
+        if not tool_name:
+            logger.warning(f"No configuration found for function: {func}")
```
prestomation: Should this be a warning?  Seems odd to me that the signature for register_api_tools has both the tools list and then statically refers to `API_TOOLS_CONFIG`

src/deadline/mcp/tools/utils.py
```diff
@@ -0,0 +1,157 @@
...
+            return json.loads(json.dumps(result, default=serializer))
+        except Exception as e:
+            return error_handler(e)
+
+    wrapper.__signature__ = signature  # type: ignore[attr-defined]
+    return wrapper
+
+
+def register_api_tools(
+    app: Any,
```
prestomation: Can we type this?

src/deadline/mcp/tools/utils.py
```diff
@@ -0,0 +1,157 @@
...
+            logger.warning(f"No configuration found for function: {func}")
+            continue
+
+        config = API_TOOLS_CONFIG[tool_name]
+        wrapper = _create_wrapper(config, serializer, error_handler)
+        wrapper.__name__ = tool_name
+        wrapper.__doc__ = config["description"]
+
+        app.tool(name=f"{prefix}{tool_name}", description=config["description"])(wrapper)
+        func._mcp_tool_registered = True  # type: ignore[attr-defined]
```
prestomation: Is this just for testing?

prestomation (dismissed):

src/deadline/client/api/_get_apis.py
```diff
@@ -0,0 +1,37 @@
...
+
+from ._session import (
+    get_boto3_client,
+)
+from .. import api
+from ..config import config_file
+
+
+@api.record_function_latency_telemetry_event()
+def get_farm(farm_id=None, config=None, **kwargs):
```
mwiebe: What's the benefit to having a wrapper for `get_farm` compared to using boto3 directly?

src/deadline/mcp/tools/utils.py
kavmur: doc strings from the original function are being overridden by the telemetry decorators. I can try to retrieve and use them, will remove descriptions here. Makes sense to have the same descriptions for models (mcp server) and humans (CLI)

src/deadline/mcp/tools/utils.py
kavmur: submit_job and download_output functions use job attachment modules that are not exposed apis/functions. And given that these are useful tools for most common workflows I added them here. So these are really exceptions unless we expose these apis. We could do a series of public deadline:Get* and s3:* calls and achieve the same result but I feel this is cleaner and we will have consistent behaviour with CLI commands.

src/deadline/mcp/tools/utils.py
kavmur: Good point, makes sense to move it to a config.py I will also add some typing. Thanks

src/deadline/mcp/tools/utils.py
kavmur: Missed this in refactoring, removed tool list from signature and will always read from config.

src/deadline/mcp/tools/utils.py
kavmur: Currently only using this for tests. I actually wanted to have the option to dynamically load/register tools to save context space and this check would help avoid re-registering the tool. But tool refreshes (hot reloads) are not fully supported by most clients, so will punt this.

src/deadline/mcp/tools/utils.py
kavmur: Thanks, added a comment.

src/deadline/mcp/server.py
kavmur: I mainly used it for quick testing during development, especially useful with Q to run the server and test with custom mcp clients. But you are right, we can also use python -m deadline.mcp or hatch shell and deadline mcp-server. I can remove this.

src/deadline/mcp/__main__.py
kavmur: No specific usecase apart from flexibility in dev and testing. Removed this as we can use hatch shell for testing during dev or simply use integration tests.

src/deadline/client/api/_get_apis.py
kavmur: yes! makes sense. Fixed. We could refactor and follow this pattern for other apis in the future.

src/deadline/client/api/_get_apis.py
kavmur: > What's the benefit to having a wrapper for `get_farm` compared to using boto3 directly?

Mostly following the existing wrapper pattern used for CLI here as this wrapper will be used both by MCP integration and CLI. In general, wrapper pattern helps us handle deadline cloud specific configuration like config, inject user context, pagination handling, telemetry over the boto interface.

pyproject.toml
kavmur: Since mcp dependencies is optional, I had the mcp integ tests skip if these were not installed and we would get warnings from the mcp modules. But rethinking this, it makes sense to always have this installed in dev env and run these tests for regression as we are reusing CLI helpers. The mcp dependencies will always be installed anyway with requirements-integ-testing.txt. Removed this supression.

DEVELOPMENT.md
```diff
@@ -181,6 +186,7 @@ Notes:
 * AWS Developers note: If testing with a non-production deployment of AWS Deadline Cloud then you will have to
 define the `AWS_ENDPOINT_URL_DEADLINE` environment variable to the non-production endpoint URL. For example,
 production endpoints look like: `export AWS_ENDPOINT_URL_DEADLINE="https://deadline.$AWS_DEFAULT_REGION.amazonaws.com"`
+*
```
crowecawcaw: Remove

README.md
```diff
@@ -296,6 +300,12 @@ Available modes:
 - `USER`: Credentials with full queue-role permissions.
 - `READ`: Credentials with read-only permissions for queue logs
 
+## Model Context Protocol (MCP) Server
+
+The AWS Deadline Cloud client includes an MCP server that enables AI assistants to interact with AWS Deadline Cloud resources through natural language. The MCP server uses FastMCP within the MCP Python SDK for simplified tool registration while maintaining full protocol compliance.
```
crowecawcaw: Nit: can remove the bit about fast mcp, not too important for the top level README

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
+# AWS Deadline Cloud MCP Server Guide
+
+**MCP (Model Context Protocol)** is an open standard that enables AI assistants to securely connect to external data sources and tools. It acts a bridge that allows Large Language Models (LLMs) like Claude, GPT, or other AI assistants to interact with applications and services through natural language.
```
crowecawcaw: Nit: add MCP link

crowecawcaw: Could be a future improvement: it'd be nice to show a few examples of things an agent can do. Like, "Submit this job template, monitor its logs, and debug any issues you find with it."

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Key Benefits of MCP:
+- **Natural Language Interface**: Control AWS Deadline Cloud using conversational commands instead of complex CLI syntax
+- **AI-Powered Workflows**: Let AI assistants help you manage rendering farms, submit jobs, and troubleshoot issues
+- **Secure Integration**: MCP provides a standardized, secure way for AI to access your cloud resources
+- **Contextual Understanding**: AI can understand your rendering workflows and provide intelligent recommendations
+
+With the AWS Deadline Cloud MCP Server, you can use natural language for various Deadline Cloud workflows such as submitting a job, reading job/farm/queue information etc.
```
crowecawcaw: I might remove this. It doesn't actually say anything more than the previous sentence.

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+
+
+## Developer Guide
+
+The AWS Deadline Cloud MCP Server follows a **public API-first strategy** to ensure maintainability, stability, and proper abstraction layers.
+The server uses FastMCP within the MCP Python SDK to provide a simplified, decorator-based approach to tool registration while maintaining full MCP protocol compliance. The server automatically registers all available Deadline Cloud API functions as MCP tools, plus additional specialized tools for job management.
```
crowecawcaw: Maybe rewrite this yourself so it's specific and clear. It reads like AI slop.

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+}
+```
+
+#### Step 3: Test the Integration
+
+1. **Unit Tests**: Add tests to `test/unit/deadline_mcp/test_mcp.py`
+2. **Integration Tests**: Add tests to `test/integ/deadline_mcp/test_mcp_server_integration.py`
+3. **Manual Testing**: Test with real MCP clients
+
+## User Guide
```
crowecawcaw, line 7: Nit: move the user guide up top so people just wanting to use it can start more easily

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+   The server will run until interrupted with Ctrl+C or Ctrl+D.
+   ```
+3. **Configure your MCP client** to connect to the server (see [Integration with MCP Clients](#integration-with-mcp-clients))
+4. **Start having conversations** with your AI assistant about your rendering workflows
+
+## Available Tools
+
+The MCP server provides access to all allowlisted Deadline Cloud API functions through automatic registration:
+
+**From `deadline.client.api._list_apis`:**
```
crowecawcaw: The dev guide says not to use internal APIs, but this list says we are using them. Maybe clarify it.

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+- Consistent error response format with error type and message
+- Logging of errors for debugging
+- Graceful handling of missing parameters
+
+### Download Tool Behavior
+The `deadline_download_job_output` tool:
+- Downloads files to their original paths by default
+- Automatically handles AWS credentials and queue permissions
+- Supports conflict resolution for existing files
+- Can download from entire jobs, specific steps, or individual tasks
```
crowecawcaw: As a general comment, there's a lot of fluffy text in this guide that seems like it's AI generated. Maybe do a pass over it and strip out stuff that's not strictly necessary. These details for example don't seem necessary for a high level README and could easily go out of date. If we do want to talk about error handling, it could probably be a bullet point somewhere intead.

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Testing the AWS Deadline Cloud MCP Server
+
+## Installation Steps
```
crowecawcaw: This looks the same as the install steps, can remove

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+      "disabled": false,
+      "autoApprove": []
+    }
+  }
+}
+```
+
+## Example Natural Language Interactions
+
+Once connected to an MCP-enabled AI assistant, you can use natural language to interact with Deadline Cloud. The AI understands context and can help you accomplish complex workflows:
```
crowecawcaw: Remove

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+```
+
+### Job Management
+```
+"Submit the render job in /path/to/my-job-bundle"
+"Submit a job with priority 80 to my render queue"
+"Show me the status of job job-3a907bac684841f69fc344867ee166de"
+"Download output from job job-3a907bac684841f69fc344867ee166de"
+"Download output from step step-render in job job-3a907bac684841f69fc344867ee166de"
+```
```
crowecawcaw: Bullet points might be better. Should also not be split up into job and farm management since that's not a meaningful user-facing boundary. I'd guess users are more likely to ask something like, "Why isn't my job being picked up?" which requires both farm and job info to answer.

docs/mcp_guide.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Troubleshooting
+
+### Common Issues
```
crowecawcaw: I don't know that any of these are too helpful. Some issues are self-explanatory because of good error messages. Others have only generic advice.

pyproject.toml
```diff
@@ -56,6 +56,9 @@ gui = [
   # If the version changes, update the version in deadline/client/ui/__init__.py
   "PySide6-essentials >= 6.6,< 6.10",
 ]
+mcp = [
+  "mcp >= 0.1.0",
```
crowecawcaw: Should this be higher? Looks like we require 0.9.0 elsewhere

src/deadline/mcp/config.py
```diff
@@ -0,0 +1,75 @@
...
+
+def get_all_tool_names() -> List[str]:
+    """Get all configured tool names."""
+    return list(API_TOOLS_CONFIG.keys())
+
+
+API_TOOLS_CONFIG: Dict[str, ToolConfig] = {
+    "list_farms": {
+        "func": api.list_farms,
+        "params": ["dryRun", "nextToken", "principalId", "maxResults"],
```
crowecawcaw: I thought we needed to define required parameters, but optional ones we could extract automatically? Does list_farms take a dry run parameter? It's not part of the public SDK.

src/deadline/_mcp/server.py
```diff
@@ -0,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from mcp.server.fastmcp import FastMCP
+
+from .utils import register_api_tools
+
+app = FastMCP("deadline-cloud")
+
+register_api_tools(app, prefix="deadline_")
+
```
crowecawcaw, line 10: Nit: can be moved into main() so the functions don't run on import

src/deadline/mcp/tools/job.py
```diff
@@ -0,0 +1,226 @@
...
+            "output": output_text.strip(),
+        }
+
+    except Exception as e:
+        return {
+            "status": "error",
+            "error": str(e),
+            "type": type(e).__name__,
+            "total_time_seconds": round(time.time() - start_time, 1),
+        }
```
crowecawcaw: Do we need this if we have the error wrapper?

src/deadline/mcp/utils.py
```diff
@@ -0,0 +1,116 @@
...
+            # Filter out empty/null values that MCP clients might send
+            filtered_kwargs = {k: v for k, v in kwargs.items() if v not in (None, "", "null")}
+            # Deadline Cloud API functions accept an optional 'config' parameter
+            # for custom ConfigParser objects. In MCP tools, we always pass None
+            # to use the default configuration (from ~/.deadline/config or environment).
+            if "config" in inspect.signature(func).parameters:
+                filtered_kwargs["config"] = None
+
+            result = func(**filtered_kwargs)
+            return json.loads(json.dumps(result, default=serializer))
```
crowecawcaw: Why do we do this? Could use a comment if it's intentional. I'd guess so the MCP server can serialize it with the default serializer later but I'm not sure.

test/integ/deadline_mcp/test_mcp_server_integration.py
```diff
@@ -0,0 +1,174 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""Integration test for the Deadline Cloud MCP Server."""
```
crowecawcaw, line 3: These look good. It'd be really great to add another test (or adapt this one) to create a real MCP client that starts the MCP server and issues commands. It would help verify that the fast MCP process is wired up correctly.

crowecawcaw: Also add one integ test for error handling

pyproject.toml
kavmur: Thanks for catching the inconsistent versions. I have updated them to the lates 1.13.0 at all places.

src/deadline/mcp/config.py
kavmur: Thanks removed dryrun! We dont need to specify optional params here, but I think optional parameters when specified do give better context in some cases. For example, next token in tool signature would help agents understand that response is paginated (before making an api call, etc. I feel this type of context can be helpful especially if the doc strings are not detailed enough.

src/deadline/mcp/tools/job.py
kavmur: Missed in refactoring, I had these tool not use the wrapper earlier. Will remove, thanks

src/deadline/mcp/config.py
kavmur: Also my understanding is that the tool signature is mostly for adding context that may help the agent choose the correct tool and not for strictly validating req parameters. Example, get_farm tool can still be called without a farm_id, and tool should still pass if it is able to retrieve the farmId from config file.

test/integ/deadline_mcp/test_mcp_server_integration.py
evanspearman-a: Probably worth checking to see if there's an existing tool for this sort of testing.

kavmur: Thanks, was easier to just write a simple client from anthropic docs. But a tool may be useful to test more complex functions in future.

test/integ/deadline_mcp/test_mcp_server_integration.py
```diff
@@ -0,0 +1,412 @@
...
+            # Verify telemetry event details
+            for call in recorded_calls:
+                event_details = call[1]["event_details"]
+                if call[1]["event_type"] == "com.amazon.rum.deadline.mcp.latency":
+                    assert "latency" in event_details, "Latency event should include latency"
+                    assert "tool_name" in event_details, "Latency event should include tool_name"
+                    assert event_details["usage_mode"] == "MCP", "Should indicate MCP usage mode"
+                    assert event_details["tool_name"] == "check_authentication_status", (
+                        "Should record correct tool name"
+                    )
```
crowecawcaw: To make sure we're not including any data we don't want to (e.g. MCP request data), it'd be nice to assert on the entire telemetry event instead of just individual properties.

Could be done here or in the unit tests.

src/deadline/client/cli/mcp_server.py
```diff
@@ -0,0 +1,38 @@
...
+
+import sys
+import click
+
+from ._common import _handle_error
+
+
+@click.command(name="mcp-server")
+@_handle_error
+def cli_mcp_server():
```
crowecawcaw: Do we get telemetry automatically whenever this command is run? It'd be interesting to know how often the MCP server is started, even if the individual tools aren't called. Optional and not blocking.

test/integ/deadline_mcp/test_mcp_server_integration.py
kavmur: Fixed, added more stronger asserts.

crowecawcaw (dismissed):

src/deadline/mcp/tools/utils.py
kavmur: Fixed

DEVELOPMENT.md
kavmur: Fixed. The integ test now uses the credentials from env.

docs/mcp_guide.md
kavmur: Added that job apis are exceptions.

src/deadline/_mcp/server.py
kavmur: It was easier to write integ tests this way.

src/deadline/mcp/utils.py
kavmur: Added comment

src/deadline/client/cli/_deadline_cli.py
```diff
@@ -20,3 +20,6 @@
     attachment_group,
     manifest_group,
 )
+
+# MCP server command
+from .mcp_server import cli_mcp_server  # noqa: F401
```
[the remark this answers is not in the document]
kavmur: import is requires so the sub command is registered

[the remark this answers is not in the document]
kavmur: import is requires so the sub command is registered

leon-li-inspire (dismissed):

docs/mcp_guide.md
```diff
@@ -0,0 +1,131 @@
...
+├── utils.py                 # Auto-registration utilities
+└── tools/                   # Tool modules
+    └── job.py               # Job management tools (submit, download)
+```
+
+### Adding New MCP Tools
+
+#### Step 1: Use public api operations
+
+Always use functions from the public API module `deadline.client.api.*`. Avoid direct boto3 calls or internal modules. Job management tools like `submit_job` and `download_job_output` are exceptions because they require job attachment functionality that are not available in the public API.
```
mwiebe: > Avoid direct boto3 calls

This isn't right. We want to avoid creating wrappers in `deadline.client.api.*` where possible, and only create a wrapper for specific reasons like how the API calls interact with the Deadline Cloud Monitor login access levels. That means we should prefer direct boto3 calls except in circumstances where there is a wrapper that handles this kind of detail.

src/deadline/client/cli/_mcp_server.py
```diff
@@ -0,0 +1,50 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mwiebe, line 1: Can you rename this file to `_mcp_server.py` so that the module is not part of the public library interface? See https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#public-contracts for more details about what is and what is not part of that public contract.

src/deadline/mcp/config.py
```diff
@@ -0,0 +1,70 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mwiebe: This directory is missing an `__init__.py` file.

Is the design for this `mcp` directory intended to be public library interface or an internal implementation detail? E.g. Developers will import from and use this code directly from Python and not just through the CLI?

From reading through all the code, I think this should be internal-only for now and we should rename the directory from `mcp` to `_mcp`.

src/deadline/mcp/config.py
```diff
@@ -0,0 +1,70 @@
...
+"""Configuration definitions for MCP tools in Deadline Cloud."""
+
+from typing import Any, Callable, List, Optional, TypedDict, Dict
+
+from ..client import api
+from .tools import job
+
+
+class ToolConfig(TypedDict):
+    """Configuration for a single MCP tool."""
```
mwiebe: This looks to me more like a callable wrapper than a configuration. What makes this a configuration?

src/deadline/mcp/config.py
```diff
@@ -0,0 +1,70 @@
...
+
+    func: Callable[..., Any]
+    params: Optional[List[str]]
```
mwiebe: Looks like this is being used for lists of parameter names (as opposed to values or definitions). Maybe `param_names` is a better variable name here.

src/deadline/mcp/utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+from mcp.server.fastmcp import FastMCP
+
+from .config import get_all_tool_names, get_tool_config, ToolConfig
+from deadline.client.api._telemetry import get_deadline_cloud_library_telemetry_client
+
+logger = logging.getLogger(__name__)
+
+
+def _get_original_docstring(func: Callable) -> Optional[str]:
```
mwiebe: Instead of creating this workaround, could you fix those decorators to forward the docstring?

test/unit/deadline_mcp/test_mcp.py
```diff
@@ -0,0 +1,113 @@
...
+        register_api_tools(mock_app, [api.list_farms])
+
+        assert mock_app.tool.call_count == 1
+        assert hasattr(api.list_farms, "_mcp_tool_registered")
+
+        # Clean up
+        delattr(api.list_farms, "_mcp_tool_registered")
+
+    def test_invalid_tools_skipped(self):
+        """Test that invalid tools are skipped."""
```
mwiebe: Should this raise an exception instead of skipping? By skipping, I think errors could slip into the implementation more easily.

mwiebe (changes requested):

src/deadline/client/cli/mcp_server.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+import sys
+import click
+
+from ._common import _handle_error
+from deadline.client.api._telemetry import get_deadline_cloud_library_telemetry_client
+
+
+@click.command(name="mcp-server")
```
mwiebe: When I cloned your branch, I found the `mcp-server` command wasn't available. I had to make the following edit:

```suggestion
from ._main import main


@main.command(name="mcp-server")
```

docs/mcp_guide.md
kavmur: Got it. I can rephrase this and remove "Avoid direct boto3 calls". My intent was to avoid CLI using a wrapper from `deadline.client.api.*` and MCP tool using a direct boto call for the same operation/API call.

src/deadline/mcp/config.py
kavmur: yes, this is internal-only hence no __init__.py. Will rename the directory.

src/deadline/mcp/config.py
kavmur: This is more for tool definition with name and params than a configuration. Will refactor.

test/unit/deadline_mcp/test_mcp.py
kavmur: Thanks, yes makes sense to raise exception instead of skipping.

src/deadline/client/cli/mcp_server.py
kavmur: Fixed

docs/mcp_guide.md
```diff
@@ -0,0 +1,131 @@
...
+
+- `deadline_submit_job()`: Submit an Open Job Description job bundle to AWS Deadline Cloud
+- `deadline_download_job_output()`: Download job output files from AWS Deadline Cloud
+
+
+## Developer Guide
+
+The MCP server exposes public Deadline Cloud operations as tools that AI assistants can call directly. Tools are defined in `config.py` where each entry maps a tool name to its corresponding function and parameters. 
+
+## Project Structure
```
mwiebe, line 88: Probably better to delete this section, everything here is visible by looking at the code.

src/deadline/mcp/utils.py
kavmur: Fixed. My initial concern was changing the telemetry wrapper would impact other components, but the change was small.

mwiebe (approved):

src/deadline/_mcp/server.py
```diff
@@ -0,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from mcp.server.fastmcp import FastMCP
+
+from .utils import register_api_tools
+
+app = FastMCP("deadline-cloud")
+
+register_api_tools(app, prefix="deadline_")
```
crowecawcaw, line 9: Do we need this function? It'd odd that it's listing a prefix for tool names when we already have a list of tool name in `TOOL_REGISTRY`. The `register_api_tools` function  has two big branches in it and is used in some test setup, so I wonder if it's being used for two separate purposes (tool registration and test setup) when it should really be two separate functions.

Not blocking though. Let's get this in then we can revisit if it needs cleanup.

crowecawcaw (approved):

PR #787, opened 2025-08-19
feat(experimental): Added integ/e2e tests for incremental download CLI

sakshie95:
Fixes: N/A

We need integration testing for the new incremental download CLI which is in experimental state

Added integration tests for the incremental download CLI using variety of job templates.
```
1. make_many_small_files - High-volume file creation template
   • Creates configurable number of small files (default: 10,000 files across 100 tasks)
   • Each task generates 100 files with unique content
   • Tests incremental download performance with large file counts
   • Validates exact file count matching (10,000 files expected)

2. dep_data_flow - Complex dependency workflow with branching/merging
   • Multi-step pipeline: Step1 → Step1-2 (frames 8-11) → Step1-2-3 & Step1-2-4 → Step1-2-34-5
   • Tests data flow through step dependencies with job attachments
   • Uses both INOUT (data_dir) and IN (input_dir) data flows
   • Validates 8 expected output files with specific content markers
   • Tests incremental download with complex dependency chains

3. dep_chain - Linear dependency chain template  
   • Sequential chain: A → B → C → D → E → F → Verify
   • Each step creates a file with specific content ("Step X is correct")
   • Tests incremental download with linear step dependencies
   • Validates 6 output files (A.txt through F.txt) with exact content matching
```
Ran the incremental download in a loop while the jobs with the above template were running and validated that all expected outputs are downloaded successfully. 

Also put another test for testing re-queued jobs, steps and tasks using the ``make_many_small_files`` template with just 2 tasks. This test is marked as ``xfail`` because there is a bug in the incremental download where ``--conflict-resolution`` method ``create_copy`` doesn't work as expected.

Testing for regressions in the flow and make sure the changes are always tested.

Ran the tests

- Have you run the unit tests? N/A testing PR
- Have you run the integration tests? Yes I added more :))

No, just tests!

No, just tests!

----


test/integ/cli/job_templates.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Add parameters if provided
+    if parameters:
+        for key, value in parameters.items():
+            cmd.extend(["--parameter", f"{key}={value}"])
+
+    # Add the bundle path
+    cmd.append(bundle_path)
+
+    result = subprocess.run(cmd, capture_output=True, text=True, check=False)
```
mwiebe: Could this call `create_job_from_job_bundle` directly instead of calling the CLI as a subprocess? That function returns the job id directly, so would simplify the code to get it from stdout.

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -0,0 +1,596 @@
...
+    """
+    Class for Incremental Download Integration Tests.
+    """
+
+    def __init__(self, farm_id: str, queue_id: str):
+        self.farm_id = farm_id
+        self.queue_id = queue_id
+        self.deadline_client = boto3.client("deadline")
+
+    def wait_for_job_completion(
```
mwiebe, line 34: Could this use the `deadline.client.api.wait_for_job_completion` function instead of implementing it? That was added recently in https://github.com/aws-deadline/deadline-cloud/pull/697

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -0,0 +1,596 @@
...
+    )
+
+
+@pytest.fixture(scope="session", autouse=True)
+def cleanup_test_files():
+    """Post-session fixture to cleanup any dangling files from incremental download tests."""
+    yield  # Run all tests first
+
+    # Cleanup patterns from test analysis
+    workspace = Path.cwd()
```
mwiebe: It would be better for these tests to create a temporary directory instead of using the CWD. By using the CWD, they can write files to locations that the dev running the tests does not expect.

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -0,0 +1,596 @@
...
+            shutil.rmtree(output_path, ignore_errors=True)
+
+
+@pytest.mark.integ
+@pytest.mark.timeout(900)  # 15 minutes timeout
+def test_incremental_download_dep_data_flow(incremental_download_test):
+    """Test incremental download with dep_data_flow template."""
+
+    # Create a unique data_dir in the workspace to avoid interference between tests
+    unique_data_dir = f"data_dir_{int(time.time())}"
```
mwiebe: `time.time` could produce the same directory name in two tests that are running concurrently, contrary to the comment. It would be better to use the Python `tempfile` module to create a unique temporary directory. You can supply a base directory to use so that all the individual directories are contained together.

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -0,0 +1,596 @@
...
+        # Verify we have the complete set of expected files
+        missing_files = set(expected_files.keys()) - set(verified_files)
+        assert not missing_files, f"Missing expected files: {missing_files}"
+
+        print(
+            f"[dep_data_flow] Successfully verified all {len(verified_files)} expected output files with correct content"
+        )
+
+    finally:
+        # Clean up test-specific output files after validation (always runs, even if test fails)
```
mwiebe: If you use the pytest tmp_path fixture (https://docs.pytest.org/en/6.2.x/tmpdir.html) and make all directories as subdirectories of that, you can remove all custom cleanup code because pytest will do that for you.

test/integ/cli/test_cli_incremental_download.py
sakshie95: Done, I still built a wrapper around it for the specific handling of errors

test/integ/cli/job_templates.py
sakshie95: Done

test/integ/cli/test_cli_incremental_download.py
sakshie95: Done. There was an issue I was facing with setting the cwd which is why I resorted to this. I've time boxed and resolved the issue now.

test/integ/cli/test_cli_incremental_download.py
sakshie95: Done

test/integ/cli/test_cli_incremental_download.py
sakshie95: Done

test/integ/cli/job_templates.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # Convert parameters to the format expected by create_job_from_job_bundle
+    job_parameters = []
+    if parameters:
+        job_parameters = [{"name": key, "value": value} for key, value in parameters.items()]
+
+    # Set farm and queue in config
+    config = config_file.read_config()
+    config.set("defaults", "farm_id", farm_id)
+    config.set("defaults", "queue_id", queue_id)
```
mwiebe: When I ran these tests, I found they submitted to the wrong queue. With this change, they went to the right one:

```suggestion
    config_file.set_setting("defaults.farm_id", farm_id, config)
    config_file.set_setting("defaults.queue_id", queue_id, config)
```

test/integ/cli/job_bundles/dep_chain/scripts/create_files.sh
```diff
@@ -0,0 +1,28 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mwiebe, line 1: When I ran the tests, this job failed with the following error:

```
2025/08/22 13:13:32-07:00 Output:
2025/08/22 13:13:32-07:00 /sessions/session-49579bde73fe4bfda1c380ef964f4a11_fmvc0ue/assetroot-47835e2f48e462ec856d/create_files.sh: line 2: $'\r': command not found
2025/08/22 13:13:32-07:00 /sessions/session-49579bde73fe4bfda1c380ef964f4a11_fmvc0ue/assetroot-47835e2f48e462ec856d/create_files.sh: line 4: set: pipefail
2025/08/22 13:13:32-07:00 : invalid option name
```

This is because the .sh files got Windows EOL when I checked out the repo, but they need to be POSIX to run on Linux workers.

You can add these files to the .gitattributes to force these files to be POSIX even on Windows. In our deadline-cloud-samples repo we set this for all .sh files, but you can be selective too: https://github.com/aws-deadline/deadline-cloud-samples/blob/938807d16fdc04242d632874831dee812d258651/.gitattributes#L2

mwiebe (changes requested):

test/integ/cli/job_templates.py
```diff
@@ -0,0 +1,137 @@
...
+def get_job_bundle_path(template_name: str) -> str:
+    """
+    Get the path to a job bundle template directory.
+
+    Args:
+        template_name: Name of the template (e.g., 'make_many_small_files')
+
+    Returns:
+        Path to the template directory
+    """
```
jericht: tiniest nit: these are job bundles, not just the templates, so template_name feels a bit weird

```suggestion
def get_job_bundle_path(name: str) -> str:
    """
    Get the path to a job bundle directory.

    Args:
        name: Name of the job bundle (e.g., 'make_many_small_files')

    Returns:
        Path to the job bundle directory
    """
```

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -0,0 +1,501 @@
...
+            result = wait_for_job_completion(
+                farm_id=self.farm_id,
+                queue_id=self.queue_id,
+                job_id=job_id,
+                timeout=timeout,
+                max_poll_interval=poll_interval,
+            )
+            return result.status == "SUCCEEDED", result.status
+        except Exception:
+            return False, "TIMEOUT"
```
jericht: Can we catch the specific timeout exception? or put the exception information in the `final_status` output string? We may lose debuggability by swallowing the original exception here.

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -0,0 +1,501 @@
...
+        self,
+        output_dir: str,
+        force_bootstrap: bool = False,
+        conflict_resolution: Optional[str] = None,
+        lookback_window: Optional[int] = None,
+        test_name: Optional[str] = None,
+    ) -> subprocess.CompletedProcess:
+        """Run the incremental download CLI command."""
+        # Use test-specific checkpoint directory to avoid conflicts between parallel tests
+        checkpoint_suffix = f"_{test_name}" if test_name else ""
```
jericht, line 60: nit: Will this be sufficient for parametrized tests? Maybe we can add a random string to the end?

or maybe we can just make sure to always pass in the full test name that pytest collects for these cases since it will include parameters in the test name for parametrized tests like `test_something[happy-path]`, `test_something[sad_path]`

.. we can do this later if/when we have parametrized tests

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -0,0 +1,501 @@
...
+        job_complete = task_run_status in ["SUCCEEDED", "FAILED", "CANCELED"]
+        if not job_complete:
+            time.sleep(5)
+
+    # Run final incremental download to ensure all files are captured
+    time.sleep(5)
+    final_result = incremental_download_test.run_incremental_download_without_storage_profiles(
+        str(tmp_path), test_name="dep_chain"
+    )
+    assert final_result.returncode == 0, f"Final incremental download failed: {final_result.stderr}"
```
jericht, line 348: nit: It looks like we have use this pattern in multiple tests. Would it be worth pulling it out into a separate function?

test/integ/cli/test_cli_incremental_download.py
sakshie95: I think it's sufficient because we have the requeue parameter in the test name and makes the directories isolated - we don't need a random string. Also to clarify the tmp_path would also be different for the different parameters in this case so it's completely isolated

test/integ/cli/test_cli_incremental_download.py
sakshie95: Yeah I thought about this.. I wanted to keep it separate instead of trying to combine it in one just because these are the core tests. I've extracted some functions out from the conflict resolution test below though that I'll re-use across tests

test/integ/cli/test_cli_incremental_download.py
sakshie95: Yes sure that makes sense, adding the exception to the ``final_status`` string. Getting the change out soon

test/integ/cli/job_templates.py
sakshie95: Sure will do :))

test/integ/cli/job_bundles/dep_chain/scripts/create_files.sh
sakshie95: That's a great call, thanks! Looks like there's an edge case for Windows that didn't get caught in my dev Mac. Fixed it now but couldn't test it on windows unfortunately. I think it makes sense to set it for all files instead of being selective, so future tests that use and *.sh files are already handled for both platforms.

As a follow up - a safeguard we should add is editing the PR template to include running the integ tests on Posix and Windows for all PRs.

test/integ/cli/job_templates.py
sakshie95: Chatted more about this offline. Looks like this happens when the defaults conflict with the env variable. Inside deadline config, ``queue_id`` is nested under ``farm_id``. Since this change set the top level queue_id, that gets ignored, and it continues using the default queue id. Didn't happen on my dev testing because the ``queue_id`` being used in the env variables was the same as the default queue in the deadline config so there was no conflict. Fixed it

test/integ/cli/job_templates.py
sakshie95: Done

test/integ/cli/test_cli_incremental_download.py
sakshie95: Done

jericht (approved):

mwiebe (approved):

PR #789, opened 2025-08-20
chore: fix typing in telemetry decorators

jericht:
Some of the typing in the telemetry decorator functions was incorrect, and the decorator to record success/fail metrics had an incorrect return value typing so usages of it required a `# type: ignore`.

The positional (`*args`) and keyword arguments (`**kwargs`) in the functions were typed incorrectly. They should be typed with the expected type of one value in the collection. See https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

- Fix the typing of the success/fail metric decorator and removed the `# type: ignore` in uses of it
- Fix the typing of positional `*args` and keyword arguments `**kwargs`

Correct typing, no functional impact.

`hatch run fmt && hatch run lint && hatch run test`

No

No

No

----

### DIFF
--- a/src/deadline/client/api/_telemetry.py
+++ b/src/deadline/client/api/_telemetry.py
@@ -16,7 +16,7 @@ from dataclasses import asdict, dataclass, field
 from datetime import datetime
 from queue import Queue, Full
 from threading import Thread
-from typing import Any, Callable, Dict, List, Optional, TypeVar, cast
+from typing import Any, Callable, Dict, Optional, TypeVar, cast
 from urllib import request, error
 
 from ...job_attachments.progress_tracker import SummaryStatistics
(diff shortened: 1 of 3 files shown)


leongdl (dismissed):
I like this change! Typing is good.

src/deadline/client/api/_telemetry.py
```diff
@@ -383,14 +383,14 @@ def get_deadline_cloud_library_telemetry_client(
     return get_telemetry_client("deadline-cloud-library", version, config=config)
 
 
-def record_success_fail_telemetry_event(**decorator_kwargs: Dict[str, Any]) -> Callable[..., F]:
+def record_success_fail_telemetry_event(**decorator_kwargs) -> Callable[[F], F]:
```
jairaws: Should `decorator_kwargs` be typed as `Any`. Doing so explicitly shows that we didn't miss this type.

Same for all other cases.

jericht: Yeah we can, but `Any` is essentially the same as no type, which is why I removed it. I can put it back if that's preferred though

jairaws: Yeah, I agree. It's the same. Just to make it explicit.

jairaws (approved):

leongdl (approved):

PR #794, opened 2025-08-22
refactor(experimental): Clean up code and add unit tests for incremental output download

mwiebe:

We identified a few bugs, discussed improved names for the `deadline queue incremental-output-download` command, and looked through the coverage report to see where the unit tests were most lacking.

Roll up these changes into one PR as follows:

* Rename `deadline queue incremental-output-download` to `deadline queue sync-output` as a more succinct but still descriptive name.
* Connect the --conflict-resolution parameter to the _incremental_download call.
* Add missing logic for jobs that do not have a storage profile when storage profiles are being used.
* Add tests using moto to simulate an S3 bucket for manifest and object downloads in the incremental download code.
* Remove the file asset_downloader_with_manifest_aggregation.py and its tests, as it is not being used anymore.
* Remove the function _update_manifest_output_paths_for_session_actions from download.py as it does not control the order of the manifests list to match the output of deadline:GetJob.
  * In the future, we can consider promoting the _add_output_manifests_from_s3 function from the _manifest_s3_downloads.py file. It takes the job dictionary as additional context in order to match the ordering. Note that this function is potentially less efficient than _update_manifest_output_paths_for_session_actions, but we have not observed performance issues in practice.

Polishing up the command towards release.

Tested manually against a Deadline Cloud queue. Added new unit tests.

Changed subcommand name is reflected in the help.

No

No, it modifies code that is not public interface yet.

No

----


mwiebe:
I checked and all the SonarCloud findings are about secure random number generation and S3 security within the tests using moto.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -368,6 +368,13 @@ def _categorize_jobs_in_checkpoint(
 
         print_function_callback(f"NEW Job: {dc_job['name']} ({job_id})")
 
+        if dc_job["storageProfileId"] is None and checkpoint.local_storage_profile_id is not None:
+            print_function_callback(
```
sakshie95, line 372: One thing we discussed during ORR was to add a regex print in the logs for such jobs for the customer to easily filter them out of the console/cron logs in case they get into an issue. This is specifically important for the case where there's a file system location name missing for a job storage profile (so a misconfigured storage profile not a missing one).

src/deadline/client/cli/_incremental_download.py
```diff
@@ -1180,7 +1188,7 @@ def _update_download_progress(
             HashAlgorithm.XXH128,
             queue,
             boto3_session_for_s3,
-            FileConflictResolution.OVERWRITE,
+            file_conflict_resolution,
```
sakshie95, line 1191: thanks for fixing this!

src/deadline/client/cli/_incremental_download.py
mwiebe: I think you're talking about this warning message, is that right?

https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_incremental_download.py#L1113

sakshie95: oh yes this looks like it. Do you mind adding some information about the unmapped paths too? Might be useful for them to debug - I'm just looking at it from the operational lens for a customer if they need to re-run some failed stuff because we aren't failing fast & continue downloading the rest if there's an unmapped path?

sakshie95 (approved):

src/deadline/client/cli/_incremental_download.py
mwiebe: Yes, makes sense. Will have to think about where we put that, whether it's documentation or also goes somewhere else.

src/deadline/client/cli/_groups/queue_group.py
```diff
@@ -268,19 +268,21 @@ def queue_get(**args):
...
     dry_run: bool,
     **args,
 ):
     """
-    BETA - Downloads job attachments output incrementally for all jobs in a queue. When run for the
-    first time or with the --force-bootstrap option, it starts downloading from --bootstrap-lookback-minutes
+    BETA - Downloads any new job attachments output for all jobs in a queue since the last run of the same command.
+
+    When run for the first time or with the --force-bootstrap option, it starts downloading from --bootstrap-lookback-minutes
     in the past. When run each subsequent time, it loads  the previous checkpoint and continues
```
jericht, line 286:
```suggestion
    in the past. When run each subsequent time, it loads the previous checkpoint and continues
```

nit: there's a double space, not from your change but just caught my eye

jericht (approved):

epmog (approved):

jericht:
> I checked and all the SonarCloud findings are about secure random number generation and S3 security within the tests using moto.

A pain in the back, but you can go into SonarCloud, login with GitHub (it will use your current login session in your browser, no need to input your credentials), then you can manually resolve the findings as false positives.

PR #807, opened 2025-08-26
test: Fix flaky sync-output test and add xfail strict false workaround for job not getting picked

sakshie95:
Test ``test_conflict_resolution_with_requeue`` is flaky because it doesn't wait for job to be re-queued successfully and its status to be updated. It is also flaky when test runs overlap because different parametrized versions read each other's outputs.

```FAILED test/integ/cli/test_cli_incremental_download.py::test_conflict_resolution_with_requeue[job] - AssertionError: Expected 20 files after initial, got 40```

There is another issue where a job doesn't get picked up in time during auto scaling.

Added wait for re-queue to be successfully done - check job status changes to one of "NOT SUCCEEDED" statuses. And also make sure each parametrized version only reads its own output dir.

Setting the failing test to ``xfail(strict=False)`` for now, we'll get a better fix out soon.

Successful integ tests

Ran the tests locally

```
600.02s call     test/integ/cli/test_cli_incremental_download.py::test_incremental_download_dependency_chain
503.35s call     test/integ/cli/test_cli_incremental_download.py::test_conflict_resolution_with_requeue[job]
336.60s call     test/integ/cli/test_cli_incremental_download.py::test_incremental_download_many_small_files
167.17s call     test/integ/cli/test_cli_incremental_download.py::test_incremental_download_dep_data_flow
109.97s call     test/integ/cli/test_cli_incremental_download.py::test_conflict_resolution_with_requeue[step]
=============================================================== 56 passed, 4 skipped, 1 xfailed in 1940.33s (0:32:20) ===============================================================
(
```

- Have you run the unit tests? No

- Are relevant docstrings in the code base updated?N/A

No

No

----

### DIFF
--- a/test/integ/cli/test_cli_incremental_download.py
+++ b/test/integ/cli/test_cli_incremental_download.py
@@ -11,6 +11,7 @@ from typing import Optional, Tuple
 
 import boto3
 import pytest
+from pathlib import Path
 
 from deadline.client.api._job_monitoring import wait_for_job_completion
 from .job_templates import (
@@ -301,6 +302,9 @@ def test_incremental_download_dep_data_flow(incremental_download_test, tmp_path)
 
 @pytest.mark.integ
 @pytest.mark.timeout(900)  # 15 minutes timeout
+@pytest.mark.xfail(
+    reason="Workaround until scheduler does not pick this job, flaky test", strict=False
+)
 def test_incremental_download_dependency_chain(incremental_download_test, tmp_path):
     """Test incremental download with dep_chain template."""
 
@@ -413,7 +417,12 @@ def test_conflict_resolution_with_requeue(incremental_download_test, requeue_lev
 
     # Download initial files
     _run_download_until_complete(
-        incremental_download_test, tmp_path, expected_initial_files, requeue_level, "initial"
+        incremental_download_test,
+        tmp_path,
+        unique_output_dir,
+        expected_initial_files,
+        requeue_level,
+        "initial",
     )
 
     # Requeue at specified level
@@ -428,12 +437,23 @@ def test_conflict_resolution_with_requeue(incremental_download_test, requeue_lev
     # Download files after requeue
     expected_final = _get_expected_file_count(requeue_level, expected_initial_files, files_per_task)
     _run_download_until_complete(
-        incremental_download_test, tmp_path, expected_final, requeue_level, "requeue"
+        incremental_download_test,
+        tmp_path,
+        unique_output_dir,
+        expected_final,
+        requeue_level,
+        "requeue",
     )
 
 
-def _run_download_until_complete(test_instance, tmp_path, expected_count, level, phase):
+def _run_download_until_complete(
+    test_instance, tmp_path, unique_output_dir, expected_count, level, phase
+):
     """Run incremental download until expected file count is reached."""
+
+    # Use the actual unique output directory passed from the test
+    test_output_dir = Path(unique_output_dir)
+
     for iteration in range(1, 11):  # Max 10 iterations
         result = test_instance.run_incremental_download_without_storage_profiles(
             str(tmp_path),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/integ/cli/test_cli_incremental_download.py [hunk 1]
```diff
@@ -414,7 +417,12 @@ def test_conflict_resolution_with_requeue(incremental_download_test, requeue_lev
 
     # Download initial files
     _run_download_until_complete(
-        incremental_download_test, tmp_path, expected_initial_files, requeue_level, "initial"
+        incremental_download_test,
```
leongdl, line 415: Can we please add mypy typing, this is really hard to process.

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -447,13 +466,44 @@ def _run_download_until_complete(test_instance, tmp_path, expected_count, level,
...
+        task_run_status = job.get("taskRunStatus")
+
+        print(f"Current job taskRunStatus: {task_run_status}")
+
+        # Job has been re-queued if it's no longer SUCCEEDED
+        if task_run_status in ["READY", "ASSIGNED", "STARTING", "SCHEDULED", "RUNNING"]:
+            print(f"Requeue took effect - job status is now: {task_run_status}")
+            return True
+
+        time.sleep(2)
```
leongdl, line 495: This seems odd and arbitary - can we use a waiter instead?

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -447,13 +466,44 @@ def _run_download_until_complete(test_instance, tmp_path, expected_count, level,
...
+
+    # If we get here, requeue didn't take effect within timeout
+    job = client.get_job(farmId=farm_id, queueId=queue_id, jobId=job_id)
+    current_status = job.get("taskRunStatus")
+    raise AssertionError(
+        f"Requeue did not take effect within {timeout}s. Job status is still: {current_status}"
     )
```
jericht, line 502: nit: does this need to be a separate function at this point? Do we plan on using it anywhere else outside the `_requeue_at_level` function?

test/integ/cli/test_cli_incremental_download.py
```diff
@@ -447,13 +466,44 @@ def _run_download_until_complete(test_instance, tmp_path, expected_count, level,
...
+        # Job has been re-queued if it's no longer SUCCEEDED
+        if task_run_status in ["READY", "ASSIGNED", "STARTING", "SCHEDULED", "RUNNING"]:
+            print(f"Requeue took effect - job status is now: {task_run_status}")
+            return True
+
+        time.sleep(2)
+
+    # If we get here, requeue didn't take effect within timeout
+    job = client.get_job(farmId=farm_id, queueId=queue_id, jobId=job_id)
+    current_status = job.get("taskRunStatus")
```
jericht, line 499: nit: Do we need to fetch the `taskRunStatus` again here? It should still be populated in `task_run_status` from the previous run, assuming the while block runs at least once. 

There is a race condition where in between the timeout and this refresh of the `taskRunStatus` that the job does transition to a requeued state and this assertion error would be confusing.

jericht (approved):

test/integ/cli/test_cli_incremental_download.py
sakshie95: sure, easy fix - I can get it up in my follow-up PR to re-introduce the xfail test above with workarounds.

test/integ/cli/test_cli_incremental_download.py
sakshie95: sure, easy fix - I can get it up in my follow-up PR to re-introduce the xfail test above with workarounds.

test/integ/cli/test_cli_incremental_download.py
sakshie95: Yes we have to add more requeue tests where this will come handy so i extracted it as a function

test/integ/cli/test_cli_incremental_download.py
sakshie95: taskRunStatus changes with the job/step/task getting requeued that's why we need to fetch it again. That was the cause of flakiness

test/integ/cli/test_cli_incremental_download.py
leongdl: ok thanks - I'll take an IOU.

test/integ/cli/test_cli_incremental_download.py
leongdl: ok thanks - I'll take an IOU.

leongdl (approved):

PR #816, opened 2025-08-29
feat(job download): Add fileCount to JSON summary for programmatic access

larrygao001:

Downstream tools (like Deadline Cloud Monitor) need to programmatically determine how many files were downloaded from the CLI's 
JSON output. Currently, they must parse the human-readable message string "Downloaded 5 files" to extract the file count.

Why this is problematic:
• String parsing is fragile and could break with message changes
• Future internationalization (i18n) would break existing parsing logic
• Not following established patterns for structured JSON API outputs
• Creates maintenance burden for downstream consumers

Requirement:
Provide a structured, backwards-compatible way for programmatic tools to access the file count without relying on string parsing.

Enhanced the JSON summary output to include a dedicated fileCount field alongside the existing human-readable message.

Implementation:
1. Enhanced _get_json_line() function:
• Added extra_properties parameter to accept additional JSON fields
• Made it extensible for future metadata additions

2. Updated JSON summary output:
• Added fileCount field to summary messages in JSON format
• Preserved existing human-readable value field for backwards compatibility

Before:
```json
{"messageType": "summary", "value": "Downloaded 5 files"}
```

After:
```json
{"messageType": "summary", "fileCount": 5, "value": "Downloaded 5 files"}
```

For Downstream Tools:
• Reliable programmatic access to file count via structured data
• No dependency on string parsing or message format
• Future-proof against i18n and message changes
• Consistent with other structured JSON outputs in the CLI

For Maintainability:
• Follows established JSON API patterns for backwards compatibility
• Reduces fragility in downstream integrations
• Enables confident CLI message updates without breaking consumers

Risk Mitigation:
• **Zero breaking changes:** Existing tools continue to work unchanged
• **Backwards compatible:** Original value field preserved
• **Extensible pattern:** Framework for adding future structured fields

Scope:
This change only affects the JSON output format (--output json flag). Human-readable output remains unchanged. The enhancement 
supports existing and future programmatic consumers while maintaining full backwards compatibility.

- Yes
- No
- Yes

N/A

No

No

----

### DIFF
--- a/src/deadline/client/cli/_groups/job_group.py
+++ b/src/deadline/client/cli/_groups/job_group.py
@@ -609,6 +609,7 @@ def _get_download_summary_message(
         return _get_json_line(
             JSON_MSG_TYPE_SUMMARY,
             f"Downloaded {download_summary.processed_files} files",
+            extra_properties={"fileCount": download_summary.processed_files},
         )
     else:
         paths_joined = "\n        ".join(
@@ -649,8 +650,15 @@ def _get_conflicting_filenames(filenames_by_root: dict[str, list[str]]) -> list[
     return conflicting_filenames
 
 
-def _get_json_line(messageType: str, value: Union[str, list[str], dict[str, Any]]) -> str:
-    return json.dumps({"messageType": messageType, "value": value}, ensure_ascii=True)
+def _get_json_line(
+    messageType: str,
+    value: Union[str, list[str], dict[str, Any]],
+    extra_properties: Optional[dict[str, Any]] = None,
+) -> str:
+    json_obj = {"messageType": messageType, "value": value}
+    if extra_properties:
+        json_obj.update(extra_properties)
+    return json.dumps(json_obj, ensure_ascii=True)
 
 
 def _get_value_from_json_line(
--- a/test/unit/deadline_client/cli/test_cli_job.py
+++ b/test/unit/deadline_client/cli/test_cli_job.py
@@ -21,7 +21,11 @@ from dateutil.tz import tzutc  # type: ignore[import]
 from deadline.client import api, config
 from deadline.client.cli import main
 from deadline.client.cli._groups import job_group
-from deadline.client.cli._groups.job_group import _get_summary_of_files_to_download_message
+from deadline.client.cli._groups.job_group import (
+    _get_summary_of_files_to_download_message,
+    _get_json_line,
+    _get_download_summary_message,
+)
 from deadline.client.exceptions import DeadlineOperationError, DeadlineOperationTimedOut
 from deadline.job_attachments.models import (
     FileConflictResolution,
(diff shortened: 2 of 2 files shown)


src/deadline/client/cli/_groups/job_group.py
```diff
@@ -609,6 +609,7 @@ def _get_download_summary_message(
         return _get_json_line(
             JSON_MSG_TYPE_SUMMARY,
             f"Downloaded {download_summary.processed_files} files",
+            fileCount=download_summary.processed_files,
```
epmog: Hmmm, I wonder if we just want to do a different message type instead of piggy backing on the JSON_MSG_TYPE_SUMMARY.

That should simplify the consuming code by not needing to add a conditional value check?

larrygao001: This is what the messages currently look like:


➜  ~ deadline job download-output --output json
```
{"messageType": "title", "value": "Keyframe-Robot-Scene Sharing Version"}
{"messageType": "progress", "value": "83"}
{"messageType": "progress", "value": "92"}
{"messageType": "progress", "value": "93"}
{"messageType": "progress", "value": "98"}
{"messageType": "progress", "value": "100"}
{"messageType": "progress", "value": "100"}
{"messageType": "summary", "value": "Downloaded 5 files"}
```

The consumer, DCM, populates the notification based on the summary stage. I think the consumer code is simple enough when determining the fileCount(`this.notification.fileCount = message.fileCount || 0;`), and current approach is better because:

1. Semantic consistency: The summary message type already represents completion information. Adding fileCount as metadata to the existing summary maintains semantic coherence - it's still summary information, just with structured data for programmatic access.

2. Backwards compatibility: The approach preserves the existing value field with human-readable text while adding machine-readable data. A new message type would require consumers to handle two different message types for essentially the same information.

3. Extensibility: The **kwargs enhancement to _get_json_line() creates a clean pattern for adding structured metadata to any message type in the future (e.g., duration, totalBytes, etc.) without proliferating message types.

4. Consumer simplicity: Consumers can choose their preferred access pattern:
   • Parse fileCount for programmatic use
   • Parse value for human display
   • Handle both in the same message handler

5. Avoids duplication: A separate message type would likely duplicate the file count information that's already in the summary text, violating DRY principles.

The current output structure is clean and purposeful:
```json
{"messageType": "summary", "fileCount": 5, "value": "Downloaded 5 files"}
```

crowecawcaw: I like the single message. As an implementation nit, it feels a bit magical that all arguments to `_get_json_line` become JSON properties. It also makes the casing on the arguments look funny. I think it'd be clearer to swap out kwargs for a named argument like:`
```python
get_json_line(
            JSON_MSG_TYPE_SUMMARY,
            f"Downloaded {download_summary.processed_files} files",
            extra_properties={"fileCount": download_summary.processed_files}
)
```

epmog: Sure, I don't have a super strong opinion here. Was more interested in discussing the differences.

larrygao001: Thanks for the feedback, pushed an update.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -649,8 +650,15 @@ def _get_conflicting_filenames(filenames_by_root: dict[str, list[str]]) -> list[
...
-def _get_json_line(messageType: str, value: Union[str, list[str], dict[str, Any]]) -> str:
-    return json.dumps({"messageType": messageType, "value": value}, ensure_ascii=True)
+def _get_json_line(
+    messageType: str,
+    value: Union[str, list[str], dict[str, Any]],
+    extra_properties: Optional[dict[str, Any]] = None,
+) -> str:
+    json_obj = {"messageType": messageType, "value": value}
+    if extra_properties:
+        json_obj.update(extra_properties)
```
epmog, line 660: unlikely, but with this folks can just override messageType/Value in the same call. I may take a look at this later by making it keyword args only.

Internal only so this doesn't really matter to address the issue DCM is facing.

epmog (approved):

crowecawcaw (approved):

PR #822, opened 2025-09-02
test: Isolate integration tests from the user's config

mwiebe:

I found that running the integration tests would fail depending on the state of my config file in ~/.deadline/config.

This modifies the integration test suite so every test function gets an isolated config file. With this change, the tests run reliably independent of how I set various options in the config.

A more reliable integration test for developers.

Ran the integration tests, saw they passed.

No

No

No

No

----

### DIFF
--- a/test/integ/cli/conftest.py
+++ b/test/integ/cli/conftest.py
@@ -1,12 +1,59 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
 from unittest.mock import MagicMock
+import tempfile
+from pathlib import Path
+import os
+
 import pytest
 from deadline.job_attachments import upload
+from deadline.client.config import config_file
 from deadline.job_attachments._aws.deadline import get_queue
 from deadline.job_attachments.asset_manifests.hash_algorithms import HashAlgorithm, hash_file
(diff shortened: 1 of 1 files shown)


test/integ/cli/conftest.py
```diff
@@ -1,14 +1,61 @@
...
 from deadline.job_attachments import upload
+from deadline.client.config import config_file
 from deadline.job_attachments._aws.deadline import get_queue
 from deadline.job_attachments.asset_manifests.hash_algorithms import HashAlgorithm, hash_file
 from deadline.job_attachments.asset_manifests.versions import ManifestVersion
 from .test_utils import JobAttachmentTest, UploadInputFilesOneAssetInCasOutputs, DeadlineCliTest
 
 
+@pytest.fixture(scope="function", autouse=True)
+def fresh_deadline_config():
```
leongdl, line 18: I thought we had a fixture for this already?

leongdl (approved):

test/integ/cli/conftest.py
mwiebe: Yes, it's in the unit tests and doesn't apply to the integration tests. If I remember right, it's also applied selectively because some tests are checking details related to the config.

epmog (approved):

PR #825, opened 2025-09-04
feat: Add 'deadline job requeue-tasks' command

mwiebe:

The Deadline CLI doesn't provide a command to requeue tasks, equivalent to functionality in Deadline Cloud monitor.
In particular, requeuing based on a set of statuses would be useful, e.g. requeuing all SUSPENDED or CANCELED tasks.

* This command lets you requeue tasks in a job that match a set of task run statuses.
* Promoted the deadline_mock fixture to the client unit test conftest.py to use the moto/deadline mocking for tests of the new command.
* Added unit tests of the new CLI command.
* Change the wording on the '--yes' option from "Skip" to "Automatically accept" so it matches the config dialog.

The Deadline CLI provides more needed functionality for managing a farm.

Wrote unit tests, and tested manually on a few jobs.

It's in the help output of the CLI.

No

No

No

----

### DIFF
--- a/src/deadline/client/cli/_groups/job_group.py
+++ b/src/deadline/client/cli/_groups/job_group.py
@@ -19,22 +19,22 @@ import textwrap
 import click
 from botocore.exceptions import ClientError
 
-from deadline.client.api._session import _modified_logging_level
-from deadline.job_attachments.download import OutputDownloader
-from deadline.job_attachments.models import (
+from ...api._session import _modified_logging_level
+from ....job_attachments.download import OutputDownloader
+from ....job_attachments.models import (
     FileConflictResolution,
     JobAttachmentS3Settings,
     PathFormat,
 )
-from deadline.job_attachments._utils import (
+from ....job_attachments._utils import (
     WINDOWS_MAX_PATH_LENGTH,
     _is_windows_long_path_registry_enabled,
 )
-from deadline.job_attachments.progress_tracker import (
+from ....job_attachments.progress_tracker import (
     DownloadSummaryStatistics,
     ProgressReportMetadata,
 )
-from deadline.job_attachments._path_summarization import (
+from ....job_attachments._path_summarization import (
     human_readable_file_size,
     summarize_path_list,
 )
@@ -45,6 +45,7 @@ from ...exceptions import DeadlineOperationError, DeadlineOperationTimedOut
 from .._common import _apply_cli_options_to_config, _cli_object_repr, _handle_error
 from .._main import main
 from ._sigint_handler import SigIntHandler
+from ...api._session import get_default_client_config
 
 logger = logging.getLogger("deadline.client.cli")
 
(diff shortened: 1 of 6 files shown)


src/deadline/client/cli/_groups/job_group.py
```diff
@@ -261,6 +262,158 @@ def job_cancel(mark_as: str, yes: bool, **args):
...
+        return
+
+    # Ask for confirmation about requeuing the selected tasks
+    if config_file.str2bool(config_file.get_setting("settings.auto_accept", config=config)):
+        click.echo(
+            f"Estimated {summary_tasks_message} ({summary_task_count_by_status}) to requeue."
+        )
+    else:
+        click.echo(
+            f"This action will requeue an estimated {summary_tasks_message} ({summary_task_count_by_status})"
```
epmog, line 347: Estimated? What sort of situations lead this to be an inaccurate count? Eventual consistency + timing between reporting and actually performing the operation?

Answered my own question. I'll resolve :)

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -261,6 +262,158 @@ def job_cancel(mark_as: str, yes: bool, **args):
...
+                    # Requeue means to set the target run status to PENDING. If a task has no dependencies,
+                    # it will switch to READY immediately.
+                    deadline_client_for_requeues.update_task(
+                        farmId=farm_id,
+                        queueId=queue_id,
+                        jobId=job_id,
+                        stepId=step["stepId"],
+                        taskId=task["taskId"],
+                        targetRunStatus="PENDING",
+                    )
```
epmog, line 411: Can the update fail aside from throttling? Can we get things like `ConflictException`s here if that task is being updated? If so, is that handled by the boto3 retry logic?

Should we continue on if any of that fails?

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -261,6 +262,158 @@ def job_cancel(mark_as: str, yes: bool, **args):
...
+                        task_summary = (
+                            ",".join(
+                                f"{param}={list(parameters[param].values())[0]}"
+                                for param in parameters
+                            )
+                            + f" ({task['taskId']})"
+                        )
+                    else:
+                        task_summary = task["taskId"]
+                    click.echo(f"    {task['runStatus']} {task_summary}")
```
epmog, line 401: aside and maybe just a future improvement: thinking about a user interfacing with this, I wonder if we should shorten some of this output by default and allow the user to specify more verbose output. Any large amount of steps/tasks will basically flood your terminal, and having that as the default behaviour doesn't feel great.

src/deadline/client/cli/_groups/job_group.py
mwiebe: Tasks might complete (or otherwise change status) between the initial and later calls, changing the set of tasks matching the filter.

src/deadline/client/cli/_groups/job_group.py
mwiebe: Probably, I'm not sure what the behavior will be. I'm inclined to follow the defaults for now, and adjust after broader usage.

src/deadline/client/cli/_groups/job_group.py
mwiebe: It's a good idea, yeah. When I was forming the output, I had a stray thought that a "list-tasks" command that outputs theses details would be great as well. I think that's better as a follow-up though.

src/deadline/client/cli/_groups/job_group.py
epmog: Gotcha, we'll end up failing on the first one if anything else goes wrong. Can easily revisit this later.

epmog (approved):

leongdl (approved):

PR #835, opened 2025-09-12
chore: add ManifestProperties.as_output_metadata for S3 metadata generation, ManifestProperties.from_dcit to create from dictionary

godobyte:
1. We need a function to construct ManifestProperties from dictionary
2. S3 output manifest metadata generation logic was scattered across different modules 
instead of being encapsulated within the relevant data model.

-  Implemented `get_output_metadata()` method in the ManifestProperties class
- Added a function to construct ManifestProperties from dictionary

1. Defines the contract for s3 metadata in ManifestProperties data class.
2. Being able to construct ManifestProperties instances from API response.

- [N/A] Have you run the integration tests?

- [Yes] Are relevant docstrings in the code base updated?
- [N/A] Has the README.md been updated? If you modified CLI arguments, for instance.

No

No

----


src/deadline/job_attachments/api/attachment.py
```diff
@@ -276,3 +243,111 @@ def _process_path_mapping(
     )
 
     return path_mapping_rule_list
+
+
+def _process_s3_metadata(s3_metadata: Dict[str, str]) -> Dict[str, str]:
```
mwiebe: I thought we only have this issue for the `asset-root` metadata. Which other keys are affected? Because this is used in an interface defined between different components, it would be better to only handle the specific keys we want this to happen for, not in general.

src/deadline/job_attachments/api/attachment.py
```diff
@@ -276,3 +243,111 @@ def _process_path_mapping(
...
+    boto3_session: boto3.Session,
+    source_root: Path,
+    s3_upload_manifest_name: str,
+    s3_upload_manifest_path: str,
+    s3_upload_manifest_metadata: Optional[Dict[str, str]] = None,
+    logger: ClickLogger = ClickLogger(False),
+) -> UploadManifestInfo:
+    """
+    This API processes individual manifests with explicit path correlation,
+    eliminating the need for hash-based correlation used in the internal APIs.
```
mwiebe: Most developers looking at this code in a context where they are thinking about the hash-based correlation idea. Could you make the description directly describe what the function does for someone who is looking at it with fresh eyes. Maybe something like "This function uploads all the files referenced by a manifest from the specified source_root file system path."

mwiebe (dismissed):

src/deadline/job_attachments/api/attachment.py
godobyte: The intention behind that is that the `_attachment_upload_single` function can process any s3 metadata, and we need to validate the input. As per synced offline, we decided to remove this new API and reuse an existing API from `upload.py`. 

The output manifest metadata handling (asset root, file system location name) now lives in the `ManifestProperties` class as the defined contract.

src/deadline/job_attachments/api/attachment.py
godobyte: Thanks for the suggestion! As per offline discussion to reuse an existing public API in `upload.py`, I've removed this new API.

src/deadline/job_attachments/models.py
```diff
@@ -215,6 +216,46 @@ def to_dict(self) -> dict[str, Any]:
...
+        return cls(
+            rootPath=data["rootPath"],
+            rootPathFormat=PathFormat(data["rootPathFormat"]),
+            fileSystemLocationName=data.get("fileSystemLocationName"),
+            inputManifestPath=data.get("inputManifestPath"),
+            inputManifestHash=data.get("inputManifestHash"),
+            outputRelativeDirectories=data.get("outputRelativeDirectories"),
+        )
+
+    def get_output_metadata(self) -> dict[str, dict[str, str]]:
```
godobyte: The logic mirrors existing behavior to ensure backward compatibility.

https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/asset_sync.py#L543-L554.

src/deadline/job_attachments/models.py
```diff
@@ -215,6 +216,46 @@ def to_dict(self) -> dict[str, Any]:
...
+        metadata = {"Metadata": {"asset-root": json.dumps(self.rootPath, ensure_ascii=True)}}
+        # S3 metadata must be ASCII, so use either 'asset-root' or 'asset-root-json' depending
+        # on whether the value is ASCII.
+        try:
+            # Add the 'asset-root' metadata if the path is ASCII
+            self.rootPath.encode(encoding="ascii")
+            metadata["Metadata"]["asset-root"] = self.rootPath
+        except UnicodeEncodeError:
+            # Add the 'asset-root-json' metadata encoded to ASCII as a JSON string
+            metadata["Metadata"]["asset-root-json"] = json.dumps(self.rootPath, ensure_ascii=True)
```
godobyte: FYI the current implementation adds json encoded string as both `asset-root` and `asset-root-json` when the path is non-ascii, see `test_get_output_metadata_non_ascii_path` test below.

jericht:
Can we update the commit title to be more focused on the customer-facing functionality/feature being added? Since this is a `feat:`, it will show in the release notes.

edit: After looking at the code, maybe this is not worth the `feat` tag. It's customer facing when looking at the package as a library, but not as an application (CLI or GUI). Maybe we can change this to a `chore:`?

src/deadline/job_attachments/models.py
jericht: nit: Function name sounds like a getter but it's really representing the data in this class in another form. How about something else like `as_output_metadata`?

test/unit/deadline_job_attachments/test_models.py
```diff
@@ -182,3 +183,190 @@ def test_manifest_snapshot_json_serialization_special_characters(self):
...
+
+        with pytest.raises(KeyError, match="rootPathFormat"):
+            ManifestProperties.from_dict(data)
+
+    def test_from_dict_invalid_path_format(self):
+        """Test ManifestProperties.from_dict raises ValueError for invalid path format"""
+        data = {"rootPath": "/test/path", "rootPathFormat": "invalid_format"}
+
+        with pytest.raises(ValueError):
+            ManifestProperties.from_dict(data)
```
jericht: Can we be more specific about the `ValueError` being raised here to make sure it's the one we expect? What error message does the exception object have? Can we assert on something in there?

test/unit/deadline_job_attachments/test_models.py
```diff
@@ -182,3 +183,190 @@ def test_manifest_snapshot_json_serialization_special_characters(self):
...
+        data = original.to_dict()
+        recreated = ManifestProperties.from_dict(data)
+
+        # Verify they are equal
+        assert recreated.rootPath == original.rootPath
+        assert recreated.rootPathFormat == original.rootPathFormat
+        assert recreated.fileSystemLocationName == original.fileSystemLocationName
+        assert recreated.inputManifestPath == original.inputManifestPath
+        assert recreated.inputManifestHash == original.inputManifestHash
+        assert recreated.outputRelativeDirectories == original.outputRelativeDirectories
```
jericht: Can we just use dataclass equality here? This test will become outdated if we forget to update it when we add more optional fields to the dataclass. dataclass automatically generates an `__eq__` method to compare the fields like this for us ([ref](https://docs.python.org/3/library/dataclasses.html#:~:text=eq%3A%20If%20true%20(the%20default)%2C%20an%20__eq__()%20method%20will%20be%20generated.%20This%20method%20compares%20the%20class%20as%20if%20it%20were%20a%20tuple%20of%20its%20fields%2C%20in%20order.%20Both%20instances%20in%20the%20comparison%20must%20be%20of%20the%20identical%20type.))
```suggestion
    def test_from_dict_roundtrip_with_to_dict(self):
        """Test that from_dict and to_dict are inverse operations"""
        # Create a ManifestProperties instance with all fields
        original = ManifestProperties(
            rootPath="/original/path",
            rootPathFormat=PathFormat.POSIX,
            fileSystemLocationName="test-location",
            inputManifestPath="s3://bucket/manifest.json",
            inputManifestHash="hash123",
            outputRelativeDirectories=["out1", "out2"],
        )

        # Convert to dict and back
        data = original.to_dict()
        recreated = ManifestProperties.from_dict(data)

        # Verify they are equal
        assert recreated == original
```

src/deadline/job_attachments/models.py
jericht: The comment is confusing:

```py
# S3 metadata must be ASCII, so use either 'asset-root' or 'asset-root-json' depending
# on whether the value is ASCII.
```

We always populate `asset-root` here, but also populate `asset-root-json` if the path is not ASCII. IMO, structuring the code like this may make it a bit easier to understand without the comment

```py
        metadata = {}
        try:
            # Set 'asset-root' metadata as the path if the path is ASCII
            self.rootPath.encode(encoding="ascii")
            metadata["Metadata"]["asset-root"] = self.rootPath
        except UnicodeEncodeError:
            # Add both 'asset-root' and 'asset-root-json' metadata encoded to ASCII as a JSON string
            json_root_path = json.dumps(self.rootPath, ensure_ascii=True)
            metadata["Metadata"]["asset-root-json"] = json_root_path
            metadata["Metadata"]["asset-root"] = json_root_path
```

A comment explaining why we need to populate both fields if the path is not ASCII would be helpful too (for backwards compatibility)

src/deadline/job_attachments/api/attachment.py
```diff
@@ -57,39 +57,6 @@ def _attachment_download(
         path_mapping_rules=path_mapping_rules
```
leongdl, line 57: nit - What does this model serve? From the conversation's readme, it doesn't give much context what / how this works.

src/deadline/job_attachments/models.py
leongdl: And this one returns a bunch of anonymous str to str dictionary of dictonaries. So how do we know what this is doing

src/deadline/job_attachments/api/attachment.py
```diff
@@ -57,39 +57,6 @@ def _attachment_download(
...
-        boto3_session,
-        file_name_manifest_dict,
-        s3_root_uri,
-        conflict_resolution,
-        path_mapping_rule_list,
-        logger,
-    )
-
-
-def _attachment_download_with_root_manifests(
```
leongdl, line 70: What does deleting these red lines do? It turned 2 functions into 1.

leongdl: What purpose did `_attachment_download_with_root_manifests` ? Missing some cntext.

godobyte: This is a cleanup and refactoring from existing code. The is not used anywhere and can be merged.

test/unit/deadline_job_attachments/test_models.py
godobyte: Sure I'll check the invalid value.

src/deadline/job_attachments/models.py
godobyte: I directly moved from original code for backward compatibility, and sure we can refactor for better readability.

src/deadline/job_attachments/api/attachment.py
godobyte: This part is not changed that's why it's not in readme. The function is processing the list of path mapping files to generate the PathMappingRule from file path.

jericht (approved):

src/deadline/job_attachments/api/attachment.py
leongdl: Yeah, thats the details we need in the PR description.

leongdl (approved):

PR #844, opened 2025-09-17
feat: Add '-p' param support to 'deadline bundle gui-submit'

mwiebe:

Sometimes, when you've got a job submission command like:

```
deadline bundle submit my_bundle -p MyParam=12
```

you want to start from there and instead use the GUI to edit more parameters instead of adding more `-p` options. This change adds support for that, so the following will work:

```
deadline bundle gui-submit my_bundle -p MyParam=12
```

Implement support for the `-p` / `--parameter` option to `deadline bundle gui-submit`.

You can pass parameter values to `deadline bundle gui-submit` to set the GUI controls for those parameters.

Manually tested different cases of job parameters vs queue parameters, and different parameter types. The GUI code doesn't have a testing framework to add unit tests in, so this change does not include unit tests.

The help output of the Deadline CLI includes this command.

No

No

No

----

### DIFF
--- a/src/deadline/client/cli/_groups/bundle_group.py
+++ b/src/deadline/client/cli/_groups/bundle_group.py
@@ -98,7 +98,11 @@ def _interactive_confirmation_prompt(message: str, default_response: bool) -> bo
     "--parameter",
     multiple=True,
     callback=validate_parameters,
-    help='The values for the job template\'s parameters. Can be provided as key-value pairs, inline JSON strings, or as paths to a JSON or YAML document. If provided more than once, the values are combined in the order that they appear. Examples: --parameter MyParam=5 -p file://parameter_file.json -p \'{"MyParam": "5"}\'',
+    help=(
+        "The values for the job template's parameters. Can be provided as key-value pairs, inline JSON strings, "
+        "or as paths to a JSON or YAML document. Later values for repeated parameter names take precedence. "
+        'Examples: --parameter MyParam=5 -p file://parameter_file.json -p \'{"OtherParam": "10"}\''
+    ),
 )
 @click.option("--profile", help="The AWS profile to use.")
 @click.option("--farm-id", help="The farm to use.")
@@ -285,6 +289,17 @@ def bundle_submit(
 
 
 @cli_bundle.command(name="gui-submit")
+@click.option(
+    "-p",
+    "--parameter",
+    multiple=True,
+    callback=validate_parameters,
+    help=(
+        "Initial values in the GUI for the job template's parameters. Can be provided as key-value pairs, inline JSON strings, "
+        "or as paths to a JSON or YAML document. Later values for repeated parameter names take precedence. "
+        'Examples: --parameter MyParam=5 -p file://parameter_file.json -p \'{"OtherParam": "10"}\''
+    ),
+)
 @click.argument("job_bundle_dir", required=False)
 @click.option(
     "--browse",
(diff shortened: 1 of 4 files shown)


src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -178,13 +252,41 @@ def on_create_job_bundle_callback(
...
+
+    job_parameters_dict = {param["name"]: param for param in (job_parameters or [])}
     for parameter in initial_settings.parameters:
+        # Overwrite the parameter values from the job bundle with values provided by job_parameters,
+        # e.g. from the CLI when this is called by the 'deadline bundle gui-submit' command.
+        if parameter["name"] in job_parameters_dict:
+            value = job_parameters_dict.pop(parameter["name"])["value"]
+            # Convert any path parameters to absolute
+            if parameter["type"] == "PATH":
+                value = os.path.abspath(value)
```
crowecawcaw, line 266: Would it make sense to move the parameter parsing logic into a common function?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -285,6 +285,13 @@ def _check_create_job_wait_canceled() -> bool:
 
 
 @cli_bundle.command(name="gui-submit")
+@click.option(
+    "-p",
+    "--parameter",
+    multiple=True,
+    callback=validate_parameters,
+    help='The values for the job template\'s parameters. Can be provided as key-value pairs, inline JSON strings, or as paths to a JSON or YAML document. If provided more than once, the values are combined in the order that they appear. Examples: --parameter MyParam=5 -p file://parameter_file.json -p \'{"MyParam": "5"}\'',
```
crowecawcaw: It'd be helpful to document how these parameters interact with the GUI. If a parameter is set, does it show up in the GUI? If a value is updated in the GUI, does the value in the command or the value in the GUI get used? I think this basically sets the default values in the GUI.

epmog: Yea, main thing to call out in the description here is that:
* they take precedence over the bundle's parameter_values,
* conflict resolution: `the values are combined in the order that they appear. `. The example doesn't actually show the result of specifying all of them. Does this mean the last value is used to populate?

src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -199,5 +301,22 @@ def on_create_job_bundle_callback(
...
+            if not _validate_and_warn_about_parameters(
+                job_parameters, initial_settings.parameters, queue_parameters, submitter_dialog
+            ):
+                # User chose to cancel, close the dialog
+                submitter_dialog.close()
+
+        # Connect to the queue parameters update signal
+        submitter_dialog.shared_job_settings._queue_parameters_update.connect(
+            validate_parameters_after_queue_load
+        )
```
epmog, line 312: How often does the queue parameter update run? Is this going to continually fire over and over again?

src/deadline/client/ui/job_bundle_submitter.py
mwiebe: Yes, I've created a new function to do this that does more thorough validation, and am using it here now. I kept the `abspath` part here, because that's doing a transformation of a path value that is context-sensitive so is better for the calling context to do.

src/deadline/client/cli/_groups/bundle_group.py
mwiebe: I've adjusted the wording for these. Have a look and let me know what you think!

src/deadline/client/ui/job_bundle_submitter.py
mwiebe: It re-runs this check if you choose 'Settings' and then update something there. E.g. if you switch queues, the validation could produce a different result because of differing queue parameter definitions.

crowecawcaw (dismissed):

epmog (approved):

crowecawcaw (approved):

PR #845, opened 2025-09-18
fix: Handled edge case for re-queued jobs handling sync output cli & some integ test tweaks

sakshie95:
1. We found an unhandled edge case with the integ tests for the sync-output a.k.a incremental download feature. In case of re-queued tasks sometimes the CLI ends up downloading the same output more than once due to the eventual consistency window if the tasks get executed and re-queued/re-executed within the same 2 min window. 
2. We found an issue with the test ``test_incremental_download_dep_data_flow`` where it looks in the wrong location for the checkpoint file and doesn't find it which results in output from the last task executed not being downloaded
3. Sometimes the integ tests also run out of memory

1. Fixed the edge case by adding the eventual consistency window to ``session_ended_threshold`` only for updated and completed jobs and not new jobs. New jobs get the eventual consistency window in accordance with the bootstrap lookback in minutes which will be None in case of re-queued jobs not intended to be downloaded again.
2. Fixed the location of the the checkpoint file in the test
3. Tried to limit the continously running loop until job completes to 10 minutes given that the test times out in 15 minutes. This might fix the memory leak but we'll keep monitoring. 

We also added the option ``--xfail-tb`` for traceback for xfail tests as we want those logs to debug while we have xfailed the tests.

Better customer experience!

No

No

----

### DIFF
--- a/test/integ/cli/test_cli_incremental_download.py
+++ b/test/integ/cli/test_cli_incremental_download.py
@@ -193,6 +193,9 @@ def test_incremental_download_many_small_files(incremental_download_test, tmp_pa
     # Run incremental download in a loop until job completes
     job_complete = False
     incremental_download_iteration_number = 0
+    start_time = time.time()
+    job_completion_timeout = 600  # 10 minutes
+
     while not job_complete:
         incremental_download_iteration_number += 1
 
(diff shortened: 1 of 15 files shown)


test/integ/cli/test_cli_incremental_download.py
```diff
@@ -217,6 +220,11 @@ def test_incremental_download_many_small_files(incremental_download_test, tmp_pa
         )
 
         job_complete = task_run_status in ["SUCCEEDED", "FAILED", "CANCELED"]
+
```
sakshie95, line 223: This is not a clean method so I want to ideally shift this to using a waiter - but I'm testing the memory leak hypothesis with this timeout handling & will shift this to a waiter in a follow-up.

sakshie95:
Hatch linting is impacted by bug https://github.com/pallets/click/issues/3065#issuecomment-3309243877

leongdl (approved):

jairaws (approved):

jairaws (commented):
There is a linting error that is making some of the auto validations fail.
```
TypeError: the JSON object must be str, bytes or bytearray, not Sentinel
```

mwiebe (approved):

PR #847, opened 2025-09-20
fix: Improved job submission performance by multithreading syscalls to stat() and…

baxeaz:
Job submission could encounter significant performance bottlenecking in stat calls when a bundle includes many files which were not cached by the OS.

Job submission encountered additional overhead by repeating this routine when calculating file sizes from manifest and when uploading objects to cas rather than using the already cached file sizes in the manifest.

verify_hash_cache_integrity method was not optimized

- Parallelize OS stat calls
- Calculate manifest input size using manifest
- Remove unnecessary stat calls from upload_object_to_cas
- Optimize verify_hash_cache_integrity

- Testing submitting a previously submitted Unreal Meerkat scene resulted in a ~48% submission speed improvement
- On my Windows test machine in Ec2 this saved ~36 seconds (76 -> 40.5)
- On my Windows laptop it saved about 26 seconds (~56 -> 30).

- ### How was this change tested?
Local testing and timing
Added unit tests, new unit tests pass
All integration tests pass

Yes

*   [X ] This PR does not add any new dependencies.

No

No

----

### DIFF
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -532,13 +532,6 @@ class S3AssetUploader:
         # Otherwise all hashes exist in S3
         return True
 
-    def _sample_cache_entries_with_limit(
-        self, cache_entries: List[S3CheckCacheEntry], limit: int = 30
-    ):
-        sampled_count: int = min(len(cache_entries), limit)
-        random.shuffle(cache_entries)
-        return cache_entries[:sampled_count]
-
     def verify_hash_cache_integrity(
         self,
         s3_check_cache_dir: Optional[str],
@@ -551,22 +544,22 @@ class S3AssetUploader:
         verifies if the cached assets exist in S3. Returns True if all sampled cached assets exist in S3, False
         otherwise.
         """
-
         # Find the list of s3 upload keys that have been cached
         s3_upload_keys: List[str] = [
             self._generate_s3_upload_key(file, manifest.hashAlg, s3_cas_prefix)
             for file in manifest.paths
         ]
-        with S3CheckCache(s3_check_cache_dir) as s3_cache:
-            cache_entries = [
-                s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}")
-                for upload_key in s3_upload_keys
-                if s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}") is not None
-            ]
 
-            # verify that a sample of the cached entries still exist in S3
-            sampled_cache_entries = self._sample_cache_entries_with_limit(cache_entries)
-            return self._check_hashes_exist_in_s3(sampled_cache_entries)
+        random.shuffle(s3_upload_keys)
+        sampled_cache_entries: List[S3CheckCacheEntry] = []
+        with S3CheckCache(s3_check_cache_dir) as s3_cache:
+            for upload_key in s3_upload_keys:
+                this_entry = s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}")
+                if this_entry is not None:
+                    sampled_cache_entries.append(this_entry)
+                    if len(sampled_cache_entries) >= 30:
+                        break
+        return self._check_hashes_exist_in_s3(sampled_cache_entries)
 
     def _separate_files_by_size(
         self,
@@ -617,13 +610,12 @@ class S3AssetUploader:
         local_path = source_root.joinpath(file.path)
         s3_upload_key = self._generate_s3_upload_key(file, hash_algorithm, s3_cas_prefix)
         is_uploaded = False
-        file_size = local_path.resolve().stat().st_size
 
         if s3_check_cache.get_entry(s3_key=f"{s3_bucket}/{s3_upload_key}"):
             logger.debug(
                 f"skipping {local_path} because {s3_bucket}/{s3_upload_key} exists in the cache"
             )
-            return (is_uploaded, file_size)
+            return (is_uploaded, file.size)
 
         if self.file_already_uploaded(s3_bucket, s3_upload_key):
             logger.debug(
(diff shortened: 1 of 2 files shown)


src/deadline/job_attachments/upload.py
```diff
@@ -551,22 +551,22 @@ def verify_hash_cache_integrity(
...
         ]
-        with S3CheckCache(s3_check_cache_dir) as s3_cache:
-            cache_entries = [
-                s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}")
-                for upload_key in s3_upload_keys
-                if s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}") is not None
-            ]
 
-            # verify that a sample of the cached entries still exist in S3
-            sampled_cache_entries = self._sample_cache_entries_with_limit(cache_entries)
```
jericht, line 568: Let's delete the `_sample_cache_entries_with_limit` function now that it's not used

src/deadline/job_attachments/upload.py
```diff
@@ -551,22 +551,22 @@ def verify_hash_cache_integrity(
...
-            return self._check_hashes_exist_in_s3(sampled_cache_entries)
+        random.shuffle(s3_upload_keys)
```
marofke, line 553: Double checked, looks like this uses [Fisher Yates](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) which is O(n), so that shouldn't impact perf too much 👍 

I like that we're randomizing the sample now, good stuff

src/deadline/job_attachments/upload.py
```diff
@@ -1292,15 +1291,17 @@ def _find_matched_root_from_local_type_locations(
...
-        total_bytes = 0
-        try:
-            for path in paths:
-                total_bytes += Path(path).resolve().stat().st_size
-        except FileNotFoundError:
-            logger.warning(
-                f"Skipping the input from total size calculation as it doesn't exist: {path}"
-            )
-        return total_bytes
+        def get_file_size(path_str: str) -> int:
```
marofke, line 1287: Just wanted to check if you did any manual verification of file sizes recorded in the manifest and the summary stats reporting (before and after your change with the same submission). It looks like it's all accurate based on unit tests, but I'm always paranoid about changes like this and like to double check that the numbers still look the same hahaha

src/deadline/job_attachments/upload.py
```diff
@@ -1292,15 +1291,17 @@ def _find_matched_root_from_local_type_locations(
...
+            try:
+                return Path(path_str).resolve().stat().st_size
+            except (FileNotFoundError, PermissionError, OSError):
+                logger.warning(f"Skipping file in size calculation: {path_str}")
```
jericht, line 1291: Should this maybe be `logging.error`? I'm assuming file size skipping calculation is OK here because if we get into any of these errors, we wouldn't be able to upload the file anyway, so the operation would fail eventually? Just want to avoid the case where we calculate a total size of files smaller than what we actually upload

jericht (commented):
Looks good! Just a couple nits

src/deadline/job_attachments/upload.py
baxeaz: Will do

src/deadline/job_attachments/upload.py
baxeaz: It all appears good from what I can tell - I did run all unit tests including the new ones, integ tests, and manual tests including successful renders and am still seeing 4.55 GB of assets reported for Meerkat both on the submitter and in download in the sync step.  Happy to do more checking if we think there could be gaps though!

src/deadline/job_attachments/upload.py
baxeaz: Hmm so this is the same logging level it had previously.  Would it actually fail later?  I'm not entirely sure there - are you saying there's a later JA step that would fail, or you're thinking the job would fail?  We don't really know if just because something is declared a n asset dependency that it's really a true "job can't run without it" dependency I think?

src/deadline/job_attachments/upload.py
marofke: I just meant doing some spot checks of the size fields of generated manifests before and after, since you changed the code around that area, if that's not too hard to check. From what I remember there should be a local copy of the manifests in your ~/.deadline directory

baxeaz: Generated a new and old manifest and they were identical except for 2 files which are expected to change between jobs

marofke: Awesome, thanks for checking!

marofke (approved):

src/deadline/job_attachments/upload.py
jericht: Oh didn't realize it was just moving the code around - we can leave it as is

jericht (approved):

src/deadline/job_attachments/upload.py
```diff
@@ -551,22 +544,22 @@ def verify_hash_cache_integrity(
...
-            sampled_cache_entries = self._sample_cache_entries_with_limit(cache_entries)
-            return self._check_hashes_exist_in_s3(sampled_cache_entries)
+        random.shuffle(s3_upload_keys)
+        sampled_cache_entries: List[S3CheckCacheEntry] = []
+        with S3CheckCache(s3_check_cache_dir) as s3_cache:
+            for upload_key in s3_upload_keys:
+                this_entry = s3_cache.get_entry(s3_key=f"{s3_bucket}/{upload_key}")
+                if this_entry is not None:
+                    sampled_cache_entries.append(this_entry)
+                    if len(sampled_cache_entries) >= 30:
```
leongdl, line 560: nit: Why 30?

leongdl: I see it came from the deleted function, just odd that we have a hard coded magic number but no explaination why, Maybe a comment?

PR #848, opened 2025-09-21
fix: Improved job submission performance by adding optional thread local cache db connections for use without requiring additional locks

baxeaz:
Our S3 cache was using a connection shared across threads which required a lock for all access, creating a bottleneck when attempting to divide up work across threads.

Add a thread local connection which can be used in cases where we can expect to benefit from reuse and lack of locking.  Use in upload_object_to_cas, verify_hash_cache_integrity, and _process_input_path.

In my testing on top of https://github.com/aws-deadline/deadline-cloud/pull/847 I was able to cut about an additional 15 seconds of my EC2 test machine time with Meerkat submissions, reducing the total time from about 40.5 seconds to around 25 seconds.  With the already uploaded Meerkat including a local cache, upload_input_files went from taking about 11 seconds to about 4 seconds.

Local testing

   Yes, existing unit tests pass, new unit tests added which pass
   Yes
   No

Yes

No

No

No

----

### DIFF
--- a/src/deadline/job_attachments/caches/cache_db.py
+++ b/src/deadline/job_attachments/caches/cache_db.py
@@ -6,6 +6,7 @@ Module for defining a local cache file.
 
 import logging
 import os
+import threading
 from abc import ABC
 from threading import Lock
 from typing import Optional
@@ -34,6 +35,8 @@ class CacheDB(ABC):
         self.cache_name: str = cache_name
         self.table_name: str = table_name
         self.create_query: str = create_query
+        self.local = threading.local()
+        self.local_connections: set = set()
 
         try:
             # SQLite is included in Python installers, but might not exist if building python from source.
@@ -81,8 +84,35 @@ class CacheDB(ABC):
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
         """Called when exiting the context manager."""
+
         if self.enabled:
+            import sqlite3
+
             self.db_connection.close()
+            for conn in self.local_connections:
+                try:
+                    conn.close()
+                except sqlite3.Error as e:
+                    logger.warning(f"SQLite connection failed to close with error {e}")
+
+            self.local_connections.clear()
+
+    def get_local_connection(self):
+        """Create and/or returns a thread local connection to the SQLite database."""
+        if not self.enabled:
+            return None
+        import sqlite3
+
+        if not hasattr(self.local, "connection"):
+            try:
+                self.local.connection = sqlite3.connect(self.cache_dir, check_same_thread=False)
+                self.local_connections.add(self.local.connection)
+            except sqlite3.OperationalError as oe:
+                raise JobAttachmentsError(
+                    f"Could not create connection to cache in {self.cache_dir}"
+                ) from oe
+
+        return self.local.connection
 
     @classmethod
     def get_default_cache_db_file_dir(cls) -> Optional[str]:
@@ -99,12 +129,23 @@ class CacheDB(ABC):
         """
         Removes the underlying cache contents from the file system.
         """
+
         if self.enabled:
+            import sqlite3
+
             self.db_connection.close()
+            conn_list = list(self.local_connections)
+            for conn in conn_list:
+                try:
+                    conn.close()
+                    self.local_connections.remove(conn)
+                except sqlite3.Error as e:
+                    logger.warning(f"SQLite connection failed to close with error {e}")
 
         logger.debug(f"The cache {self.cache_dir} will be removed")
         try:
             os.remove(self.cache_dir)
         except Exception as e:
             logger.error(f"Error occurred while removing the cache file {self.cache_dir}: {e}")
+
             raise e
--- a/src/deadline/job_attachments/caches/hash_cache.py
+++ b/src/deadline/job_attachments/caches/hash_cache.py
@@ -59,6 +59,32 @@ class HashCache(CacheDB):
             cache_dir=cache_dir,
         )
 
+    def get_connection_entry(
+        self, file_path_key: str, hash_algorithm: HashAlgorithm, connection
+    ) -> Optional[HashCacheEntry]:
+        """
+        Returns an entry from the hash cache, if it exists.
+        """
+        if not self.enabled:
+            return None
+
+        entry_vals = connection.execute(
+            f"SELECT * FROM {self.table_name} WHERE file_path=? AND hash_algorithm=?",
+            [
+                file_path_key.encode(encoding="utf-8", errors="surrogatepass"),
+                hash_algorithm.value,
+            ],
+        ).fetchone()
+        if entry_vals:
+            return HashCacheEntry(
+                file_path=str(entry_vals[0], encoding="utf-8", errors="surrogatepass"),
+                hash_algorithm=HashAlgorithm(entry_vals[1]),
+                file_hash=entry_vals[2],
+                last_modified_time=str(entry_vals[3]),
+            )
+        else:
+            return None
+
     def get_entry(
         self, file_path_key: str, hash_algorithm: HashAlgorithm
     ) -> Optional[HashCacheEntry]:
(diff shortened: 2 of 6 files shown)


src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -81,8 +84,33 @@ def __enter__(self):
...
+
         if self.enabled:
+            import sqlite3
+
             self.db_connection.close()
+            for conn in self.local_connections:
+                try:
+                    conn.close()
+                except sqlite3.Error:
+                    pass
```
jericht: Should we log a warning here? Is there a risk of building up orphaned sqlite connections?

Also lets remove the connection from `self.local_connections` after it is closed

src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -99,12 +127,23 @@ def remove_cache(self) -> None:
...
+            import sqlite3
+
             self.db_connection.close()
+            conn_list = list(self.local_connections)
+            for conn in conn_list:
+                try:
+                    conn.close()
+                    self.local_connections.remove(conn)
+                except sqlite3.Error:
+                    pass
```
jericht: Same here, should we at least log a warning in case we start accumulating orphaned connections?

src/deadline/job_attachments/caches/hash_cache.py
```diff
@@ -59,6 +59,32 @@ def __init__(self, cache_dir: Optional[str] = None) -> None:
             cache_dir=cache_dir,
         )
 
+    def get_connection_entry(
+        self, file_path_key: str, hash_algorithm: HashAlgorithm, connection
```
jericht, line 63:
```suggestion
    def get_connection_entry(
        self, file_path_key: str, hash_algorithm: HashAlgorithm, connection: sqlite3.Connection
```

src/deadline/job_attachments/caches/s3_check_cache.py
```diff
@@ -56,6 +56,29 @@ def __init__(self, cache_dir: Optional[str] = None) -> None:
             cache_dir=cache_dir,
         )
 
+    def get_connection_entry(self, s3_key: str, connection) -> Optional[S3CheckCacheEntry]:
```
jericht, line 59:
```suggestion
    def get_connection_entry(self, s3_key: str, connection: sqlite3.Connection) -> Optional[S3CheckCacheEntry]:
```

jericht: I guess this means we have to import `sqlite3` here just for typing, but if it's currently the only connection type we support I think that's fine. We can do it only for the type checking pass like:

```py
if typing.TYPE_CHECKING:
    import sqlite3
```

If we support multiple connection types, we'd probably have a common interface for connections we can use to type this.

test/unit/deadline_job_attachments/caches/test_caches.py
```diff
@@ -69,6 +70,54 @@ def test_init_throws_error_on_empty_strings(self, cache_name, table_name, create
         with pytest.raises(JobAttachmentsError):
             CacheDB(cache_name, table_name, create_query)
 
+    def test_get_local_connection_creates_new_connection(self, tmpdir):
+        """Tests that get_local_connection creates a new connection for each thread"""
```
jericht:
```suggestion
    def test_get_local_connection_same_thread(self, tmpdir):
        """Tests that get_local_connection returns the same connection for a single thread"""
```

src/deadline/job_attachments/caches/cache_db.py
baxeaz: Sure will add both

src/deadline/job_attachments/caches/cache_db.py
baxeaz: Sure, will add a warning

src/deadline/job_attachments/caches/s3_check_cache.py
baxeaz: Yeah that was the reason not to!  Sure will import it conditionally for type checking

baxeaz: Actually it's unit tests as well - probably not worth it for just the type hint in this case

jericht (dismissed):

jericht (dismissed):

src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -81,8 +84,35 @@
...
         if self.enabled:
+            import sqlite3
+
             self.db_connection.close()
+            for conn in self.local_connections:
+                try:
+                    conn.close()
+                except sqlite3.Error as e:
+                    logger.warning(f"SQLite connection failed to close with error {e}")
+                    pass
```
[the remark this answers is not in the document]
sakshie95: [1] I agree with CodeQL here :)) we don't need this pass do we?

src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -81,8 +84,35 @@ def __enter__(self):
...
+            self.local_connections.clear()
+
+    def get_local_connection(self):
+        """Create and/or returns a thread local connection to the SQLite database."""
+        if not self.enabled:
+            return None
+        import sqlite3
```
sakshie95, line 104: Is there a reason we're lazy loading sqlite3 here? Do we fear a circular dependency?

src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -99,12 +129,24 @@
...
+
             self.db_connection.close()
+            conn_list = list(self.local_connections)
+            for conn in conn_list:
+                try:
+                    conn.close()
+                    self.local_connections.remove(conn)
+                except sqlite3.Error as e:
+                    logger.warning(f"SQLite connection failed to close with error {e}")
+                    pass
```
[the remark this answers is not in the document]
sakshie95: Same as [1]

src/deadline/job_attachments/caches/hash_cache.py
```diff
@@ -59,6 +59,32 @@ def __init__(self, cache_dir: Optional[str] = None) -> None:
             cache_dir=cache_dir,
         )
 
+    def get_connection_entry(
```
sakshie95, line 62: Thanks for creating the unit tests. Were there any integ tests covering this functionality already? If not, how did you test this apart from unit testing to make sure there's no regressions?

src/deadline/job_attachments/caches/cache_db.py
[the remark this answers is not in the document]
baxeaz: We technically don't need it nope!  I've found without it or some other comment/indicator you'll sometimes get questions about "did you intend to just swallow this exception?", so just a way to communicate intent.  Happy to remove them though :)

src/deadline/job_attachments/caches/cache_db.py
baxeaz: This was an existing pattern I'm following - in [cache_db's __init__ it indicates](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/caches/cache_db.py#L39) it's to support source builds of Python

src/deadline/job_attachments/caches/cache_db.py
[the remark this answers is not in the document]
baxeaz: Removed!

src/deadline/job_attachments/caches/hash_cache.py
baxeaz: https://github.com/aws-deadline/deadline-cloud/blob/mainline/test/integ/deadline_job_attachments/test_job_attachments.py has a fair amount of tests which hit these changes yep!  I also did a fair bit of manual testing as I was validating the performance improvements

jericht (approved):

sakshie95 (approved):

PR #855, opened 2025-09-23
fix: Job submission warning dialogs are too long

### LINKED ISSUE #785: Bug: Job attachments upload summary is too large that it does not show the submit button on Mac
### Describe the bug

For scenes with numerous attachments, the job attachments upload summary dialog box was too long, causing the "Do you wish to proceed?" question and accompanying buttons to be hidden from view. This issue prevents users from submitting jobs.

Image

### Expected Behaviour

 
- The dialog box should be resizable and scroll-able to allow users to take appropriate action.

- The input directories should include line breaks after listing input directories for neater presentation.

### Current Behaviour

Right now the dialog box is not scroll-able nor is it resize-able. 

There appears to be a new feature that makes it way more easier to read and allowing prompt defaults to be automatically accepted. However, this feature is located at the bottom instead of the top of the dialog.

I believe customers who are in a rush would miss this feature, as I did. The following feedback is based solely on my experience.

- The "known asset paths" message is located at the very end of the dialog. Customers are likely to miss it, assuming it's the same prompt they've seen before.
- The message states: "To enable submission without user input, add directory locations containing the unknown paths to either: 1. The list of known asset paths in the local Deadline Cloud configuration." However, the location of this configuration isn't immediately clear, which could confuse customers. It would be helpful if known paths could be added directly through the dialog itself.
- The message in the dialog is massive. Reducing its length would improve the customer experience.

### Reproduction Steps

- Install Cinema 4D and our submitter on a Mac 
- Create a scene file with a lot of assets (the scene that I had 100s of attachments)
- Try submitting a scene. 

### Environment

At minimum:
1. Operating system: Mac
2. Output of `python3 --version`: Latest
3. Output of `deadline --version`: 0.51.1
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

AWS-Samuel:
Fixes: #785 

For scenes with numerous attachments, the job attachments upload summary dialog box was too long, causing the "Do you wish to proceed?" question and accompanying buttons to be hidden from view. This issue prevents users from submitting jobs.

It is not possible to implement scroll bars in a QMessageBox, so I needed to create a new QDialog class which I could add a scrollable text section to. The new dialog maintains the existing interface in `handle_request_warning_dialog()` and automatically resizes depending on the length of the content. 

If the length of the warnings passed to `handle_request_warning_dialog()`  is extremely long, the dialog will appropriately scale and provide a scrollable area if applicable. 

    - yes
   -  yes
  - no
  
  
  I tested this against two modes of this dialog which is 
  1) With provides default option of accepting which then presents a `Do not ask again` button and sets the default button to `OK`

        - <details>
          <summary>Original Dialog with Default</summary>
          original info
          </details>

        - <details>
          <summary>New Dialog with Default</summary>
          new info
          </details>

  2) Without a default option which only yields the `Cancel`/`OK` buttons and sets the default option to `Cancel`. This is the mode which can be extremely long.
 
        - <details>
          <summary>Original Dialog without Default</summary>
          warning original
          </details>

        - <details>
          <summary>New Dialog without Default</summary>
          warning new
          </details>

Yes through the commit title.

No

No

----

### DIFF
--- a/src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
+++ b/src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
@@ -12,21 +12,23 @@ from typing import Any, Optional
 from functools import partial
 import time
 
-from qtpy.QtCore import Qt, Signal
-from qtpy.QtGui import QCloseEvent
+from qtpy.QtCore import Qt, Signal, QSize
+from qtpy.QtGui import QCloseEvent, QFontMetrics
 from qtpy.QtWidgets import (  # pylint: disable=import-error; type: ignore
     QApplication,
     QDialog,
     QDialogButtonBox,
     QFormLayout,
     QGroupBox,
+    QHBoxLayout,
     QLabel,
-    QMessageBox,
     QPushButton,
     QProgressBar,
+    QStyle,
     QTextEdit,
     QVBoxLayout,
     QWidget,
+    QSizePolicy,
 )
 
 from .. import CancelationFlag
@@ -230,26 +232,20 @@ class SubmitJobProgressDialog(QDialog):
         self.button_box.rejected.connect(self.close)
 
     def handle_request_warning_dialog(self, message: str, default_response: bool):
+        """Presents a dialog with the provided message.
+
+        Args:
+            message (str): The message to present to the user.
+            default_response (bool):
+                True if the default is to continue. This adds a "Do not ask again" button as well.
+                False if the default is to Cancel.
+        """
         # Build the UI for user confirmation
-        message_box = QMessageBox(self)
-        if not default_response:
-            message_box.setIcon(QMessageBox.Warning)
+        dialog = _JobSumissionWarningDialog(message, default_response)
 
-        self.submission_log.append(f"{message}\n")
-        message_box.setText(message)
-        message_box.setStandardButtons(QMessageBox.Ok | QMessageBox.Cancel)
-        message_box.setDefaultButton(QMessageBox.Ok if default_response else QMessageBox.Cancel)
+        selection = dialog.exec()
 
-        if default_response:
-            # If the default response is to continue, add the "Do not ask again" button
-            dont_ask_button = QPushButton("Do not ask again", self)
-            dont_ask_button.clicked.connect(lambda: set_setting("settings.auto_accept", "true"))
-            message_box.addButton(dont_ask_button, QMessageBox.ActionRole)
-
-        message_box.setWindowTitle("Job attachments file upload confirmation")
-        selection = message_box.exec()
-
-        if selection == QMessageBox.Cancel:
+        if selection == QDialog.Rejected:
             self._warning_dialog_canceled = True
 
         self._warning_dialog_completed = True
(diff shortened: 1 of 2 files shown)


src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -361,3 +349,113 @@ def _build_ui(self):
 
         self.layout.addWidget(self.progress_bar)
         self.layout.addWidget(self.progress_message)
+
+
+class ScrollableWarningDialog(QDialog):
```
jusiskin: Two things:

1.   nit: maybe we can name this class `JobSumissionWarningDialog`

    While the message is supplied by the caller, it has the ability to change the auto-accept setting which affects future job submissions.
2.  As it stands currently, this will become part of the public API. Is that intended? Do we want consumers of the python library to use this?

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -361,3 +349,113 @@ def _build_ui(self):
...
+        text_rect = font_metrics.boundingRect(0, 0, text_width, 0, Qt.TextWordWrap, self.message)
+        text_height = text_rect.height()
+
+        # Add some padding for better appearance
+        padding = 20
+        optimal_height = text_height + padding
+
+        # Set reasonable bounds
+        min_height = 100
+        max_height = 500
```
jusiskin, line 442: Should this be based on the screen size?

Those with bigger displays may want to be able to read as much of the warning as possible without needing to scroll. I doubt anyone uses a screen with less that 500 vertical pixels these days, but...

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -361,3 +349,113 @@ def _build_ui(self):
...
         self.layout.addWidget(self.progress_bar)
         self.layout.addWidget(self.progress_message)
+
+
+class ScrollableWarningDialog(QDialog):
+    """
+    Simple Dialog which functions similar to a QMessageBox, but with a scrollable text area.
+    """
+
+    def __init__(self, message: str, default_response: bool = False):
```
jusiskin, line 367: nit: minor usability improvement to the caller IMO

```suggestion
    def __init__(self, message: str, default_button: QDialogButtonBox = QDialogButtonBox.Cancel):
```

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
AWS-Samuel: I think this would fit in a larger effort of dynamically scaling the UI based on screen size. Other dialogs in this package don't seem to do this so I would want this dialog to be consistent with that. For example, the `SubmitJobToDeadline` dialog defaults to `540, 700` on all screen sizes. For those with larger screens that want to read more at once, this dialog is resizeable, but with the current bounds we can more or less guarantee the bottom of the dialog will be visible on the vast majority of displays.

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
AWS-Samuel: I had considered modifying the interface for the dialog but opted to keep it as is to align with the only caller which is a function that has the following interface

`def handle_request_warning_dialog(self, message: str, default_response: bool):`

In this case I think it makes sense to preserve the existing interface/previous behavior

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
AWS-Samuel: Renamed, added a prefixed `_` so that this is not part of the public API, since the public API should go through  the `def handle_request_warning_dialog(self, message: str, default_response: bool):` callback

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -361,3 +357,123 @@ def _build_ui(self):
...
+                True if the default response should be to continue.
+                    - This also adds a "Do not ask again" button which will set settings.auto_accept to True.
+                False if the default response should be to Cancel.
+        """
+        super().__init__()
+        self.setWindowTitle("Job Submission Confirmation")
+        self.message = message
+        self.default_response = default_response
+        self.buttons = None
+        layout = QVBoxLayout(self)
```
mwiebe, line 383: I found that when I make the dialog larger, it stretches the space holding the triangle warning "Job submission confirmation" part. Probably the right place to add stretching is between the box containing the text, and the buttons at the bottom.

Image

mwiebe: Also, in this case the text is still larger than the vertical text box. I've got plenty of vertical space available, and it would be better for it to fill how big I made the dialog.

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -361,3 +357,123 @@ def _build_ui(self):
...
+        # Set the default button after the dialog is fully shown
+        if self.buttons is not None:  # This type check is to make linting pass
+            if self.default_response:
+                ok_button = self.buttons.button(QDialogButtonBox.Ok)
+                if ok_button:
+                    ok_button.setDefault(True)
+                    ok_button.setAutoDefault(True)
+                    ok_button.setFocus()
+            else:
+                cancel_button = self.buttons.button(QDialogButtonBox.Cancel)
```
mwiebe, line 465: I found when I select the cancel button, it submits the job instead of canceling.

mwiebe (changes requested):

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
AWS-Samuel: Fixed this in the latest.

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
AWS-Samuel: Fixed this in the latest.

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
AWS-Samuel: Here's an updated, fully expanded screenshot to demonstrate

[](url)
new

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
jusiskin: Sounds good. This is easy to improve later anyways.

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
jusiskin: Got it. It is less of a concern given it has been changed to a private API. Thanks

mwiebe (approved):

jusiskin (approved):

PR #858, opened 2025-09-24
feat: download outputs from chunked session actions

ryyakobe:

Job attachments need to support downloading outputs from chunked session actions in addition to the existing task-based output downloads. 

Some session actions no longer have `taskId` because they represent a "chunk" of tasks (e.g., frame range 0-20 corresponding to 20 tasks). Previously, WorkerAgent stored manifests at S3 path `.../job_id/step_id/task_id/timestamp_sessionaction_id/` but outputs for chunked session actions are stored at `.../job_id/step_id/timestamp_sessionaction_id/` (omitting task_id).

The existing download functionality only handles task-based outputs with the pattern `task-*/*_output`, but chunked session actions generate outputs with different S3 key patterns that aren't being captured.

- **Implemented chunked manifest handling**: Added logic to detect and process chunked session action outputs alongside traditional task outputs in `_get_tasks_manifests_keys_from_s3()` to handle both task-based (`step_id/task_id`) and chunked (`step_id/session_action_id`) output patterns. Also,  updated `get_output_manifests_by_asset_root()` to chronologically merge manifests from chunked sources.
- **Added session action ID support**: Updated the CLI to extract and pass `latestSessionActionId` from task metadata to the download functions and added session action-specific retrieval with `_get_manifests_by_session_action_id()` and `_get_manifests_by_session_action_regex()` functions.

- Users can now download outputs from chunked session actions using the existing CLI commands
- WorkerAgent can sync step-step dependencies for both task-based and chunked outputs
- The system maintains backward compatibility with existing task-based downloads

* [x] Have you run the unit tests?
* [x] Have you run the integration tests?
```
49 passed, 4 skipped, 2 xfailed, 4 xpassed in 3528.71s (0:58:48)
```

* [x] Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended
```
32 passed, 13 skipped in 19.15s
```
* [x] Manually tested downloading of job/step/task outputs from `dep_chain_ja_output_test`. Also, used the same job outputs but moved manifests from `task-` folders directly into `step-` folders and confirmed that handling "chunked" paths works as expected. Got the correct output in both cases.
```
Step A is correct
Step B is correct
Step C is correct
Step D is correct
Step E is correct
Step F is correct
Step G is correct
Step H is correct
Step I is correct
Step J is correct
Step K is correct
Step L is correct
Step M is correct
Step N is correct
Step O is correct
Step P is correct
Step Q is correct
Step R is correct
Step S is correct
Step T is correct
Step U is correct
Step V is correct
Step W is correct
Step X is correct
Step Y is correct
Step Z is correct
```

* [x] Ran a small benchmarking test to compare `re.compile` vs `re.search`
```
Summary over 10 runs with 30,000 paths:

• **re.search**: avg=0.0185s, min=0.0177s, max=0.0225s
• **compiled**: avg=0.0102s, min=0.0097s, max=0.0110s  
• **Improvement: 45.1%**

The results are consistent across 10 runs, showing the compiled regex is reliably ~45% faster. The performance is stable with minimal variance.
```

* [x] Are relevant docstrings in the code base updated?

No, this is not a breaking change. The modifications extend existing functionality while maintaining backward compatibility with existing task-based downloads.

No security impact. The change only modifies how S3 manifest keys are discovered and processed, using the same existing S3 permissions and download mechanisms.

----

### DIFF
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -172,20 +172,11 @@ def _get_output_manifest_prefix(
     job_id: str,
     step_id: Optional[str] = None,
     task_id: Optional[str] = None,
-    session_action_id: Optional[str] = None,
 ) -> str:
     """
     Get full prefix for output manifest with given farm id, queue id, job id, step id and task id
     """
     manifest_prefix: str
-    if session_action_id:
-        if not task_id or not step_id:
-            raise JobAttachmentsError(
-                "Session Action ID specified, but no Task ID or Step ID. Job, Step, and Task ID are required to retrieve task outputs."
-            )
-        manifest_prefix = s3_settings.full_output_prefix(
-            farm_id, queue_id, job_id, step_id, task_id, session_action_id
-        )
     if task_id:
         if not step_id:
             raise JobAttachmentsError(
@@ -203,19 +194,18 @@ def _get_output_manifest_prefix(
     return f"{manifest_prefix}/"
 
 
-def _get_tasks_manifests_keys_from_s3(
-    manifest_prefix: str,
+def _list_s3_objects_with_error_handling(
     s3_bucket: str,
+    manifest_prefix: str,
     session: Optional[boto3.Session] = None,
-    *,
-    select_latest_per_task=True,
-) -> List[str]:
+) -> List[dict]:
     """
-    Returns the keys of all output manifests from the given s3 prefix.
-    (Only the manifests that end with the prefix pattern task-*/*_output)
+    List S3 objects with standardized error handling for job attachments.
+    Returns a list of all S3 object contents under the given prefix.
     """
-    manifests_keys: List[str] = []
     s3_client = get_s3_client(session=session)
+    all_contents = []
+
     try:
         paginator = s3_client.get_paginator("list_objects_v2")
         page_iterator = paginator.paginate(
@@ -223,21 +213,15 @@ def _get_tasks_manifests_keys_from_s3(
             Prefix=manifest_prefix,
         )
 
-        # 1. Find all files that match the pattern: task-{any}/{any}/{any}output{any}
-        task_prefixes = defaultdict(list)
         for page in page_iterator:
             contents = page.get("Contents", None)
             if contents is None:
                 raise JobAttachmentsError(
                     f"Unable to find asset manifest in s3://{s3_bucket}/{manifest_prefix}"
                 )
-            for content in contents:
-                if re.search(r"task-.*/.*/.*output.*", content["Key"]):
-                    parts = content["Key"].split("/")
-                    for i, part in enumerate(parts):
-                        if "task-" in part:
-                            task_folder = "/".join(parts[: i + 1])
-                            task_prefixes[task_folder].append(content["Key"])
+            all_contents.extend(contents)
+
+        return all_contents
 
     except ClientError as exc:
         status_code = int(exc.response["ResponseMetadata"]["HTTPStatusCode"])
@@ -266,8 +250,52 @@ def _get_tasks_manifests_keys_from_s3(
     except Exception as e:
         raise AssetSyncError(e) from e
 
+
+def _get_tasks_manifests_keys_from_s3(
+    manifest_prefix: str,
+    s3_bucket: str,
+    session: Optional[boto3.Session] = None,
+    *,
+    select_latest_per_task=True,
+) -> List[str]:
+    """
+    Retrieves manifest keys from S3 by listing objects in a given S3 prefix, usually job or step folders.
+
+    Searches for manifest files matching the regex pattern that captures both:
+    - Chunked step manifests: .../job_id/step_id/timestamp_sessionaction_id/
+    - Task-based manifests: .../job_id/step_id/task_id/timestamp_sessionaction_id/
+
+    For chunked steps (no task ID), all manifests are included.
+    For task-based manifests, behavior depends on select_latest_per_task:
+    - If True: selects only the latest session action per task (by timestamp_sessionaction_id alphabetical order)
+    - If False: includes all manifests
+    """
+    manifests_keys = []
+
+    # Separate tracking for task-based vs chunked manifests
+    task_prefixes: dict[str, list] = defaultdict(list)
+
+    all_contents = _list_s3_objects_with_error_handling(s3_bucket, manifest_prefix, session)
+
+    step_pattern = re.compile(r"step-.*/.*/.*output.*")
+    for content in all_contents:
+        key = content["Key"]
+        if step_pattern.search(key):
+            if "task-" in key:
+                parts = key.split("/")
+                for i, part in enumerate(parts):
+                    if "task-" in part:
+                        task_folder = "/".join(parts[: i + 1])
+                        task_prefixes[task_folder].append(key)
+                        break
+            else:
+                manifests_keys.append(key)
+
+    # Handle task-based steps
     if select_latest_per_task:
-        # 2. Select all files in the last subfolder (alphabetically) under each "task-{any}" folder.
+        # TODO: Select all files in the last subfolder (alphabetically) under each "task-{any}" folder.
+        # This sorts by timestamp_sessionaction_id, but timestamp comes from WorkerAgent and shouldn't be relied on.
+        # Should use S3 LastModified instead.
         for task_folder, files in task_prefixes.items():
             last_subfolder = sorted(
                 set(f.split("/")[len(task_folder.split("/"))] for f in files), reverse=True
@@ -275,9 +303,8 @@ def _get_tasks_manifests_keys_from_s3(
             manifests_keys += [f for f in files if f.startswith(f"{task_folder}/{last_subfolder}/")]
     else:
         # Include all the keys, not just the latest per task
-        manifests_keys = [f for _, files in task_prefixes.items() for f in files]
+        manifests_keys += [f for _, files in task_prefixes.items() for f in files]
 
-    # Now `manifests_keys` is a list of the keys of files in the last folder (alphabetically) under each "task-" folder.
     return manifests_keys
 
 
@@ -738,12 +765,29 @@ def get_output_manifests_by_asset_root(
     session: Optional[boto3.Session] = None,
 ) -> dict[str, list[BaseAssetManifest]]:
     """
-    For a given job/step/task, gets a map from each root path to a corresponding list of
-    output manifests.
+    Gets output manifests grouped by asset root for job, step, or task outputs, handling both chunked and non-chunked steps.
+
+    When session_action_id is provided, retrieves outputs for that specific session action
+    by searching S3 paths containing the session action ID. Session action ID is typically
+    provided with task ID but is not required.
+
+    When session_action_id is not provided, retrieves all output manifests for the specified
+    scope (job, step, or task) and merges them chronologically by asset root. This is used
+    for downloading complete job/step/task outputs or syncing step dependencies by WorkerAgent.
     """
+    # Handle specific session action ID requests
+    if session_action_id:
+        if not step_id or not task_id:
+            raise JobAttachmentsError(
+                "Session Action ID specified, but missing Step ID or Task ID. Job, Step, and Task ID are required to retrieve session action outputs."
+            )
+        return _get_manifests_by_session_action_id(
+            s3_settings, farm_id, queue_id, job_id, step_id, task_id, session_action_id, session
+        )
+
     outputs: DefaultDict[str, list[BaseAssetManifest]] = DefaultDict(list)
     manifest_prefix: str = _get_output_manifest_prefix(
-        s3_settings, farm_id, queue_id, job_id, step_id, task_id, session_action_id
+        s3_settings, farm_id, queue_id, job_id, step_id, task_id
     )
     try:
         manifests_keys: list[str] = _get_tasks_manifests_keys_from_s3(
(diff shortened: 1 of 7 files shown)


src/deadline/job_attachments/download.py
```diff
@@ -172,20 +172,11 @@ def _get_output_manifest_prefix(
     job_id: str,
     step_id: Optional[str] = None,
     task_id: Optional[str] = None,
-    session_action_id: Optional[str] = None,
```
ryyakobe, line 175: This has never been used. Better remove it, because session actions can be located either in `.../stepid/taskid/timestamp_sessionactionid` or in `.../stepid/timestamp_sessionactionid` for chunked steps.

Besides, the actual path includes `timestamp_` prefix before the session action id, so this function was incorrect.

src/deadline/job_attachments/download.py
```diff
@@ -738,34 +766,59 @@ def get_output_manifests_by_asset_root(
...
-    For a given job/step/task, gets a map from each root path to a corresponding list of
-    output manifests.
+    Gets output manifests grouped by asset root, handling both chunked and non-chunked steps.
     """
-    outputs: DefaultDict[str, list[BaseAssetManifest]] = DefaultDict(list)
-    manifest_prefix: str = _get_output_manifest_prefix(
-        s3_settings, farm_id, queue_id, job_id, step_id, task_id, session_action_id
-    )
+    # Handle specific session action ID requests
+    if session_action_id:
```
ryyakobe, line 779: When session action id is known, we don't need to list all files inside of the task folder and search for the "latest" session action based on the `timestamp_` prefix, which comes from the WorkerAgent and can't be trusted. Instead, we can get the actual latest session action id from get_task API and get the outputs for this session action id directly.

Unfortunately, we still can't "just get" the outputs, even if we know session action id, because the path contains `timestamp_` prefix, which we don't know. So we have to list objects in S3 that include the specific session action id in path.

For task-based paths we search inside of the task folder and for chunked steps we search inside of the step folder. We could just search inside of the step folder in both cases but for task-based paths this would be inefficient, especially for steps with many tasks. In fact, that would be a regression in performance, because before currently we search in task folder.

src/deadline/job_attachments/download.py
```diff
@@ -999,7 +1052,9 @@ def _merge_asset_manifests_sorted_asc_by_last_modified(
         return None
 
     # Sort manifests by timestamp (oldest first)
-    sorted_manifests_with_timestamps = sorted(manifests_with_last_modified_timestamps)
```
ryyakobe, line 1002: Bugfix: if timestamps are the same, we'd try to compare `BaseAssetManifest` but this type doesn't support comparison.

src/deadline/job_attachments/models.py
```diff
@@ -330,6 +330,30 @@ def partial_session_action_manifest_prefix(
             f"{_float_to_iso_datetime_string(time)}_{session_action_id}",
         )
 
+    @staticmethod
+    def partial_session_action_manifest_prefix_without_task(
```
ryyakobe, line 377: WorkerAgent would need this to store chunked session actions in a path without task id.

src/deadline/job_attachments/download.py
mwiebe: Also, this is a parameter to an internal-only function prefixed with `_`, not part of the library's public interface.

src/deadline/job_attachments/download.py
```diff
@@ -1313,3 +1357,133 @@ def download_job_output(
...
+    s3_client = get_s3_client(session=session)
+
+    try:
+        paginator = s3_client.get_paginator("list_objects_v2")
+        page_iterator = paginator.paginate(
+            Bucket=s3_bucket,
+            Prefix=manifest_prefix,
+        )
+
+        regex_pattern = rf".*{re.escape(session_action_id)}.*output.*"
```
mwiebe: You're already doing pre-processing, best to use https://docs.python.org/3/library/re.html#re.compile and avoid re-compiling the regex repeatedly for every `re.search` call.

mwiebe (dismissed):

src/deadline/job_attachments/download.py
```diff
@@ -266,18 +250,61 @@ def _get_tasks_manifests_keys_from_s3(
     except Exception as e:
         raise AssetSyncError(e) from e
 
+
+def _get_tasks_manifests_keys_from_s3(
+    manifest_prefix: str,
+    s3_bucket: str,
+    session: Optional[boto3.Session] = None,
+    *,
```
leongdl, line 258: Can we explain what variable args are supported?

src/deadline/job_attachments/download.py
```diff
@@ -266,18 +250,61 @@ def _get_tasks_manifests_keys_from_s3(
...
+                        task_folder = "/".join(parts[: i + 1])
+                        task_prefixes[task_folder].append(key)
+                        break
+            else:
+                manifests_keys.append(key)
+
+    # Handle task-based steps
     if select_latest_per_task:
-        # 2. Select all files in the last subfolder (alphabetically) under each "task-{any}" folder.
+        # TODO: Select all files in the last subfolder (alphabetically) under each "task-{any}" folder.
```
leongdl, line 296: This todo is quite important - especially for re-queue cases of a task?

src/deadline/job_attachments/download.py
```diff
@@ -266,18 +250,61 @@ def _get_tasks_manifests_keys_from_s3(
...
+        # Should use S3 LastModified instead.
         for task_folder, files in task_prefixes.items():
             last_subfolder = sorted(
                 set(f.split("/")[len(task_folder.split("/"))] for f in files), reverse=True
             )[0]
             manifests_keys += [f for f in files if f.startswith(f"{task_folder}/{last_subfolder}/")]
     else:
         # Include all the keys, not just the latest per task
-        manifests_keys = [f for _, files in task_prefixes.items() for f in files]
+        manifests_keys += [f for _, files in task_prefixes.items() for f in files]
```
leongdl, line 306: Just something to look out for from other reviewers, we still need to inspect the manifests themselves right?

EG: if the first chunk rendered [1...10], while a requeue chunk ran [5..7] we need to pick the correct requeued outputs.

src/deadline/job_attachments/download.py
```diff
@@ -752,20 +796,39 @@ def get_output_manifests_by_asset_root(
...
+                if not asset_root:
+                    raise MissingAssetRootError(
+                        f"Failed to get asset root from metadata of output manifest: {manifests_keys[i]}"
+                    )
+                by_root[asset_root].append((last_modified, manifest))
+
+    # Merge each asset root chronologically
+    # We must merge here while we have LastModified timestamps, since the returned manifests
+    # lose this metadata and downstream callers can't merge chronologically.
+    for asset_root, manifest_list in by_root.items():
```
leongdl, line 828: This is a really key change here, I want to make sure we have great unit test coverage for the business logic.

src/deadline/job_attachments/download.py
```diff
@@ -1313,3 +1378,95 @@ def download_job_output(
...
+    session: Optional[boto3.Session] = None,
+) -> List[str]:
+    """
+    Get manifest keys matching a specific session action ID using regex search.
+
+    Lists all objects under the given prefix and uses regex to find paths containing
+    the session action ID. This is necessary because the exact timestamp prefix in
+    the path is unknown.
+    """
+    all_contents = _list_s3_objects_with_error_handling(s3_bucket, manifest_prefix, session)
```
leongdl: Overtime, we may end up with code that calls this list function under the hood.

Do we want to consider a LRU_CACHE, or make this a dependency injected argument to reverse this as an input instead? I see very similar code above for the task filtering version.

src/deadline/job_attachments/models.py
leongdl, line 377: nit, should this function be `_`? Its internal still for us, and worker agent.

src/deadline/job_attachments/download.py
leongdl: Just browsing the change here, can we please have more coverage?

src/deadline/job_attachments/download.py
leongdl: Is the * even needed?

src/deadline/job_attachments/download.py
leongdl: Did we implement this manifest level merge somewhere to merge the contents of the manifest(s) for the session case? I'm trying to find the logic and corresponding test case.

src/deadline/job_attachments/download.py
ryyakobe: In this PR I tried to keep the changes as small as possible. This is an existing function and I didn't chnage the signature. However, I added a docstring, just below, to explain what this flag does:
```
For task-based manifests, behavior depends on select_latest_per_task:
  - If True: selects only the latest session action per task (by timestamp_sessionaction_id alphabetical order)
```

src/deadline/job_attachments/download.py
ryyakobe: Thank you for highlighting this! You're absolutely right about its importance.

This TODO identifies an existing bug in the current download functionality for job attachments (the bug is not introduced in this PR) and I've added it specifically to increase visibility of the issue.

I'm planning to create a dedicated GitHub issue to track and address this bug properly. Since this particular code path isn't related to the chunked steps implementation, I suggest handling it separately from this PR to maintain focus and scope.

src/deadline/job_attachments/download.py
ryyakobe: This particular code path is specifically for task-based paths, not for chunked executions. The manifest handling for chunks actually happens earlier in the code flow:

- For task-based paths: keys are collected in task_prefixes
- For chunked paths: keys go directly to manifests_keys via the `else` branch:

```python
if "task-" in key:
    task_prefixes[task_folder].append(key)
else:
    manifests_keys.append(key)
```

Since this PR focuses on adding new chunking functionality, I kept the existing task-based logic unchanged.

The manifest merging logic for chunked executions that you're asking about is handled separately later.

src/deadline/job_attachments/download.py
ryyakobe: Thank you for thinking about optimization opportunities. While we could implement caching or dependency injection here, I believe we should take a different strategic approach:

Instead of optimizing the current S3 listing approach, we should move towards using the session action APIs (get/list) to obtain manifests directly. This would be both more efficient and more reliable than S3 listing.

I explored this better approach during this PR's development, but implementing it now would require significant changes to the download-output command and related code, which currently relies heavily on asset_root mappings and interactions with users, and doesn't support automatic path-mapping. After discussion with Mark, we decided to:

1. Make minimal changes required for chunking support now
2. Plan a separate effort to modernize the entire asset handling approach

This way, we can unblock chunking functionality while preparing for a more comprehensive improvement later.

src/deadline/job_attachments/models.py
ryyakobe: I like keeping this as a public function without the underscore prefix for a few reasons:

1. Consistency: The related function (above) `partial_session_action_manifest_prefix` follows the same public naming pattern

2. Library Design: deadline-cloud is designed as a proper library that's consumed by deadline-cloud-worker-agent (among others), so these functions are part of our public API contract

3. Developer Experience: Making it public clearly signals that this is an actively used API that external clients (like WorkerAgent) depend on, preventing accidental removal during future refactoring (especially since it's not used in deadline-cloud codebase)

src/deadline/job_attachments/download.py
ryyakobe: Absolutely, I agree this is a critical change that deserves thorough testing. While I've already added some coverage with tests like `test_get_tasks_manifests_keys_chunked_and_task_based_with_latest` (which covers the interaction between older/newer manifests for task-based paths), I see the gap in testing the chunked steps scenarios. I'll add those tests in the next revision.

src/deadline/job_attachments/download.py
```diff
@@ -1313,3 +1378,95 @@ def download_job_output(
...
+def _get_manifests_by_session_action_id(
+    s3_settings: JobAttachmentS3Settings,
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    step_id: str,
+    task_id: str,
+    session_action_id: str,
+    session: Optional[boto3.Session],
+) -> dict[str, list[BaseAssetManifest]]:
```
godobyte, line 1392: nit - The asset root to list of manifests dict is widely used, we may want to consider a dedicated data model class for that.

src/deadline/job_attachments/download.py
```diff
@@ -999,7 +1062,9 @@ def _merge_asset_manifests_sorted_asc_by_last_modified(
         return None
 
     # Sort manifests by timestamp (oldest first)
-    sorted_manifests_with_timestamps = sorted(manifests_with_last_modified_timestamps)
+    sorted_manifests_with_timestamps = sorted(
+        manifests_with_last_modified_timestamps, key=lambda x: x[0]
+    )
```
godobyte, line 1067: Could we add test for this?

src/deadline/job_attachments/download.py
```diff
@@ -1313,3 +1378,95 @@ def download_job_output(
...
+            task_prefix, s3_settings.s3BucketName, session_action_id, session
+        )
+    except JobAttachmentsError:
+        pass  # Unable to find manifests under task prefix.
+
+    # If no manifests found at task level, fall back to step level
+    if not manifests_keys:
+        step_prefix = _get_output_manifest_prefix(s3_settings, farm_id, queue_id, job_id, step_id)
+        try:
+            manifests_keys = _get_manifests_by_session_action_regex(
```
godobyte: I noticed this `_get_manifests_by_session_action_regex` is only few lines and called from this function only, could we move the function definition to within this function? That'd help with readability and maintainability.

src/deadline/job_attachments/download.py
godobyte: If we return all the session actions' manifest keys, would next step running `_get_asset_root_and_manifest_from_s3_with_last_modified` futures in `get_output_manifests_by_asset_root` return the manifest with last modified?

ryyakobe: Yes, it would. And maybe that's what we should do (but not in scope of this PR):
1. get manifests for all the session actions within the task
2. get lastModified for all the manifests
3. merge all those manifests by last modified

Something to consider:
* right now we only get manifests for one session action per task. If task was re-queued 100 times, we'll get 100x more manifests per task. There might be performance regression with this new approach, even though it should be more correct.
* manifests are in customer's S3 bucket. lastModified changes if customers manually move the manifest to a different location in S3, even if the content of the manifest remains the same. I don't think we expect customers to manually move the files, so I wouldn't worry about that right now, but it shows that lastModified is not 100% ideal solution. I think the better approach would be to get the aggregated manifest for the whole job or step through the API and those manifests would be aggregated based on the outputs of the latest session actions per task.

src/deadline/job_attachments/download.py
ryyakobe: I could add a specific test for this but I think it's already tested indirectly by some of the tests in this PR. I caught the issue because the test was using 2 items with the same timestamps and failed while attempting to compare BaseAssetManifest next.

src/deadline/job_attachments/download.py
ryyakobe: Indeed it's widely used! I could add that type (I think simple dict does the trick here), but we actually want to move away from asset_root-based mapping everywhere, because it doesn't always make sense. So I'm not sure if we should add a new data type for something we want to avoid and potentially make it even more widely used.

src/deadline/job_attachments/download.py
ryyakobe: Function definition within other function makes the outer function much longer and decreases readability to me, so I think it's a matter of a personal taste. I understand the scoping benefit, so if inner functions are considered a better practice in python, I don't mind moving it inside.

src/deadline/job_attachments/download.py
leongdl: ok thanks -

src/deadline/job_attachments/download.py
leongdl: Yeah, anything that is not service managed can have odd failure modes if the customer changed something.

That being said, I think we (JA team) should fix this bug ASAP, since we have a forked behavior between the download CLI, worker agent and regular job download.

src/deadline/job_attachments/download.py
leongdl: Got it.

src/deadline/job_attachments/download.py
leongdl: Thanks!! Super appreciate it.

ryyakobe: Added `test_get_output_manifests_by_asset_root_chronological_merge_chunked`, `test_get_output_manifests_by_asset_root_multiple_asset_roots` and a few other unit tests to cover more changes from this PR.

src/deadline/job_attachments/download.py
ryyakobe: Added `test_merge_asset_manifests_sorted_same_timestamp` specifically for this change.

src/deadline/job_attachments/download.py
ryyakobe: Moved `def get_manifests_by_regex` inside of `_get_manifests_by_session_action_id` as you suggested.

src/deadline/job_attachments/download.py
ryyakobe: Switched to `re.compile`.

src/deadline/job_attachments/download.py
leongdl: thank you

leongdl (approved):

src/deadline/job_attachments/download.py
godobyte: TY!

godobyte (approved):

mwiebe (approved):

Issue #849, opened 2025-09-22
Maintenance: Unbounded lru_cache


joel-wong-aws:
### Description

There are unbounded `lru_cache`s in `deadline-cloud`. Unbounded caches can lead to high memory utilization and crashes.

https://github.com/aws-deadline/deadline-cloud/blob/a70f0b7adeba4434247f2de29cc2dc3453322f27/src/deadline/client/api/_session.py#L128

https://github.com/aws-deadline/deadline-cloud/blob/a70f0b7adeba4434247f2de29cc2dc3453322f27/src/deadline/client/api/_telemetry.py#L335

### Solution

Set a maximum size for all `lru_cache`s

ttblanchard:
the `lru_cache` decorator has a [default max size of 128](https://docs.python.org/3/library/functools.html#functools.lru_cache). 

We could likely tune this down if significant memory usage from these functions becomes an issue.

joel-wong-aws:
Thanks @ttblanchard!

The default max size is reasonable. I reviewed the repo code and confirmed that all `lru_cache`s use the default (or a fixed) max size. Resolving!

PR #860, opened 2025-09-26
perf: speed up job bundle submission by reducing redundant stat calls

baxeaz:
When creating and uploading a job bundle performance is reduced by repeated stat calls to the same files

Add a stat cache of previously retrieved stat results

Faster job bundle creation.  In my own testing with Meerkat on top of the previously committed improvements I saw another ~20-28% improvement - reducing the creation time to 16-16.5 seconds.

New unit tests added which pass
Integration tests pass
Manual testing

Yes

This library is designed to be integrated into third-party applications that have bespoke and customized deployment environments. Adding dependencies will increase the chance of library version conflicts and incompatibilities. Please evaluate the addition of new dependencies. See the [Dependencies](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies) section of DEVELOPMENT.md for more details.

*   [ X] This PR does not add any new dependencies.

No

No

----

### DIFF
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -12,9 +12,11 @@ from contextlib import contextmanager
 import errno
 import logging
 import os
+import stat
 import sys
 import time
 from datetime import datetime
+from functools import lru_cache
 from io import BufferedReader, BytesIO
 from math import trunc
 from pathlib import Path, PurePath
@@ -81,6 +83,44 @@ from ._utils import (
 )
 
 logger = logging.getLogger("deadline.job_attachments.upload")
+
+
+class _FileStatCache:
+    """Private cache for file stat results to avoid redundant filesystem calls"""
+
+    @lru_cache(maxsize=1024)
+    def _get_stat(self, path_str: str) -> Optional[os.stat_result]:
+        """Get cached stat result for a path string"""
+        try:
+            return Path(path_str).stat()
+        except (FileNotFoundError, PermissionError, OSError):
+            return None
+
+    def exists(self, path: Path) -> bool:
+        """Check if path exists, using cache when possible"""
+        stat_result = self._get_stat(str(path))
+        if stat_result is not None:
+            return True
+        # Fall back to direct exists() call if stat failed
+        return path.exists()
+
+    def is_dir(self, path: Path) -> bool:
+        """Check if path is directory, using cache when possible"""
+        stat_result = self._get_stat(str(path))
+        if stat_result is not None:
+            return stat.S_ISDIR(stat_result.st_mode)
+        # Fall back to direct is_dir() call if stat failed
+        return path.is_dir()
+
+    def get_size(self, path: Path) -> int:
+        """Get file size using cached stat"""
+        stat_result = self._get_stat(str(path))
+        if stat_result is not None:
+            return stat_result.st_size
+        # Log warning for missing files and return 0
+        logger.warning(f"Skipping file in size calculation: {path}")
+        return 0
+
 
 # The default multipart upload chunk size is 8 MB. We used this to determine the small file threshold,
 # which is the chunk size multiplied by the small file threshold multiplier.
@@ -1008,6 +1048,7 @@ class S3AssetManager:
         self.session = session
 
         self.manifest_version: ManifestVersion = asset_manifest_version
+        self._stat_cache = _FileStatCache()
 
     def _process_input_path(
         self,
@@ -1145,7 +1186,7 @@ class S3AssetManager:
         for _path in input_paths:
             # Need to use absolute to not resolve symlinks, but need normpath to get rid of relative paths, i.e. '..'
             abs_path = Path(os.path.normpath(Path(_path).absolute()))
-            if not abs_path.exists():
+            if not self._stat_cache.exists(abs_path):
                 if require_paths_exist:
                     missing_input_paths.add(abs_path)
                 else:
@@ -1154,7 +1195,7 @@ class S3AssetManager:
                     )
                     referenced_paths.add(_path)
                 continue
-            if abs_path.is_dir():
+            if self._stat_cache.is_dir(abs_path):
                 misconfigured_directories.add(abs_path)
                 continue
 
@@ -1292,11 +1333,7 @@ class S3AssetManager:
 
     def _get_total_size_of_files(self, paths: list[str]) -> int:
         def get_file_size(path_str: str) -> int:
-            try:
-                return Path(path_str).resolve().stat().st_size
-            except (FileNotFoundError, PermissionError, OSError):
-                logger.warning(f"Skipping file in size calculation: {path_str}")
-                return 0
+            return self._stat_cache.get_size(Path(path_str))
 
         with concurrent.futures.ThreadPoolExecutor(max_workers=8) as executor:
             sizes = list(executor.map(get_file_size, paths))
--- a/test/unit/deadline/job_attachments/test_file_stat_cache.py
+++ b/test/unit/deadline/job_attachments/test_file_stat_cache.py
@@ -1,0 +1,111 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import stat
+from pathlib import Path
+from unittest.mock import patch, MagicMock
+
+from deadline.job_attachments.upload import _FileStatCache
+
+
+class TestFileStatCache:
+    def test_get_stat_caches_result(self, tmp_path):
+        """Test that stat results are cached and not called multiple times"""
+        cache = _FileStatCache()
+        test_file = tmp_path / "test.txt"
+        test_file.write_text("test")
+
+        with patch.object(Path, "stat") as mock_stat:
+            mock_stat.return_value = MagicMock()
+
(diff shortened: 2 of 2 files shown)


src/deadline/job_attachments/upload.py
```diff
@@ -82,6 +83,50 @@
...
+
+    def get_stat(self, path: Path) -> Optional[os.stat_result]:
+        """Get cached stat result or perform stat and cache it"""
+        path_str = str(path)
+        if path_str not in self._cache:
+            try:
+                self._cache[path_str] = path.stat()
+            except (FileNotFoundError, PermissionError, OSError):
+                self._cache[path_str] = None
+        return self._cache[path_str]
```
jusiskin: You could optimize this further using [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache).

Regardless of that, we should consider is having an unbounded cache - since this should be a best-effort optimization and not cause out of memory crashes.

There was a recent issue (#849) related to this same concern.

src/deadline/job_attachments/upload.py
```diff
@@ -82,6 +83,50 @@
 
 logger = logging.getLogger("deadline.job_attachments.upload")
 
+
+class FileStatCache:
```
jusiskin: This is becoming part of the [public API contract](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#public-contracts) based on the import path and class name. Is that intended?

jusiskin (changes requested):

src/deadline/job_attachments/upload.py
baxeaz: Sure makes sense, switched and set a maxsize

src/deadline/job_attachments/upload.py
baxeaz: No reason for it to be currently - updated

src/deadline/job_attachments/upload.py
```diff
@@ -1293,11 +1334,12 @@ def _find_matched_root_from_local_type_locations(
     def _get_total_size_of_files(self, paths: list[str]) -> int:
         def get_file_size(path_str: str) -> int:
             try:
-                return Path(path_str).resolve().stat().st_size
+                return self._stat_cache.get_size(Path(path_str))
```
jericht: Double checked that `Path.stat()` follows symlinks by default, we are good to remove `.resolve()` call here: https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat

src/deadline/job_attachments/upload.py
```diff
@@ -82,6 +84,44 @@
...
+        # Fall back to direct exists() call if stat failed
+        return path.exists()
+
+    def is_dir(self, path: Path) -> bool:
+        """Check if path is directory, using cache when possible"""
+        stat_result = self._get_stat(str(path))
+        if stat_result is not None:
+            return stat.S_ISDIR(stat_result.st_mode)
+        # Fall back to direct is_dir() call if stat failed
+        return path.is_dir()
```
jericht, line 113: Could we just return the stat result and the caller can query whatever information they want from it?

e.g. if the caller wants to know if it's a file, I think having another method for `is_file()` would be overcomplicating this class.

src/deadline/job_attachments/upload.py
```diff
@@ -82,6 +84,44 @@
...
+class _FileStatCache:
+    """Private cache for file stat results to avoid redundant filesystem calls"""
+
+    @lru_cache(maxsize=1024)
+    def _get_stat(self, path_str: str) -> Optional[os.stat_result]:
+        """Get cached stat result for a path string"""
+        try:
+            return Path(path_str).stat()
+        except (FileNotFoundError, PermissionError, OSError):
+            return None
```
jericht, line 97: Have we considered potential issues with stale data in the cache? (file metadata changes after the initial stat call) Let's double check that the lifetime of `S3AssetManager` instances is short enough that we aren't concerned about stale cache data.

baxeaz: I have yes.  S3AssetManager lives for the duration of the submission.  Changing the data while submitting isn't supported, but wasn't supported previously either - we'd need to implement some other system to copy out or otherwise lock files from changes.

leongdl: Just a thought, what if the cache was a nested element where the caller determines if it caches or not? Otherwise this decorator maybe forgotten and becomes a weird edge case as Jericho described.

test/unit/deadline/job_attachments/test_file_stat_cache.py
```diff
@@ -0,0 +1,110 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import stat
+from pathlib import Path
+from unittest.mock import patch, MagicMock
+
+from deadline.job_attachments.upload import _FileStatCache
+
+
+class TestFileStatCache:
```
leongdl, line 10: One test related question - on windows we have some edge case code to handle long long paths, eg `\\c:\` when the system does not have the long path registry on. Would the cache class have any issues?

leongdl (commented):
Nit comments only.

src/deadline/job_attachments/upload.py
baxeaz: We could do that - I've tried that change now for a bit, I'm not sure it actually comes out cleaner.  It ends up pushing some of these hidden internals out if you need to either handle the raised exceptions if I change it to truly just try to return stat or at least deal with the None case in the current form.  We also have unit tests that currently rely on this fallback behavior that would need to be changed to support that which isn't great.  Normally I'd agree we don't really want to have a wrapper for everything stat does, but because this is pretty clearly designed to be an internal helper for very specific use cases I'm a bit less concerned

src/deadline/job_attachments/upload.py
baxeaz: Not quite sure what you meant there - as in folks forget that it has the caching behavior?  You could always of course just call os.stat directly if you don't want to use the cache

test/unit/deadline/job_attachments/test_file_stat_cache.py
baxeaz: I could add a test if you can tell me a bit more about the case you're thinking of - at the end of the day though as long as we resolve the path in a uniform way it would return the string equivalent cache version - if it resolved to "something else" it would just be a cache miss and re resolve, no worse than it was before

src/deadline/job_attachments/upload.py
leongdl: Mostly from the caller side, not every caller will want cached data vs live data. So a boolean can easily select from both and versions as needed. 

Its ok , only thinking of code structure to reduce the number of if-else long term.

EG: 
`
if use_cache:
   do_this()
else
   do_that()
`

test/unit/deadline/job_attachments/test_file_stat_cache.py
leongdl: np - just thinking about all the cases we have to handle to not introduce regressions. The default case should have long path enabled so it is ok. Can be a long term follow up.

leongdl (dismissed):

src/deadline/job_attachments/upload.py
jusiskin: It seems easy enough for a caller to not use the `_FileStatCache`. This seems like a convenience for the caller, so I'd not consider this a blocker.

src/deadline/job_attachments/upload.py
```diff
@@ -82,6 +84,44 @@
...
+        return path.is_dir()
+
+    def get_size(self, path: Path) -> int:
+        """Get file size using cached stat"""
+        stat_result = self._get_stat(str(path))
+        if stat_result is not None:
+            return stat_result.st_size
+        # Log warning for missing files and return 0
+        logger.warning(f"Skipping file in size calculation: {path}")
+        return 0
```
jusiskin, line 122: I'm curious about why we have fall-backs for `is_dir` and `exists` but not here for `get_size`?

Aren't all of these high-level wrappers for `stat()` calls? I'd imagine they would either all fail (regardless of calling `stat()` or the higher-level function) or succeed. The outcome for having fall-backs is that we pay a double `stat()` penalty when the underlying `stat()` call fails. Unless there is a timing issue where the first `stat()` fails, but the subsequent one may pass - but I'm wondering if this is something that was observed.

Maybe we should just remove the fallback logic and let the caller handle exceptions. What do you think?

baxeaz: So if we remove the fallback logic at a minimum we'd need to rewrite one or two existing unit tests that rely on mocking exists calls - maybe that's ok, usually that would be something that might cause some pushback at review time though.  I could add the fallback logic to get_size as well though, that might be simpler - there's currently no use case that needs it but would keep things consistent?

baxeaz: Talked offline - Removed the try/except from _get_total_size_of_files and added cache specific testing for this warning

leongdl (approved):

jusiskin (approved):

Issue #821, opened 2025-09-02
Bug: concurrent job bundle submissions fail


joel-wong-aws:
### Describe the bug

When I attempt to submit two jobs in parallel, I get one of the two following errors:
`sqlite3.OperationalError: database is locked`
`sqlite3.OperationalError: table hashesV3 already exists`

### Expected Behaviour

Parallel job submissions should be handled in a concurrency safe way, e.g. parallel submissions are both allowed and concurrency-safe database handling, or the later submission waits until the first one is done

### Current Behaviour

The submission failed for the second job

### Reproduction Steps

Call `deadline bundle submit <job identifiers>` multiple times in parallel with large files.

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): Submitted from Amazon Linux 2023, affects all operation systems
2. Output of `python3 --version`: python 3.13.7
3. Output of `deadline --version`: 0.51.1
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? N/A

Please share other details about your environment that you think might be relevant to reproducing the bug.

epmog:
Edit: Synced up with Joel offline, seems to be around tests submitting jobs and having multiple runners on the same machine.

My guess is that two different processes are attempting to create/modify the hash cache db for job attachments at the same time. I'll add a job attachments label and poke them.

leongdl:
I have added code level retries to avoid multi-process access, and also isolated test cases via `fresh_deadline_config` so paralleled unit tests do not collide and cause flaky tests.

PR #862, opened 2025-09-26
fix: concurrent job bundle submissions fail due to sqlite3 table already exists

godobyte:
Fixes: *[821](https://github.com/aws-deadline/deadline-cloud/issues/821)*

When user attempt to submit two jobs in parallel on a new machine never ran this job bundle submit before, they can  get the following error:

sqlite3.OperationalError: table hashesV3 already exists

Only create the table if it doesn't exist already.

Concurrent initial submission wouldn't get error due to table already exists.

No

No

----

### DIFF
--- a/test/unit/deadline_job_attachments/caches/test_caches.py
+++ b/test/unit/deadline_job_attachments/caches/test_caches.py
@@ -187,6 +187,76 @@ class TestHashCache:
                 )
                 assert hc.get_entry("/no/file", HashAlgorithm.XXH128) is None
 
+    def test_table_creation_idempotent(self, tmpdir):
+        """
+        Tests that creating the hash cache table multiple times doesn't cause errors
+        """
+        cache_dir = tmpdir.mkdir("cache")
+
+        # Create the cache and table first time
+        with HashCache(cache_dir) as hc1:
+            test_entry = HashCacheEntry(
+                file_path="/test/file",
+                hash_algorithm=HashAlgorithm.XXH128,
+                file_hash="abc123",
+                last_modified_time="1234.56",
+            )
+            hc1.put_entry(test_entry)
+            retrieved_entry = hc1.get_entry("/test/file", HashAlgorithm.XXH128)
+            assert retrieved_entry == test_entry
+
+        # Create the cache again with the same directory - should not fail
+        with HashCache(cache_dir) as hc2:
+            # Should be able to retrieve the previously stored entry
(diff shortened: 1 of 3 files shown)


test/unit/deadline_job_attachments/caches/test_caches.py
```diff
@@ -187,6 +187,84 @@ def test_enter_sqlite_import_error(self, tmpdir):
...
+            )
+            hc2.put_entry(new_entry)
+            retrieved_new_entry = hc2.get_entry("/test/file2", HashAlgorithm.XXH128)
+            assert retrieved_new_entry == new_entry
+
+    def test_table_already_exists_no_error(self, tmpdir):
+        """
+        Tests that no error is raised when trying to create a table that already exists
+        This specifically tests the 'IF NOT EXISTS' clause in the CREATE TABLE statement
+        """
```
jericht, line 229: Do we need this test case? I think the test case we are adding above is sufficient.

test/unit/deadline_job_attachments/caches/test_caches.py
```diff
@@ -312,3 +390,73 @@ def test_get_connection_entry_returns_none_for_expired_entry(self, tmpdir):
...
+            )
+            s3c2.put_entry(new_entry)
+            retrieved_new_entry = s3c2.get_entry("bucket/Data/another-hash")
+            assert retrieved_new_entry == new_entry
+
+    def test_table_already_exists_no_error(self, tmpdir):
+        """
+        Tests that no error is raised when trying to create a table that already exists
+        This specifically tests the 'IF NOT EXISTS' clause in the CREATE TABLE statement
+        """
```
jericht, line 421: Same here, I think the test case above is sufficient.

test/unit/deadline_job_attachments/caches/test_caches.py
godobyte: The test above doesn't simulate the case create_query got run multiple times, which is when `with` entering the Exception block at the same time.

As per synced offline, modified the test to explicitly running the same query multiple times.

test/unit/deadline_job_attachments/caches/test_caches.py
godobyte: The test above doesn't simulate the case create_query got run multiple times, which is when `with` entering the Exception block at the same time.

As per synced offline, modified the test to explicitly running the same query multiple times.

benl-2023 (approved):

jericht (approved):

PR #871, opened 2025-10-02
test(docker): upgrade base image and match required coverage to what is tested

jericht:
The Docker tests were not working due to using an obsolete base image and the coverage for the default pytest configuration is too high

- Change the base Docker image to `python:3.9-bullseye`
- Update the test script to have a minimum coverage that reflects what is actually being tested by the docker tests.

Docker tests are passing again

Ran the Docker tests following its README.md

N/A

No

No

No

----

### DIFF
--- a/testing_containers/localuser_sudo_environment/Dockerfile
+++ b/testing_containers/localuser_sudo_environment/Dockerfile
@@ -1,6 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-FROM python:3.9-buster
+FROM python:3.13-bullseye
 
 # Set environment variables, and let our tests know that we are in an
 # environment that can run the sudo tests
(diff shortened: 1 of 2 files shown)


jairaws (dismissed):

kenyrish-amazon (dismissed):

testing_containers/localuser_sudo_environment/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-FROM python:3.9-buster
+FROM python:3.9-bullseye
```
leongdl: Is Python 3.9 deprecated? Do we need to bump it to >3.9?

jericht: Bumped to Python 3.13

leongdl (approved):

kenyrish-amazon (approved):

PR #887, opened 2025-10-10
RFC: i18n support

crowecawcaw:
Publishing as a draft for high level comments on what adding UI translations might look like. This 100% AI generated code with no cleanup, so don't bother reviewing the specifics too closely.


epmog:
I haven't done too much with i18n in client software previously, though if we're only looking to internationalize the GUI aspects this seems like a reasonable approach

PR #888, opened 2025-10-13
fix: Job history error after more than 100 submissions

mwiebe:

When submitting more than 100 jobs with largely the same name, the job numbering for the history directory stopped at 100 so repeated values and threw an error.

Change the logic to take the max of all numbers, not assume they have 2 digits.

Correct numbering of job directories within the monthly directories after job 100.

Confirmed that the fix addressed the failing case. Added a unit test.

N/A

No

No

No

----

### DIFF
--- a/src/deadline/client/job_bundle/__init__.py
+++ b/src/deadline/client/job_bundle/__init__.py
@@ -48,10 +48,15 @@ def create_job_history_bundle_dir(submitter_name: str, job_name: str) -> str:
 
     # Index the files so they sort in order of submission
     number = 1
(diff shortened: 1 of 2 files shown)


crowecawcaw (approved):

baxeaz (approved):

epmog (approved):

epmog:
What kind of error message did the user see? Some sort of "Directory already exists" exception? Did this prevent submission? Or did it pop up afterwards

mwiebe:
> What kind of error message did the user see? Some sort of "Directory already exists" exception? Did this prevent submission? Or did it pop up afterwards

Yes, they get an exception message that the directory already exists, and it prevents the submission.

PR #890, opened 2025-10-15
chore(ci): Adds github action to check for changes to public API

marofke:
Fixes: No associated github issue

While the Deadline Cloud client library does have a [public contract](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#public-contracts), we don't currently have any automated way of tracking changes to it.

There's a great CLI we're already importing due to a recent docs change, called [griffe](https://mkdocstrings.github.io/griffe) which is made to scan for API breakages. I've added a CI action that runs this tool against the target branch of the PR to see for any API breakages. 

Further, we should notify developers of any new APIs they may be inadvertently adding, so I've added some scripts that leverage griffe to check this (I've also added associated `hatch run docs:` commands to make this easier). The two step program is:
1. Generate a temporary JSON snapshot file of the existing API using a new script `generate_api_snapshot`
2. Then call another script `validate_api_snapshot` comparing a diff of the API snapshot of the working changes with the temporary JSON file

If there are any functional changes to the API, the CI step will fail, notifying the author. This helps catch any unintentional API changes.

Also updated the DEVELOPMENT.md and the scripts/README.md to reflect this.

Code authors will now be notified if they break / change the existing public interface.

- `hatch build && hatch run fmt && hatch run lint && hatch run test`
- Verified the first griffe CLI check in the CI catches breaking changes
  - Reordered optional parameters in a public function, confirmed this caused the CI to catch a breaking API change: https://github.com/aws-deadline/deadline-cloud/actions/runs/18598952853/job/53032401169?pr=890
  - Locally ran hatch run docs:check-api-breaks --against origin/mainline with changes to a public function that I added a new required parameter to, and confirmed it failed the check
- Added a temporary commit to confirm the snapshot checker could handle catching cases the base griffe check couldn't, see the run here: https://github.com/aws-deadline/deadline-cloud/actions/runs/18759660247/job/53520525551?pr=890job/53340855332?pr=890
  - manually added a new public class with some public functions (and others prefixed with `_`) and confirmed that only the public functions caused the CI to fail
  - added a class with `_` prefix, confirmed it didn't flag a failure
  - added classes and variables to a module prefixed with `_` and confirmed it didn't flag a failure
  - added a global variable (`TEST="test"`) to a public python file, confirmed it failed, prefixed it with `_`  (`_TEST="test"`), confirmed it succeeded
  - added a new optional parameter, confirmed this got flagged

- Yes, updated DEVELOPMENT.md and scripts/README.md

*   [x] This PR does not add any new dependencies. Griffe is already being pulled in through our docs requirements https://github.com/aws-deadline/deadline-cloud/blob/mainline/requirements-docs.txt

No

No

----


jusiskin:
One class of public interface that this may miss is when these APIs accept parameters or return values of a type that is not marked public according to our definition. For example:

```py
# Import some types not marked as public based on our public interface spec
from ._nested._module import ArgumentType, ReturnType


def public_interface_function(
    # public API accepts non-public type as an argument
    arg1: SomeType,
# public API returns non-public type
) -> ReturnType:
   # ...
```

The current test will not catch this leaky public interface. We would need to ensure that the argument types and return types are marked as public too, since customers need to supply and use them to use the public API method.

Can we expand the test to identify these and ensure that these types are captured in the explicit public interface?

scripts/api_snapshots/API_SNAPSHOTS.md
```diff
@@ -0,0 +1,52 @@
...
+# Check for breaking changes
+hatch run docs:griffe check -s src deadline
+
+# Check for any API changes using hatch command
+hatch run docs:compare-api-snapshots
+```
+
+### Generating New Snapshots
+
+If you are intentionally making breaking changes to the API, you will need to update the snapshot and mark your commit as a breaking change:
```
jusiskin: Can we generalize this? Right now it seems to be written towards intentionally breaking changes, but there is an additional case here where there we are intentionally adding new APIs.

Maybe the section could be "Making Intentional API Changes"? We could update the paragraph similarly.

scripts/api_snapshots/API_SNAPSHOTS.md
```diff
@@ -0,0 +1,84 @@
...
+   git commit -m "feat!: Modified create job API to add new parameter"
+   ```
+
+### Testing Locally
+
+Before pushing, you can test locally:
+
+```bash
+# Check for breaking changes against mainline branch using hatch command (recommended)
+hatch run docs:check-api-breaks --against origin/mainline
```
leongdl: Can this just be part of `hatch run test` or `hatch run integ:test`

This should just be part of normal testing.

leongdl: If we are relying on people to run this tool, then at least add this a condition to the Github PR description that they have run and executed this.

DEVELOPMENT.md
```diff
@@ -225,6 +225,13 @@ For the Python library interface:
 * Changing the location that a file or directory is created should be considered to be a breaking change. These locations have a tendancy to become
   de-facto parts of the public contract as users build automation that assumes these locations is unchanged.
 
+Note that we enforce our public contract through GitHub actions. See the [API Snapshots readme](scripts/api_snapshots/API_SNAPSHOTS.md) for more information.
+
+You can test API changes locally using these hatch commands:
```
leongdl: Please update the default PR description template too - no one reads this all the time :)

scripts/api_snapshots/API_SNAPSHOTS.md
marofke: It runs as part of the new CI step I added, so there's no need to add a checkbox to the PR template. I figured just documenting it if you want to run locally was good enough since it runs automatically in CI. I could try shoving it into the 'hatch run test' command though if you think that's necessary

DEVELOPMENT.md
marofke: This runs as part of CI so I don't think we need to add more noise to the PR template

scripts/api_snapshots/API_SNAPSHOTS.md
leongdl: Got it, I was hoping developers can run this as part of their work flow even before pushing it up to github / CI.

leongdl: Sounds good on the CI part and the github readme.

leongdl (dismissed):

leongdl:
LGTM, please merge this and get the CI ball rolling and advertising this feature!

We can always add it to the hatch run test command after the tests are proven stable and we have the process ironed out.

scripts/api_snapshots/API_SNAPSHOTS.md
```diff
@@ -0,0 +1,84 @@
...
+
+If you are intentionally making breaking changes to the API, you will need to update the snapshot and mark your commit as a breaking change:
+
+   ```bash
+   # Update the snapshot using hatch command
+   hatch run docs:generate-api-snapshot
+   
+   # Commit the updated snapshot, using ! to denote a breaking change
+   git add scripts/api_snapshots/api_snapshot.json
+   git commit -m "feat!: Modified create job API to add new parameter"
```
jusiskin: I don't believe we need the breaking change indicate (the `!` in `feat!:`) when adding a new public API, right?

DEVELOPMENT.md
```diff
@@ -225,6 +225,13 @@ For the Python library interface:
...
+* `hatch run docs:check-api-breaks --against origin/mainline` - Check for breaking changes against the mainline branch
+* `hatch run docs:validate-api-snapshot` - Validate current API against the baseline snapshot
+* `hatch run docs:generate-api-snapshot` - Generate a new API snapshot (when making intentional API changes)
```
jusiskin: nit: should we just have a summary and link to the `API_SNAPSHOTS.md` instead? to avoid repeating ourselves and having to maintain this documentation in two spots.

scripts/generate_api_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+                        if isinstance(path_str, str):
+                            try:
+                                abs_path = Path(path_str)
+                                if abs_path.is_absolute() and abs_path.is_relative_to(project_root):
+                                    normalized_paths.append(str(abs_path.relative_to(project_root)))
+                                else:
+                                    normalized_paths.append(path_str)
+                            except (ValueError, OSError):
+                                # If path conversion fails, keep original
+                                normalized_paths.append(path_str)
```
jusiskin: I'm not sure we want this fallback behavior. This could cause different snapshots on different machines depending on the checkout path, right?

scripts/validate_api_snapshot.py
```diff
@@ -0,0 +1,368 @@
...
+    # If we have the object, also check the is_public flag from griffe
+    if obj and isinstance(obj, dict):
+        # Griffe provides is_public flag - use it if available
+        if "is_public" in obj:
+            return obj["is_public"]
+
+    return True
+
+
+def normalize_paths_in_snapshot(snapshot_data, project_root: Path):
```
jusiskin: nit: this looks like a duplicate from `scripts/generate_api_snapshot.py`, right? Can we de-duplicate?

scripts/validate_api_snapshot.py
```diff
@@ -0,0 +1,368 @@
...
+            return json.load(f)
+    except FileNotFoundError:
+        print(f"Error: Snapshot file not found: {snapshot_path}")
+        sys.exit(1)
+    except json.JSONDecodeError as e:
+        print(f"Error: Invalid JSON in snapshot file: {e}")
+        sys.exit(1)
+
+
+def generate_current_snapshot() -> Dict[str, Any]:
```
jusiskin, line 34: Same comment as above. Looks like a duplicate

scripts/generate_api_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+                        else:
+                            normalized_paths.append(path_str)
+                    result[key] = normalized_paths
+                else:
+                    result[key] = value
+            elif key == "source_link" and isinstance(value, str):
+                result[key] = None  # Remove source links entirely to avoid user-specific info
```
mwiebe: nit: These could probably be `del result[key]` to reduce the size of the resulting json

mwiebe (dismissed):

jusiskin (changes requested):

scripts/api_snapshots/API_SNAPSHOTS.md
jusiskin: Sorry this one is outdated. Resolving.

scripts/generate_api_snapshot.py
marofke: Good call, we now skip these fields entirely

scripts/api_snapshots/API_SNAPSHOTS.md
marofke: Right, for some reason I thought we needed that, but I guess the important part is the `feat` part right? I've updated the comments where necessary

DEVELOPMENT.md
marofke: Good call, I've just retained the link to API_SNAPSHOTS.md

scripts/generate_api_snapshot.py
marofke: Removed!

scripts/generate_api_snapshot.py
```diff
@@ -0,0 +1,158 @@
...
+                                    normalized_paths.append(str(abs_path.relative_to(project_root)))
+                                # Skip non-relative paths to ensure consistent snapshots
+                            except (ValueError, OSError):
+                                # Skip paths that can't be processed to ensure consistent snapshots
+                                pass
+                        else:
+                            normalized_paths.append(path_str)
+                    result[key] = normalized_paths
+                else:
+                    result[key] = value
```
mwiebe: nit: What cases does this cover? Should it raise or skip instead?

.github/workflows/api-change-detection.yml
```diff
@@ -0,0 +1,80 @@
+name: API Change Detection
+
+on:
+  pull_request:
+    branches: [ mainline, release, feature_assets_cli, 'patch_*' ]
```
jusiskin: should this be `feature_*`? Currently the repository has two feature branches (`feature_attributions`, `feature_assets_cli`).

scripts/api_snapshots/api_snapshot.json
jusiskin, line 1: I see this file is 1.93MB. Git stores the full version of each file in the repository. Each time we evolve the API it will generate a new version of this file and I'm wondering about the impact on the git operations.

A few alternative approaches come to mind:

1.  **Generate snapshots on-demand for the merge target (recommended)**

    Do not store the previous snapshot in version control. Instead we would generate the snapshot on-demand based on the target branch of the PR.

    Trade-offs:
    *   No additional process for contributors
    *   No additional git storage
    *   We need to establish a process for maintainers to accept/acknowledge intended API changes

2.  **Remove unused data**

    I have not looked closely, at the `api_snapshot.json` schema, but perhaps some of it may not be needed for the use-case (detecting new public APIs being added).

2.  **Compress this file**

    Using the `gzip` CLI utility I compressed from 2MB &rarr; 92KB (~95% reduction).

    Trade-offs:
    *   Harder to reason about changes using git tooling, but the scripts this PR introduces could handle the [de]compression to support this use-case
    *   Faster repository checkouts
    *   Less storage

4.  **Git LFS**

    Turn on git LFS and mark this file as LFS-enabled.

    Trade-offs:
    *   More efficient storage for large JSON snapshots
    *   More overhead for onboarding
        *   Contributors need to install git-lfs on their developer machine
        *   Contributors need to setup LFS for each checkout of `deadline-cloud` (`git lfs install`)

scripts/api_snapshots/API_SNAPSHOTS.md
```diff
@@ -0,0 +1,84 @@
...
+If you are intentionally making breaking changes to the API, you will need to update the snapshot and mark your commit as `feat` following [conventional commit sytax](CONTRIBUTING.md#conventional-commits):
+
+   ```bash
+   # Update the snapshot using hatch command
+   hatch run docs:generate-api-snapshot
+   
+   # Commit the updated snapshot, and make sure to mention the new API
+   git add scripts/api_snapshots/api_snapshot.json
+   git commit -m "feat: Modified create job API to add new parameter"
+   ```
```
jusiskin: This may need one more pass. The heading is about making API changes, but the paragraph begins with

>   If you are intentionally making **breaking** changes to the API ..."

*   **For new APIs OR breaking changes:** the contributor must update the snapshot
*   **For breaking changes:** the contributor must structure their commit message according to the conventional commit spec for breaking changes.

mwiebe (dismissed):

jusiskin (changes requested):
Thanks for incorporating my feedback from the last pass. After thinking a bit more about this, I think we should consider the long-term impact of storing the 2MB snapshots with each API change.

I suggested an approach where we generate the snapshot from the merge target on-demand. What do you think?

scripts/README.md
```diff
@@ -1,5 +1,108 @@
...
+#### Usage Workflow
+
+##### When Making API Changes
+
+When making any API changes, follow these guidelines:
+
+**For new APIs (additions):**
+- Use conventional commit syntax to mark new features:
+  ```bash
+  git commit -m "feat: Add new job submission API endpoint"
```
jusiskin: We require commits have a developer certificate of origin, so maybe we should change to:

```suggestion
  git commit -s -m "feat: Add new job submission API endpoint"
```

same applies to the `git commit ...` commands below

.github/workflows/api-change-detection.yml
jusiskin: It appears this suggestion is still unaddressed. Should we generalize to all `feature_*` branches?

jusiskin (changes requested):
Looks great. Just two small suggestions remaining.

jusiskin (approved):
Great work here! It will be super helpful to have automated linting for public API changes worked into our CI.

This is a super valuable contribution, thanks @marofke!

leongdl (approved):

PR #896, opened 2025-10-27
perf: Adding thread local s3 cache db connections to fix performance …

baxeaz:
Note this is a re-revert of a previous commit/PR with additional testing and improvements from adding WAL mode - https://github.com/aws-deadline/deadline-cloud/pull/848

Adding back the comments and updating each with the additional bottleneck uncovered and solution.

Our S3 cache was using a connection shared across threads which required a lock for all access, creating a bottleneck when attempting to divide up work across threads.

With the original fix this uncovered further bottlenecks when attempting to write and read concurrently to larger DB files in particular.  The original code prevented us from seeing this bottleneck because we were sharing a single connection which was protected by a lock.  Concurrent reading was not possible due to the shared connection and shared self.db_lock - another thread attempting to read through one of our methods would simply sit on the lock until it was released.  With the original fix this bottleneck was removed, and instead in cases particularly with larger DB files we could see our DB connection timer (Set as the default 5 seconds) time out due to missing additional configuration settings preventing concurrent readers.

Add back the original fix and enable [Write Ahead Logging](https://sqlite.org/wal.html)

Add additional unit tests to deadline-cloud to better stress test concurrent reading/writing with larger DB files.

Improved DB performance.  Testing with the Meerkat scene in Unreal:

(Deleting my cache DB files before each test:

**5 runs with Current Code:**
19.715
19.078
19.499
19.531
19.191
**Avg: 19.4028**

**5 runs with this change:**
14.392
14.337
14.286
14.559
14.253
**Avg: 14.3654**

**~26% improvement**

**New Unit Test** - Note you can see test_large_db_concurrent_operations_expose_timeout_issues catch the timeout error if you remove the pragma setting WAL mode in the DB.

New and existing unit tests pass.

When starting from a large DB file which previously failed in Worker Agent tests:
ev-dsk-baxelson-2b-c90f05c5 % ls -al ~/.deadline/cache
total 65404
drwxr-xr-x 3 baxelson amazon     4096 Oct 25 14:14 .
drwxr-xr-x 3 baxelson amazon     4096 Oct 17 19:47 ..
drwxr-xr-x 2 baxelson amazon     4096 Oct 17 21:16 bak
-rw-r--r-- 1 baxelson amazon 65358848 Oct 25 14:14 hash_cache.db
-rw-r--r-- 1 baxelson amazon  1592320 Oct 25 14:14 s3_check_cache.db

(25-10-25 14:37:29) <0> [~/clients/deadline-cloud-worker-agent]

Worker agent Linux tests pass:
================================================= slowest 5 durations ==================================================
648.55s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_shuts_down_host_machine_if_configured
253.08s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_requires_no_instance_profile
228.66s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
214.67s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_local_session_logs_can_be_turned_off
213.22s setup    test/e2e/test_cap_kill.py::test_cap_kill_not_inherited_by_running_jobs
========================== 45 passed, 15 skipped, 1 xpassed, 1 warning in 4094.63s (1:08:14) ===========================

And Windows tests pass:
================================================= slowest 5 durations ==================================================
679.29s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_shuts_down_host_machine_if_configured
343.84s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
329.59s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_requires_no_instance_profile
328.70s setup    test/e2e/test_credential_handling.py::test_access_worker_credential_file_from_job_windows
325.79s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_session_root_dir
================================ 43 passed, 18 skipped, 1 warning in 5141.65s (1:25:41) ================================

Yes

*   [ X] This PR does not add any new dependencies.

No

No

----

### DIFF
--- a/src/deadline/job_attachments/caches/cache_db.py
+++ b/src/deadline/job_attachments/caches/cache_db.py
@@ -6,6 +6,7 @@ Module for defining a local cache file.
 
 import logging
 import os
+import threading as _threading
 from abc import ABC
 from threading import Lock
 from typing import Optional
@@ -34,6 +35,8 @@ class CacheDB(ABC):
         self.cache_name: str = cache_name
         self.table_name: str = table_name
         self.create_query: str = create_query
+        self._local = _threading.local()
+        self._local_connections: set = set()
 
         try:
             # SQLite is included in Python installers, but might not exist if building python from source.
@@ -64,6 +67,7 @@ class CacheDB(ABC):
                 self.db_connection: sqlite3.Connection = sqlite3.connect(
                     self.cache_dir, check_same_thread=False
                 )
+                self.db_connection.execute("PRAGMA journal_mode=WAL")
             except sqlite3.OperationalError as oe:
                 raise JobAttachmentsError(
                     f"Could not access cache file in {self.cache_dir}"
@@ -81,8 +85,35 @@ class CacheDB(ABC):
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
         """Called when exiting the context manager."""
+
         if self.enabled:
+            import sqlite3
+
             self.db_connection.close()
+            for conn in self._local_connections:
+                try:
+                    conn.close()
+                except sqlite3.Error as e:
+                    logger.warning(f"SQLite connection failed to close with error {e}")
+
+            self._local_connections.clear()
+
+    def get_local_connection(self):
+        """Create and/or returns a thread local connection to the SQLite database."""
+        if not self.enabled:
+            return None
+        import sqlite3
+
+        if not hasattr(self._local, "connection"):
+            try:
+                self._local.connection = sqlite3.connect(self.cache_dir, check_same_thread=False)
+                self._local_connections.add(self._local.connection)
+            except sqlite3.OperationalError as oe:
+                raise JobAttachmentsError(
+                    f"Could not create connection to cache in {self.cache_dir}"
+                ) from oe
+
+        return self._local.connection
 
     @classmethod
     def get_default_cache_db_file_dir(cls) -> Optional[str]:
@@ -99,12 +130,23 @@ class CacheDB(ABC):
         """
         Removes the underlying cache contents from the file system.
         """
+
         if self.enabled:
+            import sqlite3
+
             self.db_connection.close()
+            conn_list = list(self._local_connections)
+            for conn in conn_list:
+                try:
+                    conn.close()
+                    self._local_connections.remove(conn)
+                except sqlite3.Error as e:
+                    logger.warning(f"SQLite connection failed to close with error {e}")
 
         logger.debug(f"The cache {self.cache_dir} will be removed")
         try:
             os.remove(self.cache_dir)
         except Exception as e:
             logger.error(f"Error occurred while removing the cache file {self.cache_dir}: {e}")
+
             raise e
--- a/src/deadline/job_attachments/caches/s3_check_cache.py
+++ b/src/deadline/job_attachments/caches/s3_check_cache.py
@@ -56,6 +56,32 @@ class S3CheckCache(CacheDB):
             cache_dir=cache_dir,
         )
 
+    def get_connection_entry(self, s3_key: str, connection) -> Optional[S3CheckCacheEntry]:
+        """
+        Returns an entry from the hash cache, if it exists.  This is the "lockless" (Doesn't take
+        the main db_lock protecting db_connection) version of get_entry which expects a connection
+        parameter for the connection which will be used to read from the DB - this can generally
+        be the thread local connection returned by get_local_connection()
+        """
+
+        entry_vals = connection.execute(
+            f"SELECT * FROM {self.table_name} WHERE s3_key=?",
+            [s3_key],
+        ).fetchone()
+        if entry_vals:
+            entry = S3CheckCacheEntry(
+                s3_key=entry_vals[0],
+                last_seen_time=str(entry_vals[1]),
+            )
+            try:
+                last_seen = datetime.fromtimestamp(float(entry.last_seen_time))
+                if (datetime.now() - last_seen).days < self.ENTRY_EXPIRY_DAYS:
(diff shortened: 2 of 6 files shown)


src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -34,6 +35,8 @@ def __init__(
         self.cache_name: str = cache_name
         self.table_name: str = table_name
         self.create_query: str = create_query
+        self.local = threading.local()
```
marofke: Nit: can these be "private" members, i.e. prefixed with `_` (`self._local`)

src/deadline/job_attachments/caches/s3_check_cache.py
```diff
@@ -56,6 +56,29 @@ def __init__(self, cache_dir: Optional[str] = None) -> None:
             cache_dir=cache_dir,
         )
 
+    def get_connection_entry(self, s3_key: str, connection) -> Optional[S3CheckCacheEntry]:
```
marofke, line 59: Nit: similar comment, should this function name be prefixed with `_`? It looks like it's more of an implementation detail for `get_entry`, do we really need to add this as a new API that we need to support for consumers of the library?

src/deadline/job_attachments/caches/cache_db.py
baxeaz: Yeah local can definitely be _local, will switch

src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -6,6 +6,7 @@
 
 import logging
 import os
+import threading
```
marofke: Nit: we're now trying to be more intentional with what we expose in our public interface. Can you import this with a private alias so we're not exposing this? I.e. `import threading as _threading`

src/deadline/job_attachments/caches/s3_check_cache.py
baxeaz: get_connection_entry is called directly as well - it's the "lockless" equivalent of get_entry - get_entry calls it so we maintain backwards compatibility, but our existing cases were generally changed over to get_connection_entry to skip the lock

src/deadline/job_attachments/caches/cache_db.py
baxeaz: Oh is that what it was upset about?  Yeah I can change that - Are there no other solutions here?  This seems like it'll be a common point of friction for contributors

marofke: Yea unfortunately we need to do this for now, as we haven't been as disciplined about our imports as other libraries like OJD where everything is private by default and then we just specify what to make public through the __init__.py file: https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/src/openjd/model/__init__.py

I don't want to ask you to change up the entire file as part of this change though. And yea, we'll need to work out ways to do this better as we start to clean up the code base, but not adding new things to our public interface will help us keep this library more stable for consumers going forward.

src/deadline/job_attachments/caches/s3_check_cache.py
marofke: Ah ok I see thanks for clarifying. I think we'll need to include this in the commit message then so it's clear we added a new public function and what it's intended to be used for, so it shows up in the changelog when we do the next release.

Same goes for any other public function / etc. you are intentionally keeping public

marofke: It would be helpful to have a more descriptive docstring here explaining the tradeoffs of calling this directly over `get_entry` (and we should probably update the docstring for `get_entry` as well) now that we have auto-generated docs https://aws-deadline.github.io/deadline-cloud/code_reference/job_attachments_caches/#deadline.job_attachments.caches

src/deadline/job_attachments/caches/cache_db.py
marofke: Thanks!

marofke (approved):
Thanks for your patience updating this PR for our new standards. LGTM!

src/deadline/job_attachments/caches/cache_db.py
```diff
@@ -81,8 +85,35 @@ def __enter__(self):
...
+
+    def get_local_connection(self):
+        """Create and/or returns a thread local connection to the SQLite database."""
+        if not self.enabled:
+            return None
+        import sqlite3
+
+        if not hasattr(self._local, "connection"):
+            try:
+                self._local.connection = sqlite3.connect(self.cache_dir, check_same_thread=False)
```
leongdl, line 109: how come we don't need `self.db_connection.execute("PRAGMA journal_mode=WAL")` here.

baxeaz: I could add it there but we only need to guarantee that this is set once (And it then applies to future connections and it's saved to the DB file) - so the _enter_ context where we take care of table creation and set up our "write connection" seems cleanest at the moment.  Down the line we may want to get rid of that "shared write connection" as well as its lock and move _everything_ over to the thread local connections.  Outside the scope of my change here, and certainly that would have been a breaking change, but maybe worth considering

leongdl: Got it, so it is purely a bootstrap operation.

leongdl: Worth adding a comment next time you are working in this repo, but not blocking!

leongdl (approved):

PR #901, opened 2025-10-30
feat: add foundations for future non-english UI translations

crowecawcaw:
We want to support multiple languages in the GUI submitter.

- Use Qt's translations feature.
- Add `tr` translation util
- Wrap all UI strings in `tr`
- Add a short translations README
- Add a translations file for English. Other languages will be added in other PRs

No customer visible change yet, but when we add translation files, other languages will be supported.

Screenshot of a the GUI submitter with a Japanese translation file (not included in this PR):

Screenshot 2025-10-30 at 8 34 50 AM

- Ran unit tests.
- Tested it locally
- Added some unit tests. Unit tests are skipped for the moment since this PR doesn't include translations in other languages. 

Yes

It is expected to flag the interface change alert. Not immediately breaking, but it will be when we add non-English translations. Once we add translations, if someone's system locale is not English, the GUI submitter will appear in a different language. 

No
----


leongdl (dismissed):

joel-wong-aws (dismissed):

src/deadline/client/ui/_utils.py
```diff
@@ -150,6 +184,8 @@ def gui_context_for_cli(automatically_install_dependencies: bool):
         icon = QIcon(str(Path(__file__).parent.parent / "ui" / "resources" / "deadline_logo.svg"))
         app.setWindowIcon(icon)
 
+        load_translations()
```
jusiskin: We've instrumented this here in the `gui_context_for_cli` function. I presume this is called by the `deadline bundle gui-submit` code path, which matches the PR title.

Two questions:


## 1. Library Consumers

Do we envision consumers of the `deadline` Python library (e.g. integrated DCC submitters) to integrate with this translation support in the future? If so, what is the intended API usage - do callers need to call `load_translations` before using any `deadline` GUI APIs? How do we anticipate downstream Python applications/libraries to provide their own unique translations?

## 2. Other CLI-based GUIs

Looking at the name of this function and its callers, it seems all CLI->GUI features benefit from translations here? For example:

```sh
deadline config gui
```

Maybe we should rename the PR and commit (for the changelog). How about:

```
feat: add foundations for future non-english UI translations
```

My rationale is that currently the title may be misinterpreted as providing other language translations.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling", "hatch-vcs"]
+requires = ["hatchling", "hatch-vcs", "PySide6-essentials >= 6.6,< 6.11"]
```
epmog: based on the build environment this will most likely fail. Chatting internally

jusiskin (changes requested):

src/deadline/client/ui/_utils.py
crowecawcaw: 1. Great callout. Refactored so the translation loading doesn't require a separate call.

2. Updated!

pyproject.toml
crowecawcaw: Refactored the code to remove Pyside/Qt. Translations are now stored in JSON files. Missing translations are detected using typing. Seems simpler and avoids dependencies.

src/deadline/client/ui/translations/README.md
```diff
@@ -0,0 +1,82 @@
+# UI Translations
```
mwiebe, line 1: Would this be suitable to put in the top level `docs` directory?

crowecawcaw: It looks like the `docs/` content currently targets CLI/library consumers and is related to the public interfaces and library usage. The content in this README is more for developers making changes to this library and want to know how translations are handled internally. 

Do you still think it makes sense in `docs/`? Happy to move it if so.

src/deadline/client/ui/_utils.py
```diff
@@ -1,9 +1,46 @@
...
+# Import TranslationKey type only during type checking to avoid runtime errors
+# if _translation_keys.py doesn't exist (it's generated during build)
+if TYPE_CHECKING:
+    from ._translation_keys import TranslationKey
+else:
+    TranslationKey = str
+
+
+@lru_cache(maxsize=1)
+def get_translations() -> Dict[str, str]:
```
jusiskin: Nice solution to the self-bootstrapping!

test/unit/deadline_client/ui/test_translations.py
```diff
@@ -0,0 +1,74 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import json
+import os
+import pytest
+from pathlib import Path
+import sys
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent.parent.parent / "src"))
```
jusiskin: Is this necessary? Hatch will install the code in the hatch environment so it should work already using:

```sh
hatch run test
```

without needing this import. Modifying the module search path like this may cause unexpected issues for the test suite - especially since this happens when this test module is imported.

If there's something I am missing, we should at least provide a code comment to explain why this is here.

jusiskin (dismissed):

test/unit/deadline_client/ui/test_translations.py
crowecawcaw: This was not actually necessary. Removed!

crowecawcaw:
This PR was refactored from its initial design in order to remove Qt-related dependencies. The first attempt used Qt's built in translation tooling including a runtime translation function and some build-time tools for extracting strings and compiling them into a compressed translations format. This new revision replaces Qt tooling with  JSON for translation files, a simple function to map translation keys to the localized string, and generated mypy types to verify at build time that every string we attempt to translate has a valid translation. It now has no external dependencies and IMO is also easier to follow and understand.

leongdl (approved):

src/deadline/client/ui/_utils.py
leongdl: Oh, your PR is failing for API check, this API should be private with `_`

leongdl (changes requested):
Please move the translation functions to be private?

leongdl (approved):

jusiskin (approved):

PR #905, opened 2025-11-03
chore: Explicitly call main in deadline_cli_main.py to fix pyinstaller executables

evanspearman-a:

Pyinstaller executables for deadline-cloud were completely broken and not doing anything.

Running a build built with pyinstallers debug option showed that deadline_cli_main.py was being run, but nothing was happening. I added an explicit call to main() from here, which caused it to work again.

Executables will work again

I build a pyinstaller executable and ran `deadline --version` and verified that it now printed the version rather than doing nothing. I also ran the same command through a normal Python interpreter to make sure it didn't break anything when not running through pyinstaller.

N/A

No, this is a fixing change

No

----

### DIFF
--- a/src/deadline/client/cli/deadline_cli_main.py
+++ b/src/deadline/client/cli/deadline_cli_main.py
@@ -8,3 +8,6 @@ Required by pyinstaller, do not delete.
 from deadline.client.cli._main import deadline as main
 
 __all__ = ["main"]
+
+if __name__ == "__main__":
+    main()


src/deadline/client/cli/deadline_cli_main.py
```diff
@@ -8,3 +8,5 @@
 from deadline.client.cli._main import deadline as main
 
 __all__ = ["main"]
+
+main()
```
AWS-Samuel: Should this be wrapped in an `if __name__ == "__main__":`? Wondering if this will cause anything that loads the CLI module to call this function when we don't necessarily want it to

evanspearman-a: I can try that. I _think_ this file is only used for pyinstaller though, so it might not matter.

evanspearman-a: Still works with that, so I've added it to be safe.

AWS-Samuel (approved):

src/deadline/client/cli/deadline_cli_main.py
AWS-Samuel: Thanks!

mwiebe (approved):

PR #909, opened 2025-11-05
feat(mcp): add get_session_logs to mcp server

shidarin:
One of the endpoints needed to fulfill the user
story "tell me why my job is failing"

In order to eventually fulfill the user story of a user asking an AI agent "why is my job failing", the agent will need access to the get_session_logs() function. This PR exposes that functionality.

get_session_logs was added to the tool registry. This required the removal of `from __future__ import annotations` from `_job_monitoring` due to how FastMCP handles annotations,  but that future behavior did not appear to be required in this module.

Agentic AIs can now get session logs.

Yes, on Python 3.8 and 3.11

No.

No.

The mcp_guide.md file was updated

No

No

----

### DIFF
--- a/src/deadline/client/api/_job_monitoring.py
+++ b/src/deadline/client/api/_job_monitoring.py
@@ -4,7 +4,6 @@
 Functions for monitoring job status and retrieving job information.
 """
 
-from __future__ import annotations
 import datetime
 import time
 from typing import List, Dict, Any, Optional, Callable
--- a/docs/mcp_guide.md
+++ b/docs/mcp_guide.md
@@ -61,6 +61,8 @@ With the AWS Deadline Cloud MCP Server, you can use natural language for various
 - "Submit the render job in /path/to/my-job-bundle"
 - "Submit a job with priority 80 to my render queue"
 - "Show me the status of job job-3a907bac684841f69fc344867ee166de"
+- "Get the logs for session session-abc123 to see why my task failed"
+- "Show me the CloudWatch logs for session session-xyz789"
 - "Download output from job job-3a907bac684841f69fc344867ee166de"
 - "Download output from step step-render in job job-3a907bac684841f69fc344867ee166de"
 ```
@@ -76,7 +78,7 @@ The MCP server provides access to all allowlisted/configured Deadline Cloud API 
 - `deadline_list_storage_profiles_for_queue()`: List storage profiles for a queue
 
 - `deadline_check_authentication_status()`: Check current authentication status
-
+- `deadline_get_session_logs()`: Get CloudWatch logs for a specific session
 - `deadline_submit_job()`: Submit an Open Job Description job bundle to AWS Deadline Cloud
 - `deadline_download_job_output()`: Download job output files from AWS Deadline Cloud
 
--- a/src/deadline/_mcp/registry.py
+++ b/src/deadline/_mcp/registry.py
@@ -59,6 +59,18 @@ TOOL_REGISTRY: Dict[str, ToolDefinition] = {
         "func": api.check_authentication_status,
         "param_names": None,
     },
+    "get_session_logs": {
+        "func": api.get_session_logs,
+        "param_names": [
+            "farm_id",
+            "queue_id",
+            "session_id",
+            "limit",
+            "start_time",
+            "end_time",
+            "next_token",
+        ],
+    },
     "submit_job": {
         "func": job.submit_job,
         "param_names": None,
--- a/test/unit/deadline_mcp/test_mcp.py
+++ b/test/unit/deadline_mcp/test_mcp.py
@@ -39,7 +39,7 @@ class TestToolRegistry:
 
     def test_expected_functions(self):
         """Test that expected functions are in the registry."""
-        expected = ["list_farms", "submit_job", "download_job_output"]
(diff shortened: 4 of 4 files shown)


leongdl (approved):

shidarin:
I'll fix the linting

leongdl (changes requested):
- Please run lint.
```
Successfully installed deadline-0.0.post1+g32f549b97 psutil-7.1.3 qtpy-2.4.3 xxhash-3.6.0
Checking dependencies
cmd [1] | ruff check .
All checks passed!
cmd [2] | ruff format --check --diff .
2 files would be reformatted, 281 files already formatted
--- src/deadline/_mcp/registry.py
+++ src/deadline/_mcp/registry.py
@@ -61,7 +61,15 @@
```

leongdl (approved):

src/deadline/client/api/_job_monitoring.py
```diff
@@ -4,7 +4,6 @@
 Functions for monitoring job status and retrieving job information.
 """
 
-from __future__ import annotations
```
AWS-Samuel, line 7: I don't think this was meant to be removed

AWS-Samuel (changes requested):

crowecawcaw (approved):

src/deadline/client/api/_job_monitoring.py
shidarin: If you meant by me, it was absolutely intentional. If you mean in general, I'd have to take your word for it. I don't believe there's any delayed annotation references in this module that would require its inclusion, and it's breaking FastMCP behavior with datetime on the function we're trying to expose.

AWS-Samuel: I see now, in the description you called this out.  Do you mind sharing the error you were getting? I want to make sure that removing this import is the correct fix because it will change how annotations in this module are treated at runtime.

seant-aws (approved):

src/deadline/client/api/_job_monitoring.py
shidarin: Without removing the annotation, the MCP server closes immediately. During debugging, you'll see this error:

``` python
./.venv/bin/python -c "from deadline._mcp.server import main; print('Import successful')" 2>&1                             
  File "<string>", line 1, in <module>
  File "/Users/snw/PycharmProjects/deadline-cloud/src/deadline/_mcp/server.py", line 9, in <module>
    register_api_tools(app, prefix="deadline_")
  File "/Users/snw/PycharmProjects/deadline-cloud/src/deadline/_mcp/utils.py", line 186, in register_api_tools
    app.tool(name=f"{prefix}{tool_name}", description=description)(wrapper)
  File "/Users/snw/PycharmProjects/deadline-cloud/.venv/lib/python3.12/site-packages/mcp/server/fastmcp/server.py", line 421, in decorator
    self.add_tool(
  File "/Users/snw/PycharmProjects/deadline-cloud/.venv/lib/python3.12/site-packages/mcp/server/fastmcp/server.py", line 366, in add_tool
    self._tool_manager.add_tool(
  File "/Users/snw/PycharmProjects/deadline-cloud/.venv/lib/python3.12/site-packages/mcp/server/fastmcp/tools/tool_manager.py", line 55, in add_tool
    tool = Tool.from_function(
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/snw/PycharmProjects/deadline-cloud/.venv/lib/python3.12/site-packages/mcp/server/fastmcp/tools/base.py", line 67, in from_function
    if issubclass(param.annotation, Context):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: issubclass() arg 1 must be a class
```

The issue is that FastMCP's Tool.from_function() is trying to inspect the parameter annotations, our _create_wrapper function tries to handle this but would need changes to handle datetime..

I actually had those changes in, I even had changes in _job_monitoring allowing for get_session_logs to take either a datetime or a string, but simply removing the from `__future__ import annotations` simplifies our code. [PEP-563](https://peps.python.org/pep-0563/) is not slated to become the default behavior in python- instead [PEP-649](https://peps.python.org/pep-0649/) and [749](https://peps.python.org/pep-0749/) were implemented for py3.14+. The behavior in `__future__ import annotations` is slated to be deprecated and removed (but not until 3.13 itself is EOL). 

None of the type hints in this module required the deferred evaluation behavior it gives us, and while it exists in this module, it isn't imported in most of the modules in deadline-cloud. It either should never have been added here, or should have been removed when the code requiring a deferred annotation was removed.

Is there a test we can run or write to ensure that removing annotations from this module doesn't break code we're not currently testing?

AWS-Samuel: Thanks for explaining this and for linking the error 

> Is there a test we can run or write to ensure that removing annotations from this module doesn't break code we're not currently testing?

The unit tests passing in all python versions should be enough here. Like you said, none of the type hints in that module depend on the deferred annotation. 

My preference would be to fix the annotations bug with `datetime` in our wrapper. If we end up adding a function to the MCP server which we can't remove `__future__` from it sounds like we'll need to do that anyway. For now, to get this in, I think it's fine to go with the quick fix. 

Thanks!

AWS-Samuel (approved):

PR #910, opened 2025-11-06
rfc: chore: inform user that GUI dependencies will be installed

crowecawcaw:
Note: Demoing an idea to collect feedback.

When the `--install-gui` flag is used like `deadline bundle gui-submit --install-gui`, the library automatically installs PySide6-essentials. Several submitters use the GUI submitter and this `--install-gui` flag to function. When they run for the first time, the submitter takes several seconds to load without the user understanding why. In these use cases, the terminal output is not usually visible.

Idea for a solution: show an alert that the dependencies will be installed so that 1) the user knows to expect a delay and 2) the user approves the install. The alert uses tkinter which is part of the standard library.

Screenshot 2025-11-05 at 2 54 17 PM

Better visibility for users.

n/a

n/a

Ish. It doesn't break anything, but it does change the user experience of submitters that use the GUI submitter.

No

----

### DIFF
--- a/src/deadline/client/ui/_utils.py
+++ b/src/deadline/client/ui/_utils.py
@@ -82,7 +82,22 @@ def gui_context_for_cli(automatically_install_dependencies: bool):
     has_pyside6 = importlib.util.find_spec("PySide6")
     has_pyside2 = importlib.util.find_spec("PySide2")
     if not (has_pyside6 or has_pyside2):
-        if not automatically_install_dependencies:
+        if automatically_install_dependencies:
+            # Show GUI confirmation dialog
+            import tkinter as tk
+            from tkinter import messagebox
+
+            root = tk.Tk()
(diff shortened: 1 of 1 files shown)


src/deadline/client/ui/_utils.py
```diff
@@ -82,7 +82,22 @@ def gui_context_for_cli(automatically_install_dependencies: bool):
     has_pyside6 = importlib.util.find_spec("PySide6")
     has_pyside2 = importlib.util.find_spec("PySide2")
     if not (has_pyside6 or has_pyside2):
-        if not automatically_install_dependencies:
+        if automatically_install_dependencies:
+            # Show GUI confirmation dialog
+            import tkinter as tk
```
edwards-aws, line 87: I'll have to double check. I don't think we ship tcl/tk right now. So this would be a new dependency.

PR #912, opened 2025-11-07
feat(mcp): Add get_job_logs to deadline cloud mcp server

kavmur:
The MCP server needed a get_job_logs tool to retrieve CloudWatch logs for AWS Deadline Cloud job sessions. This enables users to debug and monitor job execution through the MCP interface without switching to the AWS console or CLI.

Added a new get_job_logs MCP tool that:
- Retrieves CloudWatch logs for a specific session or auto-selects the most recent session for a job
- Accepts farm_id, queue_id, job_id, or session_id as parameters
- Supports filtering by time range (start_time, end_time) and pagination (next_token, limit)
- Formats timestamps in UTC or local timezone based on user preference
- Prioritizes ongoing sessions over completed ones when auto-selecting

Agents can now directly use the mcp to retrieve job logs and use for monitoring/troubleshooting workflows.

No

No

----

### DIFF
--- a/src/deadline/_mcp/tools/job.py
+++ b/src/deadline/_mcp/tools/job.py
@@ -4,6 +4,7 @@
 Deadline Cloud Job tools.
 """
 
+import datetime
 import io
 import json
 import os
@@ -11,9 +12,14 @@ import time
 from contextlib import redirect_stdout
 from typing import Any, Dict, Optional
 
-from ...client.api import create_job_from_job_bundle
+from ...client.api import (
+    create_job_from_job_bundle,
+    get_session_logs,
+    list_sessions_for_job,
+)
 from ...client.cli._groups.job_group import _download_job_output
 from ...client.config import config_file
+from ...common.session_utils import get_session_sort_key
 
 # TODO: Make submit_job tool async once progress reporting feature is supported in clients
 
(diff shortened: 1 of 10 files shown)


src/deadline/_mcp/tools/job.py
```diff
@@ -196,3 +198,122 @@ def download_job_output(
...
+    if not session_id:
+        paginator = deadline.get_paginator("list_sessions")
+        sessions = [
+            s
+            for page in paginator.paginate(farmId=farm_id, queueId=queue_id, jobId=job_id)
+            for s in page.get("sessions", [])
+        ]
+
+        if not sessions:
+            raise DeadlineOperationError(f"No sessions found for job {job_id}")
```
crowecawcaw: Could we consolidate some of this logic with the command line equivalent? https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/job_group.py#L1161

src/deadline/_mcp/tools/job.py
```diff
@@ -196,3 +198,122 @@ def download_job_output(
...
+    start_time: Optional[str] = None,
+    end_time: Optional[str] = None,
+    next_token: Optional[str] = None,
+    timezone: Optional[str] = "utc",
+) -> Dict[str, Any]:
+    """
+    Get CloudWatch logs for a specific session.
+
+    If session_id is not provided, auto-selects the most recent session
+    (prioritizing ongoing sessions over completed ones).
```
crowecawcaw, line 232: This is a bit confusing with the get_session_logs function. Should we drop get_session_logs if this does the same thing?

Also could we clarify the doc string? "Get CloudWatch logs for a specific session" does not align with the get_job_logs function/tool name. Maybe something like "Get latest logs for a job. Defaults to the latest session by default, or specify a specific session to see its logs."

src/deadline/_mcp/tools/job.py
```diff
@@ -196,3 +202,102 @@ def download_job_output(
         "total_time_seconds": round(total_time, 1),
         "output": output_text.strip(),
     }
+
+
+def _parse_time(time_str: Optional[str]) -> Optional[datetime.datetime]:
+    return datetime.datetime.fromisoformat(time_str.replace("Z", "+00:00")) if time_str else None
```
AWS-Samuel, line 208: what does a parsing error here look like? Is it possible?

src/deadline/_mcp/tools/job.py
```diff
@@ -196,3 +202,102 @@ def download_job_output(
...
+    # Auto-select session if not provided (job_id is guaranteed to exist here)
+    if not session_id:
+        result = list_sessions_for_job(
+            config=config, farmId=farm_id, queueId=queue_id, jobId=job_id
+        )
+        sessions = result["sessions"]
+
+        # Select the most recent session (prefer ongoing over completed)
+        sessions.sort(key=get_session_sort_key)
+        session_id = sessions[0]["sessionId"]
```
AWS-Samuel, line 270: Is there guaranteed to be at least one session?

src/deadline/_mcp/tools/job.py
```diff
@@ -196,3 +202,102 @@ def download_job_output(
...
+
+    # Auto-select session if not provided (job_id is guaranteed to exist here)
+    if not session_id:
+        result = list_sessions_for_job(
+            config=config, farmId=farm_id, queueId=queue_id, jobId=job_id
+        )
+        sessions = result["sessions"]
+
+        # Select the most recent session (prefer ongoing over completed)
+        sessions.sort(key=get_session_sort_key)
```
AWS-Samuel, line 269: using `min()` here would be preferable to avoid an unnecessary sort.

src/deadline/_mcp/tools/job.py
kavmur: Possible but unlikely as the format is in doc strings and the chances that agent would miss the tool metadata is low.

src/deadline/_mcp/tools/job.py
kavmur: Thanks, I will check if there is atleast 1 session.

src/deadline/_mcp/tools/job.py
AWS-Samuel: We should confirm that whatever error does pop up is easily parsable by an agent if it does miss the format so that recovery is quick

PR #919, opened 2025-11-12
feat: report paths of downloaded files in json summary

crowecawcaw:
DCM uses the `deadline job download-output --output json` command to download job outputs. Currently the notification just says, "Downloaded 3 files" but I'd like to know which files were downloaded and where.

As the first step, report the downloaded file paths in the JSON report.

Enables programmatic parsing of download file paths from the `download-output` command.

Unit tests and manual testing.

No, not needed.

No

No

No
----

### DIFF
--- a/src/deadline/client/cli/_groups/job_group.py
+++ b/src/deadline/client/cli/_groups/job_group.py
@@ -789,7 +789,10 @@ def _get_download_summary_message(
         return _get_json_line(
             JSON_MSG_TYPE_SUMMARY,
             f"Downloaded {download_summary.processed_files} files",
-            extra_properties={"fileCount": download_summary.processed_files},
+            extra_properties={
+                "fileCount": download_summary.processed_files,
+                "files": download_summary.downloaded_files,
+            },
         )
     else:
         paths_joined = "\n        ".join(
--- a/src/deadline/job_attachments/progress_tracker.py
+++ b/src/deadline/job_attachments/progress_tracker.py
@@ -8,7 +8,7 @@ import time
 from dataclasses import asdict, dataclass, field
 from enum import Enum
 from threading import Lock
-from typing import Callable, Dict, Optional, Union
+from typing import Callable, Dict, List, Optional, Union
 
 from ._path_summarization import human_readable_file_size
 
@@ -76,6 +76,7 @@ class DownloadSummaryStatistics(SummaryStatistics):
     """
 
     file_counts_by_root_directory: Dict[str, int] = field(default_factory=dict)
+    downloaded_files: List[str] = field(default_factory=list)
 
     def aggregate(self, other: SummaryStatistics) -> SummaryStatistics:
         """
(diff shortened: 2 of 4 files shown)


src/deadline/client/cli/_groups/job_group.py
```diff
@@ -789,7 +789,10 @@ def _get_download_summary_message(
         return _get_json_line(
             JSON_MSG_TYPE_SUMMARY,
             f"Downloaded {download_summary.processed_files} files",
-            extra_properties={"fileCount": download_summary.processed_files},
+            extra_properties={
+                "fileCount": download_summary.processed_files,
+                "files": download_summary.downloaded_files,
```
mwiebe, line 794: How many files have you tested this with? Can you test with 100,000 files, does it behave reasonably well?

leongdl (dismissed):

src/deadline/client/cli/_groups/job_group.py
leongdl: There's probably a print line limit too to console?

epmog: +1, I know there was some work put into the summarization of job attachments, maybe that's what we should be doing here as well?

src/deadline/job_attachments/progress_tracker.py
```diff
@@ -76,6 +76,7 @@ class DownloadSummaryStatistics(SummaryStatistics):
     """
 
     file_counts_by_root_directory: Dict[str, int] = field(default_factory=dict)
+    downloaded_files: List[str] = field(default_factory=list)
```
leongdl, line 79: Just curious if we should add any details for path mapping for different roots. EG; submitter was on workstation A, while downloader is on workstation B.

Can be different OSes etc.

crowecawcaw: My motivation for adding this was so in DCM we can show where files were downloaded to. We don't need path mapping rules for that, and the non-JSON output doesn't include it.

Do you think we'd need to restructure this files list to account for it? If not, we could leave the path mapping details for later.

src/deadline/client/cli/_groups/job_group.py
crowecawcaw: Tested with 100k files and it worked fine. For my test case of 100k files, it generated about 5MB of JSON - not unreasonable. The primary use case of JSON output is for programmatic parsing (e.g. DCM), but terminals should also be able to handle MBs of data without issue.

src/deadline/job_attachments/progress_tracker.py
mwiebe: This list is always the local file paths, right? I don't think we'd want to change that. In the auto-download implementation, we print a log of the path mapping rules that are applied - that's probably a nice thing to add in a future addition.

mwiebe (dismissed):

leongdl (approved):

mwiebe (approved):

PR #920, opened 2025-11-14
fix: Job progress dialog no longer moves to back.

### LINKED ISSUE #260: Maintenance: UI Submission Progress Window Hidden Behind Submission Widget
### Description

When submitting a job (with a warning window):

1. User clicks submit on the main submission widget
2. User clicks 'OK' on the warning window
3. The submission progress window appears but is positioned behind the submission widget

This might creates confusion as users cannot see the submission progress and may be uncertain if their job was successfully submitted.

Image shows submission widget in front of progress window

Image shows after moving submission widget to the side, the progress window shows up

### Solution

Submission progress window should appear on top of the submission widget.

erico-aws:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-blender/issues/260

After the warning dialog is closed, the progress dialog moved behind the submitter dialog. 

I believe the issue was happening because of an interaction between an unparented modal dialog (the warning dialog) and the non-modal progress dialog. I've set the parent of the warning dialog to be the progress dialog and set the progress dialog to be modal. It should have always been modal because we don't want a user to be changing settings while a job is being submitted.   

ProgressStaysInFront

The progress dialog doesn't jump to the back after the warning dialog is shown.

I deployed these changes to my local install of the Blender submitter. 

No

No

No

----


jusiskin:
Is the parent submission form dialog modal? Does parenting the job submission progress dialog make it modal?

My concern is whether this prevents the user from continuing their work inside the content creation application while the submission is in progress? Submissions can take a long time depending on the number/size of job attachments and the network connection speed.

erico-aws:
> Is the parent submission form dialog modal? Does parenting the job submission progress dialog make it modal?
> 
> My concern is whether this prevents the user from continuing their work inside the content creation application while the submission is in progress? Submissions can take a long time depending on the number/size of job attachments and the network connection speed.

I checked the behaviour of the progress dialog without my changes. It won't let me minimize it or interact with Blender until I close the dialog. I think that's the behaviour we want. I feel like we don't want folks to change their scene while a submission is taking place.

miabatta (approved):
It sounds like this dialog blocks users from interacting with Blender with and without these changes. I am approving in that case, if we receive feedback about this behaviour we can revisit.

jusiskin (approved):

PR #924, opened 2025-11-19
perf: avoid S3 HEAD requests when downloading job outputs

crowecawcaw:
Job attachments uses the s3transfer library to download files from S3. When it estimates whether to use multipart downloads or a single download, it makes a HEAD request to get the file size. Job attachments already knows the file size though from the manifest. For small files, the HEAD requests take about half the total download time.

Provide the file size to s3transfer.

Roughly halves the transfer time for small files.

I created a job that generates 10k small files. It took 3m52s to download before this change and 2m04s after.

n/a

No

No

No
----

### DIFF
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -21,7 +21,7 @@ import boto3
 from boto3.s3.transfer import ProgressCallbackInvoker
 from botocore.client import BaseClient
 from botocore.exceptions import BotoCoreError, ClientError
-from s3transfer.subscribers import BaseSubscriber
+from s3transfer.subscribers import BaseSubscriber as _BaseSubscriber
 
 from .asset_manifests.base_manifest import BaseAssetManifest, BaseManifestPath as RelativeFilePath
 from .asset_manifests.hash_algorithms import HashAlgorithm
@@ -79,7 +79,7 @@ from threading import Lock
 download_logger = getLogger("deadline.job_attachments.download")
 
 
-class _FileSizeSubscriber(BaseSubscriber):
+class _FileSizeSubscriber(_BaseSubscriber):
     """Subscriber that provides file size to skip HEAD requests."""
 
     def __init__(self, size):
--- a/src/deadline/job_attachments/download.py
+++ b/src/deadline/job_attachments/download.py
@@ -21,6 +21,7 @@ import boto3
 from boto3.s3.transfer import ProgressCallbackInvoker
 from botocore.client import BaseClient
 from botocore.exceptions import BotoCoreError, ClientError
+from s3transfer.subscribers import BaseSubscriber
 
 from .asset_manifests.base_manifest import BaseAssetManifest, BaseManifestPath as RelativeFilePath
 from .asset_manifests.hash_algorithms import HashAlgorithm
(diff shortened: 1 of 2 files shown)


src/deadline/job_attachments/download.py
```diff
@@ -77,6 +78,22 @@
...
+
+    def __init__(self, size):
+        self._size = size
+
+    def on_queued(self, future, **kwargs):
+        future.meta.provide_transfer_size(self._size)
+        # Provide a dummy etag to skip HEAD request if the method exists (added in s3transfer 0.6.0).
+        # For downloads from CAS, we don't need etag validation since files are content-addressed.
+        # Older s3transfer versions don't have this method, so we check before calling.
+        if hasattr(future.meta, "provide_object_etag"):
```
baxeaz, line 93: Is there a way to just check this once and cache it?  (Or just do a version check of some sort) just to avoid looking it up for every on_queued where it shouldn't be changing?

src/deadline/job_attachments/download.py
```diff
@@ -77,6 +78,22 @@
...
+            future.meta.provide_object_etag("dummy-etag")
```
jericht, line 94: Setting this value skips HeadObject requests? Can we link to docs for reference in the code comment above?

test/unit/deadline_job_attachments/test_download.py
```diff
@@ -1784,7 +1784,7 @@ def test_download_file_error_message_on_access_denied(self):
         s3_client = boto3.client("s3")
         stubber = Stubber(s3_client)
         stubber.add_client_error(
-            "head_object",
+            "get_object",
```
leongdl, line 1787: Nit - is this due to small size at like 1797?

src/deadline/job_attachments/download.py
```diff
@@ -77,6 +78,22 @@
 
 download_logger = getLogger("deadline.job_attachments.download")
 
+
+class _FileSizeSubscriber(_BaseSubscriber):
```
leongdl, line 82: Can we please run the integration tests?  This code path does change both deadline job download, incremental downloads. 

I need to check how this impacts the worker agent too, if its in the call path we need to run the integration tests there.

crowecawcaw: I ran the integration tests, they passed.

test/unit/deadline_job_attachments/test_download.py
crowecawcaw: s3transfer calls HEAD to get the file size to determine whether it's small enough to get in a signle request or if it should use its multirequest approach. Since we provide the size, it never makes a HEAD request. The first request that fails when permissions are missing is the GET.

src/deadline/job_attachments/download.py
crowecawcaw: The s3transfer library actually does not have documentation: https://github.com/boto/s3transfer/

This behavior was discovered in the Python version matrix testing. In lieu of documentation, this change has unit tests which verify the behavior we want (i.e. not calling HEAD).

src/deadline/job_attachments/download.py
crowecawcaw: Each future is a distinct instance, though surely if one has this attribute, they all would. Do you think this check would be slow enough to warrant optimization though? There will be 1 future per file.

baxeaz: Yeah I was wondering that as well.  It might not be terribly noticeable tbh?  It seems a bit inefficient when every future will be the same but maybe not enough to really care about a hash/lookup per file

src/deadline/job_attachments/download.py
leongdl: One concern from the team is this seems to be an internal implementation detail of S3 transfer. I appreciate the improvements, but also need to be careful end to end (JA, worker running jobs) to avoid another regression. 

JA is a bit fragile as you have noticed so we want to check everything first.

leongdl: Can we do what was implemented for Incremental downloads here: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/_incremental_downloads/_manifest_s3_downloads.py#L533-L537

It will not depend on an internal API.

crowecawcaw: Sure, that would work. It adds some complexity of having two download paths and two download managers (s3transfer and an internal thread pool), but it does avoid touching s3transfer details. If the JA team prefers that, I can close this PR.

crowecawcaw:
Closing this PR in favor of one on s3transfer itself: https://github.com/boto/s3transfer/pull/363

PR #927, opened 2025-11-19
chore: fix flaky unit test

crowecawcaw:
One of the unit tests is failing because the command output has a different space character than expected. The test fails on the current mainline code. I could not find a root cause in a reasonable amount of time. Could be a dependency update.

The test assertion was unnecessarily pedantic to begin with though.

Compare the JSON data of the command result instead of the particular string. The command under test is outputting JSON for programmatic parsing so whitespace doesn't matter.

Fixes the test.

Running the test

n/a

No

No

No
----

### DIFF
--- a/test/unit/deadline_client/cli/test_cli_job.py
+++ b/test/unit/deadline_client/cli/test_cli_job.py
@@ -808,35 +808,43 @@ def test_cli_job_download_output_stdout_with_json_format(
             session=ANY,
         )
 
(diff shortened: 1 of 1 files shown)


sakshie95 (approved):

evanspearman-a:
I was working a fix to the same problem, but took a slightly different approach. I like your fix better, so I will close mine and approve this.

evanspearman-a (approved):

PR #931, opened 2025-11-24
Add Error Handling Decorator to Telemetry Client Functions

dwrench07:

Telemetry functions in the client could potentially throw exceptions that would affect the
main application flow. The telemetry system should be resilient and never interfere with 
core functionality, even when telemetry operations fail.

Added a @telemetry_safe decorator that wraps telemetry functions with exception handling. 
The decorator catches all exceptions, logs them at debug level, and returns None instead 
of propagating the error. Applied this decorator to key TelemetryClient methods including 
set_opt_out, initialize, _get_prefixed_endpoint, _get_telemetry_identifier, _start_threads
, _get_system_metadata, and _exit_cleanly. Also removed redundant try-catch blocks in the 
initialize method since the decorator now handles error suppression.

This change makes the telemetry system more robust by ensuring telemetry failures don't 
crash or interrupt the main application. Users will experience more reliable operation 
even when telemetry encounters issues. Failed telemetry operations are logged for 
debugging purposes but don't affect core functionality.

See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainline/
DEVELOPMENT.md#testing) for information on running tests.

• [x ] Have you run the unit tests?
• [ x] Have you run the integration tests?
• [ ] Have you made changes to the download or asset_sync modules? If so, then it is 
highly recommended

• [] Are relevant docstrings in the code base updated?
• [ ] Has the README.md been updated? If you modified CLI arguments, for instance.

• [ ] This PR adds one or more new dependency Python packages. I acknowledge I have 
reviewed the considerations for adding dependencies in [DEVELOPMENT.md](https://github.com
/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies).
• [x] This PR does not add any new dependencies.

No, this is not a breaking change. The change only affects internal error handling in 
telemetry functions and doesn't modify any public APIs or contracts.

No, this change does not impact security. It only adds error handling to telemetry 
functions and doesn't create, modify, or access any files or directories with security 
implications.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

By submitting this pull request, I confirm that you can use, modify, copy, and 
redistribute this contribution, under the terms of your choice.

### DIFF
--- a/src/deadline/client/api/_telemetry.py
+++ b/src/deadline/client/api/_telemetry.py
@@ -40,6 +40,21 @@ logger = logging.getLogger(__name__)
 F = TypeVar("F", bound=Callable[..., Any])
 
 
+def telemetry_safe(func: F) -> F:
+    """
+    Decorator that catches all exceptions in telemetry functions to prevent
+    telemetry issues from affecting the main application flow.
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:
+            logger.debug(f"Telemetry function {func.__name__} failed: {str(e)}")
+            return None
+    return cast(F, wrapper)
+
+
 def get_deadline_endpoint_url(
     config: Optional[ConfigParser] = None,
 ) -> str:
@@ -110,6 +125,7 @@ class TelemetryClient:
         self.set_opt_out(config=config)
         self.initialize(config=config)
 
+    @telemetry_safe
     def set_opt_out(self, config: Optional[ConfigParser] = None) -> None:
         """
         Checks whether telemetry has been opted out by checking the DEADLINE_CLOUD_TELEMETRY_OPT_OUT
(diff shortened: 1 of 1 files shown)


ttblanchard (approved):

AWS-Samuel (changes requested):
Overall looks good. Two things

1. You'll want to sign your commit with `git commit -s` so that the `DCO` check passes
2. There's a few functions which don't have the decorator, that I think we want it on
    * each of the `record_...()` functions
    * `_record_summary_statistics()`
    * `_put_telemetry_record()`
    * `record_event()`
    * `get_account_id()`
    * `update_common_details()`

dwrench07:
Thanks for the comment. I will fix it.

PR #933, opened 2025-11-27
ci: specify permissions that workflows pass to jobs/actions

moorec-aws:
We need to pass the specific permissions required by the re-usable workflows

Add the permissions required by to run the re-usable workflows.

https://github.com/aws-deadline/.github/tree/mainline/.github/workflows

Permissions are scoped to the requirement

N/A - Will be confirmed after merge

N/A

No

No

----

### DIFF
--- a/.github/workflows/build_installers.yml
+++ b/.github/workflows/build_installers.yml
@@ -32,10 +32,10 @@ jobs:
     if: (github.repository == 'aws-deadline/deadline-cloud')
     name: Build Installer
     uses: aws-deadline/.github/.github/workflows/reusable_build_installers.yml@mainline
-    secrets: inherit
     permissions:
-      id-token: write
       contents: read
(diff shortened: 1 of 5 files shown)


.github/workflows/build_installers.yml
```diff
@@ -32,6 +32,9 @@ jobs:
     if: (github.repository == 'aws-deadline/deadline-cloud')
     name: Build Installer
     uses: aws-deadline/.github/.github/workflows/reusable_build_installers.yml@mainline
+    permissions:
+      contents: read
+      id-token: write
```
epmog, line 37: already have this 2 lines down

moorec-aws: Fixed

leongdl (approved):

epmog (approved):

PR #934, opened 2025-11-28
perf(installer): reduce installation size by not duplicating files

epmog:

PyInstaller 6 makes use of symlinks on Unix (in our case mostly macOS here). However, our installation from the installer was swapping the symlink with a copy of the file it pointed to and kept the target file as well. Thus duplicating files.

`shutil.make_archive` does not respect symlinks, while `shutil.unpack_archive` appears to sometimes? In this case, we manually create the zip with `zipfile` and unpack with `unzip`.

Symlinks are respected as the build artifacts make their way though the pipeline.

Deployed these changes and built the installer myself. Confirmed that the installed artifact no longer contained duplicate files.

N/A

N/A

N/A

N/A

----


scripts/pyinstaller/make_exe.py
```diff
@@ -89,6 +90,37 @@ def clean_pyinstaller_build_dirs():
...
+            # Add files
+            for file_name in files:
+                file_path = os.path.join(root, file_name)
+                arcname = os.path.relpath(file_path, source_dir)
+
+                if os.path.islink(file_path):
+                    # Create a ZipInfo for the symlink
+                    info = zipfile.ZipInfo(arcname)
+                    info.create_system = 3  # Unix
+                    info.external_attr = 0o120777 << 16  # symlink file type
```
moorec-aws: Looked briefly at the permissions here, 755 is usually best practice, but symlink permissions are ignored on Unix/Mac - the target file's permissions are used instead. Do you know what happens for Windows?

scripts/pyinstaller/allowlist.py
```diff
@@ -25,7 +25,6 @@
         "deadline",
         "deadline.exe",
         "_internal/THIRD_PARTY_LICENSES",
-        "_internal/Python.framework/Python",
```
moorec-aws, line 28: This change is because pyinstaller already bundles everything we need?

epmog: Yea, nothing actually referenced this one. It's symlinked further in

scripts/pyinstaller/make_exe.py
epmog: Great question, currently pyinstaller doesn't leverage symlinks for windows so this isn't actually hit on that code path. I'll take a look regardless so that we're protected in case it changes.

moorec-aws (dismissed):

scripts/pyinstaller/make_exe.py
```diff
@@ -89,6 +93,37 @@ def clean_pyinstaller_build_dirs():
...
+        for root, dirs, files in os.walk(source_dir):
+            # Add directories
+            for dir_name in dirs:
+                dir_path = os.path.join(root, dir_name)
+                if os.path.islink(dir_path):
+                    # Handle symlinked directories
+                    arcname = os.path.relpath(dir_path, source_dir)
+                    info = zipfile.ZipInfo(arcname + "/")
+                    info.create_system = 3  # Unix
+                    info.external_attr = 0o120777 << 16  # symlink file type
```
jusiskin: This Python docs do not explain these very well:

*   [`create_system`](https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.create_system)

    >   System which created ZIP archive.
*   [`external_attr`](https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.external_attr)

    >   External file attributes.

Do you have any other documentation that we can link to that helps document the values here?

I see the Python docs link to https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT which do document a "version made by"

        4.4.2.2 The current mappings are:

         0 - MS-DOS and OS/2 (FAT / VFAT / FAT32 file systems)
         1 - Amiga                     2 - OpenVMS
         3 - UNIX                      4 - VM/CMS
         5 - Atari ST                  6 - OS/2 H.P.F.S.
         7 - Macintosh                 8 - Z-System
         9 - CP/M                     10 - Windows NTFS
        11 - MVS (OS/390 - Z/OS)      12 - VSE
        13 - Acorn Risc               14 - VFAT
        15 - alternate MVS            16 - BeOS
        17 - Tandem                   18 - OS/400
        19 - OS X (Darwin)            20 thru 255 - unused

Should we use `create_system = 19` for MacOS?

For the `external_attr` I presume this is the `st_mode` value returned from a `stat()` syscall. I've found [this documentation](https://man7.org/linux/man-pages/man7/inode.7.html) for the `S_IFLNK` constant:

>   S_IFLNK    0120000   symbolic link

I wonder if we should be using Python's [`stat.S_IFLNK`](https://docs.python.org/3/library/stat.html#stat.S_IFLNK) for this? Or maybe just use `stat()` and use the `st_mode` attribute to preserve the symlink bit and permission bits.

scripts/pyinstaller/make_exe.py
```diff
@@ -89,6 +93,37 @@ def clean_pyinstaller_build_dirs():
...
 
+def zip_with_symlinks(source_dir: Path, output_zip: Path) -> None:
+    """Create a zip archive that preserves symlinks."""
+    output_zip.parent.mkdir(parents=True, exist_ok=True)
+    with zipfile.ZipFile(output_zip, "w", zipfile.ZIP_DEFLATED) as zipf:
+        for root, dirs, files in os.walk(source_dir):
+            # Add directories
+            for dir_name in dirs:
+                dir_path = os.path.join(root, dir_name)
+                if os.path.islink(dir_path):
```
jusiskin: Do we need an `else` for this?

I see there is a [`ZipFile.mkdir(zinfo_or_directory, mode=511)`](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.mkdir). Based on your testing, I'm assuming the directory is created implicitly, but I suppose this would preserve the permission bits on UNIX systems.

jusiskin (changes requested):

scripts/pyinstaller/make_exe.py
epmog: This probably actually isn't tested all that well, since the builds we use don't actually create symlinked folders. We only have symlinked files.

scripts/pyinstaller/make_exe.py
epmog: Python itself only cares about 0 or 3:

https://github.com/python/cpython/blob/3.14/Lib/zipfile/__init__.py#L468-L472

100% I can make use of stat though, since we will need the `777` for symlinks (it's the target file permissions that matter)

scripts/pyinstaller/make_exe.py
epmog: Windows no longer calls this code and just makes use of shutil since it currently suits our needs

scripts/pyinstaller/make_exe.py
epmog: ZipInfo handles the create, and we can rely on lstat for getting the existing permissions on the symlink (which don't actually matter). This should be much cleaner now.

scripts/pyinstaller/make_exe.py
epmog: In regards to your original question we don't, since as you mentioned any file written will have its directory created.

moorec-aws (dismissed):

scripts/pyinstaller/make_exe.py
```diff
@@ -89,6 +94,25 @@ def clean_pyinstaller_build_dirs():
...
+    with zipfile.ZipFile(output_zip, "w", zipfile.ZIP_DEFLATED) as zipf:
+        for root, dirs, files in os.walk(source_dir):
+            for name in dirs + files:
+                path = os.path.join(root, name)
+                arcname = os.path.relpath(path, source_dir)
+
+                if os.path.islink(path):
+                    info = zipfile.ZipInfo(arcname)
+                    stat_info = os.lstat(path)
+                    info.external_attr = stat_info.st_mode << 16
```
jusiskin: Do we need to mask before bit shifting like this:

https://github.com/python/cpython/blob/9bb27cd1051e0e1ee2022fcaedb828643a84a30e/Lib/zipfile/__init__.py#L648
```suggestion
                    info.external_attr = (stat_info.st_mode & 0xFFFF) << 16
```

jusiskin: From offline discussions, we realized that this would mask out the `S_IFLINK` bit (`0o120000`) which would defeat the purpose of this PR.

jusiskin (approved):

scripts/pyinstaller/make_exe.py
epmog: correction it won't mask the file type bits. I've added it in to match cpython. Building the installer still shows the correct symlink + permissions.

moorec-aws (approved):

PR #940, opened 2025-12-08
feat!: add an 'About' dialog to show version information

lucaseck:
Many users have requested an easier way to see what version of the submitter they are using. In the past it has also been challenging to see what versions of DCCs, deadline-cloud, the submitters, Python and Qt were being used.

* Added a new "About" dialog which will display all of the version information mentioned above
  * Added translations for all of the new buttons 
* Added ability for the submitters to pass in additional information like the DCC name and version and the submitter name, package and version along with anything additional that may be useful in the context of that submitter.
  * This can be done directly through the `SubmitJobToDeadlineDialog` or the CLI for submitters that just call `deadline bundle gui-submit` to open the window.
    * For the CLI I added a new `--submitter-info` option to pass in. I've left `--submitter-name` still separately, but with a new deprecation warning. If we remove that entirely then submitters like KeyShot and AE could break if the new deadline-cloud is installed on the system since those versions are very loosely coupled. Those submitters just expect deadline to be on the PATH. They will likely want to start checking the version of deadline installed and changing the command appropriately to pass in these new fields if the version of deadline is new enough.
* Changed the title bar to display the name and version of the submitter if provided so that it's extremely clear what version is being used

Example of the UI from some testing I did to hookup the Blender submitter to pass along the information to the UI:
blender
blender_submitter_about

* Easier to gather debug version information when diagnosing problems
* Users know more easily if they are on the latest submitters or not

  - Yes they all passed and I included new ones for the CLI parsing and validation
  - No

#### Manual Testing
* Cloned my branch on Linux(RHEL9), Windows and MacOS and launched the gui from the CLI while testing passing in the new `--submitter-info` parameter with mock details. I verified that on all OSes the button could be clicked and the dialog displayed correctly.
  * I verified that both the key value pair and inline JSON options worked for passing in the info from the CLI
  * Ran with both Python 3.8 and a newer Python on each OS
  * Tested with both PySide2 and PySide6 on each OS
* On Mac I modified the Blender submitter to use this new library version and pass along the info from Blender to populate the `SubmitterInfo` and pass it to the `SubmitJobToDeadlineDialog` which then displayed the information correctly in the dialog within Blender.

  - Yes 
  - No, but I don't see any existing references where this needs to be documented in the README. My understanding is the user guide will pull in the doc strings once this PR is merged and released.

**YES**, this is a breaking change. I renamed the submitter_name parameter in multiple functions to submitter_info and changed it to accept an instance of `SubmitterInfo` instead of just a string name. This was done to avoid adding multiple more related parameters separately to the function signatures and instead use the `SubmitterInfo` dataclass going forward to collect them together as once paramter.

No

----


src/deadline/client/ui/dataclasses/_environment_info.py
```diff
@@ -0,0 +1,129 @@
...
+            deadline_dep_versions["deadline"] = version
+        except Exception as e:
+            logger.debug(f"Failed to retrieve deadline-cloud version: {e}")
+            deadline_dep_versions["deadline"] = "Unknown"
+
+        # Get direct dependency versions from package metadata
+        try:
+            deadline_deps = requires("deadline")
+            if deadline_deps:
+                for dep in deadline_deps:
```
jusiskin: This looks like it will not consider transitive dependencies. Should we be including those too?

src/deadline/client/ui/dialogs/_about_dialog.py
```diff
@@ -0,0 +1,196 @@
...
+        Format submitter and version information for display.
+
+        Returns:
+            Formatted string with submitter and version information
+        """
+
+        formatted_dict: dict[str, Any] = {}
+        formatted_dict["System Information"] = {
+            "Operating System": f"{self.environment_info.os_version} ({self.environment_info.os_architecture})"
+        }
```
jusiskin: Should this behavior be a part of the `_EnvironmentInfo` class?

If the goal is to produce well-structured YAML, that would be business logic and we would want to test that behavior independent of the UI code.

Another consideration is if we should just use `dataclasses.asdict()` so the YAML matches our class definition. This would mean less code to maintain. The trade-off is that the human-readability would degrade since we have more human-friendly keys here.

Same feedback for the `submitter_info` below.

src/deadline/client/ui/dialogs/_about_dialog.py
```diff
@@ -0,0 +1,196 @@
...
+            submitter_dict = {}
+
+            if self.submitter_info.submitter_name:
+                submitter_dict["Submitter Name"] = self.submitter_info.submitter_name
+
+            if self.submitter_info.submitter_package_name:
+                if self.submitter_info.submitter_package_version:
+                    submitter_dict["Submitter Package"] = (
+                        f"{self.submitter_info.submitter_package_name} {self.submitter_info.submitter_package_version}"
+                    )
```
jusiskin: Should we keep the version separate? It seems we have chosen to structure this as YAML, so having the version easier to extract seems like a benefit to keeping it separate.

src/deadline/client/ui/dialogs/_about_dialog.py
```diff
@@ -0,0 +1,196 @@
...
+        self.setModal(True)
+        self.setMinimumWidth(500)
+
+        self._build_ui()
+
+    def _build_ui(self):
+        """Build the dialog UI layout."""
+        layout = QVBoxLayout()
+
+        self.version_text = QTextEdit()
```
jusiskin, line 71: nit: since the contents are YAML, how hard is it to add syntax highlighting?

I imagine some customers may provide screenshots instead of copying the text. Syntax highlighting would improve the visual legibility for such cases.

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -81,6 +83,85 @@ def validate_parameters(ctx, param, value):
...
+                    f'Submitter info must be in format "key=value" Invalid: {item}'
+                )
+
+            key, val = match.groups()
+            if key not in valid_fields:
+                raise click.BadParameter(
+                    f"Unknown field '{key}'. Valid fields are: {', '.join(sorted(valid_fields))}"
+                )
+
+            info_dict[key] = val
```
jusiskin: nit: we can simplify this a little, but totally optional:

```suggestion
            if '=' not in item:
                raise click.BadParameter(
                    f'Submitter info must be in format "key=value" Invalid: {item}'
                )

            key, val = item.split('=', maxsplit=1)
            if key not in valid_fields:
                raise click.BadParameter(
                    f"Unknown field '{key}'. Valid fields are: {', '.join(sorted(valid_fields))}"
                )

            info_dict[key] = val
```

Otherwise we will want to anchor the match:

```suggestion
            match = re.match(r"^([^=]+)=(.*)$", item)
            if not match:
                raise click.BadParameter(
                    f'Submitter info must be in format "key=value" Invalid: {item}'
                )

            key, val = match.groups()
            if key not in valid_fields:
                raise click.BadParameter(
                    f"Unknown field '{key}'. Valid fields are: {', '.join(sorted(valid_fields))}"
                )

            info_dict[key] = val
```

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -321,8 +402,7 @@ def _check_create_job_wait_canceled() -> bool:
 )
 @click.option(
     "--submitter-name",
-    type=click.STRING,
-    help="Name of the application submitting the bundle. If a name is specified, the GUI will automatically close after submitting the job.",
+    help="[DEPRECATED] Use --submitter-info submitter_name=<name> instead. The name of the submitter.",
 )
```
jusiskin, line 367: It appears that `click.option` has a `deprecated` flag ([docs](https://click.palletsprojects.com/en/stable/api/#click.Option)).

src/deadline/client/dataclasses/submitter_info.py
```diff
@@ -0,0 +1,54 @@
...
+    It's used by the _AboutDialog GUI to display version information.
+
+    Attributes:
+        submitter_name: Short name of the submitter (required, e.g., "Blender", "CLI")
+        submitter_package_name: Name of the submitter package using the library (optional, e.g., "deadline-cloud-for-blender")
+        submitter_package_version: Version of the submitter package (optional)
+        host_application_name: Name of the host application (optional, e.g., "Maya", "Blender")
+        host_application_version: Version of the host application (optional)
+        additional_info: Optional dictionary for arbitrary nested data that integrations can use
+                        to pass additional metadata (optional). Supports nested structures.
```
jusiskin: nit: can we add these as docstrings to the attributes below?

This improves the dev experience in supported IDEs that will show the help text. For example in VSCode when hovering over the attribute.

An example docstring for an attribute is:

```py
    submitter_name: str
    """Short name of the submitter (required, e.g., "Blender", "CLI")"""
```

src/deadline/client/dataclasses/submitter_info.py
```diff
@@ -0,0 +1,54 @@
...
+            }
+        )
+    """
+
+    submitter_name: str
+    submitter_package_name: Optional[str] = None
+    submitter_package_version: Optional[str] = None
+    host_application_name: Optional[str] = None
+    host_application_version: Optional[str] = None
+    additional_info: Optional[dict[str, Any]] = None
```
jusiskin: Do we want to constrain the types of the dict values? e.g `int` / ` float` / `decimal` / `str` / `bool`?

src/deadline/client/ui/dataclasses/_environment_info.py
```diff
@@ -0,0 +1,129 @@
...
+            python_version = platform.python_version()
+        except Exception as e:
+            logger.warning(f"Failed to retrieve Python version: {e}")
+            python_version = "Unknown"
+
+        # Collect OS name and version
+        try:
+            system = platform.system()
+            if system == "Darwin":
+                os_version = f"macOS {platform.mac_ver()[0]}"
```
jusiskin: nit: do we want these to be combined? It makes extracting each field slightly more difficult.

src/deadline/client/ui/dataclasses/_environment_info.py
```diff
@@ -0,0 +1,129 @@
...
+        except Exception as e:
+            logger.warning(f"Failed to retrieve OS name and version: {e}")
+            os_version = "Unknown"
+
+        # Collect OS architecture
+        try:
+            os_architecture = platform.machine()
+        except Exception as e:
+            logger.warning(f"Failed to retrieve OS architecture: {e}")
+            os_architecture = "Unknown"
```
jusiskin, line 132: I wonder if we want to use the same values as OpenJD here. See this code in the worker agent:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/c5b1c2d128d0015badf5d060bc7fed9f74920eee/src/deadline_worker_agent/capabilities.py#L58-L66

src/deadline/client/ui/dataclasses/_environment_info.py
```diff
@@ -0,0 +1,129 @@
...
+                    # Extract just the package name
+                    match = re.match(r"^([a-zA-Z0-9_-]+)", dep)
+                    if not match:
+                        continue
+                    package_name = match.group(1)
+
+                    try:
+                        deadline_dep_versions[package_name] = package_version(package_name)
+                    except Exception as e:
+                        logger.debug(f"Could not retrieve version for {package_name}: {e}")
```
miabatta: nit, does this need to be nested? I'm pretty confident this will behave fine (i.e., log messages for each unknown package version without logging the "Could not retrieve deadline dependencies" message) but I guess the whole block being wrapped in `try` makes me itch a little.

src/deadline/client/ui/dialogs/_about_dialog.py
```diff
@@ -0,0 +1,196 @@
...
+        Args:
+            submitter_info: SubmitterInfo object containing submitter details
+            parent: Parent widget (optional)
+        """
+        super().__init__(parent=parent)
+
+        self.submitter_info = submitter_info
+        self.environment_info = _EnvironmentInfo.collect()
+
+        self.setWindowTitle(tr("About"))
```
jusiskin: nit: maybe we can include the submitter name here if available?

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -385,6 +406,13 @@ def on_settings_button_clicked(self):
         if DeadlineConfigDialog.configure_settings(parent=self):
             self.refresh_deadline_settings()
 
+    def _on_about_button_clicked(self):
+        """Show the About dialog with submitter information."""
+        try:
+            _AboutDialog._show_about(self.submitter_info, parent=self)
+        except Exception as e:
+            logger.error(f"Failed to show About dialog: {e}")
```
jusiskin: It appears that `_AboutDialog._show_about` catches all exception so this code is unused.

src/deadline/client/ui/dialogs/_about_dialog.py
```diff
@@ -0,0 +1,196 @@
...
+        try:
+            dialog = _AboutDialog(submitter_info, parent=parent)
+            dialog.exec_()
+        except Exception as e:
+            logger.error(f"Failed to create AboutDialog: {e}")
+            QMessageBox.critical(
+                parent,
+                "Error",
+                f"Failed to display About dialog: {str(e)}",
+            )
```
jusiskin: Maybe we should just put this in the calling code?

It is marked as private (leading `_`) and there is only one caller in `SubmitJobToDeadlineDialog`.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -114,16 +116,32 @@ def __init__(
...
+        if submitter_info and submitter_info.submitter_package_name:
+            # Format package name: split on "-", capitalize each word
+            # e.g., "deadline-cloud-for-maya" -> "Deadline Cloud For Maya"
+            formatted_name = " ".join(
+                word.capitalize() for word in submitter_info.submitter_package_name.split("-")
+            )
+            if submitter_info.submitter_package_version:
+                window_title = f"{formatted_name} {submitter_info.submitter_package_version}"
+            else:
+                window_title = f"{formatted_name}"
```
jusiskin: How does this work with language translations? Is it the caller's responsibility to translate the `submitter_package_name` when constructing the `SubmitterInfo` instance?

jusiskin (changes requested):

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -114,16 +116,32 @@ def __init__(
...
         super().__init__(parent=parent, f=f)
-        self.setWindowTitle(tr("Submit to AWS Deadline Cloud"))
+
+        # Set window title with submitter package info if available
+        window_title = tr("Submit to AWS Deadline Cloud")
+        if submitter_info and submitter_info.submitter_package_name:
+            # Format package name: split on "-", capitalize each word
+            # e.g., "deadline-cloud-for-maya" -> "Deadline Cloud For Maya"
+            formatted_name = " ".join(
+                word.capitalize() for word in submitter_info.submitter_package_name.split("-")
```
miabatta: Nit that this will produce incorrect capitalization for certain integrations (e.g. KeyShot vs Keyshot, Cinema 4D vs Cinema 4d). I don't know if there is a good alternative for formatting the name but it is probably in our best interest to provide consumers of the library more customization of how to format the name.

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -81,6 +83,85 @@ def validate_parameters(ctx, param, value):
     return parameters_split
 
 
+def _validate_submitter_info(ctx, param, value):
+    """
+    Validate provided --submitter-info value and convert to SubmitterInfo object.
+
+    Supports two input formats (mutually exclusive):
```
miabatta: Is there a reason these have to be mutually exclusive? We do something similar in the [OpenJD CLI](https://github.com/OpenJobDescription/openjd-cli/blob/057e686711f6fe162ffdea04e5a01fa884cff9de/src/openjd/cli/_common/_job_from_template.py#L59) to check for parameters (test cases [[1](https://github.com/OpenJobDescription/openjd-cli/blob/057e686711f6fe162ffdea04e5a01fa884cff9de/test/openjd/cli/test_common.py#L209)] [[2](https://github.com/OpenJobDescription/openjd-cli/blob/057e686711f6fe162ffdea04e5a01fa884cff9de/test/openjd/cli/test_common.py#L315)]) that IIRC allows parameters to be specified in different types as long as the parameter names don't overlap.

I also think restructuring this validation a bit to follow the OJD code could avoid duplicating the key validation, which feels bad here. Of course that uses `argparse` and this is `click` but I don't think the usages are terribly different

src/deadline/client/ui/dialogs/_about_dialog.py
```diff
@@ -0,0 +1,196 @@
...
+                submitter_dict["Additional Information"] = self.submitter_info.additional_info
+
+            if submitter_dict:
+                formatted_dict["Submitter Information"] = submitter_dict
+
+        formatted_dict["Python Information"] = {
+            "Python Version": self.environment_info.python_version,
+            "Deadline Library Dependencies": self.environment_info.deadline_dep_versions,
+            "Qt Version": self.environment_info.qt_version,
+        }
```
miabatta: Is there a reason we can't translate these?

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -81,6 +83,85 @@ def validate_parameters(ctx, param, value):
...
+            raise click.BadParameter("additional_info must be an object")
+
+        info_dict = data
+
+    else:
+        # key=value format - check that none start with {
+        for item in value:
+            if item.startswith("{"):
+                raise click.BadParameter(
+                    "Cannot specify multiple inline JSON values or both inliune JSON and key=value pairs. "
```
crowecawcaw: Nit: inluine typo

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -81,6 +83,85 @@ def validate_parameters(ctx, param, value):
...
+    All keys must be valid SubmitterInfo fields. Unknown keys will raise an error.
+    """
+    if not value:
+        return None
+
+    # Get valid field names from SubmitterInfo dataclass
+    valid_fields = {field.name for field in fields(SubmitterInfo)}
+
+    # Check if using JSON format (only one value allowed and it must start with {)
+    if len(value) == 1 and value[0].startswith("{"):
```
crowecawcaw: Might be simpler to commit to one data format instead of accepting JSON and key-value pairs.

src/deadline/client/ui/dataclasses/_environment_info.py
lucaseck: I had purposefully left them out. In chatting with Morgan it sounded like the top level direct dependencies are what we care about most. Could add more, but my concern is they would be nothing but clutter at that point. Any dependency issues I've run into so far have always been one of these main culprits that we have updated or have a mismatched build of for the native dependencies.

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -321,8 +402,7 @@ def _check_create_job_wait_canceled() -> bool:
 )
 @click.option(
     "--submitter-name",
-    type=click.STRING,
-    help="Name of the application submitting the bundle. If a name is specified, the GUI will automatically close after submitting the job.",
+    help="[DEPRECATED] Use --submitter-info submitter_name=<name> instead. The name of the submitter.",
```
crowecawcaw: Why not keep `--submitter-name` and have `--additional-submitter-info`? It's backwards compatible and submitter name is treated differently than the other submitter info.

Not a blocking comment if you prefer deprecation.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
lucaseck: I would say yes. For our own submitters there are no translations we would make other than maybe the word `for` otherwise Deadline Cloud and DCC name aren't things we will translate at all.

test/unit/deadline_client/cli/test_cli_bundle_submit.py
```diff
@@ -755,3 +756,307 @@ def test_bundle_submit_without_target_task_run_status(
     deadline_mock.create_job.assert_called_once()
     _, kwargs = deadline_mock.create_job.call_args
     assert "targetTaskRunStatus" not in kwargs
+
+
+@patch.object(deadline.client.ui, "gui_context_for_cli")
+def test_bundle_gui_submit_submitter_info_key_value_pairs(_mock_context):
```
crowecawcaw, line 762: Could use some unit tests for how the titles are formatted. E.g. with a version number, without a version number, etc.

src/deadline/client/cli/_groups/bundle_group.py
lucaseck: Good catch, hadn't seen that but will switch to use it.

src/deadline/client/dataclasses/submitter_info.py
lucaseck: The idea here was that submitters can continue to specify nested attributes multiple layers deep for whatever information they might want to display. e.g. loaded plugins or anything else they want to keep structuring in a nested fashion. That's why I had used Any there.

src/deadline/client/cli/_groups/bundle_group.py
lucaseck: It would be simpler, but I think it's worthwhile to support both and possibly add the ability to take in a file as well. Inline JSON is a pain to write by hand from the CLI, but way easier to do through code like will need to be done for KeyShot and AE. The key value pairs are much easier to write by hand. To Mia's point above this is what we do already in the openjd-cli: https://github.com/OpenJobDescription/openjd-cli/blob/057e686711f6fe162ffdea04e5a01fa884cff9de/src/openjd/cli/_common/_job_from_template.py#L59

src/deadline/client/cli/_groups/bundle_group.py
lucaseck: I prefer deprecation personally and to keep this related information together. We in the past have just kept adding more and more parameters and tacking things on as needed and I like the idea of collecting some of this related info together and giving it a separate interface. Unfortunately this deprecated option in the CLI will live on for some time since the existing submitters that call bundle gui-submit don't check deadline versions so removing it entirely could break old submitters when using new deadline-cloud versions.

src/deadline/client/ui/dataclasses/_environment_info.py
lucaseck: I can split up the try blocks, but they do need to be separate. Without the inner one any single failure to parse a dependencies version will stop it from collecting for the remaining ones. e.g. when it tries on Mac to collect pywin that will raise an exception and be caught, but the versions of the next dependencies in the list will still be collected

src/deadline/client/ui/dataclasses/_environment_info.py
lucaseck: I was biasing for human readability here. Even though the details get dumped as yaml, the intent was always for them to be purely human readable since this only appears in the GUI context. That being said though, to future proof it I like the idea of making them separate fields and have done so.

src/deadline/client/ui/dataclasses/_environment_info.py
jusiskin: That's fair and this is something that is easy to change later. Happy to revisit this in a future PR

src/deadline/client/dataclasses/submitter_info.py
jusiskin: I suppose the concern is that the code tries to encode this as YAML and not all objects can necessarily be converted.

Catching this in type-checking is better than at runtime.

src/deadline/client/ui/dialogs/_about_dialog.py
lucaseck: Cool idea! It looks like you can do this, but you have to roll it yourself. Qt provides a `QSyntaxHighlighter` class for doing this, but you have to define all of the regex expressions for the matching yourself: https://doc.qt.io/qtforpython-6/examples/example_widgets_richtext_syntaxhighlighter.html. So for this current change I'm going to say out of scope

src/deadline/client/ui/dialogs/_about_dialog.py
lucaseck: Spoke offline, but agreed for this first pass we will just use asdict and forgo the formatting to keep things more simple. Formatting can be added later as the main intention of this information is to be copy/pasted and sent to us when debugging. YAML will still be used for now because of the additional_info and ease of using yaml to print the possible nested structure of that.

src/deadline/client/ui/dialogs/_about_dialog.py
lucaseck: Will be covered by the move to using asdict, this formatting will be removed.

src/deadline/client/ui/dialogs/_about_dialog.py
lucaseck: This won't be a problem for this PR after the next revision as I'm removing all of the formatting for this first pass. To answer your question still though it's a judgement call. If the intent of this information is for it to be given to us to help debug, then we would probably want it all untranslated so that we can understand what we are being given more easily. The downside is then for the user it might not be as easy to read.

src/deadline/client/cli/_groups/bundle_group.py
crowecawcaw: Is there a use case for someone writing a CLI command by hand that has the submitter info? Seems like it'd be used exclusively for when the CLI is called from code where JSON is the easier format.

src/deadline/client/cli/_groups/bundle_group.py
lucaseck: Ended up taking a suggestion from a comment Mia had https://github.com/aws-deadline/deadline-cloud/pull/940#discussion_r2603869511. Copied the logic we use in the openjd CLI to clean things up a bit and now leaves the option in the future to add the same for parsing the job parameters.

src/deadline/client/dataclasses/submitter_info.py
lucaseck: Added a custom type hint to check for valid YAML types

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
lucaseck: Correct, took your other suggestion and moved the launching directly into here now so no longer an issue.

src/deadline/client/ui/dialogs/_about_dialog.py
lucaseck: Moved it over, for our other dialogs this is the way we do it, but given this is private and only intended to be used here right now I agree with the move.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
lucaseck: Changed to instead format as `Deadline Cloud <submitter_name> Submitter <submitter_version>` That puts the formatting of the submitter_name directly in the integrations hands.

src/deadline/client/cli/_groups/bundle_group.py
lucaseck: Great suggestion, I've implemented it similarly here now as well.

test/unit/deadline_client/cli/test_cli_bundle_submit.py
lucaseck: I did simplify the naming logic slightly so I'm not sure if there's much value in testing two if statements. As well, we don't have unit tests for any of these dialogs right now I believe because of how much you have to mock out to have a chance at testing anything because of all the Qt hookups.

src/deadline/client/cli/_groups/bundle_group.py
lucaseck: Testing is the main one that comes to mind. I did end up implementing a suggestion that Mia had to make this logic match what we do in the OpenJD CLI: https://github.com/aws-deadline/deadline-cloud/pull/940#discussion_r2603869511

The job parameter validation from the CLI can make use of this in the future as well.

miabatta (dismissed):

miabatta (dismissed):

src/deadline/client/cli/_common.py
```diff
@@ -194,6 +202,109 @@ def _cli_object_repr(obj: Any) -> str:
     return deadline_yaml_dump(obj)
 
 
+def _parse_file_parameter(file_path: Path) -> dict:
```
jusiskin, line 205: We should probably have some unit tests for these new functions (`_parse_file_parameter` and `_parse_multi_format_parameters`). What do you think?

src/deadline/client/ui/dialogs/_about_dialog.py
jusiskin: Sounds good 👍

src/deadline/client/cli/_groups/bundle_group.py
jusiskin: Discussed offline, but this is only available in click >= 8.2 which does not support Python 3.8

test/unit/deadline_client/cli/test_cli_bundle_submit.py
```diff
@@ -755,3 +756,603 @@ def test_bundle_submit_without_target_task_run_status(
...
+    """
+    Test that --submitter-info works with multiple key=value pairs.
+    """
+    mock_job_bundle_submitter = Mock()
+    sys.modules["deadline.client.ui.job_bundle_submitter"] = mock_job_bundle_submitter
```
jusiskin, line 767: I don't see the test code cleaning up this mock. Best-practice is to use:

1.  [`unittest.mock.patch(...)`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch)
2.  pytest's [`monkeypatch` fixture](https://docs.pytest.org/en/stable/how-to/monkeypatch.html)

to guarantee the mock gets removed after the test, otherwise there could be side-effects that impact tests that run after.

same feedback applies to the other new tests

jusiskin (changes requested):

test/unit/deadline_client/cli/test_cli_bundle_submit.py
lucaseck: I looked into this a little bit, this is the same mocking as is already done in the test file and one of the tests has a comment:
```
 # Unconventional mocking pattern because of how the function is imported in code
    mock_job_bundle_submitter = Mock()
    sys.modules["deadline.client.ui.job_bundle_submitter"] = mock_job_bundle_submitter
    mock_job_bundle_submitter.show_job_bundle_submitter
```
I believe what that's referring to is how `show_job_bundle_submitter` is imported only after the `gui_context_for_cli` is setup [here](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/bundle_group.py#L362). The tests all have to patch out `gui_context_for_cli` before the import occurs which possibly has to do with why we aren't able to patch it in a "normal" manner.

src/deadline/client/cli/_common.py
lucaseck: Good catch, I had moved those out to the common file after writing the initial tests for the overall validation in the bundle_group. I've added some new tests for those specific functions.

miabatta (approved):

test/unit/deadline_client/cli/test_cli_bundle_submit.py
jusiskin: Ah okay. It looks like a lazy import. In that case, we can have the fixture do:

```py
@pytest.fixture(autouse=True)
def mock_job_bundle_submitter() -> Generator[MagicMock, None, None]:
    mock_job_bundle_submitter = MagicMock()
    with patch.dict(sys.modules, {"deadline.client.ui.job_bundle_submitter": mock_job_bundle_submitter}):
        yield mock_job_bundle_submitter
```

I tested this locally and it works. I'm okay if we postpone this to a separate PR since there is precedent already in this file before your changes.

lucaseck: Lets do a separate PR. This change is already quite large and this mock style is used in 3 separate files for a lot of tests. Keeps the change smaller and then if there are any issues with changing the mock it's a self contained change to revert.

jusiskin (approved):

PR #945, opened 2025-12-11
fix: move bundle loading button when using gui-submit --browse

crowecawcaw:
When using `deadline bundle gui-submit --browse`, there's an awkward button under job specific settings to swap job bundles. It looks weird.

The option was added a couple years ago in this PR: https://github.com/aws-deadline/deadline-cloud/pull/35

Screenshot 2025-12-11 at 10 41 57 AM

Move the button to the bottom of the window next to the "Export bundle" button.

Screenshot 2025-12-11 at 12 51 39 PM

Cleaner bundle submitter.

Ran the command, saw the button was moved. Added unit tets

n/a

No

No

No

----


crowecawcaw:
API changes not actually API changes and safe to ignore. No one should be depending on these imports even if they were not marked as internally only.

leongdl (approved):
API change was detected, but I don't think it applies in this UI situation.

moorec-aws (approved):

PR #947, opened 2025-12-11
test: move bundle submit UI mocking to fixture

lucaseck:
In a previous [PR](https://github.com/aws-deadline/deadline-cloud/pull/940#discussion_r2608515182) it came up that new tests were being added and introducing more places were we were doing some weird mocking to get around a lazy import not being patchable the usual ways.

Moved into an autouse fixture for the test file which pre sets the module to a MagicMock()

Cleaner mocking and all tests still pass

All of the unit tests passed locally

No

No

No

No
----

### DIFF
--- a/test/unit/deadline_client/cli/test_cli_bundle_submit.py
+++ b/test/unit/deadline_client/cli/test_cli_bundle_submit.py
@@ -7,7 +7,8 @@ Tests for the CLI job bundle commands.
 import os
 import sys
 import json
-from unittest.mock import ANY, patch, Mock, call
+from typing import Generator
+from unittest.mock import ANY, patch, call, MagicMock
 
 import boto3  # type: ignore[import]
 from click.testing import CliRunner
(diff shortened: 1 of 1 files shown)


AWS-Samuel (dismissed):

test/unit/deadline_client/cli/test_cli_bundle_submit.py
```diff
@@ -982,13 +973,13 @@ def test_bundle_gui_submit_submitter_info_multiple_json_allowed(_mock_context):
 
 
 @patch.object(deadline.client.ui, "gui_context_for_cli")
-def test_bundle_gui_submit_submitter_name_overrides_submitter_info(_mock_context):
+def test_bundle_gui_submit_submitter_name_overrides_submitter_info(
+    _mock_context, mock_job_bundle_submitter
+):
```
jusiskin, line 986: nit: we don't need to add `mock_job_bundle_submitter` here because

1.  the test doesn't reference it
2. the fixture is set to `autouse=True`

same feedback to other tests below

jusiskin (dismissed):

test/unit/deadline_client/cli/test_cli_bundle_submit.py
lucaseck: Ah my bad, was under the impression even with autouse you had to. I'll fix that

lucaseck: Ended up having to leave it in, the linting was failing without and I had to remove the autouse or other tests were breaking in the CI on old python versions with:
```
ImportError: PyO3 modules compiled for CPython 3.8 or older may only be initialized once per interpreter process
```

AWS-Samuel (approved):

jusiskin (approved):

PR #953, opened 2025-12-23
feat!: Update hash cache to support byte ranges for file chunking support

mwiebe:

The hash cache only supports caching the hashes of whole files. We'd like to make a manifest format update that chunks large files, and it will need a hash cache for the chunks.

By extending the hash cache to support arbitrary byte ranges, it is ready for any chunking scheme. The update follows the table versioning precedent, creating a new V4 table for the hash cache.

The existing interface for whole file hash caching is unchanged, and new range start/end parameters are available for byte-range hash caching.

Implemented unit tests for the change, confirmed all the existing unit tests pass unchanged.

Updated the docstrings.

No

Yes, it updates the hash cache DB schema from V3 to V4.

No

----

### DIFF
--- a/src/deadline/job_attachments/caches/hash_cache.py
+++ b/src/deadline/job_attachments/caches/hash_cache.py
@@ -2,6 +2,13 @@
 
 """
 Module for accessing the local file hash cache.
+
+Supports two types of hash entries:
+1. Whole-file hashes: range_start=0, range_end=-1 (WHOLE_FILE_RANGE_END)
+2. Byte-range hashes: range_start >= 0, range_end > 0, defining the range [start, end)
+
+The range parameters allow caching hashes for arbitrary byte ranges of files,
+which is useful for caching hashes for any chunking scheme without modifying the cache.
 """
 
 import logging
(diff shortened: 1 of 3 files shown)


baxeaz (dismissed):

src/deadline/job_attachments/caches/hash_cache.py
```diff
@@ -14,24 +21,39 @@
...
+    """
 
     # The file_path is stored as a BLOB in sqlite, encoded with utf-8 and the "surrogatepass"
     # error handler, as file names encountered in practice require this.
     file_path: str
     hash_algorithm: HashAlgorithm
     file_hash: str
     last_modified_time: str
+    range_start: int = 0
+    range_end: int = WHOLE_FILE_RANGE_END
```
godobyte, line 43: Should we validate that `range_end > range_start` for byte-range entries, unless range_end is WHOLE_FILE_RANGE_END? Wondering if it's worth catching invalid ranges early in `__post_init__`, or if the caller is expected to handle that.

test/unit/deadline_job_attachments/caches/test_caches.py
```diff
@@ -267,6 +268,251 @@ def test_enter_sqlite_import_error(self, tmpdir):
...
+        """
+        # GIVEN
+        cache_dir = tmpdir.mkdir("cache")
+        expected_entry = HashCacheEntry(
+            file_path="large_file.bin",
+            hash_algorithm=HashAlgorithm.XXH128,
+            file_hash="chunk_hash_1",
+            last_modified_time="1234.5678",
+            range_start=0,
+            range_end=268435456,  # 256MB
```
godobyte, line 283: Do we want to cover the case where `range_start > range_end`? Could help clarify whether that should error or be silently accepted.

src/deadline/job_attachments/caches/hash_cache.py
mwiebe: Good idea, I've added this validation.

test/unit/deadline_job_attachments/caches/test_caches.py
mwiebe: Added a test to ensure this case raises.

baxeaz (approved):

godobyte (approved):

godobyte (approved):

PR #961, opened 2026-01-13
feat: add help button and documentation link to the job bundle GUI submitter

joel-wong-aws:
Artists were confused about how to setup their submitter. We want to provide them an easy way to find documentation.

Changed the About button to a Help button. Added a link to https://aws-deadline.github.io in the help dialog. Refactored some related code.

`hatch shell && deadline bundle gui-submit <test_scene_path>`

With the default documentation link:
`deadline bundle gui-submit job_bundles\1_task_job`
image

No

No

----


src/deadline/client/dataclasses/submitter_info.py
```diff
@@ -59,6 +59,9 @@ class SubmitterInfo:
     submitter_package_version: Optional[str] = None
     """Version of the submitter package (optional)"""
 
+    documentation_link: Optional[str] = None
+    """Links to the documentation for this submitter (optional)"""
+
```
epmog: Love this idea, but it shouldn't be part of the information we're trying to gather from users when they're reporting issues. I think we can make room in the dialog to showcase this information above the version information

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -233,7 +233,7 @@ def _build_ui(
         self.settings_button = QPushButton(tr("Settings..."))
         self.settings_button.clicked.connect(self.on_settings_button_clicked)
         self.button_box.addButton(self.settings_button, QDialogButtonBox.ResetRole)
-        self.about_button = QPushButton(tr("About") + "...")
+        self.about_button = QPushButton(tr("Help") if self.submitter_info.documentation_link else tr("About"))
```
epmog: Don't think we should change the button name based on args. It should either always be About or Help. I don't have a strong opinion on either. I think both on common nomenclature

src/deadline/client/dataclasses/submitter_info.py
epmog: We could probably also default the value to the top-level docs if you do'nt put something in?

https://aws-deadline.github.io/

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
joel-wong-aws: Changed to Help! We will also always have at least one documentation link pointing to the [user guide](https://aws-deadline.github.io)

src/deadline/client/dataclasses/submitter_info.py
joel-wong-aws: Changed this to include https://aws-deadline.github.io/ by default and accept any number of links.

karthikbekalp:
Changes look good to me although there is a failing check. Can you fix it? @joel-wong-aws

karthikbekalp (dismissed):
Spoke with @joel-wong-aws , turns out the failing check is expected. Approving! Great work!

src/deadline/client/ui/dialogs/_help_dialog.py
```diff
@@ -119,7 +133,11 @@ def _format_version_info(self) -> str:
 
         env_data = asdict(self.environment_info)
         submitter_data = asdict(self.submitter_info)
-        submitter_dict = {k: v for k, v in submitter_data.items() if v is not None}
+
+        # Remove documentation_links from the main display since they're shown separately as hyperlinks
+        submitter_dict = {
+            k: v for k, v in submitter_data.items() if v is not None and k != "documentation_links"
+        }
```
epmog: This filter is highlighting some code smell for me. Is this the right API for passing doc links? `--submitter-info` was meant for adding metadata/diagnostic information that and, although the CLI help text doesn’t really describe that right now, this is explicitly not that.

Taking a step back, for the integrations we own and any bundle gui-submits we can just point to our top-level docs page. It’s got clear guideposts to the right integration which would remove the need to pass in any information. Are there any issues with that?

My own thoughts are that we should probably keep this simple, not pass any info in, and just link to the top-level docs where each integration can add their ‘Known Issues’, etc. A user of the library can always override the docs link and in the future we could add a separate CLI option if an external user wants to add their own docs link that’s different from our general one.

test/unit/deadline_client/dataclasses/test_submitter_info.py
```diff
@@ -0,0 +1,157 @@
...
+        """Test SubmitterInfo with various submitter names."""
+        info = SubmitterInfo(submitter_name=submitter_name)
+        assert info.submitter_name == submitter_name
+        # Should have default documentation links
+        assert info.documentation_links == [
+            DocumentationLink(name="User Guide", url="https://aws-deadline.github.io")
+        ]
+
+    def test_dataclass_equality(self):
+        """Test that SubmitterInfo instances can be compared for equality."""
```
epmog: This test doesn't seem to be needed. It's effectively a singleton for the instance of the submitter.

test/unit/deadline_client/dataclasses/test_submitter_info.py
```diff
@@ -0,0 +1,157 @@
...
+Tests for the SubmitterInfo dataclass.
+"""
+
+import pytest
+
+from deadline.client.dataclasses.submitter_info import SubmitterInfo, DocumentationLink
+
+
+class TestSubmitterInfo:
+    """Test cases for SubmitterInfo dataclass."""
```
epmog, line 11: Some of these tests seem like we're just testing python's `dataclass`. It's probably worth looking over these again to see what value we're getting out of them, if any

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class TestSubmitJobDialogHelpButton:
+    """Test cases for help button functionality in SubmitJobToDeadlineDialog."""
+
+    def test_help_button_text_logic_with_documentation_links(self):
+        """Test the logic for determining help button text when documentation_links is provided."""
+        # Since documentation_links always has at least 1 link, button text is always "Help"
+        button_text = "Help"
+        assert button_text == "Help"
```
epmog: I do'nt think this test is doing what it says it's doing :eyes: Perhaps a leftover artifact from a previous iteration

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
```diff
@@ -0,0 +1,189 @@
...
+
+            # Verify error message box was shown
+            mock_msgbox.critical.assert_called_once()
+            critical_call_args = mock_msgbox.critical.call_args[0]
+            assert critical_call_args[0] == mock_dialog  # parent
+            assert "Error" in critical_call_args[1]  # title
+            assert "Failed to display Help dialog" in critical_call_args[2]  # message
+
+    def test_window_title_formatting_logic(self):
+        """Test window title formatting logic with submitter info."""
```
epmog: This test doesn't appear to be testing any actual submitter logic? It's setting some values and assert those are the values. It's basically just the dataclass all over again..

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
```diff
@@ -0,0 +1,189 @@
...
+
+        assert final_submitter_info is not None
+        assert final_submitter_info.submitter_name == "TestSubmitter"
+        # Should have default documentation links
+        assert final_submitter_info.documentation_links == [
+            DocumentationLink(name="User Guide", url="https://aws-deadline.github.io")
+        ]
+
+    def test_multiple_documentation_links_support(self):
+        """Test support for multiple documentation links."""
```
epmog: This looks to be duplicate from the dataclass test

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Test that all links are preserved
+        assert submitter_info.documentation_links is not None
+        assert len(submitter_info.documentation_links) == 3
+        assert submitter_info.documentation_links[0]["name"] == "User Guide"
+        assert submitter_info.documentation_links[1]["name"] == "API Reference"
+        assert submitter_info.documentation_links[2]["name"] == "Troubleshooting"
+
+    def test_minimum_documentation_links_requirement(self):
+        """Test that documentation_links always has at least one link."""
```
epmog: Also appears to be a duplicate

test/unit/deadline_client/ui/dialogs/test_about_dialog.py
```diff
@@ -1,21 +1,24 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
epmog, line 2: Could we add some CLI tests that actually make use of this change? Would be good to see how it looks like.

epmog (changes requested):

epmog:
The main comment is this one: https://github.com/aws-deadline/deadline-cloud/pull/961#discussion_r2705532008 once we decide there then the other comments may be worth looking at.

src/deadline/client/ui/dialogs/_help_dialog.py
```diff
@@ -68,6 +71,17 @@ def _build_ui(self):
         """Build the dialog UI layout."""
         layout = QVBoxLayout()
 
+        # Add documentation links section
+        for link_info in self.submitter_info.documentation_links:
+            name = link_info["name"]
+            url = link_info["url"]
+            doc_label = QLabel()
+            doc_label.setText(f"{name}: <a href='{url}'>{url}</a>")
```
epmog: I think the text should just be the name with that as the hyper link like [name](https://127.0.0.1), rather than "Docs: [https://127.0.0.1](https://127.0.0.1)". Thoughts?

test/unit/deadline_client/dataclasses/test_submitter_info.py
joel-wong-aws: Removed!

test/unit/deadline_client/dataclasses/test_submitter_info.py
joel-wong-aws: I've cleaned up these tests to reduce duplication and ensure the tests add value.

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
joel-wong-aws: Removed!

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
joel-wong-aws: Removed!

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
joel-wong-aws: Agreed, removed

test/unit/deadline_client/ui/dialogs/test_submit_job_dialog_help_button.py
joel-wong-aws: Ah, yes, unexpected artifact, removed

src/deadline/client/ui/dialogs/_help_dialog.py
joel-wong-aws: Good point, updated!

karthikbekalp (approved):
The changes look good to me! 

A word of advice, its generally best not to use the `--amend` option as its difficult to find out what the changes were between revisions. :)

test/unit/deadline_client/ui/dialogs/test_about_dialog.py
epmog: No longer relevant for this CR

src/deadline/client/ui/dialogs/_help_dialog.py
epmog: Chatted with Joel, we'll punt this to a later change if desired.

epmog (approved):

PR #966, opened 2026-01-16
[DRAFT] chore: Moved _hash_attachments into Job Attachments

evanspearman-a:
NOTE: The API Change Detection CI task will fail. This is intentional because it always fails when a new public API is added. This does not block merging.

Job Attachments and the rest of the client are circularly dependent on each other. The rest of the Client should depend on Job Attachments, but Job Attachments should not depend on the rest of the client. One instance of this circular dependency is the `_hash_attachments` function which is used by Job Attachments but lives in the client.

Move `_hash_attachments` into Job Attachments and remove calls to other client code from it. Replace these calls with parameters and callbacks.

We make progress to remove the circular dependency.

- Have you run the unit tests? __Yes__
- Have you run the integration tests? __Yes__
  that you ensure that the docker-based unit tests pass. __No__

- Are relevant docstrings in the code base updated? __TODO__
- Has the README.md been updated? If you modified CLI arguments, for instance. __N/A__

No

No

----

### DIFF
--- a/src/deadline/client/api/_submit_job_bundle.py
+++ b/src/deadline/client/api/_submit_job_bundle.py
@@ -55,10 +55,15 @@ from ...job_attachments.models import (
 from ...job_attachments.progress_tracker import ProgressReportMetadata, ProgressStatus
 from ...job_attachments.upload import S3AssetManager
 from ._session import session_context
-from ._job_attachment import _hash_attachments  # type: ignore[import]
 from ...job_attachments._path_summarization import human_readable_file_size, summarize_path_list
+from ...job_attachments.api import hash_attachments
+from ...job_attachments.upload import SummaryStatistics
 
 logger = logging.getLogger(__name__)
+
+
+def hashing_telemetry_callback(hashing_summary: SummaryStatistics):
+    api.get_deadline_cloud_library_telemetry_client().record_hashing_summary(hashing_summary)
 
 
 def _summarize_asset_paths(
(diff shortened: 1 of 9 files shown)


src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -712,13 +719,17 @@ def create_job_from_job_bundle(
...
+
+            hashing_summary, asset_manifests = hash_attachments(
                 asset_manager=asset_manager,
                 asset_groups=upload_group.asset_groups,
                 total_input_files=upload_group.total_input_files,
                 total_input_bytes=upload_group.total_input_bytes,
                 print_function_callback=print_function_callback,
                 hashing_progress_callback=hashing_progress_callback,
+                hash_cache_dir=hash_cache_dir,
+                telemetry_callback=hashing_telemetry_callback,
```
leongdl, line 732: Nice! Love this refactoring.

src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -712,13 +719,17 @@ def create_job_from_job_bundle(
                     print_function_callback("Job submission canceled (user input).")
                     raise UserInitiatedCancel()
 
-            _, asset_manifests = _hash_attachments(
+            hash_cache_dir = config_file.get_cache_directory()
+
+            hashing_summary, asset_manifests = hash_attachments(
```
leongdl, line 724: Question - should we stick with `_hash_attachments` for now? We can add an alias to `hash_attachment` in the `__init__.py` export when we are done with the refactoring work. We're not quite ready yet to make this a public interface.

src/deadline/job_attachments/api/_hashing.py
```diff
@@ -1,29 +1,34 @@
...
+from deadline.job_attachments.models import (
+    AssetRootGroup,
+    AssetRootManifest,
+)
+from deadline.job_attachments.progress_tracker import ProgressReportMetadata, ProgressStatus
+from deadline.job_attachments.upload import S3AssetManager, SummaryStatistics
 
 
-def _hash_attachments(
+def hash_attachments(
```
leongdl, line 15: See comment above, can we keep `_` ?

PR #972, opened 2026-01-21
ci: set mainline integration test concurrency to 1

lucaseck:
Failing integration tests it looks like because runs are clobbering each other and clearing each others queues from a recent change to deadline-cloud-test-fixtures.

Set the workflow concurrency to one for now

Test pass again hopefully

It wasn't

No

----


lucaseck:
Closing this, will just revert to previous deadline-cloud-test-fixtures version for now.

PR #979, opened 2026-01-22
refactor: introduce controller-based architecture for Qt UI async operations

justinsaws:
There were a number of threading bugs in the UI code that would sometimes cause cascading error dialogs to appear or cause the app to freeze up. This was due to using Python's threading module intermixed with the Qt UI code and not taking the appropriate steps to update UI code in a thread safe manner. There was also a large amount of business code in our shared UI constructs that could be in a central location for better error handling and refreshing.

Replace direct Python threading with DeadlineUIController singleton pattern for async AWS API operations. Adds AsyncTaskRunner for automatic cancellation of superseded requests, JobSubmissionWorker for job submission, and dedicated DeadlineThreadPool.

Refactors deadline_config_dialog, submit_job_progress_dialog, and shared_job_settings_tab to use the new controller pattern, improving thread safety and simplifying cancellation handling.

 - The dialog updates correctly and no longer has errors when switching higher order resource (AWS profiles, Farms, Queues)
 - Makes making async UI elements easier to expand and create going forward.a

Old behavior:

https://github.com/user-attachments/assets/45fb8cba-af2f-4b9c-ad7c-1c9710196339

New behavior:

https://github.com/user-attachments/assets/dc03153a-1f63-4dcd-9ad7-8b02d4925ee3

    - Yes, they pass.
    - Yes, they pass. 
    - N/A

N/A this is a refactor.

*   [X] This PR adds one or more new dependency Python packages. I acknowledge I have reviewed the considerations for adding dependencies in [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies).

No, this just refactors current functionality.

No.

----

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -21,3 +21,7 @@ pyinstrument == 5.*
 mcp >= 1.13.0; python_version >= '3.10'
 # OpenJD model library for job template parsing in mock backend
 openjd-model >= 0.8.0; python_version >= '3.9'
+
+# GUI testing dependencies
+pytest-qt == 4.*; platform_system in "Windows Darwin" and python_version >= "3.9"
+PySide6-essentials >= 6.6,< 6.11; platform_system in "Windows Darwin" and python_version >= "3.9"
--- a/src/deadline/client/ui/controllers/_async_runner.py
+++ b/src/deadline/client/ui/controllers/_async_runner.py
@@ -1,0 +1,161 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+High-level async task management with automatic cancellation.
+"""
+
+from logging import getLogger
+from typing import Any, Callable, Dict, Optional
+
+from qtpy.QtCore import QObject, Qt, Signal
+
+from ._async_task import AsyncTask
+from ._thread_pool import DeadlineThreadPool
+
+
+logger = getLogger(__name__)
+
+
+class AsyncTaskRunner(QObject):
+    """
+    Manages async task execution with automatic cancellation of superseded operations.
+
+    When a new task is started with the same operation_key, any previous task
+    with that key is automatically canceled. This prevents race conditions
+    where an older, slower request returns after a newer one.
+
+    All signal connections use Qt.QueuedConnection to ensure thread-safe
+    delivery from background threads to the main thread.
+
+    Example::
+
+        from deadline.client.ui.controllers import AsyncTaskRunner
+
+        runner = AsyncTaskRunner()
+        runner.run(
+            operation_key="list_farms",
+            fn=api.list_farms,
+            on_success=self._handle_farms,
+            on_error=self._handle_error,
+            config=config,  # kwargs passed to fn
+        )
+
+    Signals:
+        task_error: Emitted when any task encounters an error.
+                   Args: (operation_key, exception)
+    """
+
+    task_error = Signal(str, BaseException)
+
+    def __init__(self, parent: Optional[QObject] = None) -> None:
+        super().__init__(parent)
+        self._thread_pool = DeadlineThreadPool.instance()
+        self._active_tasks: Dict[str, AsyncTask] = {}
+        self._operation_counter = 0
+
+    def run(
+        self,
+        operation_key: str,
+        fn: Callable[..., Any],
+        on_success: Optional[Callable[[Any], None]] = None,
+        on_error: Optional[Callable[[BaseException], None]] = None,
+        *args: Any,
+        **kwargs: Any,
+    ) -> int:
+        """
+        Run a function asynchronously in the dedicated thread pool.
+
+        Args:
+            operation_key: Unique key identifying this operation type.
+                          If a task with the same key is running, it
+                          will be canceled before starting the new one.
+            fn: The function to execute in the background
+            on_success: Callback invoked with fn's return value on success.
+                       Called in the main thread via queued connection.
+            on_error: Callback invoked with the exception on failure.
+                     Called in the main thread via queued connection.
+            *args: Positional arguments passed to fn
+            **kwargs: Keyword arguments passed to fn
+
+        Returns:
+            Operation ID for tracking this specific task instance
+        """
+        # Cancel any existing task with the same key
+        self.cancel(operation_key)
+
+        # Generate unique operation ID
+        self._operation_counter += 1
+        operation_id = self._operation_counter
+
+        # Create the task
+        task = AsyncTask(fn, *args, operation_id=operation_id, **kwargs)
+
+        # Connect success callback with queued connection for thread safety
+        if on_success is not None:
+            task.signals.result.connect(on_success, Qt.QueuedConnection)
+
+        # Connect error callbacks
+        if on_error is not None:
+            task.signals.error.connect(on_error, Qt.QueuedConnection)
+
+        # Always emit to our error signal for centralized logging
+        # Use a closure to capture operation_key
+        def emit_task_error(e: BaseException) -> None:
+            self.task_error.emit(operation_key, e)
+
+        task.signals.error.connect(emit_task_error, Qt.QueuedConnection)
+
+        # Clean up tracking when task finishes
+        # Use a closure to capture operation_key
+        def cleanup() -> None:
+            self._active_tasks.pop(operation_key, None)
+
+        task.signals.finished.connect(cleanup, Qt.QueuedConnection)
+
+        # Track and start
+        self._active_tasks[operation_key] = task
+        self._thread_pool.start(task)
+
+        logger.debug(f"Started async task: {operation_key} (id={operation_id})")
+        return operation_id
+
+    def cancel(self, operation_key: str) -> bool:
+        """
+        Cancel a running task by operation key.
+
+        Args:
+            operation_key: The key of the operation to cancel
+
+        Returns:
+            True if a task was canceled, False if no task was running
+        """
+        task = self._active_tasks.pop(operation_key, None)
+        if task is not None:
+            task.cancel()
+            logger.debug(f"Canceled async task: {operation_key}")
+            return True
+        return False
+
+    def cancel_all(self) -> int:
+        """
+        Cancel all running tasks.
+
+        Returns:
+            Number of tasks that were canceled
+        """
+        count = len(self._active_tasks)
+        for task in self._active_tasks.values():
+            task.cancel()
+        self._active_tasks.clear()
+        if count > 0:
+            logger.debug(f"Canceled {count} async tasks")
+        return count
+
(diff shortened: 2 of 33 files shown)


src/deadline/client/ui/dialogs/deadline_config_dialog.py
```diff
@@ -833,22 +862,83 @@ def job_history_dir_changed(self):
...
+        if farm_id is None:
+            return
+        self.changes["defaults.farm_id"] = farm_id
+        # Clear the queue_id since the old queue doesn't exist in the new farm
+        self.changes["defaults.queue_id"] = ""
+        # Clear storage profile lists - they depend on queue which we're about to refresh
+        self.default_storage_profile_box.clear_list()
         self.refresh()
+        # Trigger cascading refresh - queues will load, then storage profiles
+        self._awaiting_queues_for_cascade = True
```
larrygao001, line 877: Could there be a race condition? Can the flags properly track which cascade corresponds to which user action? Similar for line 852, and line 233-235.

src/deadline/client/ui/controllers/_async_runner.py
```diff
@@ -0,0 +1,161 @@
...
+
+    task_error = Signal(str, BaseException)
+
+    def __init__(self, parent: Optional[QObject] = None) -> None:
+        super().__init__(parent)
+        self._thread_pool = DeadlineThreadPool.instance()
+        self._active_tasks: Dict[str, AsyncTask] = {}
+        self._operation_counter = 0
+
+    def run(
```
larrygao001, line 56: What if tasks hang? Will they be cleaned up?

src/deadline/client/ui/dialogs/_job_submission_worker.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # Reset state for new confirmation
+        self._confirmation_result = None
+        self._confirmation_event.clear()
+
+        # Request confirmation from main thread
+        self.confirmation_requested.emit(message, default_response)
+
+        # Block until main thread responds or cancellation
+        self._confirmation_event.wait()
```
larrygao001, line 155: So if the main thread is busy or (if possible) the dialog is closed, the worker could hang forever?

src/deadline/client/ui/controllers/_async_task.py
```diff
@@ -0,0 +1,126 @@
...
+        1. Check if canceled before starting
+        2. Execute the function
+        3. Check if canceled before emitting result/error
+        4. Emit finished signal (if not canceled)
+        """
+        if self._is_canceled:
+            return
+
+        try:
+            result = self.fn(*self.args, **self.kwargs)
```
larrygao001, line 118: Wouldn't this function still run even if canceled?

src/deadline/client/ui/controllers/_async_runner.py
justinsaws: If tasks hang, they will just lock up the thread forever. That is no different than the previous implementation and how threads work in general. You could force terminate the thread but that could have other consequences  and isn't generally advisable.

src/deadline/client/ui/controllers/_async_task.py
justinsaws: Yes, the cancellation check is only done on the intro to this function. Cancellation of any thread depends on callbacks like this and in this case the operation continuing to run in the background isn't a problem. The returned results being thrown away is what we would want. We do not want to force the thread to terminate because that can lead to undefined behavior.

src/deadline/client/ui/dialogs/_job_submission_worker.py
justinsaws: This is only done when we are waiting on user input so this is desired behavior. If the main thread is blocked for some reason and not showing the dialog then the application will show as hung and get killed. Any potentially blocking operation shouldn't be being done in the main thread already so this case shouldn't come up.

src/deadline/client/ui/dialogs/deadline_config_dialog.py
justinsaws: What's the race condition that you are seeing? The updating is handled by the controller class and they should cancel requests that would overlap.

requirements-testing.txt
```diff
@@ -21,3 +21,7 @@ pyinstrument == 5.*
 mcp >= 1.13.0; python_version >= '3.10'
 # OpenJD model library for job template parsing in mock backend
 openjd-model >= 0.8.0; python_version >= '3.9'
+
+# GUI testing dependencies
```
waninggibbon, line 25: Why only Windows and Darwin, do we not perform GUI tests in Linux?

justinsaws: The Linux test runners don't have any windowing system or UI libraries installed so Qt will crash the second it tries to create the event queue. Adding and setting up those deps on the linux runners would be non-trivial and the difference between our three platforms is essential none in the UI. 

I think our time would be better spent implementing validation testing of the UI elements instead of worrying about unit tests here.

src/deadline/client/ui/dialogs/deadline_config_dialog.py
larrygao001: Thanks for the explanation. I understand the controller cancels overlapping requests, but my concern is specifically about the boolean flags like _awaiting_queues_for_cascade. If a user rapidly switches profiles
twice, could the flag set by action 1 still be true when action 2's callback fires, causing it to incorrectly trigger a cascade? Would a request ID or counter be more robust than a boolean?

src/deadline/client/ui/controllers/_async_runner.py
larrygao001: Fair point that this matches previous behavior. Would it be worth adding a debug log when a task exceeds some threshold (e.g., 30s)? That would help with diagnosing issues in the field without changing the threading model.

src/deadline/client/ui/dialogs/_job_submission_worker.py
larrygao001: I understand this is intentional blocking for user input. My concern is if the dialog gets closed unexpectedly (crash, force quit, etc.), the worker thread would hang indefinitely. Would adding a timeout with a
check for dialog validity be a reasonable safeguard, or is that overkill for this use case?

src/deadline/client/ui/controllers/_async_task.py
larrygao001: Makes sense - discarding the result is sufficient for our use case. No further concerns here

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -113,7 +113,7 @@ def __init__(
         attachments: AssetReferences,
         on_create_job_bundle_callback: OnCreateJobBundleCallback,
         parent: Optional[QWidget] = None,
-        f: Qt.WindowFlags = Qt.WindowFlags(),
+        f: Any = Qt.WindowFlags(),
```
andychoquette, line 116: nit - this likely could've remained Qt.WindowFlags

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
```diff
@@ -321,11 +277,12 @@ def closeEvent(self, event: QCloseEvent) -> None:
...
-            logger.info("Canceling submission...")
-            self.status_label.setText(tr("Canceling submission..."))
-            if self.__submission_thread is not None:
-                while self.__submission_thread.is_alive():
+            if self._worker is not None:
+                self._worker.cancel()
+                logger.info("Canceling submission...")
+                self.status_label.setText(tr("Canceling submission..."))
+                # Wait for worker to finish with event processing
+                while self._worker.isRunning():
```
andychoquette, line 285: Nit: When this busy wait is running while submitting, will this freeze up the UI? Would it make more sense to add a QThread.wait(timeout) with a fallback?

andychoquette (approved):

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
waninggibbon: Instead of adding a timeout here, we could get around this issue all together with a signal based approach like we are doing elsewhere with the AsyncTaskRunner since JobSubmissionWorker is a qt thread, right?

src/deadline/client/ui/controllers/_async_runner.py
justinsaws: Additional logging could be helpful but not sure if it's needed in this case. Threaded logging if we aren't logging to individual logs can get messy and we weren't doing it before. This isn't complex threading logic for a long running application but just getting API responses from the service so that feels like additional complexity that probably won't add much. We will already get the logging from boto3 if the call to the API fails.

src/deadline/client/ui/dialogs/_job_submission_worker.py
justinsaws: The way this is parented in Qt, if the dialog gets closed this will get terminated and cleaned up. It is owned by the dialog. It's using the Qt constructs appropriately so it should unblock on deletion.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
justinsaws: Yeah, I don't know why it would have changed this. I'll swap that back.

src/deadline/client/ui/dialogs/submit_job_progress_dialog.py
justinsaws: The line below is what prevents this from blocking the UI. It releases control to the event queue to process background events and then comes back to re-evaluate this condition. Timeouts don't make sense here because we don't want to force terminate threads. That can leave the application in an undefined state and isn't advised on desktop applications.

A signal based approach wouldn't work here because this is on a destruction/close event. This object is going away so it needs to clean-up now. The threads are just making API calls so I think the risk here is low. Also, this isn't a long running application.

waninggibbon: Makes sense, thanks for clarifying.

waninggibbon (approved):

src/deadline/client/ui/dialogs/deadline_config_dialog.py
justinsaws: The AsyncTaskRunnner will cancel the previous call when a new call comes in so data corruption shouldn't happen. There is a potential race condition here that could be solved with the call id or generation counter that would prevent the data from flipping on the user if they are changing the selection very rapidly.

I'm not sure if the additional complexity is warented here because this is a very edge case scenario but if you feel the change is needed for the optimal experience then I'd be fine with adding it in.

PR #982, opened 2026-01-24
feat: add search term support to `deadline job get` command

mwiebe:

Users must know the exact job ID to use commands like `deadline job get --job-id job-xxx`. This is cumbersome because:
- Job IDs are opaque 32-character hex strings
- Users typically remember job names, not IDs
- Finding a job ID requires running `deadline job list` first

Enable users to find jobs using a search term instead of requiring exact job IDs. `deadline job get "search term"` searches for jobs matching that term across multiple fields, displaying full details if one matches or a summary list if multiple match.

- Add positional SEARCH_TERM argument to job get command
- Single result shows full job details via GetJob API
- Multiple results show summary with job name, ID, status, timestamp, and task counts
- Job ID pattern (job-xxx) as positional arg bypasses search
- No args uses default job_id from config (preserves original behavior)
- Add helper module _job_helpers.py with search and display functions
- Timestamps displayed in local timezone with UTC offset
- Task summary shows running, ready, pending, succeeded, failed, canceled counts
- Limit results to 5 most recent matches

Better UX when querying the status of jobs from the CLI.

Added unit tests for the new cases. Tested the command on a farm with a variety of test jobs, e.g. with longer names to see how the results look.

Docstrings for the CLI command are updated, so it shows up in CLI help.

No

No

No

----

### DIFF
--- a/docs/design/cli-job-get-with-search-term.md
+++ b/docs/design/cli-job-get-with-search-term.md
@@ -1,0 +1,118 @@
+# Design: `deadline job get` with Search Term
+
+## Overview
+
+Enable users to find jobs using a search term instead of requiring job IDs. `deadline job get "search term"` searches for jobs matching that term across multiple fields, displaying full details if one matches or a summary list if multiple match.
+
+## Problem Statement
+
+Users must know the exact job ID to use commands like `deadline job get --job-id job-xxx`. This is cumbersome because:
+- Job IDs are opaque 32-character hex strings
+- Users typically remember job names, not IDs
+- Finding a job ID requires running `deadline job list` first
+
+## CLI Interface
+
+```bash
+# Search by name (single result shows full details, multiple shows summary)
+deadline job get "Blender Render"
+
+# Job ID as positional argument
+deadline job get job-eee5a7898b6c493d84aaa12ea384156b
+
+# Prefix of Job ID as positional argument
+deadline job get job-eee5a
+
+# Job ID with --job-id flag
+deadline job get --job-id job-eee5a7898b6c493d84aaa12ea384156b
+
+# No arguments uses default job ID from config
+deadline job get
+```
+
+## Command Behavior
+
+1. If `SEARCH_TERM` matches job ID pattern (`job-[0-9a-f]{32}`): get full job details
+2. If `SEARCH_TERM` is any other string: search jobs with that term
+   - Exactly one match: get full job details via `GetJob` API
+   - Multiple matches: display summary table
+   - No matches: display "No jobs found" message
+3. If no `SEARCH_TERM` provided: use `--job-id` or default job ID from config
+
+## Output Format
+
+**Single result or job ID:** Full job details from `GetJob` API
+
+**Multiple results:**
+```
+Found 10 job(s) matching "job", showing most recent 5:
+
+  Turntable with Maya/Arnold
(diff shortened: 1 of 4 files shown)


docs/design/cli-job-get-with-search-term.md
```diff
@@ -0,0 +1,115 @@
+# Design: `deadline job get` with Search Term
```
leongdl, line 1: I like that we check in these small design docs for each new feature now. Its super helpful for new members to onboard, and also AI to help code.

docs/design/cli-job-get-with-search-term.md
```diff
@@ -0,0 +1,115 @@
...
+- Finding a job ID requires running `deadline job list` first
+
+## CLI Interface
+
+```bash
+# Search by name (single result shows full details, multiple shows summary)
+deadline job get "Blender Render"
+
+# Job ID as positional argument
+deadline job get job-eee5a7898b6c493d84aaa12ea384156b
```
leongdl, line 21: I think even job-eee5a will work as a great example. Similar to the git short hash.

leongdl (dismissed):

docs/design/cli-job-get-with-search-term.md
mwiebe: Indeed, I tried this out and it works great.

leongdl (dismissed):

baxeaz (dismissed):

src/deadline/client/cli/_groups/_job_helpers.py
```diff
@@ -0,0 +1,149 @@
...
+    canceled = task_counts.get("CANCELED", 0)
+    not_compatible = task_counts.get("NOT_COMPATIBLE", 0)
+
+    parts = []
+    if ready:
+        parts.append(f"{ready} ready")
+    if running:
+        parts.append(f"{running} running")
+    if interrupting:
+        parts.append(f"{running} interrupting")
```
godobyte: nit - should be `parts.append(f"{interrupting} interrupting")`?

docs/design/cli-job-get-with-search-term.md
```diff
@@ -0,0 +1,118 @@
...
+## Output Format
+
+**Single result or job ID:** Full job details from `GetJob` API
+
+**Multiple results:**
+```
+Found 10 job(s) matching "job", showing most recent 5:
+
+  Turntable with Maya/Arnold
+    job-ca34b643089742609bdf36391b4f0fe0  SUCCEEDED     2026-01-23 17:09:04 -0800
```
godobyte, line 51: Question - have you considered adding the job name, would what be interesting to see for customers?

src/deadline/client/cli/_groups/_job_helpers.py
mwiebe: Yes, thanks for catching that

docs/design/cli-job-get-with-search-term.md
mwiebe: The job name is on the previous line. I found that fitting the job name with the other information on the same line was too long, so I split the summaries to multiple lines.

baxeaz (approved):

docs/design/cli-job-get-with-search-term.md
godobyte: Oh I see, didn't realize they were job names. That looks great!

godobyte (approved):

leongdl (approved):

PR #984, opened 2026-01-29
test: Add a mock deadline backend and tests for trace-schedule

mwiebe:

The 'deadline job trace-schedule' command had a single test, and noted that improved testing would involve simulating service behavior more.

This change adds a mock deadline service backend that does so for a subset of Deadline Cloud, deletes that test, and creates new tests for the trace-schedule command.

Creates the `MockDeadlineBackend` that simulates a part of the Deadline Cloud service to simplify unit testing with moto. We will be able to extend this to more service functionality and use it to make the unit tests more effective.

Added tests specifically to test the `MockDeadlineBackend` behavior.

Added a short design doc about the class.

*   [x] This PR adds a dependency to the testing requirements, not to the runtime use of the library.

No.

No.

----

### DIFF
--- a/test/unit/deadline_client/cli/test_cli_job_trace_schedule.py
+++ b/test/unit/deadline_client/cli/test_cli_job_trace_schedule.py
@@ -1,0 +1,332 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Tests for trace-schedule using MockDeadlineBackend.
+"""
+
+from __future__ import annotations
+
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+from click.testing import CliRunner
+
+from deadline.client import config
+from deadline.client.cli import main
+from ..mock_deadline_backend import MockDeadlineBackend
+
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (3, 9), reason="openjd-model requires Python 3.9+"
+)
+
+SINGLE_TASK_TEMPLATE = """
+specificationVersion: jobtemplate-2023-09
+name: Test
+steps:
+  - name: Render
+    script:
+      actions:
+        onRun:
+          command: echo
+          args: ["test"]
+"""
+
+SIX_TASK_TEMPLATE = """
+specificationVersion: jobtemplate-2023-09
+name: Test
+steps:
+  - name: Render
+    parameterSpace:
+      taskParameterDefinitions:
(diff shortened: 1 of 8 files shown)


test/unit/deadline_client/mock_deadline_backend.py
```diff
@@ -0,0 +1,475 @@
...
+        deadline_mock.get_farm.side_effect = self.get_farm
+        deadline_mock.create_queue.side_effect = self.create_queue
+        deadline_mock.get_queue.side_effect = self.get_queue
+        deadline_mock.create_job.side_effect = self.create_job
+        deadline_mock.get_job.side_effect = self.get_job
+        deadline_mock.get_step.side_effect = self.get_step
+        deadline_mock.get_task.side_effect = self.get_task
+        deadline_mock.list_sessions.side_effect = self.list_sessions
+        deadline_mock.list_session_actions.side_effect = self.list_session_actions
+        deadline_mock.search_jobs.side_effect = self.search_jobs
```
crowecawcaw: Another way we could set this up is to run it as a local webserver, set the AWS_DEADLINE_ENDPOINT env var to `localhost`, then run the actual CLI without any internal mocking. It could also be reused for things like worker agent tests.

crowecawcaw (dismissed):

test/unit/deadline_client/cli/test_cli_job_trace_schedule.py
```diff
@@ -0,0 +1,323 @@
...
+        onRun:
+          command: echo
+          args: ["test"]
+"""
+
+
+class TestTraceScheduleWithMockBackend:
+    """Tests for trace-schedule command using MockDeadlineBackend."""
+
+    def setup_job(self, backend, deadline_mock, template):
```
leongdl: Can we get kiro to generate the mypy stubs please

mwiebe: Type annotations are added

test/unit/deadline_client/mock_deadline_backend.py
mwiebe: That sounds like something to add directly to the `moto` library so that https://docs.getmoto.org/en/latest/docs/server_mode.html works with it.

leongdl (approved):

crowecawcaw (approved):

PR #986, opened 2026-01-31
feat: add submission hooks support for job bundles

rickrams:
Add configurable pre-submission and post-submission hooks that execute during job submission workflow. Hooks are defined in hooks.yaml or hooks.json in the job bundle directory.

Features:
- Pre-submission hooks run before hashing/uploading and can modify payload
- Post-submission hooks run after job creation (non-blocking on failure)
- Hooks receive metadata via stdin JSON and DEADLINE_* environment variables
- Asset references from hooks are merged with existing references
- Configurable timeout per hook (default 60s)
- Works with both CLI (deadline bundle submit) and GUI (deadline bundle gui-submit)

Documentation:
- Added docs/submission-hooks.md with full usage guide and examples
- Updated README.md with submission hooks section

Testing:
- Added 45 unit tests covering models, validation, merging, and execution

Being able to execute code before and after job submission. Use cases like setting variables based on the environment or doing introspection for assets for job attachment.

A new job bundle file called "hooks" that species a command and arguments (similar to how open job description templates do)

I manually tested both CLI and gui submit with a custom script that inspect vrscene files and finds and adds additional job attaachments.

- Has the README.md been updated? YEs

### Is this a breaking change? No

### Does this change impact security? Potentially. It executes scripts that could be malicous or potentially misused or hijacked.

----


docs/submission-hooks.md
```diff
@@ -0,0 +1,286 @@
+# Submission Hooks
```
leongdl, line 1: Always tell kiro to run "hatch run fmt" and "hatch run lint" before pushing. Its failing CI on lint and format.

rickrams:
Updated to allow for two independent hook sources, each with its own security setting:

### Hook Sources

1. **Bundle Hooks** - `hooks.yaml` in job bundles
  - Controlled by `settings.allow_bundle_hooks` (default: `false`)
  - For workflows where hooks are bundled with job templates

2. **Environment Hooks** - `DEADLINE_HOOKS_DIR` environment variable
  - Controlled by `settings.allow_environment_hooks` (default: `false`)
  - For studios that want to enforce hooks across all submissions without modifying job bundles or submitter code

### Security Model

Both sources are disabled by default (opt-in). When enabled, users see a confirmation prompt showing exactly which commands will run before execution (unless `auto_accept=true`).

### Studio Deployment Example

```bash
# IT configures workstations once
deadline config set settings.allow_environment_hooks true

# Per-application launcher scripts set the hooks location
# blender_launcher.sh
export DEADLINE_HOOKS_DIR=/studio/pipeline/hooks/blender
exec blender "$@"

# maya_launcher.sh  
export DEADLINE_HOOKS_DIR=/studio/pipeline/hooks/maya
exec maya "$@"

File: docs/submission-hooks.md [hunk 2]
```diff
@@ -0,0 +1,376 @@
...
+- Validate job configuration
+- Discover and add additional input files
+- Modify job parameters (priority, etc.)
+- Enforce studio policies
+
+Pre-submission hooks **block submission** if they fail (non-zero exit code or timeout).
+
+### Post-Submission Hooks
+
+Run **after** the job is successfully created. Use these to:
```
crowecawcaw: Nit: clarify that this is after the job is created successfully in Deadline (i.e. the async processing and validation succeeded, job status is now READY or PENDING)

docs/submission-hooks.md
```diff
@@ -0,0 +1,376 @@
...
+```
+
+### Hook Definition Fields
+
+| Field | Required | Default | Description |
+|-------|----------|---------|-------------|
+| `command` | Yes | - | Executable or interpreter (e.g., `python3`, `bash`) |
+| `args` | No | `[]` | Command-line arguments |
+| `timeout` | No | `60` | Maximum execution time in seconds |
+| `env` | No | `{}` | Additional environment variables |
```
crowecawcaw, line 118: Do the submitters env vars get pass in as well? (I'm assuming so, worth clarifying.)

docs/submission-hooks.md
```diff
@@ -0,0 +1,376 @@
...
+
+- **Absolute paths**: Used as-is
+- **Relative paths**: Resolved relative to the job bundle directory
+- **Command names**: Searched in system PATH
+
+## Hook Input
+
+Hooks receive job metadata in two ways:
+
+### Environment Variables
```
crowecawcaw: Why pass in data as env vars in addition to STDIN? Because they're easier to work with? If they have the same data and env vars aren't required, I'd lean towards only supporting STDIN.

docs/submission-hooks.md
```diff
@@ -0,0 +1,376 @@
...
+    "attachments": {
+        "assetReferences": {
+            "inputFilenames": ["/path/to/discovered/texture.exr"]
+        }
+    }
+}
+print(json.dumps(output))
+```
+
+Asset references are **merged** (not replaced) - your additions combine with existing files.
```
crowecawcaw: Suggest the top level keys are merged but the nested data is replaced, so that we can un-attach files as well. It also makes eliminates some ambiguity and edge cases. I think hook code will look about the same.

docs/submission-hooks.md
rickrams: good point - added.

docs/submission-hooks.md
rickrams: Changed from union to replace-at-nested-key (hooks can now remove files too)

docs/submission-hooks.md
rickrams: yeah, calrified in docs

docs/submission-hooks.md
rickrams: yeah just for being easier to work with - simple hooks become much simpler. A bash hook that just logs the job name is echo $DEADLINE_JOB_NAME vs parsing JSON from stdin with jq or python. For studios writing quick shell-script hooks, that's a meaningful difference.

But would be ok removing them - it's not functionally required

docs/submission-hooks.md
```diff
@@ -0,0 +1,380 @@
+# Submission Hooks
+
+Submission hooks allow you to run custom scripts during the job submission workflow. You can use hooks to validate job configurations, discover additional assets, modify submission parameters, or integrate with external systems.
+
+## Quick Start
+
+Create a `hooks.yaml` (or `hooks.json`) file in your job bundle directory:
```
baxeaz: What's my job bundle directory here? Unless I'm resubmitting an existing bundle isn't that often getting created at submission time? Or does it mean some parent folder in the job history cache? I may discover this later in the PR but probably worth clarifying here

docs/submission-hooks.md
```diff
@@ -0,0 +1,380 @@
...
+- Update tracking systems
+- Log submission details
+
+Post-submission hooks **do not block** - failures are logged as warnings.
+
+## Configuration
+
+### hooks.yaml or hooks.json
+
+Place either `hooks.yaml` or `hooks.json` in your job bundle directory alongside `template.yaml`. If both exist, an error is raised.
```
baxeaz: Ah yeah, so it is in the actual bundle before submitting - Maybe I'll see how that's intended to get populated in new bundles further in the PR but not currently clear.

Reading back through the history it looks like maybe this was solved in a later iteration and there's ways of predefining hook directories now, maybe we just need to update some of this documentation that looks like it expects hooks to always be sitting in the bundle already

src/deadline/client/job_bundle/_hooks/_merger.py
```diff
@@ -0,0 +1,43 @@
...
+    Keys not present in modified are preserved from original.
+    """
+    original = original or {}
+    modified = modified or {}
+    result = dict(original)
+    result.update(modified)
+    return result
+
+
+def merge_payload(original: _Dict[str, _Any], modified: _Dict[str, _Any]) -> _Dict[str, _Any]:
```
baxeaz, line 25: So the intent is to change or add but never remove?  I saw a later comment that seemed to say it did support removing by taking a copy first and just returning everything except the removed element but it looks like it would be preserved from the original in that case?  Maybe not following exactly

docs/submission-hooks.md
```diff
@@ -0,0 +1,380 @@
...
+    }
+  ]
+}
+```
+
+### Hook Definition Fields
+
+| Field | Required | Default | Description |
+|-------|----------|---------|-------------|
+| `command` | Yes | - | Executable or interpreter (e.g., `python3`, `bash`) |
```
baxeaz, line 115: I guess direct bash/generic command support was a hard requirement rather than just saying "hooks are in python" and then we could simplify a lot of the subprocess stdin/stdout stuff?

crowecawcaw:
Some ideas from offline conversations:
- add a version number or schema or similar to hooks so we can safely update their structure in the future
- what about hooks for submitters running inside DCCs? (probably out of scope)

docs/submission-hooks.md
```diff
@@ -0,0 +1,380 @@
...
+
+### Pre-Submission Hooks
+
+Run **before** job attachments are hashed and uploaded. Use these to:
+- Validate job configuration
+- Discover and add additional input files
+- Modify job parameters (priority, etc.)
+- Enforce studio policies
+
+Pre-submission hooks **block submission** if they fail (non-zero exit code or timeout).
```
crowecawcaw, line 48: Could also add a pre-submitter hook that runs before the GUI is shown, e.g. to patch the job template or to make some parameters hidden or to fetch default values from some place.

docs/submission-hooks.md
```diff
@@ -0,0 +1,380 @@
...
+
+### Post-Submission Hooks
+
+Run **after** the CreateJob API call returns successfully (i.e., the job has been accepted by AWS Deadline Cloud). Use these to:
```
crowecawcaw, line 52: Note: should have test cases for if CreateJob fails AND if CreateJob succeeds but then the job is not created successfully (e.g. bad template).

docs/submission-hooks.md
rickrams: good point - in the DCC use case it's created at submission time. In the deadline client use case outside a DCC the bundle is pre-existing.  and as your later comment indicates support was later added for hooks outside the bundle for the IT admin use cases, so may just be a docs clarity thing.

docs/submission-hooks.md
rickrams: yeah, but potentially debatable. it does make some easy things really easy. i'm not strongly opinionated on how it complicates the implementation.

docs/submission-hooks.md
rickrams: reworked the docs to address this

docs/submission-hooks.md
rickrams: Good idea. different PR or a follow on makes sesnse to me though.

docs/submission-hooks.md
rickrams: addresed

rickrams:
> Some ideas from offline conversations:
> 
>     * add a version number or schema or similar to hooks so we can safely update their structure in the future
> 
>     * what about hooks for submitters running inside DCCs? (probably out of scope)

Added a version field.  Hooks running in DCCs should work shouldn't they? the bundle ones wouldn't since bundle is made by the integration but the IT Admin ones would?

leon-li-inspire (approved):

baxeaz (approved):

PR #991, opened 2026-02-05
feat: add deprecation warnings for legacy sync input and output in Job Attachments

godobyte:

The `AssetSync` class contains three methods (`attachment_sync_inputs`, `sync_inputs`, and `sync_outputs`) that are no longer used by the worker agent after the `ASSET_SYNC_JOB_USER_FEATURE` was fully released. These methods represent the legacy approach to attachment synchronization that has been replaced by the OpenJD task-based implementation using `AttachmentDownloadAction` and `AttachmentUploadAction`.

While these methods are only used in tests and scripted testing tools, they remain part of the public API and could potentially be used by external consumers. To safely deprecate them without breaking existing code, proper deprecation warnings are needed.

Added formal deprecation warnings to three methods in `asset_sync.py`:

1. **`attachment_sync_inputs()`** - Added `warnings.warn()` with `DeprecationWarning`
2. **`sync_inputs()`** - Added `warnings.warn()` with `DeprecationWarning`
3. **`sync_outputs()`** - Added `warnings.warn()` with `DeprecationWarning`

Each method now:
- Emits a `DeprecationWarning` when called
- Includes deprecation notice in docstring using `.. deprecated::` format
- Provides guidance: "Use other public APIs under job attachment instead."
- Uses `stacklevel=2` to show the warning at the caller's location

Added `import warnings` to the module imports.

**Positive Impacts:**
- **Clear migration path**: Users are now warned when using deprecated methods
- **Backwards compatible**: Methods still work, no breaking changes
- **Better documentation**: Docstrings clearly indicate deprecation status
- **Prepares for future cleanup**: Enables safe removal in a future major version

**User Impact:**
- Users calling these methods will see deprecation warnings in their logs
- No functional changes - all methods continue to work as before
- Provides time for users to migrate to alternative APIs

- [x] Have you run the unit tests? - Yes, all linting and type checking passes
- [ ] Have you run the integration tests? - Not run yet
- [ ] Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended that you ensure that the docker-based unit tests pass. - Yes, changes to `asset_sync` module. Docker tests should be run.

**Testing performed:**
- ✅ All linting checks pass (ruff, mypy)
- ✅ Syntax validation successful
- ✅ Type checking passes with no errors

- [x] Are relevant docstrings in the code base updated? - Yes, added `.. deprecated::` sections to all three method docstrings
- [ ] Has the README.md been updated? - No changes needed, this is an internal API deprecation

This library is designed to be integrated into third-party applications that have bespoke and customized deployment environments. Adding dependencies will increase the chance of library version conflicts and incompatibilities.

*   [x] This PR does not add any new dependencies. - Only uses Python's built-in `warnings` module

**Not yet** - This is not a breaking change. The deprecated methods continue to function exactly as before. The only change is that they now emit `DeprecationWarning` messages to inform users that these methods will be removed in a future version.

Users can:
- Continue using the methods (with warnings)
- Suppress the warnings if needed using `warnings.filterwarnings()`
- Migrate to alternative APIs at their convenience

This follows standard Python deprecation practices and provides a graceful migration path.

No, this change only adds deprecation warnings to existing methods without modifying their behavior or file/directory handling.

----

### DIFF
--- a/src/deadline/job_attachments/asset_sync.py
+++ b/src/deadline/job_attachments/asset_sync.py
@@ -9,6 +9,7 @@ import shutil
 import sys
 import time
 import json
+import warnings as _warnings
 from io import BytesIO
 from logging import Logger, LoggerAdapter, getLogger
 from math import trunc
@@ -394,6 +395,10 @@ class AssetSync:
             VIRTUAL: downloads a manifest file and mounts a Virtual File System at the
                        specified asset root corresponding to the manifest contents
 
+        .. deprecated::
+            attachment_sync_inputs is deprecated and will be removed in a future version.
+            Use other public APIs under job attachments instead.
+
         Args:
             s3_settings: S3-specific Job Attachment settings.
             attachments: an object that holds all input assets for the job.
@@ -419,6 +424,12 @@ class AssetSync:
             VIRTUAL: same as COPIED, but the summary statistics will be empty since the
                        download hasn't started yet.
         """
+        _warnings.warn(
+            "attachment_sync_inputs is deprecated and will be removed in a future version. "
+            "Use other public APIs under job attachments instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
 
         if not s3_settings:
             self.logger.info(
@@ -739,7 +750,6 @@ class AssetSync:
                 f"Total file size required for download ({input_size_readable}) is larger than available disk space ({disk_free_readable})"
             )
 
-    # This is on deprecation path, please use attachment_sync_inputs instead.
     def sync_inputs(
         self,
         s3_settings: Optional[JobAttachmentS3Settings],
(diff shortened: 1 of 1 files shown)


leongdl (dismissed):

src/deadline/job_attachments/asset_sync.py
```diff
@@ -9,6 +9,7 @@
 import sys
 import time
 import json
+import warnings
```
mwiebe: It's a little tricky to find in the CI run, but the reason for teh API change warning is this line:

🆕 ADDED APIs (1):
  + deadline.job_attachments.asset_sync.warnings (alias)

You could change this to `import warnings as _warnings`

godobyte: Ah the validator detected `warnings` is now accessible as `deadline.job_attachments.asset_sync.warnings`, making it appear as a new public API export. 

I thought the CI is informational in this case due to deprecation note, thanks for catching it!

leongdl (approved):

mwiebe (approved):

PR #994, opened 2026-02-05
fix: use string list filter for job download

kenyrish-amazon:
The sync-output CLI command was making two separate SearchJobs API calls to find active jobs - one for statuses READY/ASSIGNED/STARTING and another for SCHEDULED/RUNNING. If a job transitioned between these status groups between the two calls, it could be missed, causing incorrect output downloads. 

Use the stringListFilter with ANY_EQUALS operator to query all 5 active statuses in a single API call, eliminating the race condition.

Also upgraded required boto3 version to be compatible with release that added support for stringlistfilter.

Jobs will no longer be missed during incremental output downloads when they transition between active statuses.

```
Required test coverage of 80.0% reached. Total coverage: 83.25%
================================================================================================================ slowest 5 durations ================================================================================================================
4.61s call     test/unit/deadline_client/api/test_job_bundle_submission_asset_refs.py::test_create_job_from_job_bundle_with_all_asset_ref_variants
4.32s call     test/unit/deadline_client/cli/test_cli_handle_web_url.py::test_cli_handle_web_url_download_output_only_required_input
4.29s call     test/unit/deadline_client/api/test_api_farm.py::test_list_farms_paginated
4.26s call     test/unit/deadline_client/cli/test_cli_job.py::test_cli_job_download_output_stdout_with_only_required_input
3.16s call     test/unit/deadline_job_attachments/test_upload.py::TestUpload::test_asset_management_no_outputs_large_number_of_inputs_already_uploaded[2023-03-03-200]
=================================================================================================== 2145 passed, 68 skipped, 1 xfailed in 44.13s ====================================================================================================

```

Yes
N/A

No

No

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ classifiers = [
 # Applications that consume this library should be the ones that are more strictly
 # limiting dependencies if they want/need to.
 dependencies = [
-  "boto3 >= 1.39.10; python_version >= '3.9'",
+  "boto3 >= 1.42.29; python_version >= '3.9'",
   "boto3 >= 1.36.8; python_version < '3.9'",
   # Click 8.2 dropped python 3.8/3.9 support
   "click >= 8.1.7",
--- a/src/deadline/client/cli/_incremental_download.py
+++ b/src/deadline/client/cli/_incremental_download.py
@@ -95,6 +95,7 @@ def _get_download_candidate_jobs(
     # Construct the full set of jobs that may have new available downloads.
     # - Any active job (job with taskRunStatus in READY, ASSIGNED,
     #   STARTING, SCHEDULED, or RUNNING), that has at least one SUCCEEDED task.
+    # Use stringListFilter with ANY_EQUALS to query all statuses in a single API call
     download_candidate_jobs = {
         job["jobId"]: job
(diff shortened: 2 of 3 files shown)


leongdl (dismissed):
Thank you for fixing this.

mwiebe (dismissed):

leongdl:
I think we should bump the boto3 so the correct version with the updated service model is part of the dependency. Otherwise, the user may have an older service model and fail to use the updated string list filter.

sakshie95 (dismissed):
Awesome, thank you for fixing this! I would keep an eye out on some of our xfail tests as a follow-up to see if it helps them https://github.com/aws-deadline/deadline-cloud/blob/5bba469e07e64ed228a218e0b37743e4455bde53/test/integ/cli/test_cli_incremental_download.py#L249

leongdl:
https://github.com/boto/boto3/blob/develop/CHANGELOG.rst#L158

> I think we should bump the boto3 so the correct version with the updated service model is part of the dependency. Otherwise, the user may have an older service model and fail to use the updated string list filter.

leongdl (approved):

sakshie95 (approved):

mwiebe (approved):

PR #998, opened 2026-02-11
chore: add test verifying we only use specific Qt modules

crowecawcaw:
We're considering bundling Qt with some distributions of this software. Qt uses many licenses across its modules. We need to know which modules we're using and be alerted if we're every adding another one.

Add a unit test.

More confidence in the Qt modules we're depending on 

Removed a module from the allow list and verified the test failed.

n/a

No

No

No

----


test/unit/deadline_client/ui/test_only_qtcore_qtwidgets_qtgui_used.py
```diff
@@ -0,0 +1,68 @@
...
+    disallowed = imported - ALLOWED_QT_MODULES
+    assert not disallowed, f"Disallowed Qt modules imported: {disallowed}"
+
+
+class TestOnlyQtCoreWidgetsGuiUsed:
+    """Tests verifying only QtCore, QtWidgets, and QtGui are used."""
+
+    def test_config_dialog_imports_only_allowed_qt_modules(self):
+        """Verify deadline_config_dialog module only imports QtCore, QtGui, QtWidgets."""
+        imported = _get_imported_qt_modules("dialogs.deadline_config_dialog")
```
joel-wong-aws: Instead of manually adding the paths here, can we list the contents of the UI directory and include every file within it? Then we don't have to remember to add this test for additional modules.

test/unit/deadline_client/ui/test_only_qtcore_qtwidgets_qtgui_used.py
```diff
@@ -0,0 +1,67 @@
...
+This ensures we can distribute a minimal Qt build without unnecessary modules.
+"""
+
+import subprocess
+import sys
+
+import pytest
+
+try:
+    from deadline.client.ui.dialogs import deadline_config_dialog  # noqa: F401
```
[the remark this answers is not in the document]
waninggibbon: Can you expand on why we are importing this? The noqa comment makes me think this is intentional but it isn't clear why we would need this for this test.

test/unit/deadline_client/ui/test_only_qtcore_qtwidgets_qtgui_used.py
```diff
@@ -0,0 +1,78 @@
...
+except ImportError:
+    pytest.importorskip("deadline.client.ui.dialogs.deadline_config_dialog")
+
+ALLOWED_QT_MODULES = {"QtCore", "QtGui", "QtWidgets", "QtOpenGL", "QtOpenGLWidgets"}
+
+# Test imports in subprocess to get clean module state
+_TEST_SCRIPT = """
+import sys
+module_to_import = sys.argv[1]
+exec(f"import {module_to_import}")
```
larrygao001, line 28: exe() on sys.argv doesn't look like a good pattern, can we do something like this:

```python
importlib.import_module(module_to_import)
```

test/unit/deadline_client/ui/test_only_qtcore_qtwidgets_qtgui_used.py
```diff
@@ -0,0 +1,78 @@
...
+from typing import List, Set
+
+import pytest
+
+try:
+    from deadline.client.ui.dialogs import deadline_config_dialog  # noqa: F401
+except ImportError:
+    pytest.importorskip("deadline.client.ui.dialogs.deadline_config_dialog")
+
+ALLOWED_QT_MODULES = {"QtCore", "QtGui", "QtWidgets", "QtOpenGL", "QtOpenGLWidgets"}
```
larrygao001, line 22: This is different than what line 4 claims.

test/unit/deadline_client/ui/test_only_qtcore_qtwidgets_qtgui_used.py
```diff
@@ -0,0 +1,78 @@
...
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        pytest.fail(f"Import of {fully_qualified_module} failed: {result.stderr}")
+    output = result.stdout.strip()
+    return set(output.split(",")) if output else set()
+
+
+_ui_modules = _get_ui_module_paths()
```
larrygao001, line 62: Maybe move this inside a fixture or use pytest_generate_tests?

test/unit/deadline_client/ui/test_only_qtcore_qtwidgets_qtgui_used.py
```diff
@@ -0,0 +1,78 @@
...
+    return modules
+
+
+def _get_imported_qt_modules(fully_qualified_module: str) -> Set[str]:
+    """Run import in subprocess and return set of Qt modules that were loaded."""
+    result = subprocess.run(
+        [sys.executable, "-c", _TEST_SCRIPT, fully_qualified_module],
+        capture_output=True,
+        text=True,
+    )
```
larrygao001, line 55: need a timeout in case it hangs

test/unit/deadline_client/ui/test_only_qtcore_qtwidgets_qtgui_used.py
```diff
@@ -0,0 +1,78 @@
...
+
+def _get_ui_module_paths() -> List[str]:
+    """Discover all importable Python modules under deadline.client.ui."""
+    ui_pkg = importlib.import_module("deadline.client.ui")
+    assert ui_pkg.__file__ is not None
+    ui_dir = Path(ui_pkg.__file__).parent
+    modules = []
+    for py_file in sorted(ui_dir.rglob("*.py")):
+        if py_file.name.startswith("_"):
+            continue
```
larrygao001, line 42: Can we document why this is skipped?

crowecawcaw:
We have other mechanisms in place now. Closing.

PR #1013, opened 2026-02-26
fix: handle-web-url installs from command on PATH on linux

crowecawcaw:
On Linux, `deadline handle-web-url --install` detected the wrong path for the `deadline` command if `deadline` invoked via PATH instead of executed from an absolute or relative path. As a result, the handler did nothing.

Update the path detection. Removed a duplicate `Type=Application` line.

Fixes the URL handler when `deadline` is run via PATH.

Added unit tests. Manually installed the URL handler when running `deadline` via PATH, clicked the Download job output button in DCM, and verified a terminal window popped up to download the outputs.

n/a

No

No

No
----

### DIFF
--- a/src/deadline/client/cli/_deadline_web_url.py
+++ b/src/deadline/client/cli/_deadline_web_url.py
@@ -184,10 +184,13 @@ def install_deadline_web_url_handler(all_users: bool) -> None:
                 f"Failed to install the handler for {DEADLINE_URL_SCHEME_NAME} URLs: update-desktop-database is not installed."
             )
 
-        # Get the CLI program path. Use shutil.which to resolve the full path when
-        # invoked via PATH (e.g. bare "deadline"), falling back to os.path.abspath
-        # for relative or absolute paths (e.g. "./deadline" or "/opt/bin/deadline").
-        deadline_cli_program = shutil.which(sys.argv[0]) or os.path.abspath(sys.argv[0])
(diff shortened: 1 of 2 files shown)


src/deadline/client/cli/_deadline_web_url.py
```diff
@@ -176,16 +176,18 @@ def install_deadline_web_url_handler(all_users: bool) -> None:
...
             raise DeadlineOperationError(
                 f"Failed to install the handler for {DEADLINE_URL_SCHEME_NAME} URLs: update-desktop-database is not installed."
             )
 
-        # Get the CLI program path
-        deadline_cli_program = os.path.abspath(sys.argv[0])
+        # Get the CLI program path. Use shutil.which to resolve the full path when
+        # invoked via PATH (e.g. bare "deadline"), falling back to os.path.abspath
+        # for relative or absolute paths (e.g. "./deadline" or "/opt/bin/deadline").
+        deadline_cli_program = shutil.which(sys.argv[0]) or os.path.abspath(sys.argv[0])
```
mwiebe: Won't the `shutil.which` handle anything `os.path.abspath` is for? On Windows it's normal to run programs from the current working directly, but I think that's not a usual behavior for Linux and we probably don't want to add that.

mwiebe (approved):

Cherie-Chen (approved):

PR #1015, opened 2026-02-27
feat: auto-select farm and queue when only one is available

rickrams:
When no default farm or queue is configured, the CLI and GUI submit dialog now automatically select the farm/queue if only one is available to the user. If multiple exist, the existing behavior is preserved.

CLI: auto-selection in _apply_cli_options_to_config before raising the missing farm/queue error.

GUI: background thread in the submit dialog checks on startup and sets defaults via a signal to refresh the UI.

new installs had a poor experience as they had to go into settings and pick a farm and queue. 

As most new installs will have one farm and one queue, and generally many installs have one farm, this adds automatic selection of defaults when not set for both CLI and GUI usage.

Tested manually against various scenarios.  Includes added tests

Yes
No
No

### Does this change impact security? No

----


src/deadline/client/cli/_common.py
```diff
@@ -87,6 +88,44 @@ def wraps(ctx: click.Context, *args, **kwargs):
     return wraps
 
 
+_SETTING_FARM_ID = "defaults.farm_id"
```
leongdl: Super nit - https://github.com/search?q=repo%3Aaws-deadline%2Fdeadline-cloud%20defaults.farm_id&type=code

These constants are everywhere and could use a good refactoring to 1 constant place :) 

Out of scope for Rick though.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -49,6 +50,9 @@
 
 logger = logging.getLogger(__name__)
 
+_SETTING_FARM_ID = "defaults.farm_id"
```
leongdl: Now these consts are repeated, which over time ends up as drifted code

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -299,6 +307,51 @@ def refresh_deadline_settings(self):
...
+        """Auto-select farm/queue in a background thread if only one is available."""
+        if self.deadline_authentication_status.api_availability is not True:
+            return
+        if get_setting(_SETTING_FARM_ID) and get_setting(_SETTING_QUEUE_ID):
+            return
+        if getattr(self, "_auto_select_in_progress", False):
+            return
+        self._auto_select_in_progress = True
+
+        _threading.Thread(target=self._do_auto_select, daemon=True).start()
```
leongdl: Interesting change, I was expecting with the QSignal we would do an async signal slot to set this. (Very Qt thing)

leongdl: I was thinking we can fire a signal, and then there's a slot connected just like the change at 234 above. This is how Qt handles asynchronous patterns via the run loop instead of threads.

src/deadline/client/cli/_common.py
rickrams: Updated anyway to improve.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -299,6 +306,51 @@ def refresh_deadline_settings(self):
...
+
+    def _do_auto_select(self):
+        """Background worker that auto-selects farm/queue if only one exists."""
+        try:
+            changed = self._try_auto_select_farm()
+            changed = self._try_auto_select_queue() or changed
```
larrygao001: Nit: This is logically correct but the naming is very confusing. Can we do something like this:

```python
 farm_changed = self._try_auto_select_farm()
 queue_changed = self._try_auto_select_queue()
 if farm_changed or queue_changed:
     self._auto_select_complete.emit()
```

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -299,6 +306,51 @@ def refresh_deadline_settings(self):
...
 
+    def _auto_select_defaults(self):
+        """Auto-select farm/queue in a background thread if only one is available."""
+        if self.deadline_authentication_status.api_availability is not True:
+            return
+        if get_setting(_SETTING_FARM_ID) and get_setting(_SETTING_QUEUE_ID):
+            return
+        if getattr(self, "_auto_select_in_progress", False):
+            return
+        self._auto_select_in_progress = True
```
larrygao001: Nit: There's a theoretical race window between the flag check and set, but given this is triggered by user actions with natural delays, it's unlikely to be hit in practice. No action needed unless you want to be
extra defensive.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
rickrams: discussed separately this is an established pattern as is.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
rickrams: fixed

src/deadline/client/config/config_file.py
```diff
@@ -51,6 +51,10 @@
 __config_file_path = None
 __config_mtime = None
 
+# Setting name constants
+_SETTING_FARM_ID = "defaults.farm_id"
```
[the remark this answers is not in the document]
rickrams: fixed

src/deadline/client/config/config_file.py
```diff
@@ -51,6 +51,10 @@
...
+_SETTING_QUEUE_ID = "defaults.queue_id"
```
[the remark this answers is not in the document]
rickrams: fixed

rickrams:
The test failure on window / 3.12 is in test_deadline_controller.py which was introduced by PR #979 and is unrelated to our changes.

baxeaz:
> The test failure on window / 3.12 is in test_deadline_controller.py which was introduced by PR #979 and is unrelated to our changes.

@rickrams FYI Justin pushed a fix a bit ago - https://github.com/aws-deadline/deadline-cloud/pull/1031 - I merged into your branch and things now seem to be passing

andychoquette (commented):
When I explicitly select a queue, then attempt to change the farm, i see a permutation of the following error:

```
Encountered an error:
An error occurred (ResourceNotFoundException) when calling the ListStorageProfilesForQueue operation: Resource of type queue with id queue-04f30a1cf3654131afc63af1830bc621 does not exist.
```

https://github.com/user-attachments/assets/8b47d3ab-5e4e-45d7-a4c5-acea114dc299

Not 100% sure on the root cause of this, and I can't pin it down to a specific change in this CR.

rickrams:
> When I explicitly select a queue, then attempt to change the farm, i see a permutation of the following error:
> 
> ```
> Encountered an error:
> An error occurred (ResourceNotFoundException) when calling the ListStorageProfilesForQueue operation: Resource of type queue with id queue-04f30a1cf3654131afc63af1830bc621 does not exist.
> ```
> 
> Screen.Recording.2026-03-10.at.9.44.36.AM.mov
> 
> Not 100% sure on the root cause of this, and I can't pin it down to a specific change in this CR.

not sure how this could be this CR? Likely cause is that a persisted queue got deleted at some point for that farm?

crowecawcaw:
Pushed a follow-up commit (`aa6b57f`) addressing the open review feedback, plus a merge of latest `mainline` so CI runs against the `test_deadline_controller` fix from #1031.

### GUI: switch from `threading.Thread` to the `AsyncTaskRunner` signal/slot pattern
Per @leongdl's comments and the "Qt and Calling AWS APIs" guidance in `DEVELOPMENT.md`, the auto-select no longer spins up a raw daemon thread and emits a signal from it. It now uses the existing `AsyncTaskRunner`:

- The background worker (`_resolve_auto_select`) **only calls AWS APIs and returns plain data** — it never reads/writes settings or touches widgets.
- The result is applied on the **Qt main thread** in `_on_auto_select_resolved` via a queued connection.
- A single `operation_key` gives **latest-wins cancellation**, replacing the `_auto_select_in_progress` flag (which @larrygao001 flagged for a theoretical check-then-set race).

### Hardening against stale / out-of-order results
Because farm and queue selection is a stateful two-step flow, the result slot re-checks current settings before writing:
- A resolved farm is discarded if a farm was set in the meantime.
- A resolved queue is applied **only if it was resolved under the farm that is still configured** — so a queue from farm A can never be written while farm B is selected. This is the class of bug behind the `ResourceNotFoundException ... queue ... does not exist` that @andychoquette saw when switching farms. (The config dialog's own `self.changes` queue-clearing on farm change is separate/pre-existing and out of scope here.)
- An already-configured queue is never overwritten.

### Setting-name constants
Promoted the duplicated `"defaults.farm_id"` / `"defaults.queue_id"` literals to public `config.SETTING_FARM_ID` / `SETTING_QUEUE_ID` and referenced them from both the CLI and UI (@leongdl's drift comment). This also resolves the code-scanning "unused global variable" warnings, since the constants are now the canonical names used throughout `config_file.py` and re-exported from the `config` package.

### Tests
- **New pytest-qt tests** (`test_submit_dialog_auto_select.py`) drive the real `AsyncTaskRunner` + thread pool + signal/slot path: single farm/queue, farm-only, multiple/zero no-ops, API error, API-unavailable, already-configured — plus the race / unexpected-event-sequence cases: no duplicate concurrent run, stale farm result discarded, stale queue-for-old-farm discarded, queue applied only when the farm still matches, queue not overwritten when already set.
- **CLI integration tests** for auto-select through `_apply_cli_options_to_config` (single farm+queue auto-filled; multiple-farms and multiple-queues still raise; explicit `--farm-id` skips the lookup).

`hatch run lint` (ruff + mypy) is clean and the unit/UI suites pass locally.

crowecawcaw:
Follow-up on CI (force-pushed `0bd0de1`):

- **`check-api-changes`** initially failed because the first revision re-exported the setting-name constants as public `config.SETTING_*` and imported `AsyncTaskRunner` under a public name in the (public) submit-dialog module — that registered 7 new public API entries. Fixed by keeping the constants internal: they're consolidated onto the single `config_file._SETTING_FARM_ID` / `_SETTING_QUEUE_ID` and referenced via underscore-prefixed names in the CLI and UI, and `AsyncTaskRunner` is imported as `_AsyncTaskRunner`. `check-api-changes` now passes (verified locally with `validate-api-snapshot`: "No API changes detected"). This still satisfies the "one canonical place" drift concern and clears the code-scanning unused-global warnings.

- **`UI Tests (Windows)`** is failing on an unrelated, pre-existing xa11y flake (`PlatformError 0x80040201 "An event was unable to invoke any of the subscribers"` / `TimeoutError: No accessibility app found for PID ...`) in `TestHostRequirementsControls::test_default_and_custom_radios_present` and `test_config_gui_log_level`. Neither test exercises auto-select or the submit dialog's farm/queue logic. The same Windows job flakes intermittently on `mainline` (e.g. commits `53b648d8`, `e90e1e7a`) and on other PRs at the same time, and `UI Tests (macOS)` passed on this exact commit. I've re-run the failed job.

The full Python unit-test matrix (all OSes + 3.8–3.14) — which is where the new pytest-qt and CLI tests run — is green.

src/deadline/client/cli/_common.py
```diff
@@ -87,6 +90,31 @@
     return wraps
 
 
+def _auto_select_farm(config: Optional[ConfigParser] = None) -> Optional[str]:
+    """Auto-select farm ID if exactly one farm is available."""
+    try:
+        farms = _api.list_farms(config=config).get("farms", [])
+        if len(farms) == 1:
+            return farms[0]["farmId"]
+    except Exception:
```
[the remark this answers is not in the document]
waninggibbon: nit: maybe add a comment about defaulting to not autoselecting if something goes wrong but this is obvious enough from the implementation so not blocking.

src/deadline/client/config/config_file.py
```diff
@@ -51,6 +51,11 @@
 __config_file_path = None
 __config_mtime = None
 
+# Canonical setting-name constants. Use these everywhere instead of repeating the
```
waninggibbon, line 54: nit: since we import these everywhere, the leading underscore seems incorrect?

waninggibbon (approved):

andychoquette (approved):

PR #1017, opened 2026-02-27
feat: Add proxy and CA bundle integration tests using network namespace isolation

prestomation:
Add a wrapper to run existing integration tests through an HTTP CONNECT proxy inside a Linux network namespace, verifying that HTTPS_PROXY and AWS_CA_BUNDLE environment variables are properly respected by all AWS SDK calls. This isn't used yet, but I want to get this code in here so we can test it and turn it on in PR with a smaller change later, mostly because integration tests don't run pre-merge right now.

The network namespace (unshare --net) has no internet routes — the only path to AWS is through a socat bridge to a lightweight Python CONNECT proxy on a Unix domain socket. Any code that bypasses HTTPS_PROXY fails immediately with 'Network is unreachable'. Pre-flight checks verify isolation before tests run.

A proxy report prints per-host connection counts and bytes relayed after tests complete, failing if zero connections went through the proxy.

New files:
- scripts/run_proxy_integ_tests.sh: self-contained wrapper that starts proxy, creates namespace, bridges credentials, runs pytest
- test/integ/proxy_ca_bundle/connect_proxy.py: HTTP CONNECT proxy on Unix socket with stats tracking (per-host counts, bytes relayed)
- scripts/cleanup_test_queues.py: cleanup orphaned test queues
- docs/design/proxy-and-ca-bundle-integration-tests.md: design doc

Usage:
  hatch run integ:proxy-test sudo -E env "PATH=$PATH" bash scripts/run_proxy_integ_tests.sh

We need better confidence that our library works with standard AWS proxy/cert management configuration and that we don't break this over time

See above

Higher confidence about our proxy support

In my fork I had a copy of the integ workflow that tested against my own account.
You can see that action here: https://github.com/prestomation/deadline-cloud/actions/runs/22466284530/job/65084372176

- Have you run the unit tests? y
- Have you run the integration tests? y

There is a design doc for this feature but no changes to the core code

OnlyCI dependencies

I think the answer for this question is 'no'?

no

no

----


hatch.toml
```diff
@@ -23,6 +23,7 @@ pre-install-commands = [
 
 [envs.integ.scripts]
 test = "pytest --xfail-tb --no-cov -vvv --numprocesses=1 {args:test/integ}"
+proxy-test = "sudo -E env PATH=$PATH bash scripts/run_proxy_integ_tests.sh {args:test/integ/cli/}"
```
crowecawcaw, line 26: To double check, this does not run this test automatically in CI right? It looks like this PR adds the command but does not hook it into existing test runs.

crowecawcaw (approved):

hatch.toml
prestomation: yeah, sorry that was unclear. we need a discussion about how to get these tests turned on in PR

Cherie-Chen (approved):

PR #1029, opened 2026-03-09
feat!: add storage profile path mapping to job download-output

### LINKED ISSUE #829: Enhancement: job attachments interface does not use storage profiles by default
### Describe the bug

The S3AssetManager class which is the public interface for job attachments functionality does not use storage profiles for default. 

The S3AssetManager class takes a `farm_id`, `queue_id`, and `session` in its constructor. Then later, the `prepare_paths_for_upload` method takes `storage_profile` as an optional parameter. It's not terrible difficult to fetch the storage profile and pass it in, but passing in storage profiles is easy to miss, possible to mess up, and should be done automatically. If S3AssetManager already has the queue ID, it can find the storage profile to use internally and automatically. 

https://github.com/aws-deadline/deadline-cloud/blob/dc67589b9470c3b1c60a7f538eef8529f33f9767/src/deadline/job_attachments/upload.py#L1264

https://github.com/aws-deadline/deadline-cloud/blob/dc67589b9470c3b1c60a7f538eef8529f33f9767/src/deadline/job_attachments/upload.py#L861-L868

More broadly, it'd be nice if there were a single `create_job` function that wrapped up the job inputs and job attachments together so that consumers of the libraries can simply pass in a list of paths with the job and don't need to know how job attachments, shared storage, and storage profiles all interact.

### Expected Behaviour

Job attachments uses a queue's storage profile by default.

### Current Behaviour

Job attachments needs to be opted in to using storage profiles.

### Reproduction Steps

n/a

### Environment

n/a

leongdl:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/829

### Download behavior change:
This PR is a behavioral change, not a breaking API change. When both the local machine and the job have matching storage profiles with aligned location names, download-output now applies path mapping automatically instead of prompting the user to manually type a new root path. Users who relied on the manual prompt to enter a custom path different from their storage profile can use --ignore-storage-profiles to restore the previous behavior.

Note: When matching storage profiles exist on both sides, the previous manual OS-mismatch prompt is replaced by automatic path mapping. This is an intentional UX improvement, not a regression. The --ignore-storage-profiles flag provides an explicit opt-out.

`deadline job download-output` had no storage profile support. When downloading job output submitted from a different OS (e.g., Windows → Linux), the command prompted the user to manually type a new root path for every mismatched asset root. `deadline queue sync-output` already handled this automatically via storage profiles and path mapping, but `download-output` did not.

Added `--ignore-storage-profiles` option to `deadline job download-output` and automatic storage profile resolution from the config (`settings.storage_profile_id`).

The implementation is split into single-responsibility helper functions in a new `_job_download_helpers.py` module:

- `_resolve_storage_profiles()` — fetches local and job storage profiles, returns a `ResolvedStorageProfiles` dataclass (or `None` when mapping is not needed). All mismatch cases produce warnings and fall back to the existing manual prompt rather than erroring.
- `_apply_path_mappings_to_roots()` — applies trie-based path mapping rules to remap output root directories via `OutputDownloader.set_root_path()`

Also refactored `_generate_path_mapping_rules()` in `_path_mapping.py` to accept both `StorageProfile` dataclass and raw boto3 dicts (backward compatible via `_normalize_storage_profile()`). Removed the stale TODO from `OutputDownloader` since both download commands now handle path mapping at the CLI layer.

Mismatch cases produce clear user-facing warnings:
- No local profile but job has one → warns and recommends `deadline config`
- Both profiles exist but location names don't match → warns about aligning location names
- Local profile exists but job has none → warns that mapping will be skipped

- Users downloading cross-OS job output with matching storage profiles no longer need to manually enter root paths — mapping is automatic
- Existing behavior is fully preserved: when no storage profiles are configured or profiles don't match, the manual prompt fallback works exactly as before, with added guidance
- `--ignore-storage-profiles` provides an explicit opt-out for same-machine workflows
- No changes to `OutputDownloader` public interface or `_incremental_download.py`

- **Unit tests** (`test_job_download_helpers.py`): 17 tests covering `_resolve_storage_profiles` (all decision matrix rows, API error propagation), `_apply_path_mappings_to_roots` (platform-specific POSIX/Windows tests), `_generate_path_mapping_rules` with `StorageProfile` dataclass inputs, and CLI-level `--ignore-storage-profiles` wiring
- **CLI end-to-end tests** (`test_cli_job_download_storage_profiles.py`): 7 tests covering every code path:
  - Case 1: Both profiles match → automatic mapping
  - Case 2: Both profiles exist, no matching locations → warning, no mapping
  - Case 3: No local profile, job has one → warning, manual fallback
  - Case 4: Neither has a profile → no mapping, direct download
  - Case 5: Local profile, job has none → warning, skip mapping
  - Case 6: `--ignore-storage-profiles` → skip everything
  - Case 7: Same profile both sides → no mapping needed
- **Existing tests**: all 39 `test_cli_job.py` tests pass, all 15 `test_path_mapping.py` tests pass, all 12 `test_cli_queue_incremental_download.py` tests pass
- `hatch run fmt` — clean
- `hatch run lint` (ruff + mypy) — clean, no issues in 309 source files
- `hatch run test test/unit/` — all pass, 83%+ coverage

- Design doc: `docs/design/storage-profile-cli-support.md` (includes full analysis, decision matrix, staged implementation plan, and 7 test cases with call stacks)
- Docstrings added to all new functions
- `OutputDownloader` docstring updated to replace stale TODO

No. The `_generate_path_mapping_rules()` function signature was broadened (accepts `Union[StorageProfile, dict]` instead of just `dict`) but this is backward compatible — all existing callers passing dicts continue to work unchanged. The function is underscore-prefixed (private API).

No. The change only affects path mapping logic for download destinations. No new file permissions, no new credential handling, no new network calls beyond the existing `get_storage_profile_for_queue` API.

---

---- 
### 9.1 Test Cases

Each case traces the full code path from CLI entry through to download. All cases are implemented in `test/unit/deadline_client/cli/test_cli_job_download_storage_profiles.py`.

#### Case 1: Both profiles exist, location names match → automatic mapping

Local machine has `sp-local-111` (Linux, locations: `shared=/mnt/shared`, `temp=/tmp/render`).
Job was submitted with `sp-job-222` (Windows, locations: `shared=Z:\shared`, `temp=C:\temp\render`).

```
job_download_output()
  _validate_storage_profile_options(None, False)           → passes
  _download_job_output(..., ignore_storage_profiles=False)
    _resolve_storage_profiles(config, deadline, farm, queue, job, False)
      config_file.get_setting("settings.storage_profile_id") → "sp-local-111"
      job.get("storageProfileId")                            → "sp-job-222"
      Both truthy → fetch both via api.get_storage_profile_for_queue()
      → ResolvedStorageProfiles(job_profile=sp-job-222, local_profile=sp-local-111)
    resolved is not None → enters `if resolved:` branch
    _generate_path_mapping_rules(sp-job-222, sp-local-111)
      _normalize_storage_profile() on each
      IDs differ → match locations by name
      "shared" matches → PathMappingRule(WINDOWS, "Z:\shared", "/mnt/shared")
      "temp" matches   → PathMappingRule(WINDOWS, "C:\temp\render", "/tmp/render")
      → [rule1, rule2]
    _apply_path_mappings_to_roots(downloader, roots, [rule1, rule2])
      _PathMappingRuleApplier([rule1, rule2])  → builds trie
      transform("C:\temp\render") → "/tmp/render" → set_root_path()
      transform("Z:\shared")      → "/mnt/shared" → set_root_path()
    download proceeds to mapped local paths
```

Expected: roots are remapped, user sees "Using storage profile: Local Linux Profile" and "Mapping root: ..." messages. Download goes to `/mnt/shared` and `/tmp/render`.

#### Case 2: Both profiles exist, location names don't match → no rules, no mapping

Local machine has `sp-local-111` (locations: `shared`, `temp`).
Job was submitted with `sp-job-333` (locations: `ProjectFiles`, `OutputDir`).

```
job_download_output()
  _download_job_output(..., ignore_storage_profiles=False)
    _resolve_storage_profiles(...)
      Both IDs present → fetch both → ResolvedStorageProfiles
    resolved is not None → enters `if resolved:` branch
    _generate_path_mapping_rules(sp-job-333, sp-local-111)
      IDs differ → match locations by name
      "ProjectFiles" not in {"shared", "temp"}, "OutputDir" not in {"shared", "temp"}
      → []  (empty)
    _apply_path_mappings_to_roots(downloader, roots, [])
      `if not rules:` → prints warning about mismatched location names → return
    output_paths_by_root re-read (unchanged)
    user sees root confirmation prompt (if not auto_accept)
    download proceeds to original unmapped paths
```

Expected: no mapping applied. User sees "Using storage profile: ..." and a warning about mismatched location names recommending alignment. If OS mismatch exists, the manual root editing prompt appears as before.

#### Case 3: Job has profile, local machine does not → warning, manual fallback

Local machine has no storage profile configured.
Job was submitted with `sp-job-222`.

```
job_download_output()
  _download_job_output(..., ignore_storage_profiles=False)
    _resolve_storage_profiles(...)
      config_file.get_setting("settings.storage_profile_id") → ""
      job.get("storageProfileId")                            → "sp-job-222"
      Hits: `if not local_storage_profile_id and job_storage_profile_id:`
      → click.echo("Warning: ...no local storage profile is configured...")
      → None
    resolved is None → enters `else:` branch
    if OS mismatch → manual prompt fallback
    if same OS → download to original paths
```

Expected: warning printed recommending storage profile setup, then falls back to manual prompt if OS mismatch exists. No error thrown.

#### Case 4: Neither submitter nor downloader has a profile → no mapping, direct download

Local machine has no storage profile. Job was submitted without one.

```
job_download_output()
  _download_job_output(..., ignore_storage_profiles=False)
    _resolve_storage_profiles(...)
      config_file.get_setting("settings.storage_profile_id") → ""
      job.get("storageProfileId")                            → None
      Hits: `if not local_storage_profile_id and not job_storage_profile_id:`
      → None
    resolved is None → enters `else:` branch
    for each asset_root:
      root_path_format matches host format (same OS)
      no OS mismatch → no prompt
    download proceeds to original paths
```

Expected: same-machine case. No storage profiles, no path mapping, no prompts. Identical to pre-change behavior.

#### Case 5: Local profile configured, job submitted without one → warning, skip mapping

Local machine has `sp-local-111`. Job was submitted without a storage profile (e.g., older job or tool that didn't set one).

```
job_download_output()
  _download_job_output(..., ignore_storage_profiles=False)
    _resolve_storage_profiles(...)
      config_file.get_setting("settings.storage_profile_id") → "sp-local-111"
      job.get("storageProfileId")                            → None
      Hits: `if local_storage_profile_id and not job_storage_profile_id:`
      → click.echo("Warning: ...job was submitted without one. Path mapping will be skipped.")
      → None
    resolved is None → enters `else:` branch
    if OS mismatch → manual prompt fallback
    if same OS → download to original paths
```

Expected: warning printed, no mapping. Falls back to existing behavior. If cross-OS, user gets the manual root path prompt.

#### Case 6: `--ignore-storage-profiles` flag → skip everything

Both profiles may exist, but user explicitly opts out.

```
job_download_output(..., ignore_storage_profiles=True)
  _validate_storage_profile_options(None, True) → passes
  _download_job_output(..., ignore_storage_profiles=True)
    _resolve_storage_profiles(..., ignore_storage_profiles=True)
      → return None immediately (first line of function)
    resolved is None → enters `else:` branch
    if OS mismatch → manual prompt fallback
    download to original unmapped paths
```

Expected: all storage profile logic bypassed. Same behavior as pre-change code. Use when submitting and downloading from the same machine.

#### Case 7: Same storage profile on both sides → no mapping needed

Job was submitted with `sp-local-111`. Local machine also has `sp-local-111`.

```
job_download_output()
  _download_job_output(..., ignore_storage_profiles=False)
    _resolve_storage_profiles(...)
      Both IDs present, both "sp-local-111" → fetch both → ResolvedStorageProfiles
    resolved is not None → enters `if resolved:` branch
    _generate_path_mapping_rules(sp-local-111, sp-local-111)
      storageProfileId matches → return []
    _apply_path_mappings_to_roots(downloader, roots, [])
      `if not rules: return` → no-op
    download proceeds to original paths
```

Expected: same profile means same paths. No mapping, no prompts. Download to original locations.

### DIFF
--- a/docs/design/storage-profile-cli-support.md
+++ b/docs/design/storage-profile-cli-support.md
@@ -334,7 +334,7 @@ Both download commands already handle path mapping at the CLI layer using `_Path
 - `queue sync-output` transforms every individual file path (joins `rootPath + relativePath`, then calls `strict_transform()`) inside `_download_all_manifests_with_absolute_paths()`. It does not use `OutputDownloader` at all — it has its own manifest-based download pipeline.
 - `job download-output` (this proposal) transforms root paths via `OutputDownloader.set_root_path()`, then `OutputDownloader` joins root + relative internally during download.
 
-These are equivalent in outcome because each manifest has a single root path — transforming the root produces the same final absolute paths as transforming each `root + relative` individually. The proposed approach uses `set_root_path()` because `download-output` already depends on `OutputDownloader`, and rewriting it to use the manifest pipeline would be a larger change with no functional benefit.
+These are equivalent in outcome because each manifest has a single root path — transforming the root produces the same final absolute paths as transforming each `root + relative` individually. **Update (§13):** This equivalence only holds when all rules match exactly at the root level. With nested file system locations, a rule can match deeper than the root, and root-only transformation would miss it. The implementation now uses absolute-path transformation (joining root + relative before applying rules) when path mapping rules exist, matching `sync-output` behavior. See §13 for details.
 
 Since both download commands handle path mapping externally using the same trie, this TODO is stale and should be removed as part of Stage 3 (§8.3). `OutputDownloader` remains a download-only concern with no path mapping responsibility.
 
@@ -923,3 +923,86 @@ Storage profile affects path grouping during upload:
 
 - `hash_assets_and_create_manifest()` hashes files using xxHash XXH128 with a SQLite-backed cache
 - `upload_assets()` uploads to S3 using Content-Addressable Storage (CAS), with an S3CheckCache to prevent redundant uploads
+
+
+---
+
+## 13. Revision: Absolute-Path Transformation (Option E)
+
+### 13.1 Problem
+
+During code review, @mwiebe identified that transforming only root paths via `set_root_path()` is not equivalent to transforming full absolute paths (`root + relative`). A path mapping rule can match at a depth deeper than the asset root. For example, with nested file system locations:
+
+**Source profile (Windows):**
+- "Projects": `C:\Projects`
+- "SpecialProjects": `C:\Projects\Special`
+
+**Destination profile (Linux):**
+- "Projects": `/mnt/projects`
+- "SpecialProjects": `/opt/special`
+
+Asset root: `C:\Projects`, relative path: `Special\data.txt`.
+
+- Root-only: transforms `C:\Projects` → `/mnt/projects`. Final: `/mnt/projects/Special/data.txt`. **Wrong.**
+- Absolute-path: transforms `C:\Projects\Special\data.txt` → `/opt/special/data.txt`. **Correct** (more specific rule wins).
+
+The Deadline Cloud API does not prevent nested file system locations in storage profiles, so this is a real scenario.
+
+### 13.2 Solution: Option E — Dual Code Path
+
+When path mapping rules exist (both profiles resolved, rules generated), bypass `OutputDownloader` and use the manifest-based download pipeline directly — the same approach `queue sync-output` uses:
+
+1. Call `get_output_manifests_by_asset_root()` to fetch manifests from S3 (same data `OutputDownloader.__init__` fetches internally)
+2. For each manifest path, join `root + relative` using the source OS path module (`ntpath` or `posixpath`)
+3. Call `strict_transform()` on the full absolute path (trie picks the most specific matching rule)
+4. Download via `download_files_from_manifests()` with the transformed absolute paths
+
+When no path mapping rules exist (no profiles, ignore flag, same profile, mismatch warnings), the existing `OutputDownloader` code path is completely unchanged.
+
+### 13.3 Implementation
+
+**New function:** `_transform_manifests_to_absolute_paths()` in `_job_download_helpers.py`
+
+```python
+def _transform_manifests_to_absolute_paths(
+    manifests_by_root: dict[str, list[Any]],
+    rules: list[PathMappingRule],
+    source_os_family: StorageProfileOperatingSystemFamily,
+) -> dict[str, Any]:
+```
+
+- Joins root + relative using `ntpath` (Windows source) or `posixpath` (POSIX source)
+- Applies `strict_transform()` on each absolute path
+- Merges all manifests into a single result keyed by `""` (empty string)
+- `download_files_from_manifests()` with key `""` works because `Path("").joinpath("/abs/path")` == `Path("/abs/path")`
+
+**Updated flow in `_download_job_output()`:**
+
+```
+resolved = _resolve_storage_profiles(...)
+if resolved:
+    rules = _generate_path_mapping_rules(...)
+    if rules:
+        manifests = get_output_manifests_by_asset_root(...)  # same S3 data
+        mapped = _transform_manifests_to_absolute_paths(manifests, rules, ...)
+        download_files_from_manifests(mapped)                # absolute-path download
+        return  # early return, skip OutputDownloader path
+    elif different profiles:
+        warn about no matching location names
+else:
+    # existing OutputDownloader path (manual prompt, etc.) — unchanged
+```
+
+### 13.4 Test Coverage
+
+New unit tests in `test_job_download_helpers.py`:
+
+| Test | Description |
+|------|-------------|
+| `test_basic_windows_to_posix_mapping` | Standard cross-OS mapping via absolute paths |
+| `test_nested_location_picks_most_specific_rule` | Key case: nested locations, most specific rule wins |
+| `test_unmapped_paths_are_skipped` | Paths without matching rules are excluded |
+| `test_multiple_roots_merged` | Multiple asset roots merged into single download |
+| `test_empty_manifests_returns_empty` | Edge case: no paths to map |
+
+Updated CLI test `test_case1_both_profiles_match_auto_mapping` to verify the new code path calls `get_output_manifests_by_asset_root` instead of `set_root_path`.
--- a/docs/design/storage-profile-cli-support.md
+++ b/docs/design/storage-profile-cli-support.md
@@ -1,0 +1,925 @@
+# Storage Profile Support in CLI Commands
+
+## 1. Issue Reference
+GitHub Issue: [#829 - Enhancement: job attachments interface does not use storage profiles by default](https://github.com/aws-deadline/deadline-cloud/issues/829)
+
+## 2. Executive Summary
+
+This document provides an analysis of how storage profiles are used throughout the deadline-cloud CLI, tracing code paths from CLI commands down to file hashing and S3 upload/download operations. The goal is to identify gaps in storage profile support and propose enhancements.
+
+---
+
+## 3. Background
+
+### 3.1 Storage Profile Configuration
+
+Storage profiles are configured in `~/.deadline/config` under the setting `settings.storage_profile_id`. A storage profile describes the current host's file system layout (shared mount points, local scratch paths, OS family). It is scoped under `farm_id` rather than `queue_id` because the host is the same regardless of which queue it accesses within a farm. The configuration hierarchy is:
+
+```
+defaults.aws_profile_name
+  └── defaults.farm_id
+        └── settings.storage_profile_id  (depends on farm_id, not queue — host is the same across queues)
+        └── defaults.queue_id
+```
+
+**Source:** `src/deadline/client/config/config_file.py:83-87`
+
+### 3.2 CLI Commands with Storage Profile Support
+
+| Command | `--storage-profile-id` | `--ignore-storage-profiles` | Notes |
+|---------|:----------------------:|:---------------------------:|-------|
+| `deadline bundle submit` | ✅ | N/A | Reads `settings.storage_profile_id` from config or `--storage-profile-id` CLI option. Passed to `create_job_from_job_bundle()` which attaches it to the CreateJob API call and uses it for path grouping during upload. If not set, submission proceeds without a storage profile. No ignore flag needed — uploads don't require path mapping. |
+| `deadline bundle gui-submit` | N/A | N/A | GUI auto-populates a storage profile dropdown by listing profiles for the selected queue, filtered to the current OS. Pre-selects from `settings.storage_profile_id` in config. User can change selection before submitting. No ignore flag needed — uploads don't require path mapping. |
+| `deadline queue sync-output` | ✅ | ✅ | Full support (reference implementation). See §4.2 for code details. |
+| `deadline job download-output` | ❌ | ❌ | **Gap — prompts for manual path entry on OS mismatch** |
+
+### 3.3 Existing `--ignore-storage-profiles` Implementation
+
+The `deadline queue sync-output` command already implements `--ignore-storage-profiles`:
+
+**File:** `src/deadline/client/cli/_groups/queue_group.py:282-290`
+
+```python
+@click.option(
+    "--ignore-storage-profiles",
+    is_flag=True,
+    help="Ignores the storage profile configuration. Only use if all jobs in the queue "
+         "are submitted and downloaded from the same machine. Downloads all jobs to "
+         "unmapped paths regardless of operating system.\n"
+         "Default value is False.",
+    default=False,
+)
+```
+
+This option is used when:
+- All jobs are submitted and downloaded from the **same machine**
+- No path mapping is needed (same OS, same mount points)
+- Downloads go to the original unmapped paths
+
+The proposal is to extend this existing pattern to `deadline job download-output`.
+
+For a code deep dive into the upload path (job submission), see §12 (Appendix C).
+
+---
+
+## 4. Download Code Path
+
+### 4.1 `deadline job download-output` (Single Job)
+
+**File:** `src/deadline/client/cli/_groups/job_group.py:911-985`
+
+```
+CLI Command (job_download_output)
+    │
+    ├── No --storage-profile-id option!
+    │
+    └── _download_job_output()
+            │
+            ├── deadline.get_job() → get attachments metadata
+            │
+            ├── OutputDownloader.__init__()
+            │       └── get_job_output_paths_by_asset_root()
+            │
+            ├── OS Mismatch Detection:
+            │       │
+            │       └── If rootPathFormat != host format:
+            │               └── PROMPT USER for new root path (manual entry)
+            │
+            └── OutputDownloader.download_job_output()
+```
+
+#### 4.1.1 Current OS Mismatch Handling (Lines 570-595)
+
+```python
+# Check if the asset roots came from different OS
+for asset_root in asset_roots:
+    root_path_format = root_path_format_mapping.get(asset_root, "")
+    if PathFormat.get_host_path_format_string() != root_path_format:
+        click.echo(_get_mismatch_os_root_warning(...))
+        
+        # Manual prompt - NO storage profile support!
+        new_root = click.prompt(
+            "> Please enter a new root path",
+            type=click.Path(exists=False),
+        )
+        job_output_downloader.set_root_path(asset_root, os.path.expanduser(new_root))
+```
+
+**This is the gap identified in issue #829** - no automatic path mapping via storage profiles.
+
+### 4.2 `deadline queue sync-output` — Reference Implementation
+
+This is the reference implementation for storage profile support in a download command. The proposed changes in §8 follow this same pattern for `job download-output`.
+
+**File:** `src/deadline/client/cli/_groups/queue_group.py:257-420`
+
+```
+CLI Command (sync_output)
+    │
+    ├── --storage-profile-id option ✅
+    ├── --ignore-storage-profiles option ✅
+    │
+    ├── Validate storage profile exists
+    │
+    └── _incremental_output_download()
+```
+
+#### 4.2.1 Storage Profile Validation (Lines 360-385)
+
+```python
+if ignore_storage_profiles:
+    local_storage_profile_id = None
+else:
+    local_storage_profile_id = config_file.get_setting(
+        "settings.storage_profile_id", config=config
+    )
+    if not local_storage_profile_id:
+        raise DeadlineOperationError(
+            "The sync-output operation requires a storage profile..."
+        )
+    
+    # Validate profile exists
+    local_storage_profile = deadline.get_storage_profile_for_queue(
+        farmId=farm_id,
+        queueId=queue_id,
+        storageProfileId=local_storage_profile_id,
+    )
+```
+
+### 4.3 Incremental Download Implementation
+
+**File:** `src/deadline/client/cli/_incremental_download.py`
+
+```
+_incremental_output_download()
+    │
+    ├── _get_download_candidate_jobs()
+    │       └── SearchJobs API to find jobs with outputs
+    │
+    ├── _categorize_jobs_in_checkpoint()
+    │       └── Categorize jobs: added, updated, unchanged, completed, etc.
+    │
+    ├── _get_job_sessions()
+    │       └── ListSessions API for each job
+    │
+    ├── _get_storage_profiles()
+    │       └── GetStorageProfileForQueue for each unique profile
+    │
+    ├── _create_path_mapping_rule_appliers()
+    │       │
+    │       └── For each job's storage profile:
+    │               │
+    │               └── _generate_path_mapping_rules(
+    │                       source_storage_profile=job_profile,
+    │                       destination_storage_profile=local_profile
+    │                   )
+    │
+    └── Download with path mapping applied
+```
+
+### 4.4 Path Mapping Implementation
+
+**File:** `src/deadline/job_attachments/_path_mapping.py`
+
+#### 4.4.1 Rule Generation
+
+```python
+def _generate_path_mapping_rules(
+    source_storage_profile: dict[str, Any],
+    destination_storage_profile: dict[str, Any],
+) -> list[PathMappingRule]:
+    # If same profile, no mapping needed
+    if source_storage_profile["storageProfileId"] == \
+       destination_storage_profile["storageProfileId"]:
+        return []
+    
+    # Match file system locations by NAME
+    source_locations = {loc["name"]: loc for loc in source_storage_profile["fileSystemLocations"]}
+    dest_locations = {loc["name"]: loc for loc in destination_storage_profile["fileSystemLocations"]}
+    
+    # Generate rules for matching names
+    for source_name, source_loc in source_locations.items():
+        if source_name in dest_locations:
+            rules.append(PathMappingRule(
+                source_path_format,
+                source_loc["path"],
+                dest_locations[source_name]["path"],
+            ))
+```
+
+#### 4.4.2 Rule Application (Trie-based)
+
+```python
+class _PathMappingRuleApplier:
+    """
+    Uses a trie data structure for efficient path transformation.
+    
+    Example rules:
+        '/mnt/Projects -> X:\\Projects'
+        '/mnt/Projects/Special -> Y:\\'
+    
+    Path '/mnt/Projects/Special/data.txt' maps to 'Y:\\data.txt'
+    (most specific rule wins)
+    """
+```
+
+### 4.5 Download File Operations
+
+**File:** `src/deadline/job_attachments/download.py`
+
+#### 4.5.1 OutputDownloader Class (Lines 1231-1377)
+
+```python
+class OutputDownloader:
+    def __init__(
+        self,
+        s3_settings: JobAttachmentS3Settings,
+        farm_id: str,
+        queue_id: str,
+        job_id: str,
+        step_id: Optional[str] = None,
+        task_id: Optional[str] = None,
+        session_action_id: Optional[str] = None,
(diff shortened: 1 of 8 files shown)


src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,126 @@
...
+        return None
+
+    local_storage_profile_id = config_file.get_setting("settings.storage_profile_id", config=config)
+    job_storage_profile_id = job.get("storageProfileId")
+
+    if not local_storage_profile_id and not job_storage_profile_id:
+        # Same-machine case: no profiles on either side
+        return None
+
+    if not local_storage_profile_id and job_storage_profile_id:
```
leongdl, line 79: Is this true, check the queue sync output.

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,801 @@
+# Storage Profile Support in CLI Commands
```
leongdl, line 1: Please read the PR doc for a think through of the different cases.

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,978 @@
...
+
+This document provides an analysis of how storage profiles are used throughout the deadline-cloud CLI, tracing code paths from CLI commands down to file hashing and S3 upload/download operations. The goal is to identify gaps in storage profile support and propose enhancements.
+
+---
+
+## 3. Background
+
+### 3.1 Storage Profile Configuration
+
+Storage profiles are configured in `~/.deadline/config` under the setting `settings.storage_profile_id`. The configuration hierarchy is:
```
mwiebe: A useful detail to add here is that the storage profile is about the current host. That's why the configuration is under farm_id and not under queue_id. The host is the same regardless of which queue for a particular farm it's accessing.

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Modifies the downloader in-place via set_root_path().
+    """
+    if not rules:
+        return
+
+    applier = _PathMappingRuleApplier(rules)
+    for original_root in list(output_paths_by_root.keys()):
+        mapped_root = applier.transform(original_root)
+        if str(mapped_root) != original_root:
```
larrygao001: Should we normalize it first to clear the differences (e.g., trailing slashes, case sensitivity on Windows, \\\ vs /)?

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+            "The download-output operation requires a storage profile defined in the deadline "
+            "client configuration or provided with the --storage-profile-id option.\n\n"
+            "Storage profiles are used to generate path mappings when a job was submitted from a "
+            "machine with a different operating system or file system mount locations than the "
+            "machine downloading outputs. See "
+            "https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/"
+            "modeling-your-shared-filesystem-locations-with-storage-profiles.html "
+            "for more information.\n\n"
+            "If you only submit and download jobs from the same operating system and mount "
+            "locations, you can use the --ignore-storage-profiles option."
```
larrygao001: We can improve the wording to make it more actionable and clear:
```python
raise DeadlineOperationError(
    "The job was submitted with a storage profile but no local storage profile is configured.\n\n"
    "Options:\n"
    "  1. Configure a storage profile: deadline config set settings.storage_profile_id <id>\n"
    "  2. Provide one on the command line: --storage-profile-id <id>\n"
    "  3. Skip path mapping (same-machine only): --ignore-storage-profiles\n\n"
    "See https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/"
    "modeling-your-shared-filesystem-locations-with-storage-profiles.html"
)
```

src/deadline/job_attachments/_path_mapping.py
```diff
@@ -73,6 +73,22 @@ def _generate_path_mapping_rules(
...
+    profile: Union[StorageProfile, dict[str, Any]],
+) -> dict[str, Any]:
+    """Convert a StorageProfile dataclass to the dict format, or pass through if already a dict."""
+    if isinstance(profile, StorageProfile):
+        return {
+            "storageProfileId": profile.storageProfileId,
+            "osFamily": profile.osFamily.value,
+            "fileSystemLocations": [
+                {"name": loc.name, "path": loc.path, "type": loc.type.value}
+                for loc in profile.fileSystemLocations
```
larrygao001, line 86: Do we need a fallback value here since it could be None?
`(profile.fileSystemLocations or [])`

docs/design/storage-profile-cli-support.md
leongdl: Sounds good, easy to add.

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,978 @@
...
+# the profile of the submitting machine. The OS of the *downloading machine* might be different,
+# so we need to check that and apply path mapping rules in that case.
+```
+
+Both download commands already handle path mapping at the CLI layer using `_PathMappingRuleApplier` (trie-based), not inside `OutputDownloader`:
+
+- `queue sync-output` transforms every individual file path (joins `rootPath + relativePath`, then calls `strict_transform()`) inside `_download_all_manifests_with_absolute_paths()`. It does not use `OutputDownloader` at all — it has its own manifest-based download pipeline.
+- `job download-output` (this proposal) transforms root paths via `OutputDownloader.set_root_path()`, then `OutputDownloader` joins root + relative internally during download.
+
+These are equivalent in outcome because each manifest has a single root path — transforming the root produces the same final absolute paths as transforming each `root + relative` individually. The proposed approach uses `set_root_path()` because `download-output` already depends on `OutputDownloader`, and rewriting it to use the manifest pipeline would be a larger change with no functional benefit.
```
mwiebe: When implementing the `deadline queue download` command, I considered this and realized that in general, they are not equivalent. A path mapping rule can map from `root + partial`, and if you only map `root` that rule won't be applied. If you apply the rules on absoluate paths, the rule will apply.

src/deadline/job_attachments/_path_mapping.py
leongdl: Good catch on thinking about the None case! In this instance though, StorageProfile.fileSystemLocations is typed as List[FileSystemLocation] (non-optional) with default_factory=list, so it'll always be [] at worst — never None. The guard shouldn't be needed here.

src/deadline/client/cli/_groups/_job_download_helpers.py
leongdl: Good thinking, but this is actually handled already. The _PathMappingRuleApplier.transform() does all the normalization internally — it uses PureWindowsPath/PurePosixPath to split paths and lowercases parts for Windows case-insensitivity. The str(mapped_root) != original_root check on line 129 is just detecting whether transform() applied a rule (returns a Path) or passed the string through unchanged — it's not doing path equality comparison. So trailing slashes, case, and separator differences are already accounted for inside the trie.

docs/design/storage-profile-cli-support.md
leongdl: Sounds good, I have reworked the code path so when we do have path mapping matches, it uses the queue download code path for 1:1 functionality. 

For existing code paths where the user had to do selections, or possibly automate download I kept the existing code path for backwards compatibility.

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,1008 @@
...
+Storage profiles are configured in `~/.deadline/config` under the setting `settings.storage_profile_id`. A storage profile describes the current host's file system layout (shared mount points, local scratch paths, OS family). It is scoped under `farm_id` rather than `queue_id` because the host is the same regardless of which queue it accesses within a farm. The configuration hierarchy is:
+
+```
+defaults.aws_profile_name
+  └── defaults.farm_id
+        └── settings.storage_profile_id  (depends on farm_id, not queue — host is the same across queues)
+        └── defaults.queue_id
+```
+
+**Source:** `src/deadline/client/config/config_file.py:83-87`
```
godobyte: I don't know how useful this is, the line number can quickly becomes out of date.

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,1008 @@
...
+         "are submitted and downloaded from the same machine. Downloads all jobs to "
+         "unmapped paths regardless of operating system.\n"
+         "Default value is False.",
+    default=False,
+)
+```
+
+This option is used when:
+- All jobs are submitted and downloaded from the **same machine**
+- No path mapping is needed (same OS, same mount points)
```
godobyte: Could we be more specific here (since there is also Fleet/Worker OS), same OS and mount points among submission machines?

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,1008 @@
...
+**Source:** `src/deadline/client/config/config_file.py:83-87`
+
+### 3.2 CLI Commands with Storage Profile Support
+
+| Command | `--storage-profile-id` | `--ignore-storage-profiles` | Notes |
+|---------|:----------------------:|:---------------------------:|-------|
+| `deadline bundle submit` | ✅ | N/A | Reads `settings.storage_profile_id` from config or `--storage-profile-id` CLI option. Passed to `create_job_from_job_bundle()` which attaches it to the CreateJob API call and uses it for path grouping during upload. If not set, submission proceeds without a storage profile. No ignore flag needed — uploads don't require path mapping. |
+| `deadline bundle gui-submit` | N/A | N/A | GUI auto-populates a storage profile dropdown by listing profiles for the selected queue, filtered to the current OS. Pre-selects from `settings.storage_profile_id` in config. User can change selection before submitting. No ignore flag needed — uploads don't require path mapping. |
+| `deadline queue sync-output` | ✅ | ✅ | Full support (reference implementation). See §4.2 for code details. |
+| `deadline job download-output` | ❌ | ❌ | **Gap — prompts for manual path entry on OS mismatch** |
```
godobyte: nit - maybe add `(proposed in this design)`?

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,1008 @@
...
+    │
+    ├── --storage-profile-id option ✅
+    ├── --ignore-storage-profiles option ✅
+    │
+    ├── Validate storage profile exists
+    │
+    └── _incremental_output_download()
+```
+
+#### 4.2.1 Storage Profile Validation (Lines 360-385)
```
godobyte: ditto - let's remove the line number

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,1008 @@
...
+    Path '/mnt/Projects/Special/data.txt' maps to 'Y:\\data.txt'
+    (most specific rule wins)
+    """
+```
+
+### 4.5 Download File Operations
+
+**File:** `src/deadline/job_attachments/download.py`
+
+#### 4.5.1 OutputDownloader Class (Lines 1231-1377)
```
godobyte: Same comment for all line numbers in the file.

docs/design/storage-profile-cli-support.md
```diff
@@ -0,0 +1,1008 @@
...
+
+| Local profile configured? | Job has `storageProfileId`? | Action |
+|:---:|:---:|---|
+| Yes | Yes | Generate path mapping rules, apply automatically |
+| Yes (no matching names) | Yes | Warn about mismatched location names, recommend aligning storage profiles, fall back to manual prompt |
+| Yes | No | Warn: job was submitted without a storage profile, fall back to manual prompt |
+| No | Yes | Warn: recommend configuring a storage profile, fall back to manual prompt |
+| No | No | No mapping needed, proceed as-is (same-machine case) |
+| `--ignore-storage-profiles` | Any | Skip all mapping, download to original paths |
+
```
godobyte, line 370: The PR description says "existing behavior is fully preserved" and "Is this a breaking change? No," but I think there are at least one case in the decision matrix where that's not quite true - Both profiles exist, location names match.

Before: user gets the manual OS-mismatch prompt and types a root path.
After: automatic mapping, no prompt.

godobyte (changes requested):

src/deadline/job_attachments/_path_mapping.py
```diff
@@ -178,12 +194,3 @@ def strict_transform(self, source_path: str) -> Path:
                 return result
 
         raise ValueError("No path mapping rule could be applied")
-
-    def transform(self, source_path: str) -> Union[str, Path]:
```
mwiebe, line 182: I understand this function wasn't being used, but all the changes to the unit tests and this make the diff a fair bit more cluttered than it needs to be. Can you restore this and those unit tests to simplify?

leongdl: Done - restored in the new version.

docs/design/storage-profile-cli-support.md
leongdl: Done.

mwiebe (dismissed):

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -568,34 +578,83 @@ def _check_and_warn_long_output_paths(
...
+                    queue_role_session=queue_role_session,
+                    conflict_resolution_setting=conflict_resolution,
+                    is_json_format=is_json_format,
+                )
+                click.echo(_get_download_summary_message(download_summary, is_json_format))
+                click.echo()
+                return
+            # If no files could be mapped, fall through to the OutputDownloader path
+            # which will download to original (unmapped) paths.
+        elif resolved.job_profile.storageProfileId != resolved.local_profile.storageProfileId:
```
godobyte, line 621: Should this be storageProfileId or fileSystemLocationName? The log below indicates the latter.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -568,34 +578,83 @@ def _check_and_warn_long_output_paths(
...
+                    new_root = click.prompt(
+                        "> Please enter a new root path",
+                        type=click.Path(exists=False),
+                    )
+                else:
+                    json_string = click.prompt("", prompt_suffix="", type=str)
+                    new_root = _get_value_from_json_line(
+                        json_string, JSON_MSG_TYPE_PATHCONFIRM, expected_size=1
+                    )[0]
+                    _assert_valid_path(new_root)
```
godobyte, line 649: What does `_assert_valid_path(new_root)` do, should it be called for both branches (JSON and non-JSON)? Currently it's only in the JSON path, does `click.Path(exists=False)` validate enough like the helper?

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -975,7 +1044,16 @@ def job_download_output(step_id, task_id, output, **args):
...
-        _download_job_output(config, farm_id, queue_id, job_id, step_id, task_id, is_json_format)
+        _download_job_output(
+            config,
+            farm_id,
+            queue_id,
+            job_id,
+            step_id,
+            task_id,
+            is_json_format,
+            ignore_storage_profiles=ignore_storage_profiles,
```
godobyte, line 1053: nit - positional vs keyword args inconsistency.

src/deadline/client/cli/_groups/job_group.py
AlexTranAmz: Plus 1 on this

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,251 @@
...
+                    {
+                        "messageType": _JSON_MSG_TYPE_PROGRESS,
+                        "value": str(int(download_metadata.progress)),
+                    },
+                    ensure_ascii=True,
+                )
+                click.echo(json_line)
+                return True
+
+            return download_files_from_manifests(
```
AlexTranAmz: nit: missing telemetry, like how `_download_job_output` does

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,251 @@
...
+
+    with _modified_logging_level(logging.getLogger("urllib3"), logging.ERROR):
+        if not is_json_format:
+            with click.progressbar(length=100, label="Downloading Outputs") as download_progress:  # type: ignore[var-annotated]
+
+                def _on_progress(download_metadata: ProgressReportMetadata) -> bool:
+                    new_progress = int(download_metadata.progress) - download_progress.pos
+                    if new_progress > 0:
+                        download_progress.update(new_progress)
+                    return True
```
AlexTranAmz: Old code was having a `return sigint_handler.continue_operation` which mean Ctrl+C work, where as here, it will not work

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,251 @@
...
+            def _on_progress_json(download_metadata: ProgressReportMetadata) -> bool:
+                json_line = json.dumps(
+                    {
+                        "messageType": _JSON_MSG_TYPE_PROGRESS,
+                        "value": str(int(download_metadata.progress)),
+                    },
+                    ensure_ascii=True,
+                )
+                click.echo(json_line)
+                return True
```
AlexTranAmz, line 253: Same prob where Ctrl+C will not work

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,251 @@
...
+    FileConflictResolution,
+    JobAttachmentS3Settings,
+    PathMappingRule,
+    StorageProfile,
+    StorageProfileOperatingSystemFamily,
+)
+from ....job_attachments.progress_tracker import ProgressReportMetadata
+
+# JSON message type constant — must match job_group.py's JSON_MSG_TYPE_PROGRESS
+_JSON_MSG_TYPE_PROGRESS = "progress"
```
AlexTranAmz: Nit: duplicated with Line 77 in job_group.py

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -0,0 +1,251 @@
...
+        queue_role_session: boto3 session with queue role credentials for S3 access.
+        conflict_resolution_setting: the raw config setting string for conflict resolution.
+        is_json_format: whether to emit JSON progress lines instead of a click progressbar.
+
+    Returns:
+        DownloadSummaryStatistics from the download.
+    """
+    # Determine conflict resolution: use config setting if specified, otherwise CREATE_COPY.
+    # No interactive prompt on the mapped path — the user opted into automatic mapping.
+    if conflict_resolution_setting != FileConflictResolution.NOT_SELECTED.name:
```
AlexTranAmz, line 208: Are we missing a conflict resolution prompt? this mean if we not have a conflict resolution set, it silently just copy rather than give us 3 options like normal?

src/deadline/client/cli/_groups/job_group.py
leongdl: This was just the previous block of code - https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/job_group.py#L582-L592

The not json block has a pth type check, while the json input is string check, which we have to follow up with a path check.

So in this sense it is equivalend.

src/deadline/client/cli/_groups/job_group.py
leongdl: Updated this comment.

src/deadline/client/cli/_groups/_job_download_helpers.py
leongdl: fixed

src/deadline/client/cli/_groups/_job_download_helpers.py
leongdl: fixed

src/deadline/client/cli/_groups/_job_download_helpers.py
leongdl: Fixed!

src/deadline/client/cli/_groups/job_group.py
leongdl: Fixed

src/deadline/client/cli/_groups/_job_download_helpers.py
leongdl: I will make a follow up PR to fix this - it takes quite a bit of refactoring. The original change copied what Queue download output did which led to this disparity.


The plan: Merge this PR, Have a second smaller follow up since this one is too alrge.

godobyte: Follow-up PR sounds good, as long as we merge before the next release to maintain the same behaviour.

godobyte (approved):

mwiebe (approved):

PR #1039, opened 2026-03-11
fix: use pathlib.Path.glob to include dotfiles in output manifests

crowecawcaw:

`_match_files_with_pattern` in `_glob.py` used `glob.glob()`, which silently excludes dotfiles and dot-directories. The worker output sync in `asset_sync.py` uses `pathlib.Path.glob()`, which includes them. This inconsistency meant files like `.gtn/` and its contents would exist on disk but be missing from output manifests.

Replaced `glob.glob()` with `pathlib.Path.glob()` in `_match_files_with_pattern` so both code paths handle dotfiles consistently.

Output manifests generated via `_glob_paths` will now include dotfiles and files inside dot-directories. Users who were previously missing dotfiles in their synced outputs will now get them.

- [ ] Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended

Added dotfile test fixtures (`.dotfile`, `.hidden_dir/inside_hidden.txt`) and two new test cases in `test_glob.py`: one verifying dotfiles are included by default, one verifying they can be excluded via patterns. Updated existing test counts to reflect the new fixtures.

- No docstring changes needed — the function behavior now matches its documented intent ("all files matching the include and exclude expressions").
- No README changes needed.

No. This is a bugfix. Output manifests will now include files that were previously silently dropped. No public API signatures changed.

No. This change does not create, modify, or alter permissions on any files or directories.

----

### DIFF
--- a/src/deadline/job_attachments/_glob.py
+++ b/src/deadline/job_attachments/_glob.py
@@ -1,7 +1,16 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
+"""
+Glob utilities for file matching in job attachment manifests.
+
+Uses pathlib.Path.glob which, unlike the glob module, treats dotfiles
+(files and directories starting with '.') the same as any other file.
+The '*' wildcard matches dotfiles without requiring a leading dot in
+the pattern. This is consistent with .gitignore behavior and with the
+os.walk-based file discovery used by the main job submission path.
+"""
+
 import os
-import glob
 import json
 from pathlib import Path
 from typing import List, Optional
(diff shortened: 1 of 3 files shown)


src/deadline/job_attachments/_glob.py
```diff
@@ -55,17 +54,11 @@ def _match_files_with_pattern(base_path: str, patterns: List[str]) -> set:
         Set of normalized file paths that match the patterns
     """
     matched_files = set()
+    root = Path(base_path)
```
leongdl, line 67: I understand that this will work, but do we want to change this behavior? What if the customer's job checkout a git clone and now they get a whole new hidden git tree getting uploaded?

src/deadline/job_attachments/_glob.py
```diff
@@ -55,17 +54,11 @@ def _match_files_with_pattern(base_path: str, patterns: List[str]) -> set:
...
-        # Use recursive glob for directory matching
-        for matched_path in glob.glob(full_pattern, recursive=True):
-            # Only add files, not directories
-            if os.path.isfile(matched_path):
-                # Convert to proper path format
-                normalized_path = os.path.normpath(matched_path)
-                matched_files.add(normalized_path)
+        for matched_path in root.glob(pattern):
+            if matched_path.is_file():
+                matched_files.add(os.path.normpath(str(matched_path)))
```
leongdl, line 71: I did a quick search, it seems 3.11+ the default would support the hidden folders but < 3.11 the behavior depends. So I dont' think we want variation of behaviour on different versions of python.

leongdl (changes requested):

PR #1040, opened 2026-03-13
[WIP] feat: remove direct config access in JA

AlexTranAmz:
Job Attachments directly calls `config_file.get_cache_directory()` and `config_file.get_setting(...)` in 5 files (`upload.py`, `_diff.py`, `_aws/aws_clients.py`, `api/attachment.py`, `api/manifest.py`), creating a dependency on the client's config system. This is part of the circular dependency being broken in Milestone 1 of the Client Decoupling project.

Breaking changes:
- `S3AssetUploader.__init__` now requires `s3_max_pool_connections` and `small_file_threshold_multiplier` as keyword-only arguments instead of reading them from config internally
- Removed `get_s3_max_pool_connections()` from `_aws/aws_clients.py` entirely — callers pass the value directly
- `get_s3_client()` takes an optional `s3_max_pool_connections` param (default 50)

Non-breaking changes (optional params):
- `_manifest_diff`: added optional `cache_dir` param
- `_attachment_upload`: added optional `s3_check_cache_dir` param
- `find_file_with_status`: added optional `cache_dir` param
- `S3AssetManager`, `AssetSync`: added optional config params with defaults matching current config defaults (50, 20)
- `_get_num_download_workers`: added optional `s3_max_pool_connections` param with default

All callers in the client (`_submit_job_bundle.py`, `manifest_group.py`, `attachment_group.py`, `_session.py`) now read config values and pass them explicitly.

Job Attachments no longer imports `config_file` from the client. `S3AssetUploader` and `get_s3_max_pool_connections` have breaking signature changes. All other changes are backwards compatible via optional params with defaults.

```
========================================================================== slowest 5 durations ==========================================================================
5.78s call     test/unit/deadline_job_attachments/test_upload.py::TestUpload::test_asset_management_many_inputs[2023-03-03-200]
4.99s call     test/unit/deadline_job_attachments/test_upload.py::TestUpload::test_asset_management_no_outputs_large_number_of_inputs_already_uploaded[2023-03-03-200]
3.80s call     test/unit/deadline_job_attachments/test_upload.py::TestUpload::test_get_asset_groups[input_paths5-output_paths5-referenced_paths5-local_type_locations5-shared_type_locations5-expected_result5]
3.02s call     test/unit/deadline_job_attachments/test_upload.py::TestUpload::test_asset_management_many_inputs[2023-03-03-100]
2.61s call     test/unit/deadline_job_attachments/test_upload.py::TestUpload::test_asset_management_no_outputs_large_number_of_inputs_already_uploaded[2023-03-03-100]
========================================================================== 2346 passed, 62 skipped, 1 xfailed in 61.25s (0:01:01) ==========================================================================
```
```
========================================================================== slowest 5 durations ==========================================================================
609.05s call     test/integ/cli/test_cli_incremental_download.py::test_incremental_download_dependency_chain
456.57s call     test/integ/cli/test_cli_incremental_download.py::test_conflict_resolution_with_requeue[job]
406.18s call     test/integ/cli/test_cli_incremental_download.py::test_incremental_download_dep_data_flow
334.63s call     test/integ/cli/test_cli_incremental_download.py::test_conflict_resolution_with_requeue[step]
331.06s call     test/integ/cli/test_cli_incremental_download.py::test_conflict_resolution_with_requeue[task]
========================================================================== 56 passed, 4 skipped, 1 xfailed, 4 xpassed in 2504.15s (0:41:44) ==========================================================================
```

Yes

**Yes.** `S3AssetUploader.__init__` now requires `s3_max_pool_connections` and `small_file_threshold_multiplier` as keyword-only arguments. `get_s3_max_pool_connections()` is removed from `_aws/aws_clients.py`

No

----

### DIFF
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -26,7 +26,7 @@ from botocore.exceptions import (  # type: ignore[import]
 from botocore.session import get_session as get_botocore_session
 
 from .. import version
-from ..config import get_setting
+from ..config import get_setting, config_file
 from ..exceptions import DeadlineOperationError
 from ...job_attachments._aws.aws_clients import get_s3_client
 
@@ -164,7 +164,9 @@ def get_boto3_client(service_name: str, config: Optional[ConfigParser] = None) -
     return get_session_client(session=session, service_name=service_name)
 
 
-def get_credentials_source(config: Optional[ConfigParser] = None) -> AwsCredentialsSource:
+def get_credentials_source(
+    config: Optional[ConfigParser] = None,
+) -> AwsCredentialsSource:
     """
     Returns DEADLINE_CLOUD_MONITOR_LOGIN if Deadline Cloud monitor wrote the credentials, HOST_PROVIDED otherwise.
 
@@ -284,7 +286,9 @@ def _modified_logging_level(logger, level):
         logger.setLevel(old_level)
 
 
-def check_authentication_status(config: Optional[ConfigParser] = None) -> AwsAuthenticationStatus:
+def check_authentication_status(
+    config: Optional[ConfigParser] = None,
+) -> AwsAuthenticationStatus:
     """
     Checks the status of the provided session, by
     calling the sts::GetCallerIdentity API.
@@ -371,7 +375,10 @@ def precache_clients(
         queue_display_name=queue_display_name,
     )
     # Initialize the S3 client to populate the cache
-    return deadline, get_s3_client(queue_role_session)
+    s3_max_pool_connections = int(config_file.get_setting("settings.s3_max_pool_connections"))
+    return deadline, get_s3_client(
+        queue_role_session, s3_max_pool_connections=s3_max_pool_connections
+    )
 
 
 class QueueUserCredentialProvider(CredentialProvider):
--- a/src/deadline/client/cli/_groups/attachment_group.py
+++ b/src/deadline/client/cli/_groups/attachment_group.py
@@ -230,6 +230,13 @@ def attachment_upload(
     if not s3_root_uri:
         raise MissingJobAttachmentSettingsError("No valid s3 root path available")
 
+    s3_max_pool_connections = int(
+        config_file.get_setting("settings.s3_max_pool_connections", config=config)
+    )
+    small_file_threshold_multiplier = int(
+        config_file.get_setting("settings.small_file_threshold_multiplier", config=config)
+    )
+
     _attachment_upload(
         root_dirs=root_dirs,
         manifests=manifests,
@@ -238,4 +245,7 @@ def attachment_upload(
         path_mapping_rules=path_mapping_rules,
         upload_manifest_path=upload_manifest_path,
         print_function_callback=logger.echo,
+        s3_check_cache_dir=config_file.get_cache_directory(),
+        s3_max_pool_connections=s3_max_pool_connections,
+        small_file_threshold_multiplier=small_file_threshold_multiplier,
     )
--- a/src/deadline/job_attachments/_diff.py
+++ b/src/deadline/job_attachments/_diff.py
@@ -4,9 +4,8 @@ import concurrent.futures
 import logging
 import os
 from pathlib import Path, PurePosixPath
-from typing import Any, Callable, Dict, List, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple
 from math import trunc
-from deadline.client.config import config_file
 from deadline.job_attachments.exceptions import NonValidInputError
 from deadline.job_attachments.asset_manifests.base_manifest import (
     BaseAssetManifest,
@@ -56,14 +55,13 @@ def find_file_with_status(
     root_path: str,
     update: bool,
     statuses: List[FileStatus],
+    cache_dir: Optional[str] = None,
 ) -> List[(Tuple[FileStatus, BaseManifestPath])]:
     """
     Checks a manifest file, compares it to specified root directory or manifest of files with the local hash cache, and finds files that match the specified statuses.
     Returns a list of tuples containing the file information, and its corresponding file status.
     """
-    cache_config: str = config_file.get_cache_directory()
-
-    with HashCache(cache_config) as hash_cache:
+    with HashCache(cache_dir) as hash_cache:
         with concurrent.futures.ThreadPoolExecutor() as executor:
             futures = {
                 executor.submit(
--- a/src/deadline/job_attachments/upload.py
+++ b/src/deadline/job_attachments/upload.py
@@ -27,8 +27,6 @@ import boto3
 from boto3.s3.transfer import ProgressCallbackInvoker
 from botocore.exceptions import BotoCoreError, ClientError
 
-from deadline.client.config import config_file
-
 from .asset_manifests import (
     BaseAssetManifest,
     BaseManifestModel,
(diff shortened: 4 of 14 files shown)


AlexTranAmz:
### Summary

Job Attachments imported `config_file` from the client in 5 files to read cache directories and S3 pool settings. This PR removes all `config_file` imports by making the config values parameters on the affected functions. Per the TDD, `S3AssetUploader.__init__` and `get_s3_max_pool_connections` use required parameters (not optional) because silent default changes for performance-tuning settings would be hard to debug. All other functions use optional parameters with defaults matching the current config defaults (50 for pool connections, 20 for threshold multiplier).

### Call Stack Traces

#### Case 1: Job submission — config values flow from client to S3AssetUploader

```
create_job_from_job_bundle(...)                                    # client/api/_submit_job_bundle.py
  s3_max_pool_connections = int(config_file.get_setting("settings.s3_max_pool_connections"))       # 50
  small_file_threshold_multiplier = int(config_file.get_setting("settings.small_file_threshold_multiplier"))  # 20

  asset_manager = S3AssetManager(                                  # job_attachments/upload.py
    ...,
    s3_max_pool_connections=50,                                    # NEW: passed from client
    small_file_threshold_multiplier=20,                            # NEW: passed from client
  )
    S3AssetUploader(                                               # job_attachments/upload.py
      session=session,
      s3_max_pool_connections=50,                                  # NEW: required kwarg
      small_file_threshold_multiplier=20,                          # NEW: required kwarg
    )
      self.small_file_threshold = 8MB * 20 = 160MB
      self.num_upload_workers = 50 / min(20, 10) = 5
      self._s3 = get_s3_client(session, s3_max_pool_connections=50)  # NEW: passed through
```

Expected: identical behavior to before. Config values just originate from the client caller instead of being read inside JA.

#### Case 2: `deadline manifest diff` — cache_dir flows from client to _manifest_diff

```
manifest_diff(root="/project", manifest="file.manifest", ...)     # client/cli/_groups/manifest_group.py
  differences = _manifest_diff(                                    # job_attachments/api/manifest.py
    manifest="file.manifest",
    root="/project",
    ...,
    cache_dir=config_file.get_cache_directory(),                   # NEW: client passes cache dir
  )
    if force_rehash:
      cache_config = cache_dir                                     # was: config_file.get_cache_directory()
      with HashCache(cache_config) as hash_cache:
        directory_manifest_object = asset_manager._create_manifest_file(...)
```

Expected: same cache directory used. The value just comes from the caller now.

#### Case 3: `deadline attachment upload` — cache_dir flows to _attachment_upload

```
attachment_upload(manifests=["m.json"], ...)                       # client/cli/_groups/attachment_group.py
  _attachment_upload(                                              # job_attachments/api/attachment.py
    manifests=["m.json"],
    ...,
    s3_check_cache_dir=config_file.get_cache_directory(),          # NEW: client passes cache dir
  )
    asset_uploader = S3AssetUploader(
      session=boto3_session,
      s3_max_pool_connections=50,                                  # default
      small_file_threshold_multiplier=20,                          # default
    )
    asset_uploader.upload_assets(
      ...,
      s3_check_cache_dir=s3_check_cache_dir,                      # was: config_file.get_cache_directory()
    )
```

Expected: same behavior. Cache dir passed through from client.

#### Case 4: Download path — _get_num_download_workers uses default

```
download_files_from_manifests(s3_bucket="bucket", ...)             # job_attachments/download.py
  s3_client = get_s3_client(session=session)                       # uses default s3_max_pool_connections=50
  num_download_workers = _get_num_download_workers()               # uses default s3_max_pool_connections=50
    → int(50 / 10) = 5
```

Expected: no change in behavior. Download functions use defaults matching the config defaults. No client involvement needed for downloads since they're called from both client and worker agent.

#### Case 5: Worker agent path — AssetSync uses defaults

```
AssetSync(farm_id="farm-123", boto3_session=session)               # job_attachments/asset_sync.py
  # s3_max_pool_connections defaults to 50
  # small_file_threshold_multiplier defaults to 20
  S3AssetUploader(
    session=session,
    s3_max_pool_connections=50,
    small_file_threshold_multiplier=20,
  )
```

Expected: worker agent code doesn't need changes — defaults match current config defaults. Worker agent can optionally pass custom values if needed.

#### Case 6: precache_clients — s3_max_pool_connections flows to get_s3_client

```
precache_clients(...)                                              # client/api/_session.py
  queue_role_session = get_queue_user_boto3_session(...)
  s3_max_pool_connections = int(config_file.get_setting("settings.s3_max_pool_connections"))
  return deadline, get_s3_client(                                  # job_attachments/_aws/aws_clients.py
    queue_role_session,
    s3_max_pool_connections=s3_max_pool_connections,                # NEW: passed from client
  )
```

Expected: S3 client created with the same pool connections setting as before.

src/deadline/client/api/_session.py
```diff
@@ -27,6 +27,7 @@
 
 from .. import version
 from ..config import get_setting
+from ..config import config_file
```
leongdl: Can lines 29 and 30 be combined? 
`from ..config import get_setting, config_file`

src/deadline/job_attachments/_diff.py
```diff
@@ -56,12 +55,13 @@ def find_file_with_status(
...
     update: bool,
     statuses: List[FileStatus],
+    cache_dir: Optional[str] = None,
 ) -> List[(Tuple[FileStatus, BaseManifestPath])]:
     """
     Checks a manifest file, compares it to specified root directory or manifest of files with the local hash cache, and finds files that match the specified statuses.
     Returns a list of tuples containing the file information, and its corresponding file status.
     """
-    cache_config: str = config_file.get_cache_directory()
+    cache_config: str = cache_dir if cache_dir else ""
```
leongdl: This can be dangerous to default to "", maybe use the os get temp instead?

src/deadline/job_attachments/upload.py
```diff
@@ -140,38 +138,25 @@ class S3AssetUploader:
     def __init__(
         self,
         session: Optional[boto3.Session] = None,
+        *,
+        s3_max_pool_connections: int,
```
leongdl, line 142: How come the other functions got defaults and this one did not?

AlexTranAmz: This is intentional — S3AssetUploader is the low-level class where these values directly control S3 connection pool sizing and upload chunking behavior. Making them required here forces callers to be explicit about performance-tuning settings, so silent default changes don't slip through unnoticed and become hard to debug. The defaults live one layer up in S3AssetManager (s3_max_pool_connections=50, small_file_threshold_multiplier=20) and AssetSync, which are the typical entry points for most callers.

src/deadline/job_attachments/_diff.py
AlexTranAmz: fixed in new rev

src/deadline/client/api/_session.py
AlexTranAmz: fixed in new rev

src/deadline/job_attachments/_diff.py
```diff
@@ -56,12 +56,13 @@ def find_file_with_status(
...
     update: bool,
     statuses: List[FileStatus],
+    cache_dir: Optional[str] = None,
 ) -> List[(Tuple[FileStatus, BaseManifestPath])]:
     """
     Checks a manifest file, compares it to specified root directory or manifest of files with the local hash cache, and finds files that match the specified statuses.
     Returns a list of tuples containing the file information, and its corresponding file status.
     """
-    cache_config: str = config_file.get_cache_directory()
+    cache_config: str = cache_dir if cache_dir else tempfile.gettempdir()
```
epmog: It seems like the fallback cache_dir is different in a few different places. Previously, this behaviour was always `~/.deadline/cache`: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/config/config_file.py#L43

However, now we have this one that does the tmp folder, and a few others that go to `~/.deadline/job_attachments` based on `None` being passed to the HashCache. That actual fallback has always existed so we should probably just use that, therefore we can just pass the cache_dir into the HashCache and allow it to sort it out.

I could see an argument for changing the default location in HashCache to `~/.deadline/cache` but that's more of a deadline client choice rather than a job attachments choice. The client itself should be passing it in everywhere it's required so there'll be no difference there.

src/deadline/client/cli/_groups/attachment_group.py
```diff
@@ -225,4 +236,5 @@ def attachment_upload(
         path_mapping_rules=path_mapping_rules,
         upload_manifest_path=upload_manifest_path,
         print_function_callback=logger.echo,
+        s3_check_cache_dir=config_file.get_cache_directory(),
```
epmog, line 248: any reason we're not grabbing the config values for `s3_max_pool_connections`/`small_file_threshold_multiplier`? I believe this results in a change of behaviour if a user customized those values?

epmog (changes requested):
Can we double check all usages of the functions/classes that were modified to ensure that we're passing in the values that were previously grabbed from the config? I found one instance in deadline client where we weren't in `_attachment_upload`, but there may be more.

src/deadline/client/cli/_groups/attachment_group.py
AlexTranAmz: fix in new rev

src/deadline/job_attachments/_diff.py
AlexTranAmz: fix in new rev

epmog (approved):

leongdl (approved):

PR #1042, opened 2026-03-13
 feat: add `--show-hidden-parameters` CLI option for `deadline bundle gui-submit`

epmog:

When crafting job bundles, it can be useful to create hidden job parameters to simplify the gui shown to the user by hiding values that they don't normally need to change. However, when debugging these bundles users need to be able to debug and potentially change these values. They want to see and be able to edit all job parameters without manually modifying the bundle on disk.

Additionally, there was a bug where the float spinner could lose precision from values in the template. QDoubleSpinBox defaults to 2 decimal places and truncates values on setValue() before any signals fire, so setValue(2.718) silently becomes 2.72.

Add a `--show-hidden-paramaters` CLI option to quickly allow users to display all the parameters! 

```bash
deadline bundle gui-submit --show-hidden-parameters /path/to/bundle
```

We infer the UI control from the parameter type using following table since the paramater ui type is `HIDDEN`:

| Parameter type | Has `allowedValues`? | Widget |
|---|---|---|
| `STRING` | No | `LINE_EDIT` |
| `INT` | No | `SPIN_BOX` (integer) |
| `FLOAT` | No | `SPIN_BOX` (float) |
| `PATH` (directory) | No | `CHOOSE_DIRECTORY` |
| `PATH` (file, input) | No | `CHOOSE_INPUT_FILE` |
| `PATH` (file, output) | No | `CHOOSE_OUTPUT_FILE` |
| Any type | Yes | `DROPDOWN_LIST` |

Note: controls that require an explicit `userInterface.control` hint (like `MULTILINE_EDIT` or
`CHECK_BOX`) cannot be inferred. A revealed hidden `STRING` always becomes `LINE_EDIT`, not
`MULTILINE_EDIT`, because the `HIDDEN` control replaced whatever the author might have intended.

Future change will add a config option to allow usage within DCCs!

Users can now show previously hidden parameters and modify them in the submission dialog!

```
hatch run test
```

```
$ deadline bundle gui-submit --browse test/unit/deadline_client/ui/test_data/job_bundle_with_hidden_params --show-hidden-parameters
```

image

I also changed the values of the hidden parameters in the gui and confirmed that the submitted job had those updated values.

N/A

N/A

N/A

N/A
----


docs/design/gui-show-hidden-parameters.md
```diff
@@ -0,0 +1,131 @@
...
+
+This ensures the distinction is conveyed through both visual and non-visual channels.
+
+### Float decimal precision
+
+When a `FLOAT` spin box is created from a revealed hidden parameter, the number of decimal
+places is determined by:
+1. An explicit `userInterface.decimals` value, if present alongside `control: HIDDEN`.
+2. Otherwise, inferred from the parameter's initial value. For example, a default of `2.718`
+   produces 3 decimal places.
```
mwiebe: This doesn't look like a good idea to me. If anything, it should default to a much higher number of decimals or an adaptive decimal mode. What is the default behavior for the float spinner when no decimals are provided? I think it should match that.

src/deadline/client/ui/widgets/openjd_parameters_widget.py
```diff
@@ -525,6 +561,14 @@ def _build_ui(self, parameter):
             single_step_delta = -1
             drag_multiplier = -1.0
 
+        # When decimals is not explicitly set, infer from the initial value
```
mwiebe: What's the result if we delete this logic?

epmog: Adaptive, but we lose some information. Looks like we had a bug in the spindown where the value would be truncated cause the default decimal precision was 2 even with adaptive mode.

I've updated the parent widget to fix this precision loss issue and removed this code snippet.

docs/design/gui-show-hidden-parameters.md
epmog: Removed and fixed the underlying bug (in our widget)! Sorta surprised that there was precision loss even without specifying it...

mwiebe (dismissed):

PR #1049, opened 2026-03-17
fix: remove broken maxResults parameter from MCP

### LINKED ISSUE #1020: Bug: mcp type mismatch
### Describe the bug

The `_create_wrapper` function in `src/deadline/_mcp/utils.py` assigns a `str` annotation to parameters like `maxResults` when the parameter name is not found in the original function's signature. This creates an unresolvable type conflict: Pydantic (MCP layer) expects a string, but boto3 expects an integer.

This affects all list API tools that accept `maxResults`: `list_farms`, `list_queues`, `list_jobs`, `list_fleets`, and `list_storage_profiles_for_queue`.

### Expected Behaviour

Calling an MCP tool like `deadline_list_jobs` with `maxResults=5` should successfully pass the integer value through to the boto3 API and return paginated results.

### Current Behaviour

The parameter is stuck in a type limbo — neither type works:

1. Passing `maxResults` as an integer (`5`) → Pydantic rejects it:

pydantic_core._pydantic_core.ValidationError: 1 validation error for list_jobsArguments
maxResults
 Input should be a valid string [type=string_type, input_value=5, input_type=int]

2. Passing `maxResults` as a string (`"5"`) → boto3 rejects it:

ParamValidationError: Parameter validation failed:
Invalid type for parameter maxResults, value: 5, type: <class 'str'>, valid types: <class 'int'>

### Reproduction Steps

1. Install `deadline[mcp]` from mainline
2. Authenticate via Deadline Cloud Monitor
3. Invoke any list tool with `maxResults` via `FastMCP.call_tool()` or an MCP client:

python
import asyncio
from deadline._mcp.server import app

async def main():
   # Fails with Pydantic validation error (int not accepted)
   await app.call_tool('deadline_list_jobs', {
       'farmId': '<farm-id>',
       'queueId': '<queue-id>',
       'maxResults': 5
   })

   # Fails with boto3 ParamValidationError (str not accepted)
   await app.call_tool('deadline_list_jobs', {
       'farmId': '<farm-id>',
       'queueId': '<queue-id>',
       'maxResults': '5'
   })

asyncio.run(main())

### Environment

1. macOS 15 (Apple Silicon)
2. Python 3.13.2
3. `deadline` 0.0.post51+gc7f788dbe (mainline + PR #999, but bug exists on mainline)
4. `mcp` >= 1.13.0

### LINKED ISSUE #1047: Bug: MCP commands don't handle maxResults correctly
### Describe the bug

The MCP list tools (`list_farms`, `list_queues`, `list_jobs`, `list_fleets`, `list_storage_profiles_for_queue`) accept a `maxResults` parameter but don't actually limit the number of results returned. The parameter is passed through to the boto3 API where it controls page size, but `_call_paginated_deadline_list_api` in `_list_apis.py` loops through all pages and concatenates every result regardless.

This means `maxResults=1` returns all resources — it just fetches them one at a time across many pages.

### Expected Behaviour

Calling `deadline_list_farms` with `maxResults=5` should return at most 5 farms.

### Current Behaviour

`maxResults=5` returns all farms. The parameter only controls the per-page size of the underlying boto3 API calls, but pagination continues until all pages are exhausted.

### Reproduction Steps

1. Have a Deadline Cloud account with multiple farms
2. Call the MCP tool:

python
import asyncio
from deadline._mcp.server import app

async def main():
   result = await app.call_tool('deadline_list_farms', {'maxResults': 1})
   print(result)  # Returns ALL farms, not 1

asyncio.run(main())

### Root Cause

`_call_paginated_deadline_list_api` in `src/deadline/client/api/_list_apis.py` always paginates through every page:

python
response = list_api(**kwargs)
result = {list_property_name: response[list_property_name]}

while "nextToken" in response:
   response = list_api(nextToken=response["nextToken"], kwargs)
   result[list_property_name].extend(response[list_property_name])

return result

There is no early termination when the accumulated results reach `maxResults`.

### Environment

1. macOS 15 (Apple Silicon)
2. Python 3.13
3. `deadline` mainline

crowecawcaw:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/1047
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/1020

There were a couple bugs with `maxResults` in the MCP tool. It didn't actually limit results (just changed the page size) and the type was incorrectly detected leading to a bug.

Remove the parameter. It's not necessary for the MCP to work well. The maxResults functionality is a deeper bug in the library with wider consequences, so punting on fixing it elsewhere.

Avoids MCP errors.

Unit tests. Manually probing the MCP server.

n/a

No

No. MCP tool parameters are not a public interface. MCP is experimental.

No
----


baxeaz (dismissed):

diane-hj (dismissed):
I see that some List APIs' maxResults are still exposed but some of them are deleted. Are those APIs not impacted? 👀

crowecawcaw:
@diane-hj you're right! Updated. Also confusing, it looks like different APIs use different casing (`maxResults` vs `max_results`). Looks like another issue.

diane-hj (approved):
Thanks for updating them!

test/unit/deadline_mcp/test_mcp.py
```diff
@@ -43,6 +43,24 @@ def test_expected_functions(self):
         for func_name in expected:
             assert func_name in TOOL_REGISTRY
 
+    def test_auto_paginating_tools_do_not_expose_maxResults(self):
```
waninggibbon, line 46: nit: based on sonar I suppose this should be `max_results`.

waninggibbon (approved):

PR #1051, opened 2026-03-17
chore: remove pytest-qt dependency

crowecawcaw:
Our CI builds don't have access to pytest-qt so integ tests are failing. We can fix it, but the missing package is blocking a release at the moment and the long term fix may be slow.

Skip tests that require pytest-qt.

Unblocks the release.

Ran test collection and verified the tests were skipped.

n/a

No

No

No

----


joel-wong-aws (approved):

crowecawcaw:
No longer needed

PR #1053, opened 2026-03-17
feat: add --name option to bundle gui-submit command

### LINKED ISSUE #859: Feature request: --name support for gui-submit
### Describe the problem

When calling gui-submit programmatically, it'd be nice to override the name in the provided job template with a `--name` parameter. The regular `deadline bundle submit` command supports it. Without this parameter, the job's name must be embedded in the job template. Jobs should have a unique name to be recognizable. Job templates should be generic and reusable. Putting the name in the job template cuts against the template's reusability..

### Proposed Solution

Support a `--name` parameter.

### Example Use Cases

Above.

crowecawcaw:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/859

gui-submit did not support a `--name` parameter.

Add it.

Allows processes that call the GUI submitter to pre-populate the job name in the GUI. Aligns with the non-GUI submit command.

Unit test and manual test.

Yes

No

No

No
----


baxeaz (dismissed):

src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -255,9 +256,10 @@ def on_create_job_bundle_callback(
     )
     asset_references = AssetReferences.from_dict(asset_references_obj)
 
-    name = "Job bundle submission"
-    if template:
-        name = template.get("name", name)
+    if name is None:
+        name = (
+            template.get("name", "Job bundle submission") if template else "Job bundle submission"
```
mwiebe: This logic looks a bit repetitive. Can we actually get to here without the template being available?

crowecawcaw: Agree. It was a workaround to get mypy to detect types correctly on an older version of Python.

crowecawcaw: Cleaned up some types upstream to fix this.

src/deadline/client/job_bundle/loader.py
```diff
@@ -98,6 +98,18 @@ def parse_yaml_or_json_content(file_contents: str, file_type: str, bundle_dir: s
         raise RuntimeError(f"Unexpected file type '{file_type}' in job bundle:\n{bundle_dir}")
 
 
+@overload
```
mwiebe: Is it possible to set the return type on the main `read_yaml_or_json_object` function instead of using these two decorators?

mwiebe (approved):

baxeaz (approved):

PR #1055, opened 2026-03-17
fix: cache telemetry client by package name

### LINKED ISSUE #959: Bug: TelemetryClient caching ignores package_name parameter
### Describe the bug

The get_telemetry_client() function uses a global cache that ignores the package_name parameter when returning an existing client. This causes incorrect service attribution in telemetry events when multiple components request clients with different package names.

For context - https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/857

Root Cause
• Single global __cached_telemetry_client variable shared across all package names
• Cache key doesn't include package_name, only checks if any client exists
• Class-level _system_metadata dictionary is shared across all instances, causing service name overwrites

### Expected Behaviour

- get_telemetry_client("deadline-cloud-worker-agent", version) should return a client with service="deadline-cloud-worker-agent"
- get_telemetry_client("deadline-cloud-library", version) should return a client with service="deadline-cloud-library"
- Each unique package name should have its own cached client instance

### Current Behaviour

- The first call to get_telemetry_client() creates a client and caches it globally
- Subsequent calls with different package_name parameters return the same cached client
- All telemetry events use the service name from the first client created, regardless of which component emits them

### Reproduction Steps

Example:
1. Job attachments calls get_telemetry_client("deadline-cloud-library", version) → creates client with service="deadline-cloud-library"
2. Worker agent calls get_telemetry_client("deadline-cloud-worker-agent", version) → returns same cached client with service="deadline-cloud-library"
3. Worker agent telemetry events incorrectly show service="deadline-cloud-library"

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.)
2. Output of `python3 --version`
3. Output of `deadline --version`
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

crowecawcaw:
Fixes: #959

Telemetry client caching doesn't respect different package names.

Fix the caching.

Telemetry client correctly labels the package.

Unit tests

n/a

No

No

No

----


waninggibbon (approved):

godobyte (commented):
This looks great, thanks for making the fix!

godobyte (approved):

PR #1058, opened 2026-03-18
chore: ignore errors in telemetry handlers

### LINKED ISSUE #755: Maintenance: Add Error Handling Decorator to Telemetry Client Functions
## Description

We need to implement a try/catch decorator for all functions in the [TelemetryClient ](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/api/_telemetry.py#L58)class to ensure robust error handling and prevent any telemetry-related issues from affecting the main application flow.

## Current Behavior:

Some functions in the TelemetryClient class have basic error handling, for example, the [`initialize()`](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/api/_telemetry.py#L162) method silently catches all exceptions.

However, other methods may throw uncaught exceptions that could impact the application

### Create a decorator that will:

- Catch all exceptions
- Log errors appropriately
- Prevent exceptions from propagating to calling code
- Maintain the telemetry client's non-blocking nature

### Functions requiring the decorator:

- `set_opt_out()`
- `initialize()`
- `_get_prefixed_endpoint()`
- `_get_telemetry_identifier()`
- `_start_threads()`
- `_get_system_metadata()`
- `_exit_cleanly()`
- `_send_request()`

crowecawcaw:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/755

Telemetry handlers could fail causing the library/CLI to not work correctly. Telemetry is never required to function though.\

Ignore any telemetry errors so it never breaks the main application.

Better resilience in the event of telemetry issues.

Unit tests

n/a

No

No

No

----


src/deadline/client/api/_telemetry.py
```diff
@@ -138,43 +157,41 @@ def initialize(self, config: Optional[ConfigParser] = None) -> None:
...
-            # Silently swallow any exceptions
-            return
+        self._initialized = True
+        self._start_threads()
 
     @property
     def is_initialized(self) -> bool:
         return self._initialized
 
+    @_swallow_exceptions
```
leongdl: Is a function like this really need the swallow exception decorator?

leongdl: I would think only the record functions should have ignore exceptions.

crowecawcaw: It looks safe enough to me without it. Instead of looking at each method for risks, I just applied it everywhere that it wouldn't break something for completeness.

I can scale this back and just apply it to the heavier, top-level methods.

karthikbekalp (dismissed):

karthikbekalp:
Hmm, looks this PR needs to be rebased as there a few merge conflicts.

karthikbekalp (approved):

leongdl (approved):

PR #1060, opened 2026-03-19
chore: add missing wayland modules

crowecawcaw:
During manual testing, the Linux build from the installer was missing some Wayland modules. As a result, the GUI windows were missing the window frame with its close/maximize/minimize buttons.

Include the Wayland modules.

Fixes UI bug.

I built the change locally and verified the window frame was back.

n/a

No

No

No
----


waninggibbon (approved):

leongdl (approved):

test/unit/test_copyright_headers.py
```diff
@@ -41,17 +40,7 @@ def _check_file(filename: Path) -> None:
...
-        return False
-    with open(filename) as infile:
-        lines_read = 0
-        for line in infile:
-            if _generated_by_scm.search(line):
-                return True
-            lines_read += 1
-            if lines_read > 10:
-                break
-    return False
```
karthikbekalp, line 54: Curious, do we know why we added this in the first place?

leongdl:
Super nit - not all linux distros are on Wayland :) Maybe change the title.

karthikbekalp (commented):
Added 1 comment. Let me know if it makes sense. :)

leongdl:
Also need a retrigger to pass semantic PR.

karthikbekalp (approved):
Approving as it is in test code. But not sure why we added the test before?

test/unit/test_copyright_headers.py
crowecawcaw: The overall test is to make sure almost all files have copyright headers. The `_version.py` file is generated by setuptools-scm and excluded. This condition here was to double ensure the `_version.py` file is generated before excluding it from the copyright header validation. setuptools-scm was updated today which broke the regex. Just checking the filename seems sufficient and more robust.

PR #1070, opened 2026-03-23
feat: Add newer submitter available dialog.

karthikbekalp:

Customers did not know whether they were on the latest submitter or not. This means they would not have the latest bug fixes or features to work with our product. 

This adds a dialog which checks our download manifests and if there's a new update available notifies customer  that. 

This a new feature and that's why it fails the CI step. It introduces 2 new helper functions (a dialog and utility function to check for latest submitter) that will be available for most of the submitters to use. 

#### In Blender:

image

image

#### In Cinema 4D:
image

image

Customers can activate/deactivate the feature in the settings. 

image

DCCs can start using the dialog and the helper function in `deadline-cloud` to check for the latest versions. 

Built this into Cinema 4D and Blender to test. 

Yes

Currently running 

I've not made changes to `download` or `asset_sync` modules.

I also ran the following tests in Cinema 4D and Blender PRs:

- Cinema 4D PR: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/414
- Blender PR: https://github.com/aws-deadline/deadline-cloud-for-blender/pull/329/changes 

| # | Test Case | Steps | Expected Result | Pass? |
|---|-----------|-------|-----------------|-------|
| | **Config GUI — setting visible** | | | |
| 1 | Setting in config GUI | Open Blender → Deadline Cloud → Settings | "Show submitter update notifications" checkbox is visible | YES |
| | **Config disabled — no dialog** | | | |
| 2 | Notification disabled via GUI | Uncheck "Show submitter update notifications" → Apply → close & reopen submitter | No network call, no dialog, submitter opens normally | YES |
| | **Config enabled, no internet — graceful failure** | | | |
| 3 | Re-enable notification via GUI | Check "Show submitter update notifications" → Apply | Setting updated | YES |
| 4 | No internet / firewall | Disconnect network → open submitter | Submitter opens normally, no crash, no visible error | YES |
| | **Config enabled, internet available, no update** | | | |
| 5 | Current == latest | Restore network, ensure installed version matches manifest | No dialog, submitter opens normally | YES |
| 6 | Current > latest (dev build) | Install a version newer than manifest | No dialog, submitter opens normally | YES |
| | **Config enabled, internet available, update exists** | | | |
| 7 | Current < latest | Install an older version than manifest | Update dialog appears | YES |
| 8 | Dialog content | Inspect the dialog | Shows "Blender", correct current & latest versions | YES |
| 9 | Dialog layout | Resize / check | Min width 400px, text wraps properly | YES |
| 10 | Dialog timing | Observe launch order | Dialog appears before Blender submitter window | YES |
| | **Dialog interactions — Dismiss** | | | |
| 11 | Click "Dismiss" | Click Dismiss button | Dialog closes, submitter opens normally | YES |
| 12 | Close via X | Click X button on dialog | Dialog closes, submitter opens normally | YES |
| | **Dialog interactions — Download** | | | |
| 13 | Click "View release notes" | Click the link | Blender GitHub releases page opens in browser | YES |
| 14 | Click "Download" | Click Download button | Correct installer URL opens in browser | YES |
| 15 | Restart reminder | After Download click | "Application Restart Required" info box appears | YES |
| | **Edge cases** | | | |
| 16 | Non-standard `_version.py` tuple | Modify version tuple | No crash | YES |

This is not a breaking change, if the DCC does not consume these new helpers they would not have the feature. 

No. 

----


src/deadline/client/api/_certs/AmazonRootCA1.pem
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
+MIIDQTCCAimgAwIBAgITBmyfz5m/jAo54vB4ikPmljZbyjANBgkqhkiG9w0BAQsF
```
karthikbekalp: On macOS, bundled Python environments (e.g. inside Cinema 4D) often cannot locate the system CA certificate store. So adding this here so that we can get the manifest securely.

crowecawcaw: This cert will expire over time and will be a maintenance burden. Are there other alternatives? Could we get the system CA cert store to work?

src/deadline/client/api/_update_checker.py
```diff
@@ -0,0 +1,265 @@
...
+
+class UpdateCheckStatus(Enum):
+    """Status of the update check operation."""
+
+    SUCCESS = "success"
+    NETWORK_ERROR = "network_error"
+    TIMEOUT_ERROR = "timeout_error"
+    PARSE_ERROR = "parse_error"
+    INVALID_VERSION = "invalid_version"
+    INTEGRATION_NOT_FOUND = "integration_not_found"
```
crowecawcaw, line 64: Seems like these should be exception classes instead of an enum. Besides SUCCESS, they're all failure modes.

src/deadline/client/api/_update_checker.py
```diff
@@ -0,0 +1,265 @@
...
+
+    Returns:
+        An UpdateCheckResult describing whether an update is available.
+    """
+    if not _is_update_notification_enabled():
+        return UpdateCheckResult(
+            status=UpdateCheckStatus.SUCCESS,
+            update_available=False,
+            current_version=current_version,
+        )
```
crowecawcaw: This doesn't look quite right. Should the notification logic be outside of this function? Seems like we should not call `check_for_updates()` if the notifications are not enabled.

src/deadline/client/ui/dialogs/update_available_dialog.py
```diff
@@ -0,0 +1,181 @@
...
+)
+from qtpy.QtCore import Qt  # pylint: disable=import-error
+
+from .._utils import tr
+
+logger = logging.getLogger(__name__)
+
+# AWS brand colors used for accent elements
+_COLOR_ACCENT = "#ec7211"
+_COLOR_ACCENT_HOVER = "#f59942"
```
crowecawcaw, line 31: Do we use these colors elsewhere? We should be consistent with other styles.

src/deadline/client/ui/dialogs/update_available_dialog.py
```diff
@@ -0,0 +1,181 @@
...
+        # Button box - consistent with other dialogs
+        self.button_box = QDialogButtonBox(Qt.Horizontal)
+
+        _button_base = "border: 1px solid #888; border-radius: 6px; padding: 3px 12px;"
+
+        self.dismiss_button = self.button_box.addButton(QDialogButtonBox.Close)
+        self.dismiss_button.setText(tr("Dismiss"))
+        self.dismiss_button.setStyleSheet(f"QPushButton {{ {_button_base} }}")
+
+        self.download_button = QPushButton(tr("Download"))
```
crowecawcaw: Better as "Download installer" to hint that they need to run the installer afterwards.

src/deadline/client/ui/dialogs/update_available_dialog.py
```diff
@@ -0,0 +1,181 @@
...
+            return
+
+        # Show restart reminder and close the dialog
+        QMessageBox.information(
+            self,
+            tr("Application Restart Required"),
+            tr(
+                "Please restart {integration_name} after installation to use the new version."
+            ).format(integration_name=self.integration_name),
+        )
```
crowecawcaw, line 177: Maybe say they need to run the installer and then restart. Someone might assume clicking the button automatically runs the install process as well.

test/unit/deadline_client/api/test_update_checker.py
```diff
@@ -0,0 +1,358 @@
...
+        mock_sys.platform = "darwin"
+        mock_urlopen_fn.side_effect = urllib.error.URLError("SSL error")
+
+        with pytest.raises(urllib.error.URLError):
+            _fetch_manifest()
+
+        assert mock_urlopen_fn.call_count == 2
+
+
+class TestCheckForUpdates:
```
crowecawcaw, line 175: Could we also add test cases for versions that are uncommon or malformed? We've seen issues with dev buld version numbers like `1.2.3+patch1dkj3k` in other places.

crowecawcaw (changes requested):

src/deadline/client/api/_certs/AmazonRootCA1.pem
karthikbekalp: > This cert will expire over time and will be a maintenance burden. 

This cert is valid until 2038 which I think should be sufficiently long time. 

```
openssl x509 -in "deadline-cloud/src/deadline/client/api/_certs/AmazonRootCA1.pem" -noout -dates -subject
notBefore=May 26 00:00:00 2015 GMT
notAfter=Jan 17 00:00:00 2038 GMT
subject=C = US, O = Amazon, CN = Amazon Root CA 1
```

> Are there other alternatives? 

I tried that but it would require running `pip install certifi` or something similar to keep the certificates updated. Which means it we will end up in the same spot similar to GUI deps for this one. But maybe that's ok because we need to get the manifest from internet from our domain anyway? 

> Could we get the system CA cert store to work?

Yeah so Cinema 4D has its own installation of python and with the MacOS settings it does not get access to system python.

src/deadline/client/api/_update_checker.py
karthikbekalp: Hmm, I added it as enum so that if the DCC chooses they can raise the exception on their side or just share the error on their side or just silently go forward on their side. 

If we instead raise exceptions and the DCC does not handle it then, we can cause issues no? Let me know your thoughts.

src/deadline/client/api/_update_checker.py
karthikbekalp: So I added the check here so that its easier for the DCC owners. 
All they need to do is call this helper and it already checks if `submitter_update_notification` deadline config is set or not. Just to reduce the work for DCC owners, because if they miss then they would not be respecting the config.

src/deadline/client/ui/dialogs/update_available_dialog.py
karthikbekalp: I think I got this feedback from our team's UX designers. 
I don't think we have it standard everywhere and I was told to follow the color scheme for Deadline Cloud Monitor.

src/deadline/client/ui/dialogs/update_available_dialog.py
karthikbekalp: That's not a bad idea, change coming right up.

src/deadline/client/ui/dialogs/update_available_dialog.py
karthikbekalp: That;s not a bad idea, change coming right up.

test/unit/deadline_client/api/test_update_checker.py
karthikbekalp: Sure, I'll add this right away.

src/deadline/client/ui/dialogs/update_available_dialog.py
karthikbekalp: Updated: 
image

src/deadline/client/ui/dialogs/update_available_dialog.py
karthikbekalp: Updated: 
image

src/deadline/client/api/_certs/AmazonRootCA1.pem
crowecawcaw: Found another possibility, could we try using botocore's certs?

Some agent thoughts FWIW:
```
botocore is installed (via boto3), and botocore bundles its own cacert.pem — a full CA bundle with
  130 certificates, including Amazon Root CA 1.
  So there's already a full CA bundle sitting right there in the install at botocore/cacert.pem. The PR could
  use that instead of bundling a single PEM file:

  import botocore.httpsession
  # or more directly:
  from botocore import cacert
  import os
  ca_bundle = os.path.join(os.path.dirname(cacert.__file__), "cacert.pem")```

File: src/deadline/client/api/_update_checker.py (hunk 2 above)
Comment (crowecawcaw) [reply src/deadline/client/api/_update_checker.py:64]: Makes sense. What about using exceptions to model it but having both `check_for_update` and `safe_check_for_update`? Another option is maybe a flag like `_raise_on_error` with a `False` default.

File: src/deadline/client/ui/dialogs/update_available_dialog.py (hunk 4 above)
Comment (crowecawcaw) [reply src/deadline/client/ui/dialogs/update_available_dialog.py:31]: I think we should keep the colors consistent. What about making an issue to track theming the UIs then we can add the colors consistently across them?

**Comment by crowecawcaw** [comment]
One more thing: In the update window, I think we also need a button option for "Don't remind me again" or similar that toggles the config option for update notifications.

File: src/deadline/client/api/_certs/AmazonRootCA1.pem (hunk 1 above)
Comment (karthikbekalp) [reply src/deadline/client/api/_certs/AmazonRootCA1.pem]: > could we try using botocore's certs?

I didn't know that. I'll try that now.

File: src/deadline/client/api/_update_checker.py (hunk 2 above)
Comment (karthikbekalp) [reply src/deadline/client/api/_update_checker.py:64]: > What about using exceptions to model it but having both check_for_update and safe_check_for_update? Another option is maybe a flag like _raise_on_error with a False default.

That's a good idea. Although, I wonder if we would end up adding some code that no one uses as everyone would just use `safe_check_for_update`. 

Let me know your thoughts.

File: src/deadline/client/ui/dialogs/update_available_dialog.py (hunk 4 above)
Comment (karthikbekalp) [reply src/deadline/client/ui/dialogs/update_available_dialog.py:31]: Sure, I'll create one. Maybe at least for `deadline-cloud` so that we use it consistently everywhere.

File: src/deadline/client/api/_update_checker.py [hunk 8]
```diff
@@ -0,0 +1,264 @@
...
+    try:
+        with urllib.request.urlopen(req, timeout=MANIFEST_TIMEOUT_SECONDS) as resp:
+            return json.loads(resp.read().decode("utf-8"))
+    except urllib.error.URLError:
+        if sys.platform != "darwin":
+            raise
+
+    # macOS fallback: retry with the bundled Amazon Root CA certificate.
+    logger.debug("Default SSL verification failed on macOS, retrying with bundled CA cert")
+    ctx = ssl.create_default_context()
```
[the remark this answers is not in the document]
diane-hj: Not entirely sure, so wanted to ask. Would this be acceptable in our case? 👀

src/deadline/client/ui/dialogs/update_available_dialog.py
karthikbekalp: Github issue: https://github.com/aws-deadline/deadline-cloud/issues/1084

src/deadline/client/api/_update_checker.py
```diff
@@ -0,0 +1,265 @@
...
+            current_version=current_version,
+            error_message=f"Network error: {e}",
+        )
+    except (TimeoutError, socket.timeout):
+        return UpdateCheckResult(
+            status=UpdateCheckStatus.TIMEOUT_ERROR,
+            current_version=current_version,
+            error_message="Request timed out",
+        )
+    except (json.JSONDecodeError, UnicodeDecodeError) as e:
```
diane-hj: Should we add a generic exception here so that it can handle the unexpected error gracefully?

src/deadline/client/api/_update_checker.py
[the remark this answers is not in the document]
karthikbekalp: I think I updated it to compliant solution here: https://github.com/aws-deadline/deadline-cloud/security/code-scanning/307 so if that we use the TLS version to TLS_v1_2.

src/deadline/client/api/_update_checker.py
karthikbekalp: Makes sense. I'll add it.

src/deadline/client/api/_update_checker.py
karthikbekalp: Discussed offline, we are going to rename to `safe_check_for_update` to signify that this function does not throw any exceptions.

karthikbekalp:
> Added do not remind me again button

image

src/deadline/client/api/_certs/AmazonRootCA1.pem
karthikbekalp: Updated!

crowecawcaw (approved):

diane-hj (approved):
Thanks for the feature!

PR #1072, opened 2026-03-25
chore: Update to support Python 3.14

mwiebe:

Can't install deadline-cloud on Python 3.14.

- Allow 3.14 installation, add it to the test matrix and classifier
- Remove upper bound on pywin32. See reasoning in https://github.com/OpenJobDescription/openjd-adaptor-runtime-for-python/pull/162#discussion_r1862680554

Can install in more environments.

Added Python 3.14 to the test matrix.

Added Python 3.14 to the classifier.

No

No

No

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ authors = [{ name = "Amazon Web Services" }]
 dynamic = ["version"]
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.8, <3.14"
+requires-python = ">=3.8, <3.15"
 # https://pypi.org/classifiers/
 classifiers = [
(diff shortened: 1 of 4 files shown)


leongdl (approved):

pyproject.toml
```diff
@@ -42,7 +43,7 @@ dependencies = [
   # Job Attachments
   "typing_extensions >= 4.8",
   "xxhash >= 3.4,< 3.7",
```
epmog, line 45: Do we want to adjust the upperbound for xxhash as well?

mwiebe: Could change the upper bound to 4.

epmog (approved):

PR #1073, opened 2026-03-25
feat!: add in-memory session config for gui-submit CLI overrides

epmog:

`deadline bundle gui-submit` didn't support `--farm-id`, `--queue-id`, `--profile`, or `--storage-profile-id` CLI arguments, so there was no way to pre-populate the submission dialog for a different farm or
queue. The only way to change these was through the settings dialog, which always wrote changes directly to the on-disk config (~/.deadline/config). There was no way to distinguish between "change
this for the current submission only" and "change my workstation defaults" — every edit permanently modified the workstation config.

I ran into similar issues when trying to implement a new cli option for the gui with an additional config option here: #1042, where being unable to separate the current submission's options vs the persistent disk settings made for an awful user experience.

The main idea in this change is separate the usage of the config to either be the currently loaded settings or the on-disk settings. In this way, users will be able to provide overrides that don't mess with their next submission but also be able to change their mind about values without having to restart the submission process. This required an overhaul of the set_settings/get_settings usage in deadline-cloud so that any changes we make to the session configuration flow through to the rest of the application.

While the changes was inspired by making `gui-submit` better, it also means that we have to adjust `deadline config gui` to still be intuitive for users looking to update their settings.

When session matches disk (no changes):

image

When session does not match disk or there are unsaved changes are in the dialog (does not match session or disk):

image

`deadline config gui` is the same as the above but there is no session config, since you're directly affecting the disk.

High-level changes are as follows:

1. Added `--profile`, `--farm-id`, `--queue-id`, `--storage-profile-id` CLI options to bundle_gui_submit, using `_apply_cli_options_to_config` to produce an in-memory `ConfigParser` session config.

2.`_apply_cli_options_to_config` now always returns a fresh `ConfigParser` copy — never None, never mutating the caller's config or the read_config() cache.

3. Threaded the session config through the full widget hierarchy: SubmitJobToDeadlineDialog → SharedJobSettingsWidget → DeadlineCloudSettingsWidget → DeadlineFarmDisplay / DeadlineQueueDisplay. All
 get_setting() call sites in shared_job_settings_tab.py now pass config=self._config.

4. Reworked the settings dialog button bar to: Ok, Cancel, "Config File" dropdown (Save to Disk / Load from Disk), and Apply. Without a session config (deadline config gui), Apply is permanently
disabled. With a session config (bundle gui-submit), Apply writes to the in-memory session only.

5. configure_settings() now returns a ConfigureSettingsResult dataclass instead of bool, carrying both changes_applied and the (possibly modified) session_config.

6. Added `persist_job_id(job_id, profile, farm_id, queue_id)` helper that writes the job ID to the correct hierarchical config section on disk. Both CLI and GUI callers use this after successful
submission, so the job ID is always persisted to the right section regardless of CLI overrides.

- Users can now run `deadline bundle gui-submit --farm-id X --queue-id Y` to pre-populate the submission dialog
- The settings dialog opened from the submitter now operates on the in-memory session config by default, with explicit Save to Disk / Load from Disk actions.
- Job IDs are now always persisted after submission, even when CLI overrides are used.

- [ ] Have you made changes to the download or asset_sync modules?

Added unit tests, and also ran my own manual tests to verify the two guis were behaving as expected

- [x] Relevant docstrings updated in all modified source files.
- [ ] README.md — no CLI argument documentation changes needed (the new options follow the same pattern as bundle submit).
- [x] Design doc added: docs/design/in-memory-config.md.

* [x] This PR does not add any new dependencies.

Yes, one breaking changes for UI authors:

1. DeadlineConfigDialog.configure_settings() return type changed from bool to ConfigureSettingsResult. A dataclass instance is always truthy, so if configure_settings(): will no longer work as before.
Callers must use result.changes_applied instead. Callers that discard the return value (like dev_application.py and config_group.py) are unaffected.

*Notable behavioral change (non-breaking)*: Job ID is now persisted even when CLI overrides are used. Previously deadline bundle submit --farm-id X --queue-id Y would not save the job ID.

N/A

----


docs/design/in-memory-config.md
```diff
@@ -0,0 +1,232 @@
...
+
+**`DeadlineConfigDialog.configure_settings()` return type changed from `bool` to `ConfigureSettingsResult`.**
+
+- **Who is affected**: Code that captures and inspects the return value of `configure_settings()`, including truthiness checks like `if configure_settings():`.
+- **Migration**: The `ConfigureSettingsResult` dataclass instance is always truthy (even when `changes_applied=False`), so `if configure_settings():` will no longer work as before. Replace with `if configure_settings().changes_applied:`. Code that discarded the return value is unaffected.
+
+**`_submission_succeeded_signal_receiver` now uses `persist_job_id` instead of `set_setting`.**
+
+- **Who is affected**: Code that subclasses `SubmitJobToDeadlineDialog` and overrides `_submission_succeeded_signal_receiver`, or that patches `set_setting` in tests expecting it to be called with `defaults.job_id` after submission.
+- **Migration**: Update test expectations to check for `persist_job_id` calls instead of `set_setting("defaults.job_id", ...)`.
```
epmog: I'll remove this, not a breaking change. Internal implementation detail.

crowecawcaw:
Talked offline a bit. There might be a better UX than the dropdown menu. Maybe buttons like "Apply" and "Reset" would feel more intuitive?

PR #1074, opened 2026-03-25
[WIP] ci: Add import dependency check for client/job_attachments boundary

seant-aws:

The `feature_client_decouple` branch aims to separate job_attachments from the client library, maintaining a one-way dependency (client can import job_attachments, but not vice versa). There was no automated enforcement of this architectural boundary.

Added a CI workflow using [import-linter](https://import-linter.readthedocs.io/) to enforce the one-way dependency. The check ensures that `deadline.job_attachments` does not import from `deadline.client`. Seems to be a package from an approved list.

Changes:
- `pyproject.toml`: Added import-linter configuration
- `requirements-testing.txt`: Added import-linter as a test dependency
- `.github/workflows/import-dependency-check.yml`: PR workflow that runs on feature branches

PRs that introduce imports from `deadline.client` into `deadline.job_attachments` will fail CI.

- [ ] Have you made changes to the `download` or `asset_sync` modules?

Tested locally:

**Test 1: Passing case (no forbidden imports)**
```bash
PYTHONPATH=src lint-imports
```
Output:
```
Analyzed 141 files, 475 dependencies.
-------------------------------------

Job Attachments must not import from Client KEPT

Contracts: 1 kept, 0 broken.
```

**Test 2: Failing case (injected forbidden import)**
```bash
echo "from deadline.client import api" > src/deadline/job_attachments/_test_violation.py
PYTHONPATH=src lint-imports
```
Output:
```
Analyzed 142 files, 476 dependencies.
-------------------------------------

Job Attachments must not import from Client BROKEN

Contracts: 0 kept, 1 broken.

----------------
Broken contracts
----------------

Job Attachments must not import from Client
-------------------------------------------

deadline.job_attachments is not allowed to import deadline.client:

-   deadline.job_attachments._test_violation -> deadline.client.api (l.1)
```

- [x] Configuration is self-documenting in pyproject.toml
- [ ] README.md - N/A, this is CI infrastructure

Added `import-linter == 2.*` as a **test-only** dependency (in `requirements-testing.txt`). It is not a runtime dependency and will not affect end users.

No. This adds CI enforcement only.

No. This is a static analysis tool that runs in CI.

----


leongdl (dismissed):

.github/workflows/import-dependency-check.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.13'
+
+      - name: Install dependencies
+        run: pip install import-linter
+
+      - name: Check import dependencies
+        run: PYTHONPATH=src lint-imports
```
epmog: `PYTHONPATH`? That shouldn't be a variable that we set here. There should be a config option or cli arg to specify it where to look (maybe default is cwd and that's good?

epmog (changes requested):

.github/workflows/import-dependency-check.yml
seant-aws: Ok I've tweaked it a bit to use hatch now, should be better ?

leongdl (approved):

AlexTranAmz (approved):

test/unit/test_copyright_headers.py
```diff
@@ -7,7 +7,9 @@
 _copyright_header_re = re.compile(
     r"Copyright Amazon\.com, Inc\. or its affiliates\. All Rights Reserved\.", re.IGNORECASE
 )
-_generated_by_scm = re.compile(r"# file generated by setuptools[_-]scm", re.IGNORECASE)
+_generated_by_scm = re.compile(
+    r"# file generated by (setuptools[_-]scm|vcs-versioning)", re.IGNORECASE
+)
```
epmog, line 12: is this not deleted in mainline? we probably just need to merge mainline in.

epmog (approved):

PR #1080, opened 2026-03-26
feat: add sanitized stack trace reporting to telemetry error events

vaish-gujjari:
Signed-off-by: Vaishnavi Devi Gujjari

The Deadline Cloud client telemetry system could record errors but had no way to capture sanitized stack traces alongside them. Stack traces are critical for diagnosing issues, but they can contain customer-specific file paths (home 
directories, bucket names, project paths) that must not be included in telemetry per ADR 2024-02-19.

Added a new _stack_trace_sanitizer module and a record_error_with_trace() method on TelemetryClient.

The sanitizer strips customer-specific file paths from Python tracebacks while preserving diagnostic value:
- Paths under known packages (deadline, openjd, boto3, botocore) are kept as package-relative paths (e.g. deadline/client/api/_telemetry.py)
- Paths under site-packages for unknown libraries are kept as library-relative paths
- All other paths (customer scripts, project directories) are reduced to the bare filename only
- Quoted file paths in exception messages are sanitized using the same logic

The record_error_with_trace() method captures the exception type, scope, sanitized message, and sanitized stack trace, then emits them as a com.amazon.rum.deadline.error telemetry event. It also accepts optional extra_details for additional 
context.

- Adds a new private module _stack_trace_sanitizer.py — no public API impact
- Adds a new method record_error_with_trace() to the internal TelemetryClient class — no public API impact
- Replaces existing record_error() calls in CLI and GUI submit error handlers with record_error_with_trace() to capture stack traces alongside error telemetry

Added sanitizer and telemetry tests in test_api_telemetry.py covering path sanitization, traceback sanitization, message sanitization, live exception handling, record_error_with_trace event structure, extra_details merging, and opt-out behavior.

  - Yes. Added tests in test_stack_trace_sanitizer.py (path sanitization, traceback sanitization, message sanitization, live exception handling) and test_api_telemetry.py (record_error_with_trace event structure, extra_details merging, message sanitization, path stripping, opt-out behavior).

  - N/A, no integration-level changes.
 
  - No.

  - Yes. The new module and all public functions have docstrings.
  
  - No. This is an internal telemetry change with no user-facing impact. 

No. All changes are to private/internal modules only.

No. This change specifically improves privacy by ensuring customer file paths are stripped from telemetry data before transmission. No files or directories are created or modified. No threat modeling needed.


test/unit/deadline_client/api/test_api_telemetry.py
```diff
@@ -92,6 +102,10 @@ def test_opt_out_env_var(fresh_deadline_config, monkeypatch, env_var_value):
     client.record_hashing_summary(SummaryStatistics(), from_gui=True)
     client.record_upload_summary(SummaryStatistics(), from_gui=False)
     client.record_error({}, str(type(Exception)))
+    try:
+        raise RuntimeError("opt-out test")
+    except RuntimeError as exc:
+        client.record_error_with_trace(exc, "test")
```
crowecawcaw: I don't quite get how this relates to the rest of the test. Should we raise the exception earlier before the assertions? Same for the test above.

src/deadline/client/api/_telemetry.py
```diff
@@ -164,6 +164,30 @@ def initialize(self, config: Optional[ConfigParser] = None) -> None:
...
             return
 
+    def record_error_with_trace(
+        self,
+        exc: BaseException,
+        exception_scope: str,
+        extra_details: Optional[dict] = None,
+        from_gui: bool = False,
+    ) -> None:
+        from ._stack_trace_sanitizer import sanitize_exception, sanitize_message
```
crowecawcaw: Nit: move import to the top of the file (unless there's a structural reason it can't be, like circular imports)

crowecawcaw:
Instead of getting a traceback string, could we get the traceback data and then sanitize it? I think we could have more robust sanitization because we'd get the paths in structured data instead of from parsing strings. We could also select just the elements we want (e.g. error code, file name, line number) to capture to make sure there's not unexpected sensitive data that gets added to the string.

test/unit/deadline_client/api/test_api_telemetry.py
```diff
@@ -441,3 +554,140 @@ def fake_init(self, **kwargs):
...
+        assert "deadline/client/api/_telemetry.py" in result
+
+    def test_no_path_unchanged(self):
+        msg = "ValueError: something went wrong"
+        assert sanitize_message(msg) == msg
+
+    def test_unquoted_unix_path(self):
+        msg = "Failed to process /home/user/job/input.txt"
+        result = sanitize_message(msg)
+        # Unquoted paths are not sanitized — Python exceptions typically quote paths
```
waninggibbon: This seems like a dangerous assumption to make - a third party library or dependency might throw an error message with a path in it and we shouldn't assume that the path will be in quotes.

src/deadline/client/api/_stack_trace_sanitizer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    return parts[-1]
+
+
+# Matches file paths in exception messages — paths are typically quoted with ' or "
+_MSG_PATH_RE = re.compile(r"'([^'/\\]*[/\\][^']*)'|\"([^\"/\\]*[/\\][^\"]*)\"")
+
+
+def sanitize_message(message: str) -> str:
+    """Sanitize an exception message by replacing file paths with safe versions."""
```
waninggibbon: I am worried that the error messages may contain information we don't expect that may have sensitive data beyond just the file paths. We ultimately call this with `str(exc)` which could really contain anything because we don't have control over what our dependencies put in their error messages.

Is there ever a version of this where we just log the sanitized stack trace and exception name/type but not the message?

src/deadline/client/api/_stack_trace_sanitizer.py
```diff
@@ -0,0 +1,80 @@
...
+        m = _FRAME_RE.match(line)
+        if m:
+            prefix, filepath, suffix = m.groups()
+            sanitized.append(f'{prefix}"{_sanitize_path(filepath)}",{suffix}')
+        else:
+            sanitized.append(line)
+    return "\n".join(sanitized)
+
+
+def sanitize_exception(exc: BaseException) -> str:
```
waninggibbon, line 93: Instead of this denylist approach (taking the whole exception, putting into a string and then looking for things to take out), consider an allowlist approach.  You could use something like `traceback.extract_tb()` to work with structure data instead, and only emit the fields you explicitly choose.

This would look something like:
```
def sanitize_exception(exc: BaseException) -> str:
    lines = ["Traceback (most recent call last):"]
    for frame in traceback.extract_tb(exc.__traceback__):
        safe_path = _sanitize_path(frame.filename)
        lines.append(f'  File "{safe_path}", line {frame.lineno}, in {frame.name}')
        # Intentionally omit frame.line — source code context could
        # contain credentials, customer data, or other sensitive values
    lines.append(f"{type(exc).__qualname__}: {sanitize_message(str(exc))}")
    return "\n".join(lines)
```

This gives you filename, lineno, and name as discrete values you can sanitize individually, and it avoids passing through anything unexpected.

waninggibbon: You would need to consider chained exceptions as well, but `traceback.TracebackException` should have the full chain.

src/deadline/client/api/_telemetry.py
vaish-gujjari: yep, modified

test/unit/deadline_client/api/test_api_telemetry.py
vaish-gujjari: The record calls (including record_error_with_trace) come after the assertions intentionally.. the assertions first confirm the client is in the opted-out state, then the record calls verify those 
methods don't raise when opted out. This matches the existing pattern for record_hashing_summary, record_upload_summary, and record_error right above.

src/deadline/client/api/_stack_trace_sanitizer.py
```diff
@@ -0,0 +1,76 @@
...
+"""
+Stack trace sanitizer for Deadline Cloud client telemetry.
+
+Uses an allowlist approach: only explicitly chosen fields (sanitized filename,
+line number, function name, exception type) are emitted. Source code context
+and exception messages are intentionally omitted as they could contain
+customer data.
+
+Conforms to ADR 2024-02-19: "No customer content or other information provided
+by the customer can be submitted, such as bucket names, file names, or similar."
```
crowecawcaw: Rewrite to remove internal reference. But the comment itself is good context and worth keeping, maybe labelled as a design tenet

src/deadline/client/api/_stack_trace_sanitizer.py
```diff
@@ -0,0 +1,76 @@
...
+        "boto3",
+        "botocore",
+    }
+)
+
+
+def _sanitize_path(filepath: str) -> str:
+    """Replace a full file path with the package-relative portion or bare filename."""
+    if filepath.startswith("<"):
+        return filepath
```
crowecawcaw, line 34: I don't understand this, could use a comment. Same for other conditions in this function.

test/unit/deadline_client/api/test_api_telemetry.py
```diff
@@ -530,6 +610,112 @@ def test_init_swallows_get_system_metadata_exception(self, fresh_deadline_config
             assert "version" not in client._system_metadata
 
 
+class TestSanitizePath:
```
crowecawcaw: Nit:easier to read if parameterized

test/unit/deadline_client/api/test_api_telemetry.py
```diff
@@ -281,6 +287,80 @@ def test_record_error(fresh_deadline_config, mock_telemetry_client):
...
+
+    # THEN
+    queue_mock.put_nowait.assert_called_once()
+    event: TelemetryEvent = queue_mock.put_nowait.call_args[0][0]
+    assert event.event_type == "com.amazon.rum.deadline.error"
+    assert event.event_details["exception_type"] == "ValueError"
+    assert event.event_details["exception_scope"] == "test_scope"
+    assert "ValueError" in event.event_details["stack_trace"]
+    assert "message" not in event.event_details
+    assert event.event_details["usage_mode"] == "CLI"
```
crowecawcaw: I think it'd be better to assert on the full details here because we're mainly concerned with extra data being added that we don't want. It'll be verbose but I think thats good here. The individual asserts are more like spot checks. Same thing for stack trace sanitization.

crowecawcaw (changes requested):
I think the sanitization process would benefit from some basic fuzzing. Like throwing random paths and exceptions at the function and seeing if we can get something to leak. Can be simplistic. Here's an example that we added to OJD for python which caught a couple issues that slipped through other unit tests: https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/test/openjd/model/test_fuzz.py

crowecawcaw (approved):

waninggibbon (approved):

PR #1087, opened 2026-04-01
chore: Use bundled certificate by default for checking manifest version + UX fixes for update available dialog

karthikbekalp:
### What was the problem/requirement and solution? (What/Why)

There were 2 minor issues that I wanted to fix from the previous change:
1. We were only going for fallback if the OS was Mac. But instead we could just make it consistent so that we always use the bundled certificate to keep the failure modes low. 
2. The UX was a little janky in Windows so made them fixed everywhere. 

Lower failure modes and consistent UX in all OSes. 

Ran the tests manually. 

#### Before

Mac:

image

Windows:

image

#### After 

Mac:

image

Windows:

image

This change does not require documentation. 

No

No

----

### DIFF
--- a/src/deadline/client/api/_update_checker.py
+++ b/src/deadline/client/api/_update_checker.py
@@ -113,11 +113,10 @@ def _is_update_notification_enabled() -> bool:
 def _fetch_manifest() -> Dict[str, Any]:
     """Fetch and parse the remote manifest JSON.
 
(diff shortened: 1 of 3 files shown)


diane-hj (approved):
LGTM! 🙏

crowecawcaw (approved):

PR #1093, opened 2026-04-06
docs: Add docs as to why MaxRetriesPerTask and MaxFailedTasksCount we…

### LINKED ISSUE #1085: Feature request: make default maxFailedTasks 0
### Describe the problem

Current the default value for maxFailedTasks is 5 which is gratuitous. In all likelihood, a job that fails on its first try will fail on subsequent tries. The extra attempts slow iterations if we're waiting for a job to succeed or fail.

### Proposed Solution

Make the default 0. If someone has a job that is not reliable, they can opt in to retries.

### Example Use Cases

...

karthikbekalp:
Fixes:  https://github.com/aws-deadline/deadline-cloud/issues/1085

We wanted to add documentation on why specific defaults were added and the rationale behind these defaults. 

Add the docs for this. 

Customers know why the decision was taken in the first place. 

Only docs changes. 

Yes

Nope, only docs changes. 

Nope, only docs changes. 

Nope. 

----

### DIFF
--- a/docs/decisions/job-retry-and-failure-defaults.md
+++ b/docs/decisions/job-retry-and-failure-defaults.md
@@ -1,0 +1,73 @@
+# Decision: Job Retry and Failure Default Values
+
+## Overview
+
+Deadline Cloud submitters expose two job-level settings that control how task failures
+are handled: `maxRetriesPerTask` and `maxFailedTasksCount`. Their default values were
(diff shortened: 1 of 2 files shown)


karthikbekalp:
Moving this to draft so that we can investigate recent trends.

karthikbekalp:
Closing this PR so that we can investigate the data more and come to better defaults.

PR #1098, opened 2026-04-11
fix: External tools corrupting known paths configuration

mwiebe:

Logging in with Deadline Cloud monitor on Windows rewrites the ~/.deadline/config file with a different .ini library than Python. It treats '\\' in entries differently, and breaks the known paths.

This change normalizes path configurations on Windows to write '/' instead of '\\' path separators, converting to/from normalized form in set_setting and get_setting.

Correct behavior of known paths on Windows.

Added unit tests, and confirmed the known paths were untouched after a Deadline Cloud monitor login by running `deadline config gui` before/after the login.

N/A

No

No

No

----

### DIFF
--- a/src/deadline/client/config/config_file.py
+++ b/src/deadline/client/config/config_file.py
@@ -61,6 +61,7 @@ __config_mtime = None
 SETTINGS: Dict[str, Dict[str, Any]] = {
     "deadline-cloud-monitor.path": {
         "default": "",
+        "is_path": True,
         "description": "The filesystem path to Deadline Cloud monitor, set during login process.",
     },
     "defaults.aws_profile_name": {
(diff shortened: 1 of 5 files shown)


src/deadline/client/config/config_file.py
```diff
@@ -391,9 +394,45 @@ def get_setting(setting_name: str, config: Optional[ConfigParser] = None) -> str
...
+    backslashes with forward slashes on Windows. This prevents corruption when
+    the config file is rewritten by tools (like Deadline Cloud Monitor) that may
+    interpret backslashes as escape characters in INI-style config values.
+
+    Forward slashes work correctly on all platforms including Windows.
+    Only performs replacement on Windows.
+    """
+    if os.name == "nt":
+        return path_value.replace("\\", "/")
+    return path_value
```
jaikish, line 424: This looks like a general problem and I searched if Python has inbuilt utilities and found this: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix

jaikish: Can customers choose network paths? I asked AI if `\\server\my_dir` conversion to `//server/my_dir` still works and it says the first pair of slashes still has to be `\\server/my_dir`

mwiebe: This value gets normalized to '/' when in the config file, but then normalized back to '\\' when reading, so the path it's using on Windows always starts with `\\`

jaikish (approved):

epmog (approved):

PR #1102, opened 2026-04-13
chore: add UI tests with mock Deadline server and CI workflow

crowecawcaw:

This PR introduces 3 major test components:
1. A mock Deadline Cloud service that runs as a localhost HTTP server for testing against. The mock services implemented basis create/read/update/delete/list APIs for the main Deadline resources. By setting the AWS_ENDPOINT_URL_DEADLINE env var to `localhost:8000`, we can run the client against our mock server. Using the mock endpoint gives us higher fidelity testing than unit tests (because the full client is running without any code being stubbed out). The fidelity is not as high as integ tests which hit the actual Deadline endpoint, but we don't need credentials or real AWS resources, so tests are easier to run (especially in GitHub actions) and fast.
2. UI testing using the [xa11y library](https://xa11y.dev/). xa11y is a cross-platform library for interacting with desktop accessibility APIs that I wrote. Compared to squish, it's permissively open sourced. Compared to other accessibility libraries, it has a consistent interface across platforms.
3. GitHub workflows to run the UI tests on new PRs

The initial tests in this PR cover some basic config reading and writing and exporting job bundles. The tests function more like examples in this PR. If we merge then, we can expand coverage.

We can test the UI.

CI actions!

n/a

Yes - adds new test dependencies.

No

No
----


test/ui/test_bundle_gui_submit.py
```diff
@@ -0,0 +1,214 @@
...
+            },
+        }
+    ],
+}
+
+
+def _find_app(pid: int) -> xa11y.App:
+    """Wait for the xa11y app to appear for the given PID."""
+    end = time.monotonic() + STARTUP_TIMEOUT
+    while time.monotonic() < end:
```
leon-li-inspire: This piece of code seems duplicate in `config_gui`, could we reuse it ?

test/ui/test_bundle_gui_submit.py
```diff
@@ -0,0 +1,214 @@
...
+            app.locator(f'dialog[name="{DIALOG_TITLE}"]').wait_attached(timeout=STARTUP_TIMEOUT)
+            time.sleep(2)
+
+            # Click "Export bundle"
+            app.locator('button[name="Export bundle"]').press()
+            time.sleep(ACTION_SETTLE)
+
+            # The export shows an info dialog — dismiss it
+            ok_button = app.locator('button[name="OK"]')
+            if ok_button.exists():
```
leon-li-inspire: Don't we need to raise error if the `ok_button` does not exist?

test/ui/test_bundle_gui_submit.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def test_has_submit_and_export_buttons(self, gui_submit: xa11y.App):
+        assert gui_submit.locator('button[name="Submit"]').exists()
+        assert gui_submit.locator('button[name="Export bundle"]').exists()
+
+
+class TestExportBundle:
+    """Verify that clicking 'Export bundle' saves a job bundle."""
+
+    def test_export_creates_bundle(self, bundle_dir, submitter_env, tmp_path):
```
leon-li-inspire: Why could not we use `gui_submit` in this test ?

test/ui/test_cli_mock_server.py
```diff
@@ -0,0 +1,261 @@
...
+        capture_output=True,
+        text=True,
+        env=env,
+        timeout=30,
+    )
+    assert result.returncode == 0, f"aws deadline {' '.join(args)} failed:\n{result.stderr}"
+    return json.loads(result.stdout) if result.stdout.strip() else {}
+
+
+def _deadline(*args: str, env: dict) -> str:
```
leon-li-inspire: Are we trying to test a test resource used in test here? Is it necessary ?

test/ui/test_config_gui_log_level.py
```diff
@@ -0,0 +1,232 @@
...
+
+def _cli_set(setting: str, value: str) -> None:
+    subprocess.check_call(
+        ["deadline", "config", "set", setting, value],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+
+
+def _find_app_name(pid: int) -> str:
```
leon-li-inspire: There is another piece code tries to find the app, seems like we could consolidate this piece of reusable code.

waninggibbon:
This is a huge improvement to what we have now, but thinking big picture here, how would we implement some of the more complex text cases such as DCC integration points? Do the github runners have GPUs and could we install blender, etc? Or are you imaging that those more complex test cases would not run on CI and would instead be part of the release workflow or some other pipeline?

.github/workflows/ui_tests.yml
```diff
@@ -0,0 +1,95 @@
...
+on:
+  push:
+    branches: [mainline, release]
+  pull_request:
+    branches: [mainline, release, feature_assets_cli, 'patch_*', 'feature_*']
+
+jobs:
+  ui-test-linux:
+    name: UI Tests (Linux)
+    runs-on: ubuntu-latest
```
waninggibbon, line 12: Is there any value in testing on an older instance instead of latest for validating support on older machines? 

Another thing to think about is that the underlying machines these alias's point to will sometimes change which could cause difficult to debug errors from dependencies not being available. Maybe not a huge deal because the system dependencies below seem pretty light, but calling it out as something to think about.

test/ui/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+    dialog = app.locator(f'dialog[name="{DIALOG_TITLE}"]')
+    dialog.wait_attached(timeout=STARTUP_TIMEOUT)
+
+    yield app
+
+    # Teardown: close dialog if still open, then kill process
+    try:
+        if dialog.exists():
+            app.locator('button[name="Cancel"]').press()
+            time.sleep(ACTION_SETTLE)
```
waninggibbon: Should we be validating here that the cancel button actually closed the dialog?

test/ui/test_bundle_gui_submit.py
waninggibbon: Maybe we can abstract some of these repeated test util functions into a unit test helper class?

.github/workflows/ui_tests.yml
crowecawcaw: Good point about older systems. I think that would be worth testing and GitHub offers ubuntu 2022 as a test runner option. I think OS version coverage might make sense on some sort of end to end test which also exercises the installer and DCM? I imagined these particular UI tests being more aimed at making sure the UI code is correct. I don't feel strongly and am open to ideas.

Also good point that the image will change over time and maybe the list of dependencies will not be correct. I don't think it'll be too painful since these tests run when the PR is raised so they're quick to update and validate (as opposed to e.g. CDK managed pipelines with canaries).

Also, on your comment about DCC testing: IMO it'd be great to run larger e2e or DCC tests in GitHub actions too since actions are easy to set up, easy to fork, and support all major OSes. It looks like GitHub has some test runners with GPUs but maybe they require opting in and paying. Or alternatively, there might be ways to get a soft GPU that's sufficient for opening Blender. I think tests like this would belong in a separate suite. But they would be amazing to add.

test/ui/test_cli_mock_server.py
crowecawcaw: Sort of. These tests verify that basic CLI commands like `deadline farm list` with the mock HTTP server. I think they give us a little bit more confidence for CLI behavior that unit tests since we're not patching any CLI internals. I do like that they also verify the mock server is working correctly since it's not a trivial test helper.

test/ui/test_bundle_gui_submit.py
crowecawcaw: Good call. I added some helpers and classes similar to how I've seen frontend e2e test set up where there's a class representing the page/window/widget to be tested.

.github/workflows/ui_tests.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      # xa11y requires the process calling its APIs to have the
+      # kTCCServiceAccessibility TCC entitlement. Grant it to the Python
+      # binary inside hatch's ui env (the one pytest actually runs under).
+      - name: Grant accessibility TCC permission
+        run: |
+          HATCH_PYTHON=$(hatch -e ui run python -c "import os,sys; print(os.path.realpath(sys.executable))")
+          SETUP_PYTHON=$(python3 -c "import os,sys; print(os.path.realpath(sys.executable))")
+          for PYTHON_PATH in "$SETUP_PYTHON" "$HATCH_PYTHON"; do
+            sudo sqlite3 "/Library/Application Support/com.apple.TCC/TCC.db" \
```
waninggibbon, line 78: ~~I tried to do something similar to this in CodeBuild and ran into an issue where SIPS was blocking this even with sudo. Is that not an issue here?~~ Disregard I see this ran no problem in the GitHub runner: https://github.com/aws-deadline/deadline-cloud/actions/runs/25137148429/job/73678245354?pr=1102

test/ui/test_bundle_gui_submit_controls.py
```diff
@@ -0,0 +1,56 @@
...
+            (getattr(c, "name", None) or "") in valid_states
+            or (getattr(c, "value", None) or "") in valid_states
+            for c in combos
+        ), f"No combo_box with Initial state value (READY/SUSPENDED) among {len(combos)} combos"
+
+
+class TestHostRequirementsControls:
+    """The host-requirements tab exposes the two top-level radio buttons."""
+
+    def test_default_and_custom_radios_present(self, gui_submit: SubmitterDialog) -> None:
```
waninggibbon, line 43: I'm seeing this consistently fail locally. At a quick glance, it looks like the issue is because the host requirements tab is only present when `show_host_requirements_tab=True` is passed (submit_job_to_deadline_dialog.py:~L354) but the gui submit doesn't set this true.  I see this test passed in CI which is interesting, but from what I can tell the test needs to enable the tab some how by plumbing in `show_host_requirements=True` with a flag or env var or something.

DEVELOPMENT.md
```diff
@@ -475,17 +475,8 @@ These are the manual test cases for the client software release cycle, covering
 | Pre-requisite: Uninstall any previous versions of the Deadline Cloud Submitter Installer | Update PATH if necessary. | |
 | Verify Deadline CLI can be successfully installed using the staged individual installer | Run the staged individual installer and verify that it can install. | |
 | Verify correct version of Deadline CLI is being tested | Verify correct version using `deadline --version` command. | |
-| Verify user can modify workstation configuration settings using Deadline GUI (`deadline config gui`) | Run `deadline config gui` and verify dialogue loads as expected. Verify User can authenticate using DCM Profile using Login button. Using a DCM Profile, verify correct farm/queue resources are pulled, and the existing config can be modified. Verify User can Logout of DCM Profile using Logout button. Verify User can authenticate using an AWS Profile. Using an AWS Profile, verify correct farm/queue resources are pulled, and the existing config can be modified. | |
```
waninggibbon, line 478: Should we keep this test case since the new tests don't actually complete the full login/logout flows?

leon-li-inspire (approved):

DEVELOPMENT.md
crowecawcaw: We actually have an integ test running through the full end to end login flow now! https://github.com/aws-deadline/deadline-cloud/pull/1148

This PR exercises the UI buttons against the mock backend to make sure they're wired up correctly. Between the two, I think we have decent coverage.

test/ui/test_bundle_gui_submit_controls.py
crowecawcaw: Could you share some logs? I won't block on this since it passes locally and in CI, but there could very well be an undocumented assumption about the environment.

kavmur (approved):

test/ui/test_bundle_gui_submit_controls.py
crowecawcaw: Looks like that flag is true for the CLI submitter, so it must have a different root cause: https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/ui/cli_job_submitter.py#L179

PR #1108, opened 2026-04-15
feat(cli): Add --include-path selective filtering to download-output

larrygao001:
Add selective download support for job output attachments:

- Add `--include-path` and `--include-path-stdin` options to `download-output` for downloading specific files or directory prefixes
- Add `_filter_paths()` and `_matches_any_filter()` in `download.py`
- Add `path_filters` parameter to `OutputDownloader`
- Add `_validate_and_normalize_include_paths()` with path traversal rejection, backslash-to-forward-slash conversion, and normalization
- Fix bug in `get_job_input_paths_by_asset_root()` where S3 root prefix was not prepended to input manifest keys
- Extract shared helpers: `_run_download_ux()`, `_get_job_download_context()`, `_parse_filters_and_config()`, `_handle_download_error()`
- Update `job_attachments_guide.md` to document `--include-path`
- Add unit tests for path filtering and validation

A follow-up PR will add the `download-input` command using the patterns established here.

The existing `deadline job download-output` command downloads *all* output files for a job. Users have no way to download only specific files or folders from a job's output. This makes the CLI less useful for workflows where users only need a subset of files.

- **`--include-path`** (repeatable): accepts exact relative file paths or directory prefixes (trailing `/`). Multiple values are OR'd. Without `--include-path`, behavior is unchanged — all outputs are downloaded.
- **`--include-path-stdin`**: reads one path per line from stdin until EOF or an empty line (sentinel). After consuming stdin, reopens `/dev/tty` (or `CON` on Windows) so interactive prompts still work. This is needed for the Deadline Cloud Monitor desktop app (Tauri) integration, where the Monitor writes file paths to the child process's stdin to avoid shell argument limits and Tauri's scoped command argument validation.
- **Path validation**: `_validate_and_normalize_include_paths()` rejects `..` (path traversal), converts `\` to `/` (Windows compatibility), strips `./`, collapses `//`.
- **Bug fix**: `get_job_input_paths_by_asset_root()` was not prepending the S3 root prefix + Manifests folder to input manifest keys (unlike `asset_sync.py` which does this correctly). Fixed to use `s3_settings.add_root_and_manifest_folder_prefix()`.
- **Refactoring**: Extracted shared download UX flow into `_run_download_ux()`, `_get_job_download_context()`, `_parse_filters_and_config()`, and `_handle_download_error()` to prepare for the follow-up `download-input` PR.

- `download-output` without `--include-path` behaves exactly as before — no breaking change.
- The `get_job_input_paths_by_asset_root()` bug fix changes the S3 key used to fetch input manifests. This function was not previously called by any CLI command. The worker agent uses `asset_sync.py` which already had the correct prefix logic.

  - Yes. New unit tests + existing download tests all pass (210 passed, 0 failed).
  - Manually tested end-to-end against a real job with known input/output files:
    - `download-output --include-path "output/output_0005.png"` → downloaded 1 of 25 files
    - `download-output` with 10 `--include-path` values → downloaded exactly 10 of 25 files
    - `--include-path "nonexistent.txt"` → "No output files available"
    - `--include-path "../../etc/passwd"` → rejected with "Path filter must not contain '..'"
    - `--include-path-stdin` without piped data → reads until empty line, then continues to interactive prompts
  - Yes, changes to `download.py`. Docker-based tests should be run before merge.

  - Yes. All new functions have docstrings.
  - The CLI reference (`docs/cli_reference/deadline_job.md`) is auto-generated from click decorators — new options will appear automatically.
  - `docs/job_attachments_guide.md` updated to mention `--include-path`.

No. `download-output` without `--include-path` behaves identically to before. The `get_job_input_paths_by_asset_root()` fix changes behavior of a function that was not previously called from any CLI command.

- The `--include-path` option introduces a new input vector. Path traversal is mitigated with defense in depth:
  1. `_validate_and_normalize_include_paths()` rejects any filter containing `..` at parse time.
  2. `_ensure_paths_within_directory()` validates download destination paths stay within the asset root at download time (existing check).
- No new files/directories are created beyond the existing download behavior.
- No threat modeling needed — same S3 access patterns and credential flow as existing `download-output`.

----


src/deadline/client/cli/_groups/job_group.py
```diff
@@ -915,13 +945,69 @@ def _assert_valid_path(path: str) -> None:
...
 @click.option("--farm-id", help="The farm to use.")
 @click.option("--queue-id", help="The queue to use.")
 @click.option("--job-id", help="The job to use.")
 @click.option("--step-id", help="The step to use.")
 @click.option("--task-id", help="The task to use.")
+@click.option(
+    "--path-filter",
+    multiple=True,
+    help="Download only files matching this relative path or directory prefix (trailing /). Repeatable.",
+)
```
crowecawcaw, line 997: What about "--include_path"? "filter" is ambiguous if it filters paths in or out.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -915,13 +945,69 @@ def _assert_valid_path(path: str) -> None:
...
+@click.option(
+    "--path-filter-stdin",
+    is_flag=True,
+    help="Read path filters from stdin, one per line.",
+)
```
crowecawcaw, line 1014: Do we need this one? If we do want to accept paths over stdin, seems like we should always accept them instead of requiring a flag and the stdin input. But I'd lean towards not accepting stdin if we don't need it.

crowecawcaw (changes requested):
Let's split up the download-input command from the filtering since the changes are independent and to make it easer to review. Suggest starting with the path filtering first and then adding download-input with the patterns we land on in the first PR.

src/deadline/client/cli/_groups/job_group.py
larrygao001: Good suggestion. Updated it.

src/deadline/client/cli/_groups/job_group.py
larrygao001: We need stdin support for the Deadline Cloud Monitor desktop app (Tauri) integration. When the Monitor spawns the CLI for selective downloads, it writes file paths to the child process's stdin, this avoids shell argument length limits and bypasses Tauri's scoped command argument validation (which only allows fixed argument patterns).

The explicit flag is intentional over implicit stdin detection (`isatty()`) because:

Without the flag, the CLI would need to check if stdin is a pipe on every invocation, and isatty() returns False in some non-interactive environments even when there's no piped data, causing the CLI to hang waiting for input
The flag makes the contract explicit: "I am sending paths on stdin, read them"

We use an empty-line sentinel to signal end-of-input (instead of EOF), then reopen stdin from /dev/tty so interactive prompts still work after the pipe is consumed. This lets --include-path-stdin work both with and without --yes.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -740,6 +663,113 @@ def _update_download_progress(
...
+    )
+    root_path_format_mapping: dict[str, str] = {}
+    job_attachments = job.get("attachments", None)
+    if job_attachments:
+        for manifest in job_attachments["manifests"]:
+            root_path_format_mapping[manifest["rootPath"]] = manifest["rootPathFormat"]
+    return deadline, job, queue, queue_role_session, job_attachments, root_path_format_mapping
+
+
+def _download_job_output(
```
mwiebe: How does this PR interact with the storage profile PR https://github.com/aws-deadline/deadline-cloud/pull/1029? Should we work to get that merged sooner so you could rebase?

mwiebe: For example, the download-input command should support the storage profile mapping the same way.

larrygao001: Both PRs refactor `_download_job_output`  and there is an overlap in `download.py`  too, so yes I'd need to refactor the code after PR 1029 is merged. It'd be great if we can get that PR merged sonner.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -959,13 +961,45 @@ def _assert_valid_path(path: str) -> None:
...
 
+def _validate_and_normalize_include_paths(filters: list[str]) -> list[str]:
+    """
+    Validates and normalizes include paths.
+    - Rejects filters containing '..' (path traversal prevention)
+    - Converts backslashes to forward slashes (Windows compatibility)
+    - Strips leading './'
+    - Normalizes '//' to '/'
+    """
+    normalized = []
```
leongdl: This seems like a weird function to have ?

Do we want to think about glob instead ? Or something similar to glob? This looks almost like a glob filter.

leongdl: Can we move this to a utility file?

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -1018,6 +1054,22 @@ def job_download_output(step_id, task_id, output, ignore_storage_profiles, **arg
...
+    filters = list(include_path)
+    if include_path_stdin:
+        for line in sys.stdin:
+            stripped = line.strip()
+            if not stripped:
+                break
+            filters.append(stripped)
+        try:
+            tty_path = "CON" if sys.platform == "win32" else "/dev/tty"
+            sys.stdin = open(tty_path)  # noqa: SIM115
```
leongdl: This seems really odd - can we use click?
https://click.palletsprojects.com/en/stable/prompts/

leongdl: Also, we should add checks for the `is_json` - the DCM case may fail here.

src/deadline/job_attachments/download.py
```diff
@@ -1243,6 +1244,47 @@ def mount_vfs_from_manifests(
         vfs_manager.start(session_dir=session_dir)
 
 
+def _matches_any_filter(file_path: str, filters: list[str]) -> bool:
+    """
+    Check if a file path matches any of the given filters.
+    - Exact match: filter "renders/frame_001.exr" matches only that path
+    - Directory prefix: filter "renders/" matches all paths starting with "renders/"
+    """
+    for f in filters:
```
leongdl: def matches_filter(file_path: str, filters: list[str]) -> bool:                                                           
      return any(                                                                                                           
          file_path.startswith(f) if f.endswith("/") else file_path == f                                                    
          for f in filters                                                                                                  
      )    

Maybe do this?

waninggibbon:
I merged the latest upstream/mainline into this branch and added a commit to make --include-path work with the storage profile path mapping flow from #1029.

The issue was that when storage profiles are configured, download-output takes a different code path (_transform_manifests_to_absolute_paths → _download_mapped_manifests) that
bypassed the path_filters entirely. I added _filter_manifests() in download.py to filter BaseAssetManifest objects before the path transformation, and wired it into job_group.py.

Changes in my commit (4cc2547):
- download.py: Added _filter_manifests() function
- job_group.py: Imported _filter_manifests and applied it before _transform_manifests_to_absolute_paths
- test_path_filtering.py: Added TestFilterManifests with 5 unit tests

src/deadline/client/cli/_groups/job_group.py
waninggibbon: Moved to a utility file. Do you have any specific use cases where using glob might be better here? This seems simpler to me but happy to discuss using glob if you had something in mind.

src/deadline/job_attachments/download.py
waninggibbon: Good call, fixed!

src/deadline/client/cli/_groups/job_group.py
waninggibbon: The stdin reopen is for restoring interactive prompts (conflict resolution, OS mismatch root path) after piped input is consumed.

I added a JSON check so the reopen is skipped when --output json is used. This would have already been covered in DCM because of the OSError but hopefully this way the intent is more clear.

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -305,3 +305,25 @@ def _on_progress_json(download_metadata: ProgressReportMetadata) -> bool:
...
+    Validates and normalizes include paths.
+    - Rejects filters containing '..' (path traversal prevention)
+    - Converts backslashes to forward slashes (Windows compatibility)
+    - Strips leading './'
+    - Normalizes '//' to '/'
+    """
+    normalized = []
+    for f in filters:
+        if ".." in f:
+            raise click.BadParameter(f"Path filter must not contain '..': {f}")
```
crowecawcaw: I don't think we need to reject this, even though it's odd. This is a path filter so it chooses what files are downloaded. So if someone's trying to do something malicious with a weird path, at most they'll download all files for the job. If we were attaching files where `..` could get the submitter access to unexpected files on the workstation, we'd need to think a bit more.

I think it's better to leave it out because it's simpler and so it doesn't suggest there's a security issue to defend against and which requires a robust control.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -1018,6 +1036,23 @@ def job_download_output(step_id, task_id, output, ignore_storage_profiles, **arg
...
+            stripped = line.strip()
+            if not stripped:
+                break
+            filters.append(stripped)
+        if output != "json":
+            try:
+                tty_path = "CON" if sys.platform == "win32" else "/dev/tty"
+                sys.stdin = open(tty_path)  # noqa: SIM115
+            except OSError:
+                pass  # Non-interactive environment — no TTY to reopen
```
crowecawcaw: From what I can tell, this comes in to play if the caller sends in paths via stdin then later wants to send interactive commands (e.g. confirming something) also via stdin. So this code switches stdin back to TTY after the paths finish reading from stdin. I think this logic is confusing and unnecessary. Let's not mix the two. If someone sends in paths via stdin, they will not get any interactive prompts. If there are interactions required, we should either choose a default option that makes sense or fail with an error so the caller always needs to pass in the choice explicitly.

crowecawcaw (commented):
We added a new set of testes for job attachment CLI commands in this PR: https://github.com/aws-deadline/deadline-cloud/pull/1125

The idea to test CLI commands without any internal mocking by using mock HTTP servers for S3 and Deadline. This feature would be great to cover with this new test pattern. It'd look something like creating a temp directory, creating a manifest with some real files in S3, downloading the files with this filter, and asserting the files that are created in our temp/test directory are only the ones we expect.

src/deadline/client/cli/_groups/_job_download_helpers.py
waninggibbon: Good call, I left this out and am no longer validating against it so `..` would just be treated like any other non-existant path

src/deadline/client/cli/_groups/job_group.py
waninggibbon: Makes sense, fixed in latest commit.

waninggibbon:
Added the following e2e tests in latest commit:

| Test | Scenario |
|------|----------|
| include_path | Directory prefix filter (renders/) — downloads 2 of 3 files |
| include_path_exact_file | Exact file path — downloads 1 of 2 files |
| include_path_multiple | Two --include-path values OR'd — cherry-picks from different directories |
| include_path_no_match | Filter matches nothing — reports "no output files available", nothing written |
| include_path_stdin | --include-path-stdin with piped paths + empty line sentinel — no interactive prompts |
| include_path_stdin_json | --include-path-stdin + --output json — DCM's exact invocation pattern, verifies all output is valid JSON lines |

crowecawcaw (dismissed):

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -966,6 +973,16 @@ def _assert_valid_path(path: str) -> None:
 @click.option("--job-id", help="The job to use.")
 @click.option("--step-id", help="The step to use.")
 @click.option("--task-id", help="The task to use.")
+@click.option(
+    "--include-path",
+    multiple=True,
+    help="Download only files matching this relative path or directory prefix (trailing /). Repeatable.",
+)
+@click.option(
```
leongdl, line 998: This seems a bit odd as a command line arg

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -1018,6 +1037,17 @@ def job_download_output(step_id, task_id, output, ignore_storage_profiles, **arg
     if task_id and not step_id:
         raise click.UsageError("Missing option '--step-id' required with '--task-id'")
 
+    filters = list(include_path)
+    if include_path_stdin:
```
leongdl: This is really odd - we did quite a bit of work before on how to source input glob filters. See the previous thinking in manifest snapshot.

leongdl (changes requested):

waninggibbon:
Updated to mirror the manifest style arguments (`--include` and `--exclude` with no separate stdin option)

New tests:
| Test | Scenario |
|------|----------|
| include_path | --include "renders/" — directory prefix downloads 2 of 3 files |
| include_path_exact_file | --include "renders/frame_001.exr" — exact path downloads 1 of 2 |
| include_path_multiple | Two --include values OR'd — cherry-picks from different directories |
| include_path_no_match | Filter matches nothing — graceful "no output files available" message |
| exclude | --include "renders/" --exclude "renders/draft/" — downloads renders except drafts |
| glob_pattern | --include "renders/*.exr" — fnmatch glob downloads only .exr files |
| include_exclude_config | --include-exclude-config filters.json — JSON file with include + exclude |
| include_exclude_config_large_inline_json | --include-exclude-config '{...}' — inline JSON blob selecting 150 of 200 files (DCM use case) |

waninggibbon:
Discussed this extensively offline, here is a quick summary of the latest changes:

### What changed

CLI options:
- --include (kept) — glob patterns matched against the full path (root + relative), e.g. */renders/*.png or *.exr
- --include-config (renamed from --include-exclude-config) — JSON string or file path with include patterns, e.g. {"include": ["*/renders/*.exr"]}
- --submission-path (new flag) — when set, filters match against the original submission paths instead of the local workstation paths
- --exclude — removed for now, we can add it back later if needed

Filtering behavior:
- Glob patterns now match against the full path (root + relative) instead of just the relative path. This enables patterns like */renders/*.png that match across directory structures.
- By default, filters are applied against workstation paths (after path mapping). Pass --submission-path to filter against the original submission paths instead.
- Windows backslash root paths are normalized to forward slashes for cross-OS glob matching, without affecting the actual download paths.

Code quality:
- Extracted _full_path() helper to DRY up path joining in _filter_paths and _filter_manifests
- Added OutputDownloader.apply_include_filters() method instead of reaching into internals from the CLI layer
- Renamed _parse_include_exclude → _parse_include_config

### DCM compatibility

The --include-config option accepting a file path covers DCM's use case for large filter sets (avoids Windows CreateProcess 32KB argument limit). The deadline:// protocol path is
unaffected — it only does full downloads.

### Tests

All existing tests updated, plus new coverage for:
- Full-path glob matching (including Windows root paths)
- --submission-path flag behavior
- Workstation path filtering with patterns containing root directory names

waninggibbon:
## Summary of latest changes

Simplified the CLI to just two options: --include and --submission-path.

### --include (repeatable)

Supports both glob patterns and plain relative paths:

bash
# Glob patterns
deadline job download-output --include "*.exr"
deadline job download-output --include "*/renders/*"

# Relative paths (matched as suffix against full path)
deadline job download-output --include "renders/frame_001.exr"

# Multiple filters (OR'd)
deadline job download-output -i "renders/frame_001.exr" -i "renders/frame_002.exr" -i "*.log"


Patterns are matched against the full path (root + relative), so */renders/*.png works across directory structures. Plain relative paths like renders/frame_001.exr match as a suffix —
any full path ending in /renders/frame_001.exr will match.

### --submission-path flag

By default, filters match against the local workstation paths (after storage profile mapping). Pass --submission-path to filter against the original submission paths instead.

### Why this covers all our use cases

| Use case | How it works |
|---|---|
| Human CLI | --include "*.exr" or --include "renders/frame_001.exr" |
| DCM desktop (filtered) | Multiple --include args + --submission-path. DCM has relative paths from manifests and can pass them directly. Tauri scope uses args: true to allow variable argument count. |
| Web copy/paste | Repeated --include flags in the generated command |
| Web protocol (deadline://) | Unaffected — still does full downloads, no filtering |
| Direct S3 ("Save Selected") | Unaffected — bypasses CLI entirely |

### What was removed

- --exclude — removed for now, can add back later
- --include-exclude-config / --include-config — not needed since --include handles both globs and exact paths
- --path-filter-stdin — not needed; DCM can use repeated --include args instead

### Cross-OS

Windows backslash root paths are normalized to forward slashes for matching without affecting download paths. A Windows user's --include "renders\frame.exr" works the same as a Mac
user's --include "renders/frame.exr".

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -305,3 +305,32 @@ def _on_progress_json(download_metadata: ProgressReportMetadata) -> bool:
...
+        if f.startswith("./"):
+            f = f[2:]
+        while "//" in f:
+            f = f.replace("//", "/")
+        if f:
+            normalized.append(f)
+    return normalized
+
+
+def _parse_include_filters(include: tuple[str, ...]) -> Optional[list[str]]:
```
crowecawcaw: Nit: feels like an unnecessary function that could be rolled into _normalize_filters instead.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -548,6 +552,8 @@ def _download_job_output(
         for manifest in job_attachments_manifests:
             root_path_format_mapping[manifest["rootPath"]] = manifest["rootPathFormat"]
 
+    # When --submission-path is set, filter against the submission (S3) paths.
```
crowecawcaw: Super nit: more like the relative job attachment paths. S3 is more of an implementation detail of job attachments

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -557,6 +563,7 @@ def _download_job_output(
         task_id=task_id,
         session_action_id=session_action_id,
         session=queue_role_session,
+        include_filters=include_filters if submission_path else None,
```
crowecawcaw: Nit: could use better variable names to differentiate the two include_filters variables. There's probably a 3rd meaning elsewhere for workstation filters.

src/deadline/job_attachments/download.py
```diff
@@ -1243,6 +1244,98 @@ def mount_vfs_from_manifests(
...
+    if normalized_root.endswith("/"):
+        return normalized_root + relative
+    return normalized_root + "/" + relative
+
+
+def _is_relative_filter(f: str) -> bool:
+    """Check if a filter is a plain relative path (no glob characters, not absolute)."""
+    return (
+        not any(c in f for c in ("*", "?", "[")) and not f.startswith("/") and not f.endswith("/")
+    )
```
crowecawcaw: Why can't relative filters be globs?

One idea: Auto-prepend `*/` to any relative filter (glob or not) before fnmatching, and drop the literal suffix branch entirely. It could collapse logic to one rule: "relative patterns match anywhere under the root."

waninggibbon: Good idea, that is easier than trying to maintain separate paths for glob vs relative.

src/deadline/client/cli/_groups/_job_download_helpers.py
waninggibbon: Yeah pretty redundant, I went ahead and inlined it in the new implementation.

src/deadline/client/cli/_groups/job_group.py
waninggibbon: Dropped "s3" in latest commit!

src/deadline/client/cli/_groups/job_group.py
waninggibbon: Renamed to `include_patterns` when referring to user provided patterns, and `include_filters` when referring to the actual filter that gets applied to distinguish between the two!

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -966,6 +986,21 @@ def _assert_valid_path(path: str) -> None:
...
+    "the full path (root + relative). Supports *, ?, [seq]. A trailing / matches all "
+    "files under that directory. Repeatable",
+)
+@click.option(
+    "--submission-path",
+    is_flag=True,
+    default=False,
+    help="Match include filters against the original submission paths instead of "
+    "the local workstation paths. By default, filters match against workstation paths.",
+)
```
crowecawcaw, line 1005: Could be better:
- the True/False boolean doesn't map cleanly to what this means
- currently sounds like it takes a path

What about `--filter-paths-by WORKSTATION` or `--filter-paths-by MANIFEST` or some variation on those.

waninggibbon: Updated to `--match-paths-by SOURCE|WORKSTATION` in the latest commit

src/deadline/job_attachments/download.py
```diff
@@ -1243,6 +1244,90 @@ def mount_vfs_from_manifests(
         vfs_manager.start(session_dir=session_dir)
 
 
+def _full_path(root: str, relative: str) -> str:
```
mwiebe: There's other code that joins the roots and manifest paths. Can we use the same mechanism here to make sure we don't have differences?

src/deadline/job_attachments/download.py
```diff
@@ -1243,6 +1244,90 @@ def mount_vfs_from_manifests(
...
+    A filter ending with '/' is treated as a directory prefix (matches all files under it).
+    Relative filters (not starting with '/' or '*') are auto-prepended with '*/' so they
+    match anywhere under the root — e.g. 'renders/*.exr' matches '*/renders/*.exr'.
+    The file_path should be the full path (root + relative).
+    """
+    from fnmatch import fnmatch
+
+    for f in filters:
+        if f.endswith("/"):
+            if file_path.startswith(f):
```
mwiebe: Shouldn't this still use fnmatch?

test/unit/deadline_job_attachments/test_path_filtering.py
```diff
@@ -0,0 +1,269 @@
...
+    BaseManifestPath,
+)
+from deadline.job_attachments.asset_manifests.hash_algorithms import HashAlgorithm
+from deadline.job_attachments.asset_manifests.v2023_03_03 import (
+    AssetManifest as AssetManifestv2023_03_03,
+    ManifestPath as ManifestPathv2023_03_03,
+)
+
+
+class TestMatchesAnyFilter:
```
mwiebe: Looks like this is only testing against posix paths, need windows paths with `\` separators too.

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -499,6 +501,8 @@ def _download_job_output(
     task_id: Optional[str],
     is_json_format: bool = False,
     ignore_storage_profiles: bool = False,
+    include_patterns: Optional[list[str]] = None,
+    match_paths_by: str = "LOCAL",
```
leongdl: Question, why a hard coded "LOCAL", not a string enum?

leongdl: I see this parameter is LOCAL or JOB?

leongdl: Would it cause the code to be brittle since it implies users to know this string has 2 explicit values.

src/deadline/job_attachments/download.py
```diff
@@ -1243,6 +1244,90 @@ def mount_vfs_from_manifests(
...
+def _matches_any_filter(file_path: str, filters: list[str]) -> bool:
+    """
+    Check if a file path matches any of the given filters using glob-style matching.
+    Uses fnmatch for pattern matching (supports *, ?, [seq], [!seq]).
+    A filter ending with '/' is treated as a directory prefix (matches all files under it).
+    Relative filters (not starting with '/' or '*') are auto-prepended with '*/' so they
+    match anywhere under the root — e.g. 'renders/*.exr' matches '*/renders/*.exr'.
+    The file_path should be the full path (root + relative).
+    """
+    from fnmatch import fnmatch
```
leongdl: nit :) Kiro loves inline imports, can we move it up to the top.

src/deadline/job_attachments/download.py
waninggibbon: Yes, good catch. Will fix in next commit.

src/deadline/job_attachments/download.py
leongdl: https://github.com/aws-deadline/deadline-cloud-job-attachments/blob/mainline/src/deadline/job_attachments/download.py

WARNING - we cannot merge this now.

leongdl (dismissed):
https://github.com/aws-deadline/deadline-cloud-job-attachments/blob/mainline/src/deadline/job_attachments/download.py

WARNING - this collides with the JA separation.

src/deadline/job_attachments/download.py
waninggibbon: Thanks for the heads up, I'll raise a new PR against the new repo.

src/deadline/client/cli/_groups/job_group.py
waninggibbon: Just a foolish blunder, no more reasoning than that. Will fix!

test/unit/deadline_job_attachments/test_path_filtering.py
waninggibbon: Adding in next commit!

waninggibbon:
Status update — ready for re-review

  The job-attachments separation is complete and the dependency is now live:

  - deadline-cloud-job-attachments v0.0.2 has been released with the path filtering functions (_full_path, _matches_any_filter, _filter_paths, filter_manifests, apply_include_filters)
  — see #21
  - This PR now depends on deadline-job-attachments == 0.0.2 and imports the public filtering API from the package instead of defining it inline in download.py
  - Merged upstream mainline to pick up recent changes (job attachments migration, boto3 bump, etc.)

  What's in this PR (CLI layer only):

  - --include / -i (repeatable) glob patterns for download-output, matched against full paths via fnmatch
  - --match-paths-by enum (JOB or LOCAL) to choose whether filters match against submission paths or local workstation paths
  - Relative paths auto-prepend */ for suffix matching (DCM use case)
  - Shared download helpers extracted for the follow-up download-input PR

  All previous review feedback has been addressed:

  - Replaced --include-path/--include-path-stdin with glob-based --include (per @leongdl's feedback to match manifest CLI patterns)
  - Removed --exclude and --include-config to simplify the CLI surface
  - Renamed --submission-path flag → --match-paths-by JOB|LOCAL enum (per @crowecawcaw's feedback)
  - Using posixpath.join in _full_path for consistency with _transform_manifests_to_absolute_paths (per @mwiebe's feedback)
  - fnmatch for directory prefix filters instead of startswith
  - Windows path separator tests added
  - Inline fnmatch import moved to top of file

  Tests: 2121 unit tests passing, 20 attachment e2e tests passing (including 10 download-output filter scenarios), lint and mypy clean.

crowecawcaw (dismissed):

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -966,6 +987,22 @@ def _assert_valid_path(path: str) -> None:
...
+    help="Glob pattern or relative path for files to include in download. Matched against "
+    "the full path (root + relative). Supports *, ?, [seq]. A trailing / matches all "
+    "files under that directory. Repeatable",
+)
+@click.option(
+    "--match-paths-by",
+    type=click.Choice(["JOB", "LOCAL"], case_sensitive=False),
+    default="LOCAL",
+    help="Control which paths --include filters are matched against. "
+    "JOB matches against the paths recorded at job submission. "
```
crowecawcaw, line 1003: Nit: could also be job outputs. Wording might benefit from some tightening up. Someone from JA might understand the technicalities better to offer a suggestion. Not blocking.

crowecawcaw (approved):

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -42,6 +43,13 @@
 JSON_MSG_TYPE_PROGRESS = "progress"
 
 
+class MatchPathsBy(str, Enum):
+    """Which paths --include filters are matched against."""
```
baxeaz, line 47: This could maybe use a better description for what a JOB path vs a LOCAL path is

baxeaz (approved):

PR #1117, opened 2026-04-17
feat: Update 'deadline job trace-schedule' to use batch get APIs.

mwiebe:

The `deadline job trace-schedule` command needs to get all the steps and tasks of the job it's tracing to produce an accurate visualization. For a 10K-task job, this could take about 20 minutes.

Use the new batch APIs (See https://docs.aws.amazon.com/deadline-cloud/latest/userguide/release-notes.html 2026-04-06 entry) to get the steps and tasks in batches instead of individually.

The 10K-task job took about 4 minutes to collect the trace.

Added new functions to the deadline_mock fixture (that works with moto) for the batch APIs being used, including error injection to test partial failure and retry cases. Implemented tests for the trace schedule based on this.

Compared the before/after of the test job, confirmed that the data was the same except for how it assigned the workers to process ids in the trace. Loaded the trace in the Perfetto UI to confirm it worked there.

N/A, this is a performance improvement of this command.

No

It raises an error for the trace-schedule command on Python 3.8. The command is marked as experimental, so requiring Python 3.9+ seems reasonable.

No

----


docs/design/trace-schedule-batch-apis.md
```diff
@@ -0,0 +1,220 @@
...
+
+Measured on a 10,000-task job:
+
+| | Before | After |
+|-|-|-|
+| Wall clock | ~1402 s (~23 min) | ~226 s (~4 min) |
+| `get_task` round-trips | ~10,000 | 0 |
+| `batch_get_task` round-trips | 0 | 100 |
+
+The remaining time is dominated by the serial per-session
```
baxeaz, line 19: Could these requests not be made in parallel, would that help?  (Also are we missing a Batch API here?)

mwiebe: Yes, I explicitly limited the scope of this change to be about the batch APIs, I think parallel requests should be a separate change.

crowecawcaw (approved):

baxeaz (approved):

PR #1118, opened 2026-04-17
chore: add fleet and worker command tests, add http interface for mock backend

crowecawcaw:
We manually test fleet and worker commands - tedious.

1. Extend the mock backend to also support running as a localhost HTTP server so we can run the full CLI without any patching against a local endpoint.
2. Add automated tests for fleets and workers.
 
Automates tests. Adds a new testing pattern for end-to-end CLI testing without needing credentials.

It is tests

n/a

No

No

No
----

### DIFF
--- a/test/unit/deadline_client/mock_deadline_backend.py
+++ b/test/unit/deadline_client/mock_deadline_backend.py
@@ -7,15 +7,41 @@ See docs/design/deadline-tests-mock-backend.md for design details.
 """
 
 from __future__ import annotations
+import json as _json
+import re as _re
+import threading as _threading
+import traceback as _traceback
 from datetime import datetime, timedelta, timezone
+from http.server import BaseHTTPRequestHandler as _BaseHTTPRequestHandler
+from http.server import HTTPServer as _HTTPServer
 from typing import Any
 from unittest.mock import MagicMock
+from urllib.parse import parse_qs as _parse_qs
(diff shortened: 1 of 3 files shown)


test/unit/deadline_client/mock_deadline_backend.py
```diff
@@ -222,15 +253,129 @@ def create_queue(self, *, farmId: str, displayName: str, **kwargs) -> dict:
...
+    @route("GET", "/farms/{farmId}/fleets", "ListFleets")
+    def list_fleets(
+        self, *, farmId: str, maxResults: int = 100, nextToken: str | None = None, **kwargs
+    ) -> dict:
+        params = {"farmId": farmId, "maxResults": maxResults}
+        if nextToken is not None:
+            params["nextToken"] = nextToken
+        params.update(kwargs)
+        self._validate("ListFleets", params)
+        fleets = [dict(f) for k, f in self.fleets.items() if k[0] == farmId]
```
kavmur, line 319: nit: Do we need to use maxResults here, we may want to test pagination?

kavmur (approved):

leon-li-inspire (approved):

test/unit/deadline_client/mock_deadline_backend.py
crowecawcaw: Great call! I'll merge this for now to unblock some other work and expand these tests.

PR #1122, opened 2026-04-20
ci: run Code Quality on mainline pushes to seed shared cache

crowecawcaw:
GitHub Actions scopes caches per ref. PR runs can restore caches from the default branch, but caches created on PR merge refs aren't shared across PRs. Since Code Quality only ran on pull_request, mainline had no cache and every new PR started cold.

Adding a push trigger on mainline populates a shared cache that all new PRs can restore from on their first run.

PRs can use the cache from mainline and will mostly get a cache hit for faster builds.

n/a

n/a

No

No

No

----

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -3,6 +3,9 @@ name: Code Quality
 on:
   pull_request:
     branches: [ mainline, release, feature_assets_cli, 'patch_*' , "feature_*"]
+  push:
+    # Run on mainline pushes to populate the Actions cache that PRs restore from.
+    branches: [ mainline ]
   workflow_call:
     inputs:
       branch:


leon-li-inspire (approved):

.github/workflows/code_quality.yml
```diff
@@ -3,6 +3,9 @@ name: Code Quality
 on:
   pull_request:
     branches: [ mainline, release, feature_assets_cli, 'patch_*' , "feature_*"]
+  push:
+    # Run on mainline pushes to populate the Actions cache that PRs restore from.
+    branches: [ mainline ]
```
kavmur, line 8: This also triggers for changes outside the source code (e.g., CI workflow edits, doc updates). Can we add paths filter to limit it to src/ and test/?

kavmur (approved):

.github/workflows/code_quality.yml
crowecawcaw: We'd also want to run this whenever dependencies change in `requirement.txt` files or similar. I don't think there's a downside to running these checks on every merge. Is there a downside to leaving it?

PR #1123, opened 2026-04-21
feat: support conflict resolution in job download output path mapped cases

### LINKED ISSUE #829: Enhancement: job attachments interface does not use storage profiles by default

leongdl:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/829

Fixes: https://github.com/aws-deadline/deadline-cloud/pull/1029#discussion_r3113550910

The mapped-manifest download path (used when storage profiles resolve path mapping rules) silently defaulted to `CREATE_COPY` for file conflicts without offering the user the same skip/overwrite/copy prompt that the existing OutputDownloader path provides. This inconsistency meant users on the automatic mapping path had no way to choose how to handle existing files at the destination.

Extracted the conflict resolution logic into a shared `_resolve_conflict_resolution()` function in `_job_download_helpers.py`. Both the mapped-manifest path and the existing OutputDownloader path now call this same function, which:

1. Uses the config setting if `settings.conflict_resolution` is set
2. Defaults to `CREATE_COPY` if `--yes` (auto_accept) is passed
3. Otherwise checks for conflicting files and prompts the user with skip/overwrite/copy/cancel options

Both download code paths now have identical conflict resolution behavior. Users re-downloading to a destination with existing files will see the same interactive prompt regardless of whether storage profile mapping is active.

- 5 new unit tests in `test_job_download_helpers.py` for `_resolve_conflict_resolution` (config priority, auto_accept default, no conflicts, user cancel, user overwrite)
- All existing unit tests pass (53 relevant tests, 2559 total)
- 9/9 e2e tests pass against live AWS resources

- [ ] Have you made changes to the `download` or `asset_sync` modules? No.

- Docstring added to `_resolve_conflict_resolution()`

No. The OutputDownloader path behavior is unchanged — same logic, just moved to a shared function. The mapped-manifest path gains the conflict prompt it was previously missing.

No. Only affects user-facing prompt logic for file conflict resolution.

---

### DIFF
--- a/src/deadline/client/cli/_groups/_job_download_helpers.py
+++ b/src/deadline/client/cli/_groups/_job_download_helpers.py
@@ -179,12 +179,55 @@ def _transform_manifests_to_absolute_paths(
     return result
 
 
+def _resolve_conflict_resolution(
+    conflict_resolution_setting: str,
+    auto_accept: bool,
+    conflicting_filenames: list[str],
+) -> Optional[FileConflictResolution]:
+    """Determine file conflict resolution, prompting the user if needed.
+
+    Shared logic for both the mapped-manifest and OutputDownloader download paths.
+
+    Args:
+        conflict_resolution_setting: raw config setting string (e.g. "CREATE_COPY", "NOT_SELECTED").
+        auto_accept: whether to skip interactive prompts (--yes flag).
+        conflicting_filenames: list of absolute paths that already exist locally.
+
+    Returns:
+        FileConflictResolution to use, or None if the user canceled.
+    """
+    if conflict_resolution_setting != FileConflictResolution.NOT_SELECTED.name:
+        return FileConflictResolution[conflict_resolution_setting]
+
+    if auto_accept:
+        return FileConflictResolution.CREATE_COPY
+
+    if not conflicting_filenames:
+        return FileConflictResolution.CREATE_COPY
+
+    # Lazy import to avoid circular dependency
+    from .job_group import _get_conflict_resolution_selection_message
+
+    click.echo(_get_conflict_resolution_selection_message(conflicting_filenames))
+    user_choice = click.prompt(
+        "> Please enter your choice (1, 2, 3, or n to cancel the download)",
+        type=click.Choice(["1", "2", "3", "n"]),
+        default="3",
+    )
+    if user_choice == "n":
+        click.echo("Output download canceled.")
+        return None
+
+    return FileConflictResolution(int(user_choice))
+
+
 def _download_mapped_manifests(
     mapped_manifests: dict[str, Any],
     queue: dict[str, Any],
     queue_role_session: Any,
     conflict_resolution_setting: str,
     is_json_format: bool,
+    auto_accept: bool = False,
 ) -> Any:
     """Download output files using path-mapped manifests with progress reporting.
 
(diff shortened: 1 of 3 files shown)


src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -179,12 +179,55 @@ def _transform_manifests_to_absolute_paths(
     return result
 
 
+def _resolve_conflict_resolution(
```
leongdl, line 182: Purely re-factored code, nothing new.

See below
https://github.com/aws-deadline/deadline-cloud/pull/1123/changes#diff-1ee0c7304bc12bdf60c243a04a88f93122a7196dc84e3c5a09321c035f25351eL725

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -719,29 +723,12 @@ def _check_and_warn_long_output_paths(
         click.echo("\nSummary of file paths to download:")
         click.echo(textwrap.indent(summarize_path_list(all_output_paths), "  "))
 
-    # If the conflict resolution option was not specified, auto-accept is false, and
-    # if there are any conflicting files in local, prompt users to select a resolution method.
-    # (skip, overwrite, or make a copy.)
-    if conflict_resolution != FileConflictResolution.NOT_SELECTED.name:
```
leongdl, line 725: This is the source of the function above.

src/deadline/client/cli/_groups/_job_download_helpers.py
```diff
@@ -199,16 +242,25 @@ def _download_mapped_manifests(
...
-    # Determine conflict resolution: use config setting if specified, otherwise CREATE_COPY.
-    # No interactive prompt on the mapped path — the user opted into automatic mapping.
-    if conflict_resolution_setting != FileConflictResolution.NOT_SELECTED.name:
-        file_conflict_resolution = FileConflictResolution[conflict_resolution_setting]
-    else:
-        file_conflict_resolution = FileConflictResolution.CREATE_COPY
+    from pathlib import Path as _Path
+
+    # Check for conflicting files at the mapped destinations
+    conflicting_filenames: list[str] = []
```
leongdl, line 253: At this point in the code, manifest_path.path has already been transformed by _transform_manifests_to_absolute_paths() into a full absolute local destination path (e.g., 
result.exr
). So _Path(manifest_path.path).is_file() is checking whether the file already exists at the exact location it would be downloaded to.

Compare with the OutputDownloader path's _get_conflicting_filenames():
```
for root, filenames in filenames_by_root.items():
    for filename in filenames:
        abs_path = Path(root).joinpath(filename).resolve()
        if abs_path.is_file():
```
Same idea — joins root + relative to get the absolute destination, then checks if it exists. Our mapped path just skips the join because the paths are already absolute.

mwiebe (approved):

src/deadline/client/cli/_groups/job_group.py
```diff
@@ -612,7 +613,10 @@ def _check_and_warn_long_output_paths(
                     queue_role_session=queue_role_session,
                     conflict_resolution_setting=conflict_resolution,
                     is_json_format=is_json_format,
+                    auto_accept=auto_accept,
                 )
+                if download_summary is None:
```
godobyte, line 618: Q - did we used to have print here to say nothing to be downloaded? Having some visibility might be good for user.

godobyte (approved):

src/deadline/client/cli/_groups/job_group.py
leongdl: This is just an early return, the `download_summary` is None case is reflected above:
```
   if user_choice == "n":
        click.echo("Output download canceled.")
        return None
```

So it would say "Output download canceelled" and the prints are not changed.

AlexTranAmz (approved):

PR #1135, opened 2026-04-24
docs: Added documentation on how customers can deactivate update noti…

karthikbekalp:

Just adding some public documentation on how customers can deactivate the update notification feature. 

----

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -157,6 +157,11 @@ $ deadline config gui
 
 By default, configuration of AWS Deadline Cloud is provided at `~/.deadline/config`, however this can be overridden by the `DEADLINE_CONFIG_FILE_PATH` environment variable.
 
+Submitter integrations check for newer versions on startup. To deactivate these notifications:
+```sh
+$ deadline config set settings.submitter_update_notification false
+```
+
 ## Authentication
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -157,6 +157,38 @@ $ deadline config gui
 
 By default, configuration of AWS Deadline Cloud is provided at `~/.deadline/config`, however this can be overridden by the `DEADLINE_CONFIG_FILE_PATH` environment variable.
 
+## Submitter Update Notifications
```
crowecawcaw: This feels like a lot of docs on the main page for a simple toggle flag. Would it also work someplace else that's less top level? Also could we shorted it?

crowecawcaw (changes requested):

docs/index.md
karthikbekalp: Hmm, good questions. I added it to the main page because I want IT admins who set up deadline cloud for artists to know they can deactivate the option. 

Is there a more suitable place that you can think of? 

> Also could we shorted it?

I think we can. 

How about this? 

```
Submitter integrations check for newer versions on startup. To deactivate these notifications:

$ deadline config set settings.submitter_update_notification false

```

karthikbekalp: Thanks, updated!

diane-hj (approved):

crowecawcaw (approved):

PR #1145, opened 2026-04-28
test: add DCM integration test workflows (Linux, macOS, Windows)

crowecawcaw:
We want to automated DCM + Deadline client integ tests.

Add them. New tests download DCM and then use the Deadline CLI to authenticate and call `deadline farm list`. Tests run on PR.

Less manual testing

...

n/a

No

No

No
----

### DIFF
--- a/.github/scripts/dcm_e2e_test.py
+++ b/.github/scripts/dcm_e2e_test.py
@@ -1,0 +1,204 @@
+"""E2E: deadline auth login → DCM opens browser → drive IdC sign-in via xa11y.
+
+Cross-platform. Relies on xa11y >= 0.7 for:
+  - xa11y.screenshot().save_png(path)  — native screen capture on all platforms
+  - Locator.type_text(str)              — splice text via a11y API (no keystrokes)
+"""
+
+import os
+import platform
+import subprocess
+import sys
+import time
+
+import xa11y
+
+sys.stdout.reconfigure(line_buffering=True)
+
+IS_LINUX = platform.system() == "Linux"
+
+USERNAME = os.environ["MONITOR_USERNAME"]
+PASSWORD = os.environ["MONITOR_PASSWORD"]
+
+# Browser xa11y app name and Open-Link button label vary per platform.
+BROWSER_NAME = os.environ.get("BROWSER_NAME", "Firefox")
+OPEN_LINK_LABELS = os.environ.get("OPEN_LINK_LABELS", "Open Link,Open link,Open").split(",")
+
+SCREENSHOT_DIR = os.environ.get("SCREENSHOT_DIR", "/tmp")
+
+
+def run(cmd, check=True, timeout=180):
+    print(f"$ {' '.join(cmd)}", flush=True)
+    r = subprocess.run(cmd, text=True, capture_output=True, timeout=timeout)
+    print(r.stdout, r.stderr, flush=True)
+    if check and r.returncode:
+        fail(f"failed: {cmd}")
+    return r
+
+
+def screenshot(name):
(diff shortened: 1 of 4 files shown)


baxeaz (commented):
Does this need to be merged before we see these tests running in the PR?  Or am I not looking in the right place?

crowecawcaw:
@baxeaz usually yes, but I got them running on this PR already by pushing changes to a branch in this repo and raising the PR from it. As a security feature, GitHub will not pass secrets to PR actions if the PR is being introduced in the PR and if the PR source is a fork. But it will allow it if the PR is from the same repo because it means the owner has write permissions already.

.github/scripts/dcm_e2e_test.py
```diff
@@ -0,0 +1,204 @@
...
+
+# Browser xa11y app name and Open-Link button label vary per platform.
+BROWSER_NAME = os.environ.get("BROWSER_NAME", "Firefox")
+OPEN_LINK_LABELS = os.environ.get("OPEN_LINK_LABELS", "Open Link,Open link,Open").split(",")
+
+SCREENSHOT_DIR = os.environ.get("SCREENSHOT_DIR", "/tmp")
+
+
+def run(cmd, check=True, timeout=180):
+    print(f"$ {' '.join(cmd)}", flush=True)
```
Cherie-Chen, line 31: with this print statement, are we having the risk of printing out username and password?

.github/scripts/dcm_e2e_test.py
```diff
@@ -0,0 +1,204 @@
...
+IS_LINUX = platform.system() == "Linux"
+
+USERNAME = os.environ["MONITOR_USERNAME"]
+PASSWORD = os.environ["MONITOR_PASSWORD"]
+
+# Browser xa11y app name and Open-Link button label vary per platform.
+BROWSER_NAME = os.environ.get("BROWSER_NAME", "Firefox")
+OPEN_LINK_LABELS = os.environ.get("OPEN_LINK_LABELS", "Open Link,Open link,Open").split(",")
+
+SCREENSHOT_DIR = os.environ.get("SCREENSHOT_DIR", "/tmp")
```
andychoquette, line 27: Just curious - how can we review screenshots for failed tests? Will a link to screenshots be shared in the failed test output? Or will they render somewhere in the test results?

andychoquette (commented):
LGTM

andychoquette (approved):

.github/scripts/dcm_e2e_test.py
crowecawcaw: GitHub has a hand feature it automatically redacts secrets from logs. So since the username, password, and monitor subdomain are secrets, their values will be filtered out.

leon-li-inspire (approved):

.github/scripts/dcm_e2e_test.py
crowecawcaw: They should appear as artifacts on the overall workflow, but I'm not seeing them: https://github.com/aws-deadline/deadline-cloud/actions/runs/25062315769/

I'm going to merge since they're passing but will poke around a bit more to see if there's a missing step.

crowecawcaw: That's actually the wrong run. Here's an example with a screenshot. It's just a black frame though, maybe a gap on Linux with virtual displays and screenshots. https://github.com/aws-deadline/deadline-cloud/actions/runs/25061313881#artifacts

Here's a Mac one with something useful: https://github.com/aws-deadline/deadline-cloud/actions/runs/25061314610#artifacts

PR #1151, opened 2026-04-29
fix: restore deadline auth login after STS-to-ListFarms switch

crowecawcaw:
## Summary

`check_authentication_status` (introduced in #1147) broke `deadline auth login` against Deadline Cloud monitor profiles. The DCM integration tests (`dcm_integration_test_{linux,macos,windows}.yml`) have hung for 180s on every run since then; the same hang is observable for any human running `deadline auth login` against an IdC profile.

Two bugs, both fixed here. Either one alone is insufficient.

## Bug 1: ListFarms probe was missing `principalId`

`check_authentication_status` called `deadline:ListFarms(maxResults=1)` without injecting `principalId` from the DCM profile's `user_id`. For IdC-issued sessions the service rejects the un-scoped call with AccessDenied, so `_login_deadline_cloud_monitor`'s polling loop never sees `AUTHENTICATED`.

The sibling `check_deadline_api_available` and the `api.list_farms` wrapper already inject `principalId`; this change makes both probes share that logic.

**Fix:** Extract a private `_list_farms_for_auth_probe()` helper in `_session.py` and have both `check_authentication_status` and `check_deadline_api_available` call it. Single source of truth for the principalId-injection policy.

## Bug 2: stale profile-map cache in the poll loop

After fixing bug 1, `deadline auth login` still hung — even though a separate `deadline auth status` process reported `AUTHENTICATED` / `api_availability: true`.

Root cause: `get_user_and_identity_store_id` goes through botocore's `get_scoped_config()`, which caches the parsed profile map on first read. The polling loop in `_login_deadline_cloud_monitor` probes every 0.5s. The first probe runs BEFORE DCM has written `user_id`/`identity_store_id`/the real `monitor_id` to `~/.aws/config`, so botocore caches the pre-login profile. Subsequent probes (even after DCM finishes writing) see the same cached profile → `user_id` is still `None` → ListFarms is called without `principalId` → AccessDenied → NEEDS_LOGIN forever.

This is specific to long-running processes like `deadline auth login`; fresh invocations (`deadline auth status`, `deadline farm list`) create a new session and load the profile once, so they don't hit this.

**Fix:** Invalidate the boto3 session cache each iteration of the poll loop in `_login_deadline_cloud_monitor` so each probe reads the current on-disk profile.

## Testing

- `hatch run test test/unit/` → 2085 passed, 21 skipped.
- `hatch run typing`, `ruff check` → clean.
- DCM integration tests dispatched against this branch: **all three platforms pass** (run [25170595534](https://github.com/aws-deadline/deadline-cloud/actions/runs/25170595534)):
  - linux: 1m 52s (was hanging at 4m 45s before)
  - macos: 1m 54s (was hanging at 5m)
  - windows: 2m 10s (was hanging at 5m 18s)
- manually running the auth flow locally with DCM creds 

Timings match the pre-regression baseline from 2026-04-28.

## Known pre-existing API surface

The check-api-changes job flags `deadline.client.api.Any` and `deadline.client.api.Dict` as removed. These were incidental re-exports of `typing.Any`/`typing.Dict` that leaked into the public namespace via a top-level `from typing import Any, Dict, Optional`. The inline usage was removed as part of refactoring `check_deadline_api_available` to call the shared helper; they are not legitimate Deadline APIs, so I left them removed.

## Follow-up

`check_authentication_status` (3-way enum) and `check_deadline_api_available` (bool) now share the same probe but still exist as separate public helpers with different return types and failure semantics. Consolidating them would touch the public surface (`deadline auth status` CLI, GUI status widget) and is tracked in #1152.

## Signed-off-by

All commits include `Signed-off-by` per repository policy.

### DIFF
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -297,10 +297,10 @@ def _list_farms_for_auth_probe(config: Optional[ConfigParser] = None) -> None:
     Raises whatever exception the underlying boto3 call raises; callers are
     responsible for exception handling.
     """
+    from ._list_apis import _apply_principal_id_filter
+
     list_farm_params: dict = {"maxResults": 1}
-    user_id, _ = get_user_and_identity_store_id(config=config)
-    if user_id:
-        list_farm_params["principalId"] = str(user_id)
+    _apply_principal_id_filter(list_farm_params, config=config)
     get_boto3_client("deadline", config=config).list_farms(**list_farm_params)
 
 
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -283,6 +283,27 @@ def _modified_logging_level(logger, level):
         logger.setLevel(old_level)
 
 
+def _list_farms_for_auth_probe(config: Optional[ConfigParser] = None) -> None:
+    """
+    Makes a minimal ``deadline:ListFarms`` call used as an auth/reachability
+    probe by :func:`check_authentication_status` and
+    :func:`check_deadline_api_available`.
+
+    For Deadline Cloud monitor profiles, injects ``principalId`` so the call
+    is scoped to the caller's IdC user (matching the :func:`api.list_farms`
+    wrapper). Without it, IdC-issued sessions get denied by the service and
+    the auth-login poll loop never resolves to AUTHENTICATED.
+
+    Raises whatever exception the underlying boto3 call raises; callers are
+    responsible for exception handling.
+    """
+    list_farm_params: dict = {"maxResults": 1}
+    user_id, _ = get_user_and_identity_store_id(config=config)
+    if user_id:
+        list_farm_params["principalId"] = str(user_id)
+    get_boto3_client("deadline", config=config).list_farms(**list_farm_params)
+
+
 def check_authentication_status(
     config: Optional[ConfigParser] = None,
 ) -> AwsAuthenticationStatus:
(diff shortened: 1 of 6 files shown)


src/deadline/client/api/_session.py
```diff
@@ -283,6 +283,27 @@ def _modified_logging_level(logger, level):
...
+    wrapper). Without it, IdC-issued sessions get denied by the service and
+    the auth-login poll loop never resolves to AUTHENTICATED.
+
+    Raises whatever exception the underlying boto3 call raises; callers are
+    responsible for exception handling.
+    """
+    list_farm_params: dict = {"maxResults": 1}
+    user_id, _ = get_user_and_identity_store_id(config=config)
+    if user_id:
+        list_farm_params["principalId"] = str(user_id)
```
baxeaz: I wonder if we want to centralize this at some point for the operations that need it rather than need to remember to do this on the 4? or so List operations, maybe as part of #1152

baxeaz (commented):
For Bug 2 it sounds like a race condition?  Do we want other code using that "pre login" cached stale profile map info or do we want some other mechanism for other threads to have to wait till the login is resolved?

src/deadline/client/api/_session.py
crowecawcaw: Deduped this code.

Regarding the race condition: it's not really a race condition, more like waiting for stale data to be refreshed. DCM and the Deadline Client both us the config file. Here in the Client, we're waiting for DCM to go through the sign in flow and save the user ID back to the config file. The polling is a normal part of waiting for the sign in process, not hiding a design flaw.

baxeaz (approved):

Cherie-Chen (approved):

PR #1132, opened 2026-04-23
test: add pytest-qt GUI tests 

andychoquette:
Add pytest-qt tests that replace the Squish tst_verify_settings_dialogue and tst_verify_gui_submitter_bundles test suites. These tests use MockDeadlineBackend to provide fake API responses in-process, requiring no AWS credentials, no Squish license, and no external infrastructure.

New files:
- test/gui/conftest.py: shared fixture for MockDeadlineBackend
- test/gui/test_settings_dialogue.py: 14 tests for the config dialog
- test/gui/test_gui_submitter_bundles.py: 6 tests for the GUI submitter

Changes:
- mock_deadline_backend.py: add ListFarms, ListQueues, ListStorageProfilesForQueue APIs and storage_profiles storage
- hatch.toml: add [envs.gui] with pytest-qt and PySide6 deps
- code_quality.yml: add GUITests job (Linux/macOS/Windows matrix)

Run locally with: hatch run gui:test

Fixes: *<insert link to GitHub issue here>*

### How was this change tested?

See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#testing) for information on running tests.

- Have you run the unit tests?
- Have you run the integration tests?
- Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended
  that you ensure that the docker-based unit tests pass.

### Was this change documented?

- Are relevant docstrings in the code base updated?
- Has the README.md been updated? If you modified CLI arguments, for instance.

### Does this PR introduce new dependencies?

This library is designed to be integrated into third-party applications that have bespoke and customized deployment environments. Adding dependencies will increase the chance of library version conflicts and incompatabilities. Please evaluate the addition of new dependencies. See the [Dependencies](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies) section of DEVELOPMENT.md for more details.

*   [ ] This PR adds one or more new dependency Python packages. I acknowledge I have reviewed the considerations for adding dependencies in [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies).
*   [ ] This PR does not add any new dependencies.

### Is this a breaking change?

A breaking change is one that modifies a public contract in a way that is not backwards compatible. See the 
[Public Contracts](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#public-contracts) section
[...]

PR #1157, opened 2026-05-07
test(ui): cover 'Load a different job bundle' button flow

crowecawcaw:
We have a manual test case for the "Load another bundle" button.

Automate it away.

Removes a manual test case.

PR actions

n/a

No

No

No
----


crowecawcaw:
This will make more sense using pytest-qt. Closing and will wait to build on this PR: https://github.com/aws-deadline/deadline-cloud/pull/1132/

PR #1158, opened 2026-05-07
feat: add beforeUI submission hook phase

rickrams:
Adds a new beforeUI hook phase that runs before the job submission dialog opens, allowing studios to pre-populate dialog fields such as job name, description, and parameters (including deadline: prefixed shared job properties) from pipeline context.

Hook output schema:
  name        - pre-fill job name
  description - pre-fill job description
  parameters  - pre-fill parameters by name; use deadline: prefix for
                shared job properties (priority, maxFailedTasksCount,
                maxRetriesPerTask, maxWorkerCount, targetTaskRunStatus)

beforeUI hooks are gated by the existing allow_bundle_hooks / allow_environment_hooks settings and show a confirmation prompt unless auto_accept is enabled.

When using the gui submitter there's use cases to adjust fields automatically for users before the UI is displayed, especially use cases with system defined hooks where certain things like priority, naming, or other best practices can be dynamically set as starting defaults when the UI is launched.

Adds a new beforeUI hook phase that runs before the job submission dialog opens

Locally tested a few different parameters and scenarios.

- Have you run the unit tests? Yes!
- Have you run the integration tests? No.

Yes

No

*   [*] This PR does not add any new dependencies.

No

It's adding additional script execution so there's potentially some impact. It's reusing the same basic appraoch as the existing hooks feature though so in practice should be minimal impact.

----

### DIFF
--- a/docs/submission-hooks.md
+++ b/docs/submission-hooks.md
@@ -12,6 +12,11 @@ Place a `hooks.yaml` (or `hooks.json`) in your job bundle directory alongside `t
 
 ```yaml
 version: "1.0"
+preUI:
+  - command: python3
+    args: [prefill.py]
+    timeout: 10
+
 preSubmission:
   - command: python3
     args: [validate.py]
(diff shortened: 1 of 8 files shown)


docs/submission-hooks.md
```diff
@@ -12,6 +12,11 @@ Place a `hooks.yaml` (or `hooks.json`) in your job bundle directory alongside `t
 
 ```yaml
 version: "1.0"
+beforeUI:
```
crowecawcaw: Maybe "preUI" to match the submission hooks?

docs/submission-hooks.md
```diff
@@ -37,6 +42,58 @@ Requires `settings.allow_environment_hooks` to be enabled. Both sources can be a
 
 ## Hook Types
 
+### Before-UI Hooks
+
+Run **before** the submission dialog opens. Use these to:
+- Pre-populate job name, description, and priority
+- Set parameter defaults based on the current scene or pipeline context
+- Query a project management system for task metadata
```
crowecawcaw, line 50: These can be done with CLI parameters today. I guess the hook adds another way to do it with external scripts instead of adding logic to the thing that calls the submitter. Can you give a bit more context around the use case?

rickrams: Doing it via a system hook would allow a TD / IT Admin to enforce standardization across many users. Also conditional naming based on the date or a look up in a production tracking system.

docs/submission-hooks.md
rickrams: ooh good call!

src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -309,6 +370,28 @@ def on_create_job_bundle_callback(
     for parameter in job_parameters_dict.values():
         initial_shared_parameter_values[parameter["name"]] = parameter["value"]
 
+    # Merge pre-UI hook output — hooks override CLI parameters for shared fields,
```
leon-li-inspire: I feel this comment is a little confusing and misleading because 

"hooks override CLI parameters for shared fields" — suggests hooks win for shared properties, but actually hooks are applied to shared properties only because CLI doesn't guard them (they weren't in job_parameters_dict since CLI params matching template names get consumed earlier).


Will the following comment be better ? 
```
  # Apply pre-UI hook output to pre-fill dialog fields.
  # CLI --parameter values take precedence over hook values for template and queue parameters.
  # Hook-supplied name/description always apply (no CLI precedence check).
```

or 

```
  # Apply pre-UI hook output: skip parameters already set by CLI, apply the rest.
```

src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -281,6 +313,35 @@ def on_create_job_bundle_callback(
...
     initial_settings.browse_enabled = browse
 
+    # Run pre-UI hooks to allow studios to pre-populate dialog fields.
+    pre_ui_output: dict[str, Any] = {}
+    allow_env_hooks = _config_file.str2bool(_get_setting("settings.allow_environment_hooks"))
+    allow_bundle_hooks = _config_file.str2bool(_get_setting("settings.allow_bundle_hooks"))
+    hook_manager = _HookManager(input_job_bundle_dir, logger.info)
+    hooks = hook_manager.load_hooks()
+    if hooks and hooks.pre_ui and not allow_bundle_hooks:
+        logger.warning(
```
leon-li-inspire, line 323: Is it intentional to print this warning log while `allow_env_hooks` is true so that we will still execute the hooks ?

src/deadline/client/ui/job_bundle_submitter.py
rickrams: Thanks, addressed.

src/deadline/client/ui/job_bundle_submitter.py
rickrams: ah fixed.

leon-li-inspire (approved):

leon-li-inspire:
Replaced by https://github.com/aws-deadline/deadline-cloud/pull/1178

PR #1172, opened 2026-05-13
chore(ui): add version information label to main submit dialog

jericht:
Related to: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/222

The submitter versions are currently available but are hidden away in the Help menu

Add a label to the submitter dialog with a summary of the versions of submitter installed 
- deadline-cloud
- Submitter
- DCC

It clips the version numbers to 10 chars (probably we won't go more than `99.123.123` for any version of our software) in the label while the tooltip has the full version string

There's a label with the above versions in the main submit dialog

- Added a unit test
- Installed it to local Blender submitter and verify label and tooltip look nice

image

No

No

No

No

----


jericht:
Decided not to do this, Help button has the information required.

PR #1173, opened 2026-05-14
chore: unify GUI and unit test coverage into single test suite

andychoquette:
## Summary

Adds pytest-qt based GUI unit tests for the Deadline Cloud submitter and settings dialogs, running them alongside the existing unit and CLI e2e tests under a single unified coverage measurement.

- Add GUI unit tests covering settings dialog, host requirements, job attachments, job timeouts, OpenJD parameters, shared job properties, submitter bundles, dataclasses, and utility widgets
- Include `deadline.client.ui` in the main coverage measurement (previously omitted) — combined coverage is 72.74% against a 69% threshold
- Add `features = ["gui"]` to the default hatch environment so PySide6 is available for test runs
- Add `python_version >= '3.9'` marker to PySide6 dependency to avoid CI failures on Python 3.8
- Pin `virtualenv<21` for hatch compatibility
- Remove unused imports flagged by ruff
- Update `test/README.md` with GUI unit test documentation

## Test plan

- [x] `hatch run test` passes locally with all 2460 tests (including GUI tests) passing
- [x] Combined coverage at 72.74% exceeds the 69% threshold
- [x] GUI tests run in offscreen mode without requiring a display
- [x] CI passes on all matrix combinations (Linux/macOS/Windows × Python 3.9/3.11/3.13)


crowecawcaw:
PR description suggests we want to track GUI test coverage separately and should focus on core flows only. Is that the case? Seems like if GUI code isn't as well tested, we should reflect that in the coverage and close the gaps. GUI tests with pytest-qt are probably fast right?

andychoquette:
> PR description suggests we want to track GUI test coverage separately and should focus on core flows only. Is that the case? Seems like if GUI code isn't as well tested, we should reflect that in the coverage and close the gaps. GUI tests with pytest-qt are probably fast right?

Yeah, after reflecting on this a bit, the better course of action is to hold these tests to the same coverage standard as other tests in this repo. They run quickly enough that increasing coverage really doesn't have a downside. I'll revise this PR to include new tests in the existing workflow to get the coverage up to acceptable levels.

crowecawcaw:
Converting to a draft until CI actions pass

test/unit/deadline_client/ui/gui/test_gui_dataclasses.py
```diff
@@ -0,0 +1,652 @@
...
+        assert result[0]["max"] == (2**31) - 1
+
+
+class TestCustomAmountRequirement:
+    # Note: CustomAmountRequirement.DEFAULT_VALUE is -(2**31) + 1 = -2147483647.
+    # The __post_init__ validation uses `self.min and self.max` which treats
+    # the negative default as truthy. When only min is set (positive) and max
+    # stays at DEFAULT_VALUE (negative), min > max triggers a ValueError.
+    # Therefore, setting only min (positive) without max is not valid with
+    # this class. Tests below reflect the actual behavior.
```
crowecawcaw: Maybe worth fixing instead of testing the current behavior

crowecawcaw:
LGTM, ready to review?

test/unit/deadline_client/ui/gui/test_gui_dataclasses.py
andychoquette: done - thanks for the callout.

crowecawcaw (approved):

leon-li-inspire (approved):

PR #1175, opened 2026-05-19
feat: emit process_start telemetry event on client initialization

andychoquette:
Enables accurate session counting for the FFSR metric denominator by emitting a com.amazon.rum.deadline.process_start event when the telemetry background thread starts.

We intend to track FFSR (fault free session rate) across our interfaces to understand session-level service reliability. This is calculated by: 

`FFSR = (Total Sessions - Sessions with Faults) / Total Sessions × 100%`

To properly measure this metric, we need a count of unique sessions. 

Emit a `session_start` event when clients initialize, so unique sessions can be counted. 

Added a unit test to cover the behavior

- Have you run the integration tests? No

No

No

No
----


crowecawcaw:
I think this will emit an event whenever the library inits which is either when the submitter opens or when a CLI command is invoked. Both are probably more like events in a session than a new session itself. Is there another way to count sessions, maybe starting from CreateMonitorSession or maybe using some sort of timespan (e.g. all events from the same machine in the same hour).

andychoquette:
> I think this will emit an event whenever the library inits which is either when the submitter opens or when a CLI command is invoked. Both are probably more like events in a session than a new session itself. Is there another way to count sessions, maybe starting from CreateMonitorSession or maybe using some sort of timespan (e.g. all events from the same machine in the same hour).

In that case, we can query the `session_start` event over a certain timespan to track a session, something like the following, where `telemetry_id` identifies the unique machine:

```
fields @timestamp, UserDetails.userId as telemetry_id
  | stats count(*) as event_count,
          count_distinct(UserDetails.sessionId) as process_count
    by bin(1h) as session_window, telemetry_id
```

Having a `session_start` metric feels like the right way to guarantee at least one event is logged per process invocation. Since it isn't actually measuring the session start, I can rename it to `process_start` or `process_invocation`. What do you think?

leon-li-inspire (dismissed):

waninggibbon (approved):

leon-li-inspire (approved):

PR #1181, opened 2026-05-29
feat: job bundle browser/sharing

epmog:
> 📄 **Start here:** Review the [design document](https://github.com/epmog/deadline-cloud/blob/98ebb399bdc895977647aac83f83c8cd72d94055/docs/design/job-bundle-browser.md) included in this PR
  for the full design rationale and architecture.

  When using `deadline bundle gui-submit --browse`, the "Load a different job bundle" button, or "Submit
  Job Bundle" menu item in Deadline Cloud monitor, users are presented with a native OS folder picker.
  This is inadequate because:

  - Job bundles are not distinguishable from regular folders by name alone
  - Users must already know where their bundles live and navigate there manually
  - There is no preview of what a bundle contains — users pick blindly
  - There is no way to share bundles across a team without out-of-band file sharing
  - The picker always starts at the job history directory or home, with no way to configure a default
  location

  ## What was the solution? (How)

  Replace the native folder picker with a custom job bundle browser dialog.

image

  It provides:

  1. **A navigable tree view** showing only folders and job bundles (📦/📁 icons), with a text filter for
  quick search
  2. **A preview panel** showing bundle name, description, steps, and parameters when a bundle is selected
  3. **Three browsing sources**: Local filesystem, S3 (shared team bundles), and Job History (previously
  submitted bundles)
  4. **S3 bundle sharing** via `.ojd` archives (zip format) stored in the queue's job attachment bucket
  under a `job-bundles/` prefix — no extra configuration needed
  5. **A "Save bundle as" button** in the submitter dialog to upload the current bundle to S3 for team
  access, or export locally
  6. **ETag-based caching** of S3 archives for fast repeated access
  7. **A configurable default browse directory** setting
  8. **Bundle visibility** — hide/unhide bundles on S3 per user by right-clicking the bundle and selecting the option
  9. **Background loading** — dialog appears instantly (~0.26s), S3 data loads asynchronously with
  parallelized API calls
  10. **Keyboard navigation** — source radios, filter, tree, and buttons are all tab-navigable

  Key design decisions:
  - S3 only supports `.ojd` archives (not loose folder bundles) — this gives a single canonical sharing
  format with atomic uploads and simple cache validation
  - The bundle name in the browser preview is shown unresolved (with `{{Param.X}}` intact) so users see
  the template identity rather than a specific parameterization
  - S3 bucket is derived from the queue's existing job attachment settings — zero additional configuration
  - `from_config` parallelizes the GetQueue API call with S3 credential resolution for faster cold starts
  - Pre-warms boto3 session and Deadline client before starting background thread to avoid GIL contention

  Additionally there are new CLI commands:

  | Command | Description |
  |---------|-------------|
  | `deadline bundle list [path]` | List bundles in a local directory (default: configured dir or home) |
  | `deadline bundle list --queue` | List .ojd bundles from the queue's S3 job-bundles folder |
  | `deadline bundle upload <dir>` | Package a local bundle as .ojd and upload to S3 |
  | `deadline bundle download <name>` | Download an .ojd bundle from S3 (uses ETag cache) |
  | `deadline bundle info <name>` | Show bundle details (same info as the preview panel) |
  | `deadline bundle hide <name>` | Hide a bundle from the browser and list (remains in S3) |
  | `deadline bundle unhide <name>` | Unhide a previously hidden bundle |

  Examples:
```
  $ deadline bundle list --queue
  blender-render
  maya-arnold

  $ deadline bundle upload ./my-render-job --name custom-name
  Archiving  [####################################]  100%
  Uploading  [####################################]  100%
  Uploaded bundle to s3://my-farm-bucket/DeadlineCloud/job-bundles/custom-name.ojd

  $ deadline bundle download blender-render -o /tmp/bundles
  Downloading  [####################################]  100%
  Extracting   [####################################]  100%
  Downloaded bundle to: /tmp/bundles/blender-render

  $ deadline bundle download blender-render
  Downloaded bundle to: /Users/me/.deadline/cache/job-bundles/abc123/blender-render

  $ deadline bundle info blender-render --queue
  Path: s3://my-farm-bucket/DeadlineCloud/job-bundles/blender-render.ojd
  Name: Blender Render
  Description: Renders a Blender scene file using Cycles
  Steps:
    • RenderBlender
  Parameters:
    Frames (STRING) = 1-100
    OutputDir (PATH) = /tmp/output

  $ deadline bundle list --queue --output json | jq -r '.[0].path'
  s3://my-farm-bucket/DeadlineCloud/job-bundles/blender-render.ojd
```

  New configuration:

  | Setting | Description |
  |---------|-------------|
  | `settings.job_bundle_default_directory` | Default local directory when browsing for job bundles (also in settings dialog) |

  Also exposed in the Deadline Cloud settings dialog (Settings → General settings) as a "Job bundle
  directory" picker.

  ## What is the impact of this change?

  - Users get a richer bundle selection experience with preview and filtering
  - Teams can share job bundles via S3 without manual file distribution
  - The existing `--browse` flag and "Load a different job bundle" button now open the new dialog instead
  of the native picker
  - New `.ojd` archive format introduced as the standard for shared bundles
  - Progress bars shown for archive/upload/download/extract operations (chunked for large files)
  - MCP tools (`list_shared_bundles`, `upload_bundle`, `download_bundle`) enable AI assistants to manage
  shared bundles

  ## How was this change tested?

  ## Was this change documented?

  - [x] Has the README.md been updated?

  Design document at `docs/design/job-bundle-browser.md` covers the full design. CLI commands include help
  text. Code has docstrings on all public classes and methods.

  ## Does this PR introduce new dependencies?

  ## Is this a breaking change?

  No

  ## Does this change impact security?

  The S3 archive cache at `~/.deadline/cache/job-bundles/` stores archived bundles. Input validation is performed on archives before extraction to prevent path traversal attacks (absolute paths and `../` entries are
  rejected, using `os.path.commonpath` with `os.path.realpath` for cross-platform safety). Symlinks are
  skipped during archiving (`os.walk(followlinks=False)`). The cache directory follows the same permission
  model as the existing `~/.deadline/` directory.

  S3 access uses the same boto3 session/profile as the rest of deadline-cloud — no new credential flows.

  ---

  By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this
  contribution, under the terms of your choice.


docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,417 @@
...
+7. Respects a configurable default local browse directory.
+
+## Design
+
+### Bundle Formats
+
+Job bundles can be either:
+
+- **Directories** — a folder containing `template.yaml` or `template.json` at the root, plus any scripts, data files, and `asset_references.yaml`.
+- **Archives** — an `.ojd` file (zip format under the hood) containing a job bundle. The template can be at the archive root or inside a single wrapper directory.
```
crowecawcaw, line 31: This makes sense to me. I think it should be formalized in the OJD spec as part of this project.

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,417 @@
...
+
+    def root_path(self) -> str:
+        """The starting path for browsing."""
+        ...
+```
+
+Two implementations:
+
+- `LocalBundleRepository` — walks the local filesystem. Lists directories and archive files. Directories are bundles if they contain `template.yaml`/`template.json`. Archives are always shown as bundles (validated on preview). Provides `extract_bundle()` for extracting archives to a local directory.
+- `S3BundleRepository` — lists objects and prefixes under the queue's job attachment bucket at `{rootPrefix}/job-bundles/`. Only `.ojd` archives are recognized as bundles; subfolders are shown for navigation only. Provides `resolve_bundle()` which handles archive download+cache+extract.
```
crowecawcaw: I think we should move the labelling of this feature away from "s3" and more towards "queue" or "job attachments". The bundle archives are stored on S3, but that's an implementation details. I think the more customer-facing framing is that these are bundles shared on the queue. So for example, maybe `deadline bundle list --queue` instead of `deadline bundle list --s3`

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,417 @@
...
+
+### S3 Archive Caching
+
+Archive bundles from S3 are cached locally to avoid re-downloading on repeated use.
+
+**Cache location**: `~/.deadline/cache/job-bundles/{hash}/{bundle-name}/`
+
+Where `{hash}` is a truncated SHA-256 of `{bucket}/{s3-key}` to ensure uniqueness.
+
+**Cache validation**: On each access, a single `head_object` call retrieves the archive's ETag. If it matches the cached ETag, the local copy is used directly. If it differs (or no cache exists), the archive is re-downloaded and re-extracted.
```
crowecawcaw: Also need to recheck the metadata. If the bundle's s3 metadata is renamed but the bundle hash is not, we should show the new name.

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,417 @@
...
+**Why only archives are cached**: An archive is a single S3 object with a single ETag — one `head_object` validates the entire bundle.
+
+### S3 Object Metadata for Preview
+
+When `deadline bundle upload` uploads an archive, it attaches bundle metadata as S3 user metadata on the object:
+
+- `bundle-name`: The template's `name` field
+- `bundle-description`: The template's `description` field (newlines collapsed to spaces)
+- `bundle-steps`: Comma-separated list of step names
+- `bundle-parameters`: Comma-separated `name:type` pairs
```
crowecawcaw: Some risk of the metadata not aligning with the job template. Would nice if there were a way to read this stuff from the template directly instead. Especially because HEAD and GET requests cost the same amount and take about as long, so there's not a optimization reason to use HEADs. Might might working with the zip archives more complicated though.

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,417 @@
...
+- Path display showing the current browse location.
+- Cancel and Select buttons. Select is enabled only when a valid bundle is highlighted.
+
+### Share Button
+
+The submitter dialog includes a "Share" button alongside the existing "Export bundle" and "Submit" buttons. Clicking "Share" packages the current job bundle as an `.ojd` archive and uploads it to the queue's S3 `job-bundles/` folder, making it available to the team via the browser's S3 source. The bundle name defaults to the job name, with `{{Param.X}}` references resolved using current parameter values. Spaces and slashes in the resolved name are replaced with underscores. S3 user metadata (name, description, steps, parameters) is attached for zero-download preview.
+
+Share is enabled when the API is available and a farm and queue are configured — it does not require valid queue parameters (unlike Submit), since sharing only needs S3 access, not a runnable job configuration.
+
+Note: uploading a bundle with the same name as an existing one silently overwrites it in S3.
```
crowecawcaw: Better to show a confirm dialog first

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,417 @@
...
+- Path display showing the current browse location.
+- Cancel and Select buttons. Select is enabled only when a valid bundle is highlighted.
+
+### Share Button
+
+The submitter dialog includes a "Share" button alongside the existing "Export bundle" and "Submit" buttons. Clicking "Share" packages the current job bundle as an `.ojd` archive and uploads it to the queue's S3 `job-bundles/` folder, making it available to the team via the browser's S3 source. The bundle name defaults to the job name, with `{{Param.X}}` references resolved using current parameter values. Spaces and slashes in the resolved name are replaced with underscores. S3 user metadata (name, description, steps, parameters) is attached for zero-download preview.
```
crowecawcaw: Maybe we could call this "Add to queue bundles" or similar so it clearer how the bundles are organized and how permissions work.

crowecawcaw:
Thought from offline discussion: would be great to have a way to rename, update, or delete bundles from UI, especially for the queue view.

docs/design/job-bundle-browser.md
epmog: So the future looking thing here is that job bundles (portable) will contain their own assets, and with them being part of the artifact that exists in s3 we want to avoid downloading lots of data just to populate the bundle viewer. I don't want to download 10s of GB just to know the description of the job.

docs/design/job-bundle-browser.md
epmog: I always try to err on the side of short action names to not overload the user. I think maybe the thing to ask is, does `Share` communicate what we want. Similar to `Export Bundle`

docs/design/job-bundle-browser.md
epmog: Yea that makes sense to me!

docs/design/job-bundle-browser.md
epmog: Yea, we've put off formalizing job bundles for a while now. Something we can get started

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -508,6 +512,147 @@ def on_export_bundle(self):
...
+                )
+        except Exception as exc:
+            QMessageBox.critical(self, "Share failed", f"Failed to create bundle:\n{exc}")
+            return
+
+        # Get queue S3 settings
+        try:
+            farm_id = get_setting("defaults.farm_id")
+            queue_id = get_setting("defaults.queue_id")
+            queue_obj = get_queue(farm_id=farm_id, queue_id=queue_id)
```
andychoquette: This `get_queue(...)` call doesn't pass a `session=`, and the upload below uses a bare `boto3.client("s3")`. Both fall back to boto3's default credential chain instead of the session that `deadline.client.api.get_boto3_session()` builds from `defaults.aws_profile_name` in `~/.deadline/config`. Result: clicking **Share** fails with `AccessDeniedException ... The security token included in the request is expired`, even though the rest of the app works fine.

The same issue exists in `cli/_groups/bundle_group.py`: `_get_queue_s3_settings()` calls `get_queue(...)` without a session, and `S3BundleRepository(...)` is constructed without one.

andychoquette:
a few UX thoughts: 
1. Consider allowing users to toggle visibility of hidden folders in the folder explorer. Showing them by default adds a lot of noise, and I can't imagine many users will be storing job bundles in hidden folders. 
Screenshot 2026-06-08 at 12 36 03 PM

2. Maybe we can render the bundle parameters as a table? Its challenging to visually parse data from a bulleted list where kv pairs are delimited by `=`s.
Screenshot 2026-06-08 at 12 37 12 PM

3. Since the source selection affects the folder picker component, we should flip these options. Ideally, users can achieve their goal in a form linearly, top-to-bottom, and I find myself jumping past the folder picker to select a source, just to return to the folder picker. Subjective, but worth considering. 

Screenshot 2026-06-08 at 3 33 57 PM

4. If the S3 (queue) option is disabled, due to missing permission for example, we should tell the user why in an inline alert. 

Screenshot 2026-06-08 at 3 33 39 PM

5. This is existing behavior I think, but after exporting a bundle, the window is closed automatically. This is unexpected and not communicated. 

6. I find myself wanting to browse around in the bundle view, but its only accessible via an action button. Consider giving it a dedicated tab so it can be navigated to and from with ease. This'll also allow us to eliminate the Load bundle button, too. Here's an idea:
Screenshot 2026-06-08 at 1 31 12 PM

7. The saving / sharing experience is a bit confusing, and the confusion is exacerbated by this new functionality. For example, when I export my job bundle, it isn't configurable during the interaction and closes the submitter when done.  

An option: drop the Export and Share buttons in favor of a "Save bundle" button that opens a dedicated save dialog where we can communicate where it will be saved and what the implications are (e.g. sharing to queue shares to all queue users). Example: 



We can also include an info alert that describes the implications of sharing when sharing is selected: 

Screenshot 2026-06-08 at 3 34 18 PM

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
epmog: Now pulling the session and threading it through, thanks!

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,633 @@
...
+        preview_layout.addWidget(self._preview_params_label)
+        self._preview_params = QTableWidget()
+        self._preview_params.setColumnCount(3)
+        self._preview_params.setHorizontalHeaderLabels(["Name", "Type", "Value"])
+        self._preview_params.horizontalHeader().setStretchLastSection(True)
+        self._preview_params.horizontalHeader().setSectionResizeMode(QHeaderView.ResizeToContents)
+        self._preview_params.verticalHeader().setVisible(False)
+        self._preview_params.setEditTriggers(QTableWidget.NoEditTriggers)
+        self._preview_params.setSelectionMode(QTableWidget.NoSelection)
+        preview_layout.addWidget(self._preview_params, stretch=1)
```
phil-IO-p: The parameters table is the only widget with `stretch=1`, so it's meant to absorb the preview panel's leftover height. But for a bundle with no parameters the table is hidden, leaving no stretch item. The layout then spreads the extra height between the Name/Description/Steps labels, producing big vertical gaps.

To reproduce: a bundle with no `parameterDefinitions` shows the gaps; one with parameters looks fine. A fix that worked for me is to drop the table's stretch and add a trailing spacer so the fields stay packed at the top, like so:

```python
preview_layout.addWidget(self._preview_params)
preview_layout.addStretch(1)
```

src/deadline/client/ui/dialogs/export_bundle_dialog.py
```diff
@@ -0,0 +1,142 @@
...
+        name_row.addWidget(QLabel("Name:"))
+        self._name_edit = QLineEdit(default_name)
+        name_row.addWidget(self._name_edit)
+        layout.addLayout(name_row)
+
+        # Save to
+        source_row = QHBoxLayout()
+        source_row.addWidget(QLabel("Save to:"))
+        self._radio_queue = QRadioButton(tr("Queue"))
+        self._radio_queue.setEnabled(self._queue_available)
```
phil-IO-p, line 77: The "Queue" option here can look available when it isn't. It's enabled based on `on _queue_available` though it never checks whether the user can actually write to the S3 bucket. So a user signed in through the Deadline monitor (no direct bucket access) sees Queue enabled, fills in the form, clicks Export, and only then gets a raw `AccessDenied` error. The browser dialog greys out Queue when S3 isn't reachable so could Export do the same up-front check so the two behave consistently?

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -576,42 +635,125 @@ def on_export_bundle(self):
...
-                tr("{submitter} job submission").format(
-                    submitter=self.submitter_info.submitter_name
-                ),
-                message,
-            )  # type: ignore[call-arg]
+                tr("Export bundle"),
+                f"Bundle exported to queue:\ns3://{queue_repo._bucket}/{s3_key}",
+            )
+        except Exception as exc:
+            QMessageBox.critical(self, "Export failed", f"Failed to upload bundle:\n{exc}")
```
phil-IO-p: This surfaces the raw boto exception (e.g. "...is not authorized to perform: s3:PutObject...no identity-based policy") directly to the user. That is arguably a bit too technical (i.e. not actionable) for most users. Could we catch the `AccessDenied` case and show something like "You don't have permission to share bundles on this queue" instead?

src/deadline/client/ui/dialogs/export_bundle_dialog.py
epmog: I've changed the permissions to use the queue role permissions which do have access to writing to that bucket

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -576,42 +635,125 @@ def on_export_bundle(self):
...
-            if sys.platform == "win32":
-                # Open the directory in the OS's file explorer
-                os.startfile(self.job_history_bundle_dir)
+    def _export_to_local(self, dest_dir: str, bundle_name: str):
+        """Copy the bundle to a local directory."""
+        assert self.job_history_bundle_dir is not None
+        dest_path = os.path.join(dest_dir, bundle_name)
+        try:
+            if os.path.exists(dest_path):
+                shutil.rmtree(dest_path)
```
phil-IO-p: Local export seems to delete an existing same-named folder with no confirmation. I reproduced this by exporting the same bundle to the same directory twice; the first copy was silently replaced. The Queue export path (_export_to_queue, the "already exists, overwrite?" prompt around line 722) does ask first. Could we add the same confirm here?

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,633 @@
...
+        super().__init__(parent=parent)
+        self.setWindowTitle(tr("Browse Job Bundles"))
+        self.setMinimumSize(750, 550)
+        self.resize(850, 620)
+
+        self._s3_repo: Optional[S3BundleRepository] = queue_source
+        self._s3_error = queue_error
+        self._s3_available = self._s3_repo is not None
+
+        self._local_repo = LocalBundleRepository(root=local_source, include_archives=False)
```
phil-IO-p: The GUI Local source passes `include_archives=False`, so a local `.ojd` never appears in the browser (the CLI's bundle list does show it). The issue is in the following flow: after `deadline bundle download <name>` the .ojd lands locally, but `gui-submit --browse` → Local doesn't show it, so the user can't find what they just downloaded. Is hiding archives in the Local view intended? If archives are meant to be Queue-only in the GUI, would it be better for download to extract to a directory instead of leaving a `.ojd`?

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,633 @@
...
+        if self._radio_local.isChecked():
+            self._current_repo = self._local_repo
+        elif self._radio_s3.isChecked() and self._s3_repo:
+            self._current_repo = self._s3_repo
+        elif self._radio_history.isChecked() and self._history_repo:
+            self._current_repo = self._history_repo
+        self._selected_path = None
+        self._select_button.setEnabled(False)
+        self._clear_preview()
+        self._populate_root()
```
phil-IO-p: Switching source rebuilds the tree from scratch, so the user's expanded folders and scroll position are lost every toggle; I noticed this flipping Queue to Local. Could we keep a small per-source state so that switching back restores where they were?

src/deadline/client/ui/dialogs/export_bundle_dialog.py
```diff
@@ -0,0 +1,142 @@
...
+    @property
+    def local_directory(self) -> str:
+        return self._location_edit.text()
+
+    def _build_ui(self, default_name: str):
+        layout = QVBoxLayout(self)
+
+        # Name
+        name_row = QHBoxLayout()
+        name_row.addWidget(QLabel("Name:"))
```
phil-IO-p: Can we also translate this by wrapping in `tr()`?

src/deadline/client/ui/dialogs/export_bundle_dialog.py
```diff
@@ -0,0 +1,142 @@
...
+        self._name_edit = QLineEdit(default_name)
+        name_row.addWidget(self._name_edit)
+        layout.addLayout(name_row)
+
+        # Save to
+        source_row = QHBoxLayout()
+        source_row.addWidget(QLabel("Save to:"))
```
phil-IO-p: Can we also translate this by wrapping in `tr()`?

src/deadline/client/ui/dialogs/export_bundle_dialog.py
```diff
@@ -0,0 +1,142 @@
...
+        if not self._queue_available and self._queue_error:
+            self._queue_warning.setText(f"\u26a0 <b>Queue unavailable:</b> {self._queue_error}")
+            self._queue_warning.setVisible(True)
+        else:
+            self._queue_warning.setVisible(False)
+        layout.addWidget(self._queue_warning)
+
+        # Location
+        location_row = QHBoxLayout()
+        location_row.addWidget(QLabel("Location:"))
```
phil-IO-p: Can we also translate this by wrapping in `tr()`?

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,633 @@
...
+        source_label = QLabel(tr("Source:"))
+        source_row.addWidget(source_label)
+        self._radio_s3 = QRadioButton(tr("Queue"))
+        self._radio_s3.setEnabled(self._s3_available)
+        self._radio_s3.toggled.connect(self._on_source_changed)
+        source_row.addWidget(self._radio_s3)
+        self._radio_local = QRadioButton(tr("Local"))
+        self._radio_local.toggled.connect(self._on_source_changed)
+        source_row.addWidget(self._radio_local)
+        self._radio_history = QRadioButton(tr("History"))
```
phil-IO-p, line 475: I couldn't switch the source with the keyboard while testing. Could we make these keyboard-navigable?

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,633 @@
...
+        if isinstance(self._current_repo, S3BundleRepository):
+            try:
+                self._hidden_set = self._current_repo.get_hidden_set()
+            except Exception:
+                logger.debug("Failed to fetch visibility manifest", exc_info=True)
+
+        root = self._model.invisibleRootItem()
+        for entry in self._cached_root_entries:
+            is_hidden = entry.name.startswith(".") or entry.name in self._hidden_set
+            self._add_entry_item(root, entry, is_hidden=is_hidden)
```
phil-IO-p, line 948: When a source has no bundles the tree is just blank with no explanation -- I saw this with an empty Local directory. Could we show a short empty-state message that includes an action (e.g. "Use Export or deadline bundle upload to add one")?

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,633 @@
...
+
+        Handles S3 download/cache, archive extraction, and direct directory paths.
+        Returns None if no selection.
+        """
+        if not self._selected_path:
+            return None
+
+        if self._selected_is_s3 and self._s3_repo:
+            if self._selected_is_archive:
+                return self._s3_repo.resolve_bundle(self._selected_path, "")
```
phil-IO-p: The S3 download/extract here (and the upload at `submit_job_to_deadline_dialog.py:743`, and the CLI upload/download) appear to run synchronously with no progress indication; I didn't observe any QProgress*/QThread/cursor or callback anywhere in these paths, so a large bundle would effectively freeze the dialog (or hang the CLI silently). Notably bundle submit already uses `_ProgressBarCallbackManager` for attachment upload... could the new commands reuse that, and the GUI run the transfer off the UI thread with a progress bar / Cancel (as the design doc's "Progress Indication" section describes)?

src/deadline/client/job_bundle/_repository.py
```diff
@@ -0,0 +1,761 @@
...
+
+        if head:
+            # Check local cache validity
+            cache_valid = meta and _normalize_etag(head.get("ETag")) == _normalize_etag(
+                meta.get("etag")
+            )
+
+            # Try S3 user metadata for preview (set by 'deadline bundle upload')
+            s3_metadata = head.get("Metadata", {})
+            info = _bundle_info_from_s3_metadata(s3_metadata, path)
```
phil-IO-p, line 1394: Preview appears to trust the S3 ojd-* metadata before reading the template, so if a bundle is replaced out-of-band (e.g. `aws s3 cp` overwriting the object but not the metadata) the preview could show stale name/steps/params. Also, the metadata path only carries `name:type` with no parameter values, so the Value column would be blank for those. Could the archive's template be the source of truth, with metadata just a fast hint?

phil-IO-p:
There's hide/unhide but no way to delete or rename a shared bundle. Hidden ones could accumulate in S3 indefinitely, and a mistyped name can only be worked around by uploading a second copy (?). Is delete/rename planned or intentionally out of scope for now?

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -576,42 +635,125 @@ def on_export_bundle(self):
...
+
+            QMessageBox.information(
                 self,
-                tr("{submitter} job submission").format(
-                    submitter=self.submitter_info.submitter_name
-                ),
-                message,
-            )  # type: ignore[call-arg]
+                tr("Export bundle"),
+                f"Bundle exported to queue:\ns3://{queue_repo._bucket}/{s3_key}",
```
phil-IO-p: Success/error messages still show the raw s3://… path. Could we say "Shared on queue <name>" and tuck the S3 path behind a details link, so the implementation detail isn't front-and-center for artists?

phil-IO-p:
+1 an earlier comment that recommended a dedicated tab for bundles. Selecting a bundle loads it into the other tabs. Then we could replace "Export bundle" with two explicit affordances: "Save a copy" and "Share to queue" alongside a "Use bundle" button

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,634 @@
...
+                self._mark_item_error(item)
+            self._select_button.setEnabled(False)
+            return
+
+        self._preview_name.setText(info.name)
+        self._preview_name.setStyleSheet("font-weight: bold; font-size: 14px;")
+        self._preview_name.setVisible(True)
+
+        if info.description:
+            self._preview_desc.setText(info.description)
```
phil-IO-p: The description is set verbatim, so templates that author description as a multi-line YAML block render with the hard line breaks intact, which produces awkward breaks mid-sentence in the preview (the `QLabel`'s word-wrap can't reflow across the embedded `\ns`). Could we collapse intra-paragraph line breaks so it wraps to the panel width, while preserving real paragraph breaks (blank lines)? Something like:

```python
import re

def _normalize_description(text: str) -> str:
  paragraphs = re.split(r"\n\s*\n", text.strip())
  return "\n\n".join(" ".join(p.split()) for p in paragraphs if p.strip())
```
then `self._preview_desc.setText(_normalize_description(info.description))`. This also matches how bundle upload already normalizes the description for S3 metadata `(" ".join(desc.split()))`.

phil-IO-p:
Two related issues with the "Queue unavailable" warning banner, which appears in both `job_bundle_browser_dialog.py` and `export_bundle_dialog.py` with identical styling:

1. **Contrast fails WCAG AA**. The text color `#b35900` on background `#fff3e0` is 4.41:1, below the 4.5:1 minimum for normal-weight body text. Darkening the text to `#7a4200` brings it to 7.35:1 (passes AA/AAA) on the same fill, and a darker border (`#e0a040`) makes the edge legible — the current `#ffcc80` border is only 1.35:1 against the fill.
2. **Not theme-aware**. The submitter inherits the OS/Qt palette (no in-app theme toggle), so a hardcoded light-amber banner looks out of place in dark mode. Deriving the colors from the palette lets it adapt (e.g. light mode `#7a4200`/`#fff3e0`, dark mode `#ffcc80`/`#3a2a10` (9.35:1)).

  Since the styling is duplicated across both dialogs, it'd be worth factoring into a small shared helper in `ui/_utils.py` so they stay consistent.

**Before**

Screenshot 2026-06-16 at 12 17 41 AM

**After**

Screenshot 2026-06-16 at 12 16 18 AM

phil-IO-p:
While reviewing, I prototyped some UX polish for the bundle browser preview panel, the queue-unavailable warning banner, and the tree/layout. Pushed it as a branch for reference:

https://github.com/epmog/deadline-cloud/compare/bundle-sharing...phil-IO-p:deadline-cloud:ux-preview-improvements

**Summary of changes**

- Warning banner (a11y)
  - Fixed low-contrast amber text (was 4.41:1, below WCAG AA) and made the banner theme-aware so it works in light and dark mode
  - Factored the styling into a shared helper so the browser and export dialogs stay consistent
  
- Preview panel
  - Centered empty state with a bundle icon + prompt (also fixes a bug where the placeholder could turn red after viewing an errored bundle)
  - Parameters table: sized to content (no large empty gap), long values wrap instead of getting cut off, friendlier type labels (Text/Path/Number), and a "(required)" flag for params with no default
  - Section headers (Description / Steps / Parameters) with item counts, consistent spacing/type scale, and theme-aware colors
  - Multi-line template descriptions now wrap cleanly instead of breaking mid-sentence
  - Raised, rounded panel surface with a subtle gradient to separate it from the tree

- Tree & layout
  - Folders sorted above bundles
  - Moved the read-only Path field up under the source selector, and grouped "Show hidden" with the filter

Here are some screenshots:

Bundle explorer - Empty (preview) state

Bundle explorer - Bundle details

Happy to chat about these!

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
epmog: Not necessarily Queue-only but download does already extract to a folder. The file extension/archive is sort of meant as a "here's a job bundle that's not being edited and is easiest for sharing". I think future changes will make it even more prominent as we the rest of the tooling is built to handle it and it becomes defined as part of the openjd spec. ie. submitjob right now doesn't support a `.ojd`. Easy enough to add to the clientsoftware, but rather than unarchiving on every submit (potentially expensive), it should be able to just inspect the archive looking for its info.

I'll punt that tangent for now, but yes here we should should be showing local ojd bundles. It's partially a leftover of a previous iteration before .ojd where we supported .zip/.tar.gz/etc.

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
epmog: Yea. There's a bit more I'd like to do to get the performance/feel to a better state before merging this. Like faster loading the dialog. Progress bars fit in that. I'll look into getting some changes up for that.

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
epmog: Should be doable!

src/deadline/client/job_bundle/_repository.py
epmog: Yea better to not mix/match metadata vs. actual bundle - we don't need to use the S3 metadata if we already have a valid template local though it should be the same (unless it was modified in the local cache).

Upcoming change will ensure that:
1. if the E-Tags match, it uses all information from that bundle that was downloaded locally
2. if E-Tag doesn't match, use S3 metadata as a fast hint
3. If no metadata, it can't preview until it's downloaded locally.

As far as I understand the metadata will not be maintained on a new version of an s3 object unless the user specifically says "copy metadata" ala `aws s3api copy-object --metadata-directive COPY`. So I'm not too particularly worried about that.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
epmog: Yea, this information probably isn't as important now that the user is actually choosing where it's being saved. Previously the user has 0 control, so we had to tell them somehow.

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
epmog: done!

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
epmog: image


How about something like this when there are no bundles in the location or the filter doesn't find anything

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
epmog: I've pushed a change that remembers the bundle selection for each source (and the folders expanded to reach it). Let me know your thoughts!

crowecawcaw:
Collection of UX feedback:

Bundle preview — spacing and sizing
  - Title should be the largest text
  - Parameter table should be indented
  - Parameters should come before steps (OJD convention)
  - Steps and parameters should be open by default

  Queue tab
  - Shows only a loading state when I have no queue bundles

  Upload/download
  - Can I upload and download from the GUI? Maybe right-click → upload/download

  History/source confusion
  - History seems to show history only for a queue, but the "queue" source isn't selected, so it seems like it should show all previously submitted jobs. "Local"
  also isn't queue-limited. Not sure what the right answer is, but the expectation could be clearer.

  "Show hidden" is inconsistent across tabs (nit)
  - Queue: shows bundles that were previously hidden
  - Local: shows folders and files starting with a period
  - History: effectively does nothing

  "Filter bundles" behavior
  - Not sure how it works. Searching opens sub-directories but doesn't seem to find a bundle on my desktop. Is it searching? Applying a basic filter to top-level
  items might be more intuitive.
  - After entering filter text and then deleting it, all sub-directories stay expanded.

  Show in Finder (idea)
  - In local and history browsers, add a right-click → "show in Finder" option

  Path input
  - Can't be edited. Maybe make it static text if we need it — but I'm not sure we do, since the same info is shown visually in the browser.

  Naming (nit/idea)
  - Rename "History" → "Local history"? "Queue" means remote, "local" means on the workstation, but "history" is ambiguous — history of queue submissions, or
  history from this workstation?

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,654 @@
...
+```json
+{
+  "etag": "\"d41d8cd98f00b204e9800998ecf8427e\"",
+  "last_modified": "2026-04-30T12:00:00+00:00"
+}
+```
+
+For S3 archives, `etag` is used for validation. For local archives, `mtime` (float, seconds since epoch) is stored instead. The fields present indicate the source type.
+
+**Cleanup**: `deadline bundle cache clean` removes cached bundles from this directory. There is no separate temp dir or `atexit` cleanup — all extracted archives live in the cache.
```
crowecawcaw: I don't love that we're adding a new public concept that needs to be manually maintained. What do we do for the job history directory? Could we auto-prune the cache?

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,654 @@
...
+- Path display showing the current browse location (read-only, positioned under the source selector).
+- "Show hidden" checkbox — grouped with the filter as a view control. Unchecked by default, toggling refreshes the tree to include/exclude hidden items. For Local/History sources, this means dot-prefixed directories. For the Queue source, this means bundles marked as hidden via the visibility manifest (see [Bundle Visibility](#bundle-visibility)).
+
+**Left panel** — Filter and navigable tree view:
+- A text filter at the top that narrows the tree as you type. Case-insensitive, matches against entry names. Uses recursive filtering so parent folders remain visible when a child matches. The tree auto-expands when filtering to show results.
+- Shows folders (📁) and job bundles (📦) with distinct icons. Both directory bundles and archive bundles use the 📦 icon. Local archives show the `.ojd` extension to distinguish them from directory bundles.
```
crowecawcaw, line 210: I found it slightly confusing that my bundles that are directories showed up as a file at the end of the list instead of in the middle of the other directories. Trade offs with either sorting, so ok if you decide to keep it as is too.

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,654 @@
...
+"settings.job_bundle_default_directory": {
+    "default": "",
+    "description": (
+        "The default local directory to open when browsing for job bundles. "
+        "If empty, defaults to the user's home directory."
+    ),
+}
+```
+
+This setting is also exposed in the Deadline Cloud settings dialog (Settings → General settings) as a "Job bundle directory" picker, alongside the existing "Job history directory" setting.
```
crowecawcaw, line 300: Could we make this config store the last location that was used? Seems like a UX opportunity to open to the last place. I find it annoying with our current GUI submitter that I need to keep navigating back to the same bundles folder each time.

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,654 @@
...
+    {"name": "OutputDir", "type": "PATH", "_display_value": "/tmp/output"}
+  ]
+}
+```
+
+### Bundle Visibility
+
+Bundles shared on S3 can be "hidden" without deleting them. This is the S3 equivalent of dot-prefixed hidden folders on local filesystems — bundles remain accessible but are not shown in the browser by default.
+
+**Mechanism — sidecar manifest with optimistic concurrency:**
```
crowecawcaw: I don't like this manifest very much. It adds a new public interface that we need to document and maintain and is only a performance optimization as far as I can tell. I'm not convinced performance without this manifest warrants it. Did we try running S3 HEAD requests in parallel? We could reasonably send dozens at once. We could punt on the manifest until we see customers with hundreds of bundles in one prefix who notice latency issues when loading this window.

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,654 @@
...
+  "version": 1,
+  "hidden": ["blender-render", "old-maya-job"]
+}
+```
+
+The `version` field allows evolving the format in the future (e.g. adding per-entry comments or timestamps) without breaking older clients. The `hidden` array contains bundle names (without the `.ojd` extension) — these correspond 1:1 with object keys since duplicate names are not allowed in the same folder. In Python, the array is deserialized into a `set` for O(1) lookup, and serialized back as a sorted list for stable output.
+
+**Scope:** A single manifest exists at the `job-bundles/` root and covers all bundles including those in subfolders (stored as relative paths, e.g. `"rendering/custom-renderer"`). This avoids additional S3 calls to fetch per-subfolder manifests.
+
+**Re-upload behavior:** If a hidden bundle is re-uploaded (overwritten), it remains hidden. The hide is "sticky" to the name. This is intentional — hiding expresses "this name shouldn't clutter the default view" regardless of the object's content. The export dialog handles this explicitly (see [Export Bundle](#export-bundle)).
```
crowecawcaw: What about deleting bundles by writing a 0 byte object over it? Then we wouldn't need this hiding concept.

docs/design/job-bundle-browser.md
```diff
@@ -0,0 +1,654 @@
...
+### S3 Object Metadata for Preview
+
+When `deadline bundle upload` uploads an archive, it attaches bundle metadata as S3 user metadata on the object:
+
+- `ojd-name`: The template's `name` field (hard cap: 256 chars)
+- `ojd-desc`: The template's `description` field, newlines collapsed to spaces (hard cap: 600 chars)
+- `ojd-steps`: Comma-separated list of step names (dynamically allocated)
+- `ojd-params`: Comma-separated `name:type` pairs (dynamically allocated)
+- `ojd-step-count`: Total number of steps in the template (always included, enables accurate count display when steps are truncated)
+- `ojd-param-count`: Total number of parameters in the template (always included, enables accurate count display when params are truncated)
```
crowecawcaw, line 145: S3 HEAD and GET requests cost the same in terms of $ and latency. As an idea, what if we ensured the template was in the first bytes of the ojd bundle and fetched the first bundles bytes to read these values? Instead of duplicating them in S3 metadata. Single source of true, but at the expensive of more complexity.

Also boto's S3 client incorrectly rejects some characters from S3 metadata, so not all characters will work here. Not a blocker, just a note. Might need a test case for it.

docs/design/job-bundle-browser.md
epmog: Chatted about this previously but deletion is something that I'd decided can be punted til after feature launch if it's needed.

For now I'd much rather Artists be able to easily hide bundles from view that are no longer used and have TD/IT clean-up bundles as they see fit (ie. direct s3 operations)

Having hidden bundles is also similar to a local filesystem browser, where hidden folders/files exist and a user can choose to view/inspect them.

docs/design/job-bundle-browser.md
epmog: This might now be feasible with bundles guaranteed to be a zip archive under the hood (previously it allowed other formats/compressions). A single source of truth is always nice, the main thing is that we don't want to be pulling down the whole bundle just to show a preview.

I'll take a look to see how feasible this is

docs/design/job-bundle-browser.md
epmog: You're asking for the configuration value to change to always be the last selected folder?

The way I've been using this feature, is that I've been treating this directory similarly to the set-up queue bundles. "Here's the location of all the bundles you care about, but locally". Which has completely alleviated my main pain point that was the folder dialog never defaulted to show me something useful.

There's maybe a discussion to have around having Local remember the folder you were in last time, but giving a button to reset that back to the configured value we store here. Thoughts? I'd like to punt on this and see what users think about it. Can always tweak it later.

docs/design/job-bundle-browser.md
epmog: This was feedback from a previous UX review. Seems fine to leave it as-is for now.

docs/design/job-bundle-browser.md
epmog: It was especially handy for me when I was performing testing of this feature and wanted to blow-away an existing cache/ensure I had an up-to-date cache. Adding a mechanism that performs that operation (most likely infrequent) based on the configured directory doesn't seem like a bad thing.

That being said, deadline has quite a few different caches that can all take up disk space that don't have their operations yet (history, hashcache, and s3cache. But it's mainly the history dir). So If we want to take extra time to get this right across all of those, then I can remove and punt this. The larger the bundles people upload/how frequently they update, the sooner we'll want something here to help folks manage their diskspace. But that can wait.

Some of what makes it a little more nuanced here for auto-clean-up is that you could have caches across buckets that you may not be able to verify. So you could only ever clean out stale caches (ie. no longer in the bundle folder) for the active queue unless you decided to purge based on some other factors.

TL;DR I can remove and address it later. Users can purge the cache folder themselves.

docs/design/job-bundle-browser.md
epmog: I'll double check about just sending out a lot of requests in parallel. Even without a lot of job bundles (30ish). It takes about 1-2seconds to populate based on how far away you are from your region to fully populate everything due to required sequential aws calls. Windows pops up much faster than the population and starts with `Loading...`

Currently:

0. Pre-aws calls:
    * Acquire aws credentials (potentially credential process adds latency)
1. Stage 1
    * deadline:GetQueue (s3 bucket)
    * deadline:AssumeQueueRoleForUser (permissions for bucket)
2. Stage 2
    * s3:ListObjectsV2 (`.../job-bundle`)
    * s3:GetObject (.bundle-visibility.json)
3. Browser is fully populated (perceived slowness stops here)
4. On bundle selection - HEAD object on bundle in s3

Proposed change- 0, 1 stay the same, but now we must wait for list object v2 to start polling for visibility

2. List objects in s3 `.../job-bundle`
3. Concurrently perform HEAD on all objects
4. Browser is fully populated and previews are ready from metadata (perceived slowness stops here)

When tinkering with this. The initial population is what felt very slow to me, compared to selecting a bundle and having the preview appear.

epmog: I've gone and made the change and performance has changed in a few ways:

Better:
1.  scrolling through jobs is a tiny bit faster since HEADs are no longer performed when selecting bundle. Cached for the lifetime of the browser currently

Worse:
1. If the number of bundles is more than the number of connection pools configured by the user it costs additional round trips. Around ~1 second longer from my location to populate the bundles from deadline-cloud samples with 50 connection pools.
2. Using user-defined metadata for hiding a bundle is not great since S3 actually needs to perform a copy in-place which invalidates the E-Tags (forcing re-downloads), and copy_object has a 5GB cap. Would need to add a multi-part copy to support bigger bundles.
    * If instead we use S3 object tagging for hiding specifically we no longer continuously overwrite the file in S3, but lose out on the cached HEAD calls. We also then have a requirement that the user's queue role has GetObjectTagging and PutObjectTagging (do not have by default). This would effectively double the number of s3 calls we need to make.

epmog:
> Collection of UX feedback:
> 
> Bundle preview — spacing and sizing
> 
>     * Title should be the largest text

done
>     * Parameter table should be indented

done
>     * Parameters should come before steps (OJD convention)

done
> 
>     * Steps and parameters should be open by default
done
> 
> Queue tab
> 
>     * Shows only a loading state when I have no queue bundles
Fixed the message. For what it's worth it was actually a credential issue. If you don't have bundles but have valid creds, it shows "No bundles", but it was stuck in "Loading" if you didn't have valid creds.
> 
> Upload/download
> 
>     * Can I upload and download from the GUI? Maybe right-click → upload/download
A good idea. There's a download button specifically for queue bundles now, we could have a feature request to upload local ones though.
> 
> History/source confusion
> 
>     * History seems to show history only for a queue, but the "queue" source isn't selected, so it seems like it should show all previously submitted jobs. "Local"
>       also isn't queue-limited. Not sure what the right answer is, but the expectation could be clearer.
It's based on AWS profile. Always has been. Could potentially to "Profile history" but I'm very much in favour of shorter succinct names.
> 
> "Show hidden" is inconsistent across tabs (nit)
> 
>     * Queue: shows bundles that were previously hidden
> 
>     * Local: shows folders and files starting with a period
Queue/Local match in my mind. We've created a Hide/Unhide mechanism for the queue via an S3 mechanism (something users don't need to understand). Local filesystems already have the concept of hidden folders/files so we leverage that instead.
> 
>     * History: effectively does nothing
While we don't expect users to really do anything in history (they can choose the folder but it's written to by us), you can definitely put hidden stuff in there. I don't want to change that because I value the consistency in the UI and switching between sources shouldn't change that.
> 
> 
> "Filter bundles" behavior
> 
>     * Not sure how it works. Searching opens sub-directories but doesn't seem to find a bundle on my desktop. Is it searching? Applying a basic filter to top-level
>       items might be more intuitive.
Folders (Queue/Local) are lazy loaded because it can be expensive to perform all those operations. Locally, we're looking at any folder that contains a template.yaml to determine if it's a bundle which requires you to go a few layers down. We don't explore that or S3 fully unless a user starts going down those nested folders. I've made a change to preload 1 more layer down to maybe help make that feel a little better.
> 
>     * After entering filter text and then deleting it, all sub-directories stay expanded.
Fixed. Now it does it based on if you had actually selected something.
> 
> 
> Show in Finder (idea)
> 
>     * In local and history browsers, add a right-click → "show in Finder" option
I had a similar idea in my head that I implemented. Decided to put a button in the preview to handle and show download size (Queue)
> 
> 
> Path input
> 
>     * Can't be edited. Maybe make it static text if we need it — but I'm not sure we do, since the same info is shown visually in the browser.
I made it look less editable, not sure if it looks better or not. I knowing what bundle/object i'm looking at is important to me, especially if I wanna do something outside the browser with it.
> 
> 
> Naming (nit/idea)
> 
>     * Rename "History" → "Local history"? "Queue" means remote, "local" means on the workstation, but "history" is ambiguous — history of queue submissions, or
>       history from this workstation?

See reply above

docs/design/job-bundle-browser.md
epmog: Made this moot by switching this to a local config, no limitations surfaced by changing metadata, and it loads faster like the original. Resolving

docs/design/job-bundle-browser.md
epmog: Removed these commands, punting for later

src/deadline/client/job_bundle/_repository.py
```diff
@@ -0,0 +1,1491 @@
...
+
+# ── S3 Cache ─────────────────────────────────────────────────
+
+
+def get_bundle_cache_dir() -> str:
+    """Get the root cache directory for S3 bundle archives."""
+    return os.path.join(get_cache_directory(), "job-bundles")
+
+
+def _cache_key(bucket: str, s3_key: str) -> str:
```
andychoquette, line 902: `_cache_key` never applies `sanitize_bundle_name`, so an S3 object (or local file) named `...ojd` yields cache dir <sha>/.. = the cache root. Extraction then overwrites other bundles' cached templates and the next cache miss `rmtree`s every cached bundle plus every `.visibility.json.` Writable by anyone who can share a bundle. I had an agent repro this. 

In other words, a file named `...ojd`, when previewed, gets extracted right to `~/.deadline/cache/job-bundles/`, poisoning the cache.

src/deadline/client/ui/widgets/job_bundle_settings_tab.py
```diff
@@ -76,18 +76,83 @@ def refresh_ui(self, settings: JobBundleSettings):
...
+                queue_loading=True,
+                local_source=default_dir,
+                history_source=job_history_dir,
+                parent=self,
+            )
+            # Connect the worker's result BEFORE starting it. A fast failure (e.g.
+            # not logged in) can emit ``done`` almost immediately; a cross-thread
+            # queued signal emitted before the connection exists is dropped, which
+            # would leave the browser stuck on "Loading..." forever.
+            s3_worker = _S3InitWorker()
```
andychoquette, line 136: SIGABRT kills the host app (Maya/Nuke): `_S3InitWorker` is `wait()`ed only on reject paths; picking a Local/History bundle before S3 init returns drops the last ref to a running QThread → qFatal. Same class of bug in the browser dialog, which has no closeEvent, no reject override, and no cancel flag (job_bundle_browser_dialog.py:199,810,847), so cancelling during prefetch aborts too.

andychoquette: had an agent reproduce an exit 134 here.

andychoquette (dismissed):

crowecawcaw:
Only blocking comment: suggest renaming "Open Bundle" to "Show bundle in (Explorer/Finder/desktop)". "Opening" sounds like opening the submitter for it


Non-blocking / suggestions / for consideration
---
- suggest putting path on the bottom of the window. It's a detail, not a title so much
- success window could use some policy. emoji smells like Claude
  Screenshot 2026-08-13 at 3 07 34 PM
- Bug? I have a bundle that has a non-required DataDir input. It shows as required in the local preview but option when uploaded to S3. Not sure why.
  Screenshot 2026-08-13 at 3 06 29 PM
- right click > Hide is a bit of a hidden feature. Right clicking on files in other views doesn't do anything. Maybe add a Hide button as well next to "Download bundle"?
- when queue creds are expired, there's a nice message now. Would be even better if it had a button to click to login. I had to close my window, run a command to log in, then open the browser again. 
- might be nice to have a help panel somewhere. Like if someone wonders what this is or what it means to look in the queue for bundles, we could provide a little bit of text.

Idea: we have a few buttons down here, and "Save bundle as" feels more like a local filesystem operation that a network one. I think "share" would hint more that it's to be shared with other people. What about a dropdown menu here like, "More options" -> Load bundle, Export bundle, Share bundle.
Screenshot 2026-08-13 at 3 17 48 PM

src/deadline/client/cli/_groups/bundle_group.py
```diff
@@ -556,3 +581,441 @@ def _print_response(
...
+
+    # Get file size for progress bar
+    file_size = repo.get_bundle_size(match.path)
+
+    # Download and extract are sequential inside download_full_bundle.
+    _bars: dict = {}
+
+    def _dl_callback(n):
+        if "dl" not in _bars:
+            _bars["dl"] = click.progressbar(length=file_size, label="Downloading")
```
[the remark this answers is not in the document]
epmog: no logner write to stdout when using json

epmog:
Thanks for the review!

> Only blocking comment: suggest renaming "Open Bundle" to "Show bundle in (Explorer/Finder/desktop)". "Opening" sounds like opening the submitter for it

Swapped to `Show bundle folder`. Short/sweet/easily changed in the future (and means I don't have to translate each app per operating system!)

> ## Non-blocking / suggestions / for consideration
> 
>     * suggest putting path on the bottom of the window. It's a detail, not a title so much

Might just be preferences. I definitely prefer to have that information grouped together up top.

> 
>     * success window could use some policy. emoji smells like Claude

removed!

> 
>     * Bug? I have a bundle that has a non-required DataDir input. It shows as required in the local preview but option when uploaded to S3. Not sure why.

The issue here is that including required information in the s3 metadata would take up a decent chunk of the limited space. Can punt the discussion of fail open/closed here for "required" being populated or not. I believe I had it as "unknown" or something similar before.

> 
>     * right click > Hide is a bit of a hidden feature. Right clicking on files in other views doesn't do anything. Maybe add a Hide button as well next to "Download bundle"?

I'd like to avoid adding too many buttons (we've seen this elsewhere). Might suggest a different route. Future improvement could allow multi-select and hide multiple bundles at once. Or maybe there's a toggle instead of a button. Something to think about for a future iteration of the UI.

> 
>     * when queue creds are expired, there's a nice message now. Would be even better if it had a button to click to login. I had to close my window, run a command to log in, then open the browser again.

Yea, that'd be a good addition - easy to add in a future change. A lot of the pain with not having that action available here is caused by Monitor credentials requiring the user to click "login" even though they don't actually have to login.

> 
>     * might be nice to have a help panel somewhere. Like if someone wonders what this is or what it means to look in the queue for bundles, we could provide a little bit of text.

Are you thinking of something in particular that's confusing to the user? Might be good to get feedback from additional users too
 
> 
> Idea: we have a few buttons down here, and "Save bundle as" feels more like a local filesystem operation that a network one. I think "share" would hint more that it's to be shared with other people. What about a dropdown menu here like, "More options" -> Load bundle, Export bundle, Share bundle. 

Previous feedback on a different UI change suggested not combining buttons in a similar fashion. I'm definitely down to make the UI cleaner in that regard by removing buttons. Seems like a reasonable future change.

src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py
```diff
@@ -458,86 +469,379 @@ def _on_help_button_clicked(self):
...
+        queue_error = ""
+        QApplication.setOverrideCursor(Qt.WaitCursor)
+        try:
+            queue_repo = S3BundleRepository.from_config()
+        except Exception as e:
+            queue_error = str(e)
+        finally:
+            QApplication.restoreOverrideCursor()
+
+        # Get default local directory
```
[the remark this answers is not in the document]
epmog: can do as a follow-up

src/deadline/client/ui/dialogs/job_bundle_browser_dialog.py
```diff
@@ -0,0 +1,1704 @@
...
+        self._preview_workers: set = set()
+        self._preview_request_path: Optional[str] = None
+        # Background first-level preloads (Queue source): list top-level folders'
+        # children off the UI thread, then apply on the main thread.
+        self._preload_workers: set = set()
+        self._ready = False
+
+        self._build_ui()
+        self._ready = True
+        self._populate_root()
```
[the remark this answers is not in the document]
epmog: Can address afterwards

crowecawcaw (approved):

jericht (approved):
Sat with @epmog and went through the feature together, LGTM

API breaking change is calling out removing an import for `config_file` - this is fine. It was imported and only used internally, not meant to be consumed from the `job_bundle_settings_tab` module.

PR #1184, opened 2026-06-02
feat: consume monitor session_id in telemetry for cross-component cor…

andychoquette:
## Summary

Consume the Monitor session_id from `~/.deadline/config` in CLI telemetry for cross-component session correlation.

Read the Monitor's session_id from ~/.deadline/config and include it as 'monitor_session_id' in telemetry event details. This enables correlating CLI/library telemetry events with the same IDC session used by Monitor and CTDX. Profile-scoped section is checked first, falling back to global.

Monitor and CTDX now write a shared `session_id` (derived from the IDC session token) to `~/.deadline/config` (see CR-276445103). The CLI had no way to correlate its telemetry events with the same session, making cross-component session tracking impossible.

During `TelemetryClient.initialize()`, read the `session_id` from the deadline config. Check the profile-scoped section (`[profile-{name} deadline-cloud-monitor]`) first, then fall back to the global `[deadline-cloud-monitor]` section. If found, include it as `monitor_session_id` in telemetry event common details so it appears in every outgoing RUM payload alongside the existing per-process `sessionId`.

- Telemetry events from the CLI now include a `monitor_session_id` field in event details when a Monitor session is active
- The existing per-process `sessionId` in `UserDetails` is unchanged — no loss of per-invocation granularity
- If no monitor session_id exists in config (e.g., HOST_PROVIDED credentials path), the field is simply omitted

- Have you run the unit tests? Yes — 4 new tests added covering profile-scoped lookup, global fallback, precedence, and absent config. All 64 telemetry tests pass.
- Have you run the integration tests? Tested locally with a dev build of CTDX writing session_id to config, confirmed via INFO-level logging that the CLI picks it up.

- Are relevant docstrings in the code base updated? Yes — `_get_monitor_session_id` has a docstring.
- Has the README.md been updated? No changes needed — no CLI argument or public API changes.

No. This is additive — a new field in telemetry event details. No public API or CLI contract changes.

No. The `session_id` is a one-way SHA256 hash of the IDC token (first 32 hex chars). It cannot be used to derive the original token. It is read from a file that already exists with user-only permissions.

### DIFF
--- a/src/deadline/client/api/_telemetry.py
+++ b/src/deadline/client/api/_telemetry.py
@@ -188,6 +188,11 @@ class TelemetryClient:
         if monitor_id:
             self._system_metadata["monitor_id"] = monitor_id
 
+        monitor_session_id = self._get_monitor_session_id(config=config)
+        if monitor_session_id:
+            self._common_details["monitor_session_id"] = monitor_session_id
+            logger.debug("Telemetry using monitor_session_id: %s", monitor_session_id)
+
         self._initialized = True
         self._start_threads()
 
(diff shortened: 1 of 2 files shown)


src/deadline/client/api/_telemetry.py
```diff
@@ -232,6 +237,30 @@ def _get_telemetry_identifier(self, config: Optional[ConfigParser] = None):
             config_file.set_setting("telemetry.identifier", identifier)
         return identifier
 
+    # Section name convention shared with Monitor/CTDX (see CR-276445103).
```
crowecawcaw: Nit: remove CR ID and CTDX

leon-li-inspire (dismissed):

crowecawcaw (approved):

leon-li-inspire (approved):

PR #1194, opened 2026-06-05
docs: add Python version guidance

### LINKED ISSUE #1143: Docs: formalize Python version support reasoning
### Documentation Issue

We aim to keep compatibility with older versions of Python because some people use this library with older DCCs that use older Python versions. We should document this logic along with specific info for DCCs and Python versions so that 1) we know why our version support goes back to Python 3.8 and 2) we know what we're dropping support for if we drop Python 3.8 in the future.

crowecawcaw:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/1143

It wasn't clear which Python versions we should support.

Add guidance for how to reason about version support.

Gives us a clear path forward for retiring support for old Python versions.

n/a

Yes

No

No

No
----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -25,6 +25,12 @@ Table of Contents:
       - [Private Modules](#private-modules)
       - [Public Modules](#public-modules)
       - [On `import os as _os`](#on-import-os-as-_os)
+    - [Python Version Support](#python-version-support)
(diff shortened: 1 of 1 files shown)


andychoquette (approved):
LGTM

Cherie-Chen (approved):

PR #1198, opened 2026-06-08
fix: populate host requirements from job template in gui-submit

### LINKED ISSUE #584: Bug: host requirements do not populate from job template in `deadline bundle gui-submit`
### Describe the bug

Currently, when running `deadline bundle gui-submit`, host requirements will not auto-populate. This will leave the host requirements as "run on all worker hosts" even though it will only run on certain hosts.

### Expected Behaviour

The host requirements are pre-filled from the job template.
* If there is only one step in the job template, the host requirements GUI host requirements matches the host requirements in the single step
* If there are multiple steps in the job template and all host requirements are the same, then the GUI matches the host requirements in the template
* If there are multiple steps in the job template and there are different host requirements in the template, then the host properties GUI is deactivated.

### Current Behaviour

The host requirements tab always lists "run on all worker hosts".

### Reproduction Steps

1. Create a job template with a single step and host requirements for that step
2. Run `deadline bundle gui-submit` on that job.
3. The host requirements list "run on all worker hosts"

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): Affects all OSes
2. Output of `python3 --version`: Affects all python versions
3. Output of `deadline --version`: `deadline, version 0.49.5`

crowecawcaw:
Fixes #584

When running `deadline bundle gui-submit`, the Host requirements tab always defaulted to "Run on all available worker hosts" even when the job template's steps declared hostRequirements.

Add `HostRequirements.from_dict` (the inverse of `serialize`) to parse an OpenJD step `hostRequirements` dict back into the UI dataclass, and have `show_job_bundle_submitter` resolve the requirements across the template's steps before constructing the dialog:

- single step, or multiple steps with identical hostRequirements -> pre-fill the tab from the template
- steps with differing (or partially missing) hostRequirements -> leave the tab deactivated ("Run on all available worker hosts")

Makes the GUI submitter work with host requirements.

Unit tests. Also manually opening the GUI submitter with bundles with different steps and host requirements.

n/a

No

No

No
----

### DIFF
--- a/src/deadline/client/ui/dataclasses/__init__.py
+++ b/src/deadline/client/ui/dataclasses/__init__.py
@@ -19,7 +19,7 @@ __all__ = [
 
 import os
 from dataclasses import dataclass, field
-from typing import Literal, Dict, List, Union
+from typing import Any, Literal, Dict, List, Optional, Union
 
 from ...job_bundle.parameters import JobParameter
 
@@ -347,9 +347,9 @@ class HostRequirements:
     Settings for a HostRequirementsWidget.
     """
 
-    os_requirements: OsRequirements = field(default=None)
-    hardware_requirements: HardwareRequirements = field(default=None)
-    custom_requirements: CustomRequirements = field(default=None)
+    os_requirements: Optional[OsRequirements] = field(default=None)
+    hardware_requirements: Optional[HardwareRequirements] = field(default=None)
+    custom_requirements: Optional[CustomRequirements] = field(default=None)
 
     def __post_init__(self):
         if isinstance(self.os_requirements, dict):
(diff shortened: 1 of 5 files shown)


andychoquette (approved):

leon-li-inspire (approved):

crowecawcaw:
Three types of API changes flagged:
- reexporting generic functions like `from typing import Any`. Not a real API change
- fixing type annotations that were previously incorrect e.g. `HardwareRequirements` -> `Optional[HardwareRequirements]`. Intentional
- adding the new `HostRequirements.from_dict`. Intentional.

Changes intentional. Merging.

PR #1000, opened 2026-02-13
WIP: Editable farm, queue and storage profiles in the Shared Job Settings Tab (Don't review yet)

folouiseAWS:
### What was the problem/requirement? (What/Why)
Previously, users had to navigate to the Settings dialog to select their farm, queue, and storage profile before submitting a job. This was a common source of confusion. New users didn't realize they needed to open Settings first, and experienced users found the extra step disruptive to their workflow.
### What was the solution? (How)

The Settings dialog remains available for other configuration (AWS profile, job history, etc.) but no longer contains farm/queue/storage profile dropdowns.

Farm, queue, and storage profile combo boxes with refresh buttons are now on the Shared job settings tab
Storage profile selector only appears when the selected queue has storage profiles available.

### What is the impact of this change?

Before
Screenshot 2026-02-17 at 1 55 34 PM
Screenshot 2026-02-17 at 1 55 49 PM

After
Screenshot 2026-02-17 at 1 51 01 PM
Screenshot 2026-02-17 at 1 51 16 PM

### How was this change tested?

See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#testing) for information on running tests.

- Have you run the unit tests? Yes
- Have you run the integration tests? Yes and update squish tests
- Have you made changes to the `download` or `asset_sync` modules? If so, then it is highly recommended
  that you ensure that the Docker-based unit tests pass.

### Was this change documented?

- Are relevant docstrings in the code base updated? yes
- Has the README.md been updated? If you modified CLI arguments, for instance. No

### Does this PR introduce new dependencies?

This library is designed to be integrated into third-party applications that have bespoke and customized deployment environments. Adding dependencies will increase the chance of library version conflicts and incompatabilities. Please evaluate the addition of new dependencies. See the [Dependencies](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies) section of DEVELOPMENT.md for more details.

*   [X] This PR adds one or more new dependency Python packages as dependencies . I acknowledge I have reviewed the considerations for adding dependencies in [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#dependencies). It addes required testing depending to the testing dependency
*   [ ] This PR does not add any new dependencies.

### Is this a breaking change?

Breaking changes
[...]

PR #1199, opened 2026-06-08
feat!: allow selecting farm and queue in job submission window 

crowecawcaw:
In the job submission window, users can see the farm and queue but can't change them. For that, they need to open the settings window.

Allow selecting the farm and queue in the job submission window. Also do some refactoring so the farm and queue auto-selection logic lives in a single place (was previously implemented in two locations) and expand test coverage to cover more scenarios.

The logic is, when changing profiles or farms or after signing in, select:
- the profile's default farm and queue, if set
- or the first farm and the first queue if there's only one option
- otherwise leave the selection blank

Farms and queues are easier to select.

Manually switching between farms and between profiles to verify the auto-selection logic works correctly.

n/a

No

No

No

Related: https://github.com/aws-deadline/deadline-cloud/pull/1000

----


baxeaz (commented):
A number of public symbol removals it looks like, should this be breaking?

crowecawcaw:
API changes are intentional. We're moving farm and queue selectors into the submitter window. Marking this as a breaking change to be same (`feat!`) though in practice no one should be import the particular exported being flagged.

karthikbekalp:
The changes look good to me but I wonder if this needs to be reviewed by UX SMEs first? Maybe @andychoquette  or @phil-IO-p ?

karthikbekalp (approved):
The changes look good to me but I wonder if this needs to be reviewed by UX SMEs first? Maybe @andychoquette or @phil-IO-p ?

PR #1200, opened 2026-06-08
feat!: drop python 3.8

### LINKED ISSUE #762: Maintenance: Move per-test run coverage control to .github shared workflows
### Description

In https://github.com/aws-deadline/deadline-cloud/pull/758, I worked out a way to use a lower coverage target for Python 3.8, because a feature needed Python >= 3.9 so its tests do not run there. This involved copying the file https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_python_build.yml here to https://github.com/aws-deadline/deadline-cloud/blob/mainline/.github/workflows/reusable_python_build_coverage_override.yml and adding a parameter to override the coverage option.

The result of this is that the github workflows are more complicated here, and this capability isn't available across all our repos.

### Solution

Move the ability to override the coverage fail-under back to the aws-deadline/.github repo, by making the cov-fail-under parameter optional and using conditionals in the Run Tests step in the file https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_python_build.yml. Then remove the copied workflow from deadline-cloud, and use the shared one instead.

crowecawcaw:
Closes: https://github.com/aws-deadline/deadline-cloud/issues/762

Python 3.8 is getting difficult to support. Deadline API updates are no longer published in boto3 for Python 3.8. Some core dependencies no longer provide updates for Python 3.8. Additionally, none of our target DCCs use Python 3.8 and from internal telemetry, it's extremely low usage.

See our version support guidance: https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#python-version-support

Drop Python 3.8 support.

Simpler maintenance and dependencies are up to date.

CI

Yes

No

Yes - deservers a minor version bump.

No
----


epmog (approved):
This seems reasonable to me. It's always possible to add back in if there's large demand, but from everything we've looked into this should be fine to remove. Ship it!

andychoquette (approved):

PR #1201, opened 2026-06-09
fix: consolidate auth status

### LINKED ISSUE #1152: Consolidate check_authentication_status and check_deadline_api_available
After #1151 the two helpers share a private `_list_farms_for_auth_probe()` and do the same probe with different wrappers:

| | `check_authentication_status` | `check_deadline_api_available` |
|---|---|---|
| Return type | `AwsAuthenticationStatus` enum (AUTHENTICATED / NEEDS_LOGIN / CONFIGURATION_ERROR) | `bool` |
| On failure | Splits NEEDS_LOGIN vs CONFIGURATION_ERROR via `get_credentials_source` | Logs the exception and returns False |
| Callers | `_login_deadline_cloud_monitor` poll loop | `deadline auth status` CLI, GUI status widget |

Now that the probe itself is shared, it's worth consolidating the public surface to a single auth-status helper, likely the enum-returning one, and migrating the CLI and GUI widget over. This would:
- Remove a duplicated public API.
- Give the CLI/GUI the NEEDS_LOGIN vs CONFIGURATION_ERROR signal (arguably a UX improvement — today they only see "unavailable").
- Require a decision on what to do about the exception-logging that `check_deadline_api_available` does today but `check_authentication_status` does not (the poll-loop caller deliberately avoids spam).

Scope deferred from #1151 to keep that fix minimal, since this touches the public API surface and the auth-status CLI's logging behavior.

Not urgent — tracking as tech debt.

crowecawcaw:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/1152

We consolidated the API availability and authZ checks in terms of logic, but have duplicate code paths for them.

Clean them up but don't make backwards incompatible changes. We can fully remove the redundant availability check in the future.

Cleaner code

CI

Yes

No

No

No
----


src/deadline/client/ui/deadline_authentication_status.py
```diff
@@ -46,7 +46,8 @@ class DeadlineAuthenticationStatus(QObject):
 
        status.creds_source: result of api.get_credentials_source()
        status.auth_status: result of api.check_authentication_status()
-       status.api_availability: result of api.check_deadline_api_available()
+       status.api_availability: True iff auth_status is AUTHENTICATED (derived
```
baxeaz, line 49: *if

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
```diff
@@ -92,15 +92,13 @@ class AuthenticationState(enum.Enum):
...
         AUTHENTICATED_READY: User is authenticated and has API access to list-farms
-        AUTHENTICATED_NO_API: User is authenticated but lacks API access permissions to list-farms
         NEEDS_LOGIN: User needs to log in to authenticate with a DCM profile
         CONFIGURATION_ERROR: There is a configuration issue with the AWS profile.
         UNEXPECTED_ERROR: An unknown or unexpected error occurred during authentication.
     """
 
     REFRESHING = enum.auto()
     AUTHENTICATED_READY = enum.auto()
-    AUTHENTICATED_NO_API = enum.auto()
```
baxeaz, line 103: Looks like the API checker is failing on this being removed - I think we do consider that "breaking"?

src/deadline/client/ui/deadline_authentication_status.py
crowecawcaw: "iff" is shorthand for "if and only if", somewhat of a math term

src/deadline/client/ui/widgets/deadline_authentication_status_widget.py
crowecawcaw: Will address by updating the API checker. The widget itself is a public export, but its internals like this constant are not meant to be.

waninggibbon (dismissed):

waninggibbon (approved):

baxeaz (approved):

PR #1202, opened 2026-06-09
feat!: support farms across all Deadline Cloud regions

crowecawcaw:
We want to make it easier to work with farms across multiple regions within a single profile.

Add `region` as a parameter throughout the library and CLI. Farms are now identified by `(region, farm_id)` instead of assuming a single region. 

----

### DIFF
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -169,6 +169,14 @@ def _resolve_region(
     2. The ``defaults.farm_region`` config setting, if set (non-empty).
     3. ``None`` (let boto3/the session decide, preserving single-region behavior).
 
+    Contract / region-value convention: this normalizes "no region" to ``None`` and never
+    returns ``""``. Both an explicit ``region=""`` and an empty ``defaults.farm_region`` are
+    treated as "not set" (the falsy ``if region:`` / ``if farm_region:`` checks below).
+    Callers downstream of resolution therefore test ``if region is not None:`` -- a
+    resolved region is either ``None`` (use the session/profile default) or a real region.
+    Raw, pre-resolution inputs (CLI args, config reads) may legitimately be ``""`` and use a
+    truthy ``if region:`` check instead; the two conventions are intentional, not arbitrary.
+
     Args:
         config (ConfigParser, optional): The AWS Deadline Cloud config to use.
         region (str, optional): An explicit region override.
@@ -177,8 +185,9 @@ def _resolve_region(
             differ from the default farm). When ``None``, the default farm's region is used.
 
     Returns:
-        The resolved region name, or ``None`` if nothing is configured.
+        The resolved region name, or ``None`` if nothing is configured (never ``""``).
     """
+    # Truthy (not "is not None"): an explicit region="" means "not provided", so fall through.
     if region:
         return region
 
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -26,7 +26,7 @@ from botocore.exceptions import (  # type: ignore[import]
 from botocore.session import get_session as get_botocore_session
 
 from .. import version
-from ..config import get_setting, config_file
+from ..config import get_setting, set_setting, config_file
 from ..exceptions import DeadlineOperationError
 from ...job_attachments._aws.aws_clients import get_s3_client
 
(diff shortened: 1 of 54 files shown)


src/deadline/client/cli/_groups/farm_group.py
```diff
@@ -32,27 +32,42 @@ def cli_farm():
...
     """
-    # Get a temporary config object with the standard options handled
+    # Get a temporary config object with the standard options handled.
+    # Note: --region is consumed directly here (to scope the fan-out) rather than
+    # being routed through config via _apply_cli_options_to_config.
     config = _apply_cli_options_to_config(**args)
 
     try:
-        response = api.list_farms(config=config)
+        response = api.list_farms(config=config, region=region)
```
baxeaz, line 52: Hmm will it be unexpected and potentially breaking for this behavior to change if you don't provide a region option or set it as the list of one deadline_regions?  It seems like opting **into** multi region would be a better path if we can take it?

src/deadline/client/api/_session.py
```diff
@@ -130,35 +135,104 @@ def get_default_client_config(**kwargs) -> botocore.config.Config:
...
 
     Args:
         service_name (str): The AWS service to get the client for, e.g. "deadline".
         config (ConfigParser, optional): If provided, the AWS Deadline Cloud config to use.
+        region (str, optional): The AWS region to create the client for. When omitted,
+            the region is resolved from `defaults.farm_region` (if set), otherwise the
+            session/profile region is used.
     """
 
     session = get_boto3_session(config=config)
```
baxeaz, line 242: Because we don't call this with a region it looks like every thread calling from _list_farms_for_region will potentially share the same session which isn't Thread safe - https://docs.aws.amazon.com/boto3/latest/guide/session.html#multithreading-or-multiprocessing-with-sessions

src/deadline/client/cli/_groups/farm_group.py
crowecawcaw: I think we want to support multi-region by default, if someone has farms in other regions. Someone will only experience a different output if they have farms in multiple regions in the same account. And in that case, I'd guess they'd prefer to see all farms most of the time.

I'll mark this feature as breaking because it does change behaviors in ways customers might be surprised by. The API changes aren't breaking in a low level, interface sense though - region is optional everywhere.

src/deadline/client/api/_session.py
crowecawcaw: Great catch! Addressed by creating the clients in the main threat before fanning out across regions.

src/deadline/client/config/config_file.py
```diff
@@ -38,6 +38,100 @@
...
+#
+# This hardcoded list is the default set of regions scanned during the list-farms
+# fan-out. We don't auto-discover regions at runtime (botocore's
+# get_available_regions("deadline") returns [] -- deadline is absent from its
+# endpoints.json -- and querying SSM on every CLI call isn't worth the latency).
+# Because it's hardcoded, this list may go stale as AWS Deadline Cloud launches in
+# new regions, so it MUST be updated when the service-endpoints page above changes.
+# Users can override it without waiting for a release via the DEADLINE_CLOUD_REGIONS
+# env var or the settings.deadline_regions config setting (see get_deadline_regions).
+DEADLINE_REGIONS: List[str] = [
```
leon-li-inspire, line 64: Have we updated any region build doc to update this list when there will be new regions ?

leon-li-inspire:
> We want to make it easier to work with farms across multiple regions within a single profile.

Could you elaborate the scenario for this requirement, it's not quite clear to me how this feature can help multi-region customers.

leon-li-inspire:
I notice there could be following UI changes 

  1. Farm combo box — now shows region-prefixed labels like (us-west-2) My Farm and fills incrementally as regions stream in
  2. Streaming/progressive loading — farms appear one region at a time instead of all-at-once
  3. Per-region warning tooltip — when a region fails, the combo box tooltip shows the error instead of a modal popup
  4. Config dialog — default_farm_changed now persists defaults.farm_region alongside defaults.farm_id
  5. New farms_appended signal — combo box appends items without resetting selection

Could we add some before/after screenshots in the PR description ?

crowecawcaw:
Screenshots of how this feature will appear to users.

Main difference is the region label in the farms list. 
Screenshot 2026-06-11 at 12 18 45 PM

`farm_region` added to the Deadline config file
Screenshot 2026-06-11 at 12 19 45 PM

Warning if list-farms fails to connect to one region:
Screenshot 2026-06-11 at 12 22 31 PM

leon-li-inspire (approved):

baxeaz (approved):

PR #1217, opened 2026-06-19
feat: configure HTTPS proxy and CA bundle via deadline config

### LINKED ISSUE #1046: Feature request: Configure HTTP proxy settings via deadline config
### Describe the problem

I can use HTTPS_PROXY env variable to set the proxy for deadline-cloud apps.  A few problems with this:

I have to set this in my environment which impacts all other applications. To workaround this I need to create a wrapper script that sets these

the problem with that workaround is that deadline-cloud is invoked from multiple places. not just the CLI, but also DCC integrations and from the Monitor application

Additionally this setting is not honored by deadline cloud monitor. The monitor picks up OS configuration while deadline-cloud does not. IT would be nice if these were unified

### Proposed Solution

Add proxy configuration to the deadline cloud config.

Update Deadline Cloud monitor to also honor this config

The same goes for CA certs bundles (AWS_CA_BUNDLE)

### Example Use Cases

deadline config set settings.https_proxy http://MYPROXY

crowecawcaw:
Fixes: https://github.com/aws-deadline/deadline-cloud/issues/1046

We want to support proxies better including for submitters and other software built on top of the `deadline` library.

Add proxy setting as a config option.

Makes it easier to configure proxy settings for this library.

Integ tests that run a local proxy that points to a mock endpoint and verify the library hits the mock endpoint. Also manual local testing with a local proxy.

Yes, in config settings

No

No

No

----


src/deadline/client/api/_session.py
```diff
@@ -136,32 +143,160 @@ def get_default_client_config(**kwargs) -> botocore.config.Config:
...
+):
+    """
+    Build a boto3 client applying the resolved ``(https_proxy, ca_bundle)``
+    settings: the proxy via the botocore Config ``proxies`` field and the CA
+    bundle via the client's ``verify`` kwarg. Shared by ``create_client`` and the
+    cached ``get_session_client`` so both apply the settings identically.
+    """
+    https_proxy, ca_bundle = settings_key
+    proxy_kwargs = {} if "proxies" in config_kwargs else _proxy_config_kwargs(https_proxy)
+    client_kwargs: dict = {"config": get_default_client_config(**proxy_kwargs, **config_kwargs)}
```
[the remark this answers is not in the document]
crowecawcaw: Good catch — this is a real bug, fixed in 81929d4. `_build_client` now always passes an explicit `proxies` kwarg to `get_default_client_config` (the resolved dict, or an empty dict to opt out), and `get_default_client_config` treats any explicit `proxies` key — including an empty value — as "caller is authoritative" and skips its own on-disk read. So a custom config with no proxy no longer inherits the on-disk default. Regression test: `test_get_boto3_client_custom_config_without_proxy_ignores_disk_proxy`.

src/deadline/client/api/_session.py
```diff
@@ -158,10 +226,16 @@ def get_session_client(session: boto3.Session, service_name: str, region: Option
...
+    bundle via the ``verify`` kwarg). Because this function is cached and cannot
+    accept an (unhashable) ConfigParser, both settings are read from the on-disk
+    default config; callers needing per-config resolution should use
+    ``create_client`` directly. The resulting client is cached, so a fresh
+    process / session picks up changed settings.
     """
-    if region is None:
-        return session.client(service_name, config=get_default_client_config())
-    return session.client(service_name, config=get_default_client_config(), region_name=region)
+    return create_client(session, service_name, region=region)
```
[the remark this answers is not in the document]
crowecawcaw: Addressed in 1c1360e (and hardened in 81929d4): `get_boto3_client` now resolves the proxy/CA-bundle from the caller-supplied `config` via `_client_settings_key(config)` and threads that hashable `(https_proxy, ca_bundle)` tuple into `get_session_client` as part of the lru_cache key. A custom in-memory `ConfigParser` is now honored, and clients built from different configs no longer collide in the cache. Tests: `test_get_boto3_client_honors_supplied_config`, `test_get_session_client_cache_keyed_on_settings`.

src/deadline/client/api/_job_monitoring.py
```diff
@@ -309,7 +309,7 @@ def get_session_logs(
             queue_session = get_queue_user_boto3_session(
                 deadline=deadline, config=config, farm_id=farm_id, queue_id=queue_id
             )
-            logs_client = queue_session.client("logs", config=get_default_client_config())
+            logs_client = create_client(queue_session, "logs", config=config)
```
andychoquette: would we also need to extend this client creation + proxy config setting to the [job attachments package](https://github.com/aws-deadline/deadline-cloud-job-attachments/blob/47ad19f2a9e6a5f3e760cfc03c401f15de5cb0c9/src/deadline/job_attachments/_aws/aws_clients.py#L62)? It looks like the s3 clients will ignore this and instead look to HTTPS_PROXY / AWS_CA_BUNDLE.

crowecawcaw: Definitely. I was going to wait for this one to merge, since it defines the config values, the open a similar PR over in job attachments. We'll also want to support these proxy settings in DCM to fully close the loop.

crowecawcaw: While looking into this, I found a new path forward that avoids complicating the interfaces and require separate JA changes at the cost of a not supporting configs with different proxy settings being used at the same time. Seems like a good tradeoff because proxy settings are usually set once on a workstation. Will update.

karthikbekalp (approved):

PR #1221, opened 2026-06-22
fix(security): gate preGUI bundle hooks on allow_bundle_hooks only

diane-hj:

The preGUI hook execution path in the GUI submitter (job_bundle_submitter.py) incorrectly used (allow_env_hooks OR allow_bundle_hooks) to gate bundle hook execution. This allowed untrusted bundle preGUI hooks to execute when only environment hooks were enabled (allow_environment_hooks=true, allow_bundle_hooks=false), bypassing the documented security control.

Bundle preGUI hooks are now gated correctly on allow_bundle_hooks, matching the behavior of the CLI submission path in _submit_job_bundle.py. When allow_bundle_hooks is false, loaded bundle hooks are cleared to None to prevent any execution. The unused allow_env_hooks variable was removed from this code path since preGUI hooks are only loaded from the bundle directory here (environment preGUI hooks are not implemented in this path).

Users who have allow_environment_hooks=true and allow_bundle_hooks=false will no longer have bundle preGUI hooks execute. This is the correct and expected behavior per the documentation. No change for users who have allow_bundle_hooks=true.

```
80 passed in 3.40s

====== 1 failed, 2775 passed, 20 skipped in 137.32s ======
test_cli_handle_web_url_install_open_url_uninstall_real_linux (FileNotFoundError: [Errno 2] No such file or directory: 'xdg-mime' Looks like its because my setup is different)
```
- Have you run the integration tests? No, the change is minimal and will be run once the PR is merged. 

- Are relevant docstrings in the code base updated? No docstring changes needed. This is a bugfix to match documented behavior.
- Has the README.md been updated?  No. No interface changes.

No. This change makes the code behave as documented. Users relying on the previous (buggy) behavior where allow_environment_hooks=true inadvertently allowed bundle hooks would need to explicitly set allow_bundle_hooks=true if they want bundle preGUI hooks to execute.

Yes, this is a security fix. It closes a bypass where bundle hooks could run despite allow_bundle_hooks=false.

----

### DIFF
--- a/src/deadline/client/ui/job_bundle_submitter.py
+++ b/src/deadline/client/ui/job_bundle_submitter.py
@@ -339,16 +339,17 @@ def show_job_bundle_submitter(
 
     # Run pre-GUI hooks to allow studios to pre-populate dialog fields.
     pre_gui_output: dict[str, Any] = {}
-    allow_env_hooks = _config_file.str2bool(_get_setting("settings.allow_environment_hooks"))
     allow_bundle_hooks = _config_file.str2bool(_get_setting("settings.allow_bundle_hooks"))
     hook_manager = _HookManager(input_job_bundle_dir, logger.info)
     hooks = hook_manager.load_hooks()
-    if hooks and hooks.pre_gui and not allow_bundle_hooks and not allow_env_hooks:
+
+    if hooks and hooks.pre_gui and not allow_bundle_hooks:
         logger.warning(
             "Note: Job bundle contains preGUI hooks but bundle hooks are disabled.\n"
             "Enable with: deadline config set settings.allow_bundle_hooks true"
         )
-    if (allow_env_hooks or allow_bundle_hooks) and hooks and hooks.pre_gui:
+
+    if hooks and hooks.pre_gui and allow_bundle_hooks:
         if not _config_file.str2bool(_get_setting("settings.auto_accept")):
             confirmation_msg = (
                 _generate_hooks_confirmation_message(hooks, input_job_bundle_dir)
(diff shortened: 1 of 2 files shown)


src/deadline/client/ui/job_bundle_submitter.py
```diff
@@ -339,16 +339,17 @@ def on_create_job_bundle_callback(
...
-    if hooks and hooks.pre_gui and not allow_bundle_hooks and not allow_env_hooks:
+    if hooks and hooks.pre_gui and not allow_bundle_hooks:
         logger.warning(
             "Note: Job bundle contains preGUI hooks but bundle hooks are disabled.\n"
             "Enable with: deadline config set settings.allow_bundle_hooks true"
         )
-    if (allow_env_hooks or allow_bundle_hooks) and hooks and hooks.pre_gui:
+        hooks = None
+
+    if allow_bundle_hooks and hooks and hooks.pre_gui:
```
crowecawcaw: Nit: this `if` is similar to the one above it but rearranges the conditions. Might be clearer to put them in the same order or make a temp variable for pre_gui_hooks so we don't check it twice.

diane-hj: Thank you! I refactored the if condition.

diane-hj: Tried the temp variable approach but it breaks mypy:  `_generate_hooks_confirmation_message` expects a non-optional HookConfiguration, and mypy only knows hooks isn't None if hooks is checked directly in the if condition. Therefore updated it to the same condition order. 🙏

crowecawcaw (approved):

karthikbekalp (approved):

baxeaz (approved):

PR #1233, opened 2026-06-24
fix: regionalize endpoint URL for cross-region API calls

waninggibbon:

When a profile has a non-standard endpoint override (e.g. via the [services] section in ~/.aws/config for gamma/beta environments), boto3 applies it regardless of the region_name passed to .client(). This causes SigV4 credential-scope mismatches when submitting jobs to a farm in a different region than the profile's default.

Detect the override in get_session_client and replace the session's default region in the URL with the target region so that the signing region and endpoint stay consistent.

Makes sure that cross region resource interactions work when an endpoint override is present (such as when using a DCM gamma profile).

- Ran (and added new) unit tests
- Ran integration tests
- Manually tested by submitting a job to a farm in us-east-1 using a DCM profile for a monitor based in us-west-2 with the `services` override pointing to the gamma us-west-2 endpoint.

- Added doc strings to new functions

*   This PR does not add any new dependencies.

Not a breaking change. 

No change to security posture.
----

### DIFF
--- a/src/deadline/client/api/_session.py
+++ b/src/deadline/client/api/_session.py
@@ -173,12 +173,16 @@ def _resolve_cross_region_endpoint_url(
             scoped_config = session._session.get_scoped_config()
             services_name = scoped_config.get("services")
             if services_name and isinstance(services_name, str):
-                services_defs = session._session.full_config.get("services", {})
-                service_config = services_defs.get(services_name, {}).get(service_name, {})
-                if isinstance(service_config, dict):
-                    endpoint_url = service_config.get("endpoint_url")
-        except (ProfileNotFound, KeyError):
-            # Profile may not exist or config structure may be unexpected; fall through.
+                full_config = session._session.full_config
+                if isinstance(full_config, dict):
+                    services_defs = full_config.get("services", {})
+                    if isinstance(services_defs, dict):
+                        service_config = services_defs.get(services_name, {}).get(service_name, {})
+                        if isinstance(service_config, dict):
+                            endpoint_url = service_config.get("endpoint_url")
+        except Exception:
+            # Session internals may not be accessible (e.g. profile doesn't exist,
+            # unexpected config structure, or test mocks). Fall through safely.
             pass
 
     if not endpoint_url:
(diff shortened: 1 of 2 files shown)


src/deadline/client/api/_session.py
```diff
@@ -140,6 +141,60 @@ def get_default_client_config(**kwargs) -> botocore.config.Config:
...
+            scoped_config = session._session.get_scoped_config()
+            services_name = scoped_config.get("services")
+            if services_name and isinstance(services_name, str):
+                full_config = session._session.full_config
+                if isinstance(full_config, dict):
+                    services_defs = full_config.get("services", {})
+                    if isinstance(services_defs, dict):
+                        service_config = services_defs.get(services_name, {}).get(service_name, {})
+                        if isinstance(service_config, dict):
+                            endpoint_url = service_config.get("endpoint_url")
```
crowecawcaw: Looks functional but depends on internals. An idea that just uses the public interfaces to do the same thing:

```python
def get_session_client(session, service_name, region=None):
      if region is None:
          return session.client(service_name, config=get_default_client_config())

      client = session.client(service_name, config=get_default_client_config(), region_name=region)
      resolved = client.meta.endpoint_url
      session_region = session.region_name
      # An override leaked the session's region into a cross-region endpoint.
      if region not in resolved and session_region and session_region in resolved:
          return session.client(
              service_name,
              config=get_default_client_config(),
              region_name=region,
              endpoint_url=resolved.replace(session_region, region),
          )
      return client
```

test/unit/deadline_client/api/test_api_session.py
```diff
@@ -345,6 +346,125 @@ def test_get_session_client_same_region_cached():
...
+        assert _resolve_cross_region_endpoint_url(session, "deadline", "us-east-1") is None
+
+    def test_replaces_region_in_custom_endpoint(self):
+        session = self._make_session(
+            "us-west-2",
+            services_name="deadline-gamma-us-west-2",
+            services_defs={
+                "deadline-gamma-us-west-2": {
+                    "deadline": {
+                        "endpoint_url": "https://gamma.bealine-dev.us-west-2.amazonaws.com"
```
crowecawcaw: Let's remove this endpoint

crowecawcaw (approved):

PR #1245, opened 2026-07-02
feat: add unified BaseSubmitter abstract base class

leon-li-inspire:
## Summary
Adds a unified `BaseSubmitter` abstract base class (in `deadline.client.api`), giving every DCC submitter one shared contract for producing a job bundle (job template, parameter values, asset references, queue parameters). This is the base contract that the per-DCC submitter PRs (Maya/Nuke/Blender/Houdini/…) implement, and that the AYON consumer resolves against.

## Changes
- New private module `src/deadline/client/api/_submitter.py`: the `BaseSubmitter` ABC plus `BaseSubmitterSettings` and the frozen `SubmissionContext` snapshot, and the `get_queue_parameters` helper.
- `src/deadline/client/api/__init__.py`: re-export the public surface (`BaseSubmitter`, `BaseSubmitterSettings`, `SubmissionContext`, `get_queue_parameters`).
- `get_queue_parameters` fetches the queue parameter *definitions* and resolves each `value` (from `default`, overridable via `initial_values`); raises `DeadlineOperationError` when farm/queue are unconfigured.
- `BaseSubmitter.get_asset_references` returns the typed `AssetReferences` (not a plain dict); `SubmissionContext.asset_references` is typed to match.

## Why
Today each DCC submitter has its own bespoke submission entry points, and consumers (e.g. AYON) must import each DCC submitter's internal module. A single ABC lets any host implement one interface and any consumer drive all of them uniformly.

## Public API surface
The module `_submitter.py` is private (underscore-prefixed, matching every other module in `deadline/client/api/`); the public contract is the names re-exported from `deadline.client.api`:
- `BaseSubmitter` — the ABC that DCC submitters subclass.
- `BaseSubmitterSettings` — common submission settings; DCCs subclass to add DCC-specific fields.
- `SubmissionContext` — frozen snapshot of collected submission data (settings, job_template, parameter_values, asset_references).
- `get_queue_parameters(farm_id=None, queue_id=None, initial_values=None)` — fetch + resolve queue parameter definitions; pass `initial_values` (`{parameter_name: value}`, matching the submit dialog's `initial_shared_parameter_values`) to override defaults.

## API changes
Nothing pre-existed on `mainline`, so relative to the merge-base this PR is **additive only** — the griffe breaking-change check passes (0 removals, 0 signature changes); only the additive-snapshot gate (`check-api-changes`) fires, which is expected for a PR whose purpose is to introduce new public surface and requires maintainer acknowledgment.

## Testing
- Unit tests in `test/unit/deadline_client/api/test_submitter.py` cover `get_queue_parameters` (explicit ids, `initial_values` override, configured-defaults fallback, `DeadlineOperationError` when unconfigured) and `BaseSubmitter.get_submission_context` (scene-init, arg threading, pre-fetched queue parameters, typed `AssetReferences` pass-through).
- Downstream DCC implementations built on this ABC were validated end-to-end on a live AWS Deadline Cloud farm (Maya, Nuke, Blender, Houdini) — GUI publish → BaseSubmitter → CreateJob → 10/10 SUCCEEDED renders.

Rebased on current `mainline`.


src/deadline/client/api/_submitter.py
```diff
@@ -0,0 +1,216 @@
...
+            settings = self.get_settings()
+        queue_parameters = get_queue_parameters()
+        return SubmissionContext(
+            settings=settings,
+            job_template=self.get_job_template(settings),
+            parameter_values=self.get_parameter_values(settings, queue_parameters),
+            asset_references=self.get_asset_references(settings),
+        )
+
+
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `aa4fd6c`. Corrected the `get_queue_parameters` Returns docstring: it now states that it returns the full `JobParameter` *definition* dicts (with `type`/`default`/`userInterface`/`allowedValues`/…) augmented with a resolved `value` key — i.e. DCC-submitter inputs passed to `get_parameter_values`, NOT the reduced name/value `parameterValues` that `deadline:CreateJob` accepts — and notes that the `set_*`/`append_*` helpers mutate this definition list in place. Kept the full definitions (not reduced to name/value) since the DCC submitters need the definition fields.

src/deadline/client/submitter_api.py
```diff
@@ -0,0 +1,216 @@
...
+    DCC-agnostic. Uses configured defaults if farm_id/queue_id not provided.
+
+    Args:
+        farm_id: The farm ID. If not provided, uses the default from settings.
+        queue_id: The queue ID. If not provided, uses the default from settings.
+        initial_values: Optional dict of {parameter_name: value} to override
+            default parameter values.
+
+    Returns:
+        A list of parameter definition dicts with "name" and "value" keys.
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `aa4fd6c`. `get_submission_context` now accepts `farm_id`, `queue_id`, `initial_values`, and `host_requirements` (all keyword-only), forwards `host_requirements` to `get_job_template`, and also accepts a pre-fetched `queue_parameters` list (used as-is, bypassing the fetch). The convenience path now works when the configured defaults are not the intended target.

src/deadline/client/api/_submitter.py
```diff
@@ -0,0 +1,216 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `aa4fd6c`. Added `test/unit/deadline_client/test_submitter_api.py` covering: the `set_*`/`append_*` helpers (replace-existing, append-when-absent, append-with-existing-value, append-with-empty-existing-value, append-when-absent); `get_queue_parameters` (explicit ids, `initial_values` override, fallback to configured defaults, and `DeadlineOperationError` when farm/queue are unconfigured); and `get_submission_context` (builds from scene when no settings, threads farm/queue/initial_values/host_requirements through, and uses a pre-fetched queue_parameters list).

src/deadline/client/submitter_api.py
```diff
@@ -0,0 +1,247 @@
...
+def append_rez_packages(
+    parameter_values: list[dict[str, Any]],
+    packages: str,
+) -> None:
+    """Append to existing RezPackages in parameter values list."""
+    for param in parameter_values:
+        if param.get("name") == "RezPackages":
+            existing = param.get("value", "")
+            param["value"] = f"{existing} {packages}".strip() if existing else packages
+            return
```
karthikbekalp: I believe we only provide Conda packages in our submitters, do we use Rez anywhere?

karthikbekalp (commented):
Only 1 question. Rest looks good to me.

src/deadline/client/submitter_api.py
```diff
@@ -0,0 +1,247 @@
...
+        if initial_values and param["name"] in initial_values:
+            param["value"] = initial_values[param["name"]]
+
+    return params
+
+
+# NOTE: deadline-cloud intentionally provides no discovery registry or factory
+# (no register_submitter_api / get_submitter_api). A consumer always runs inside
+# a known DCC and imports that DCC's concrete SubmitterAPI directly. See the TDD
+# "Discovery: consumer-side direct import" section for rationale.
```
crowecawcaw: Nit: does not feel relevant, or am I missing something?

src/deadline/client/submitter_api.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+@dataclass
+class SubmitterSettings:
+    """Common submission settings across all DCCs.
+
+    DCC submitters subclass this to add DCC-specific fields.
+    """
+
+    name: str = ""
```
crowecawcaw: nit: job_name?

src/deadline/client/api/_submitter.py
```diff
@@ -0,0 +1,247 @@
...
+@dataclass
+class SubmitterSettings:
+    """Common submission settings across all DCCs.
+
+    DCC submitters subclass this to add DCC-specific fields.
+    """
+
+    name: str = ""
+    description: str = ""
+    frame_list: str = ""
```
crowecawcaw, line 37: Feels like someone we should not require since it assumes a frame-based rendering workload.

src/deadline/client/api/_submitter.py
```diff
@@ -0,0 +1,247 @@
...
+    project_path: str = ""
```
crowecawcaw, line 41: Is this the input path? Or what does `project` mean here?

src/deadline/client/submitter_api.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+class SubmitterAPI(ABC):
+    """Abstract base class defining the unified interface all DCC submitters implement."""
+
+    @abstractmethod
+    def get_settings(self) -> SubmitterSettings:
+        """Create settings fully initialized from the live DCC scene.
+
+        This MUST populate frame_list, project_path, output_path,
```
crowecawcaw: Nit: nuke doesn't populate frame list. Make it optional? Also for some DCCs, an empty frame list might mean "use the frame range in the scene"

src/deadline/client/api/_submitter.py
```diff
@@ -0,0 +1,247 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
leon-li-inspire, line 1: Change the filename to `_submitter.py`, and move under `/client/api` rather than `/client`

src/deadline/client/submitter_api.py
```diff
@@ -0,0 +1,247 @@
...
+class SubmissionContext:
+    """Read-only snapshot of collected submission data."""
+
+    settings: SubmitterSettings
+    job_template: dict[str, Any]
+    parameter_values: list[dict[str, Any]]
+    asset_references: dict[str, Any]
+
+
+class SubmitterAPI(ABC):
```
leon-li-inspire: TODO: Think about the naming, e.g. "BaseSubmitter" ?

src/deadline/client/submitter_api.py
```diff
@@ -0,0 +1,247 @@
...
+    """Append to existing CondaPackages in parameter values list."""
+    for param in parameter_values:
+        if param.get("name") == "CondaPackages":
+            existing = param.get("value", "")
+            param["value"] = f"{existing} {packages}".strip() if existing else packages
+            return
+    parameter_values.append({"name": "CondaPackages", "value": packages})
+
+
+def set_rez_packages(
```
leon-li-inspire: TODO: Make them more extenable

test/unit/deadline_client/api/test_submitter.py
```diff
@@ -0,0 +1,255 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
leon-li-inspire, line 1: This test file should be moved to test/unit/deadline_client/api folder right ?

src/deadline/client/submitter_api.py
leon-li-inspire: Good catch — we only ship Conda packages, so the Rez helpers weren't pulling their weight. Removed all four per-package helpers (set/append_conda_packages, set/append_rez_packages) in `7926b0b`. Overriding any queue parameter (Conda, Rez, or custom) is now uniform via `get_queue_parameters(initial_values={name: value})`.

src/deadline/client/submitter_api.py
leon-li-inspire: Agreed, it wasn't relevant here — removed the discovery-registry NOTE in `4aa1af7`.

src/deadline/client/submitter_api.py
leon-li-inspire: Done — renamed `name` → `job_name` in `4aa1af7`.

src/deadline/client/api/_submitter.py
leon-li-inspire: Agreed — `frame_list` is optional. Documented it as such in `4aa1af7`: empty means "use the scene's frame range," and it's not required for non-frame-based workloads.

src/deadline/client/api/_submitter.py
leon-li-inspire: It's the input/project-root path (e.g. the Maya workspace root) used to anchor relative asset paths and as an input directory — distinct from the render `output_path`. Added inline docs clarifying this in `4aa1af7`.

src/deadline/client/submitter_api.py
leon-li-inspire: Made it optional — softened the `get_settings` docstring in `4aa1af7`: `frame_list` need not be populated, and an empty value means "use the scene's frame range" (e.g. Nuke doesn't set it).

src/deadline/client/api/_submitter.py
leon-li-inspire: Done in `4aa1af7` — renamed to `_submitter.py` and moved under `src/deadline/client/api/`. It's now a private module; the public surface is the re-exports from `deadline.client.api`.

src/deadline/client/submitter_api.py
leon-li-inspire: Went with `BaseSubmitter` — renamed `SubmitterAPI` → `BaseSubmitter` in `4aa1af7`.

src/deadline/client/submitter_api.py
leon-li-inspire: Rather than extend them, removed the four per-package helpers entirely in `7926b0b` — their usage was unclear and redundant. Overriding any queue parameter is now uniform via `get_queue_parameters(initial_values={name: value})` (matching the submit dialog's `initial_shared_parameter_values` shape).

test/unit/deadline_client/api/test_submitter.py
leon-li-inspire: Right — moved to `test/unit/deadline_client/api/test_submitter.py` in `00397c1`, alongside the other `api/*` module tests (`test_queue_parameters_*`, `test_api_*`).

src/deadline/client/api/_submitter.py
```diff
@@ -0,0 +1,202 @@
...
+from __future__ import annotations
+
+__all__ = [
+    "BaseSubmitter",
+    "SubmissionContext",
```
leon-li-inspire, line 7: Should `SubmissionContext` -> `BaseSubmissionContext` ?

src/deadline/client/api/_submitter.py
```diff
@@ -0,0 +1,202 @@
...
+    "SubmitterSettings",
```
leon-li-inspire: Should `SubmitterSettings` -> `BaseSubmissionContext`?

src/deadline/client/api/_submitter.py
leon-li-inspire: I'd lean toward keeping `SubmissionContext` as-is. Nothing subclasses it — across all 10 DCC submitters it's constructed and consumed directly as a frozen data snapshot (unlike `SubmitterSettings`/`BaseSubmitter`, which every DCC subclasses). A `Base` prefix usually signals "meant to be extended," so it'd be a bit misleading here. Happy to rename if you'd rather have naming symmetry — let me know and I'll sweep it (it touches all DCC PRs).

src/deadline/client/api/_submitter.py
leon-li-inspire: Assuming you mean `SubmitterSettings` -> `BaseSubmitterSettings` (the comment says `BaseSubmissionContext`). That one makes sense: every DCC subclasses it (`MayaSubmitterSettings(SubmitterSettings)`, `BlenderSubmitterSettings`, … all 10), so a `Base` prefix matches the `SubmitterAPI` -> `BaseSubmitter` rename and reads correctly. Want me to rename `SubmitterSettings` -> `BaseSubmitterSettings` here (and the DCC agents follow up in their PRs)? Confirm and I'll push it.

leon-li-inspire: Done in `4960a8d` — renamed `SubmitterSettings` -> `BaseSubmitterSettings` (matches `BaseSubmitter`; all 10 DCCs subclass it). DCC PRs will follow up with the matching subclass/import updates.

src/deadline/client/api/_submitter.py
leon-li-inspire: Left `SubmissionContext` as-is (not renamed) — nothing subclasses it; it's a frozen snapshot constructed/consumed directly, so a `Base` prefix would imply extensibility that isn't there. Renamed only `SubmitterSettings` -> `BaseSubmitterSettings` in `4960a8d`. Shout if you'd prefer symmetry and I'll rename this too.

karthikbekalp (dismissed):

src/deadline/client/api/_submitter.py
leon-li-inspire: Good call — rather than document just `project_path`, I added an attribute docstring to **every** field in both `BaseSubmitterSettings` (all 14) and `SubmissionContext` (all 4) in `e89d779`, matching the `AssetReferences` docstring style in the same package. `project_path` is now documented as the scene's root/working INPUT directory (distinct from `output_path`).

crowecawcaw (approved):

karthikbekalp (approved):

PR #1255, opened 2026-07-07
feat: add Qt-free run_pre_gui_hooks for DCC submitters

leon-li-inspire:
## What

Adds a public, Qt-free `run_pre_gui_hooks` (in a new `deadline.client.ui.pre_gui_hooks` module) that in-application (DCC) submitters — Maya, Nuke, Blender, etc. — can call to run **pre-GUI** submission hooks and pre-populate their dialog fields. Today DCC submitters cannot run pre-GUI hooks at all; this closes that gap without baking hook execution into the shared `SubmitJobToDeadlineDialog`.

## Why

Pre-GUI hooks are the intended `DEADLINE_HOOKS_DIR` mechanism for DCCs (which have no on-disk bundle at submission time), but the logic lived inline in `job_bundle_submitter.py` — coupled to `qtpy` (imported at module top) and to `JobBundleSettings`. A DCC couldn't reuse it: it holds a different settings dataclass and can't supply the inline `QMessageBox` prompt or settings-shaped metadata.

## How

- **New module `deadline/client/ui/pre_gui_hooks.py`** (no top-level `qtpy` import, so it is headless-importable):
  - `PreGuiHookContext` — dataclass bundling the hook inputs, so `run_pre_gui_hooks`' signature stays stable as pre-GUI metadata grows.
  - `run_pre_gui_hooks(context, *, confirm_callback=None)` — public, Qt-free; sources hooks from `DEADLINE_HOOKS_DIR` and (when `bundle_dir` is given) the bundle; returns merged `{name, description, parameters}`. DCC callers pass `bundle_dir=None`.
  - `qt_hook_confirmation(parent)` — the sole Qt adapter; imports Qt lazily and is passed as `confirm_callback` when the standard prompt is wanted.
- `apply_pre_gui_output` stays in `job_bundle_submitter.py` — it is `JobBundleSettings`-specific (the standalone submitter's type) and not generic across submitters.
- `show_job_bundle_submitter` refactored onto the above; **gui-submit behavior is unchanged**.

## Testing

Adds headless unit tests exercising `run_pre_gui_hooks` (env-only DCC scenario, confirm-callback accept/decline/none, scalar-override + parameter-merge across sources), `apply_pre_gui_output` (routing + CLI precedence), and `PreGuiHookContext` defaults. Existing gui-submit behavior tests are preserved as regression coverage.

## Follow-ups

DCC adoption lands in each DCC repo separately (Maya first — separate PR). Each bumps its `deadline` dependency floor to the release that ships `run_pre_gui_hooks`.

### DIFF
--- a/src/deadline/client/ui/pre_gui_hooks.py
+++ b/src/deadline/client/ui/pre_gui_hooks.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 import os
 from dataclasses import dataclass, field
 from logging import getLogger
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, Optional
 
 from ..config import config_file as _config_file
 from ..config.config_file import get_setting as _get_setting
@@ -30,6 +30,13 @@ from ..job_bundle._hooks import (
 from ._utils import tr
 
 logger = getLogger(__name__)
+
+__all__ = [
+    "PreGuiHookContext",
+    "apply_pre_gui_output",
+    "qt_hook_confirmation",
+    "run_pre_gui_hooks",
+]
 
 
 @dataclass
@@ -66,7 +73,7 @@ class PreGuiHookContext:
 def run_pre_gui_hooks(
     context: PreGuiHookContext,
     *,
-    confirm_callback: Optional[Callable[[List["_HookManager"]], bool]] = None,
+    confirm_callback: Optional[Callable[[list], bool]] = None,
 ) -> dict[str, Any]:
     """Run all allowed pre-GUI hooks and return their merged output.
 
@@ -164,7 +171,7 @@ def run_pre_gui_hooks(
     return merged
 
 
-def qt_hook_confirmation(parent: Any) -> Callable[[List["_HookManager"]], bool]:
+def qt_hook_confirmation(parent: Any) -> Callable[[list], bool]:
     """Return a ``confirm_callback`` for :func:`run_pre_gui_hooks` that shows the standard
     Qt confirmation dialog listing every hook that will run.
 
@@ -175,7 +182,7 @@ def qt_hook_confirmation(parent: Any) -> Callable[[List["_HookManager"]], bool]:
     """
     from qtpy.QtWidgets import QMessageBox  # pylint: disable=import-error
 
-    def _confirm(sources: List["_HookManager"]) -> bool:
+    def _confirm(sources: list) -> bool:
         confirmation_msg = (
             "".join(
                 _generate_hooks_confirmation_message(m.hooks, m._original_bundle_dir)
--- a/src/deadline/client/ui/pre_gui_hooks.py
+++ b/src/deadline/client/ui/pre_gui_hooks.py
@@ -194,3 +194,56 @@ def qt_hook_confirmation(parent: Any) -> Callable[[List["_HookManager"]], bool]:
         return reply == QMessageBox.Yes
 
     return _confirm
+
+
+def apply_pre_gui_output(
+    pre_gui_output: dict[str, Any],
+    initial_settings: Any,
+    initial_shared_parameter_values: dict[str, Any],
+    cli_provided_param_names: Optional[set[str]] = None,
+) -> None:
+    """Apply merged pre-GUI hook output onto a submitter's settings and its shared values.
+
+    ``name`` / ``description`` overwrite the corresponding fields on ``initial_settings``.
+    For ``parameters``: any name that matches a job-template parameter (an entry in
+    ``initial_settings.parameters``) updates that entry in place; every other name lands in
+    ``initial_shared_parameter_values`` (queue parameters, ``deadline:`` job properties, etc.).
+    CLI-supplied parameter names (``cli_provided_param_names``) always win over hook values.
+
+    This is generic across submitters:
+
+    * The standalone job-bundle submitter passes a ``JobBundleSettings`` whose ``parameters``
+      is a list of template-parameter dicts, so template params are routed onto it in place.
+    * DCC submitters (Maya, Nuke, etc.) pass their own settings dataclass, which has no
+      ``parameters`` list. With no template-parameter list, every hook parameter is treated
+      as a shared value — matching how a DCC seeds ``initial_shared_parameter_values``.
+
+    ``initial_settings`` only needs assignable ``name`` / ``description`` attributes and,
+    optionally, a ``parameters`` list of ``{"name": ..., "value": ...}`` dicts.
+    """
+    if not pre_gui_output:
+        return
+
+    cli_provided_param_names = cli_provided_param_names or set()
+    hook_params = pre_gui_output.get("parameters", {})
+    # DCC settings have no template-parameter list; treat their absence as "no template
+    # params", so every hook parameter flows to the shared values dict below.
+    template_parameters = getattr(initial_settings, "parameters", None) or []
+    template_param_names = {p["name"] for p in template_parameters}
+    for param_name, param_value in hook_params.items():
+        # CLI --parameter values take precedence over hook values.
+        if param_name in cli_provided_param_names:
+            continue
+        if param_name in template_param_names:
+            # Job template parameter — update initial_settings.parameters in-place
+            for p in template_parameters:
+                if p["name"] == param_name:
+                    p["value"] = param_value
+                    break
+        else:
+            # Shared job property (deadline: keys, queue parameters, etc.)
+            initial_shared_parameter_values[param_name] = param_value
+    if "name" in pre_gui_output:
+        initial_settings.name = pre_gui_output["name"]
+    if "description" in pre_gui_output:
+        initial_settings.description = pre_gui_output["description"]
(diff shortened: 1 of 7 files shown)


src/deadline/client/ui/pre_gui_hooks.py
```diff
@@ -0,0 +1,196 @@
...
+            prompting — callers that want the standard prompt pass ``qt_hook_confirmation``.
+
+    Returns:
+        The merged pre-GUI output (any of ``name``, ``description``, ``parameters``), or an
+        empty dict if no pre-GUI hooks ran.
+    """
+    # Source selection (which bundle/env HookManagers have runnable preGUI hooks) is
+    # Qt-free logic in the hooks package so it can be unit-tested without a GUI.
+    sources = _collect_pre_gui_hook_sources(
+        bundle_dir=context.bundle_dir or "",
```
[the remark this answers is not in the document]
leon-li-inspire: Good catch — confirmed and fixed in 0d546a5. `collect_pre_gui_hook_sources` now skips the bundle source entirely when `bundle_dir` is falsy, so a DCC submission (which passes `bundle_dir=""`) never resolves a hooks file relative to the process CWD. Added regression tests: one for the env-only DCC path, and one asserting a `hooks.yaml` placed in the CWD is not loaded when `bundle_dir=""` even with `allow_bundle_hooks` enabled.

src/deadline/client/ui/pre_gui_hooks.py
```diff
@@ -0,0 +1,249 @@
...
+from ..config import config_file as _config_file
+from ..config.config_file import get_setting as _get_setting
+from ..exceptions import DeadlineOperationCanceled as _DeadlineOperationCanceled
+from ..job_bundle._hooks import (
+    HookManager as _HookManager,
+    HookMetadata as _HookMetadata,
+    _generate_hooks_confirmation_message,
+    collect_pre_gui_hook_sources as _collect_pre_gui_hook_sources,
+)
+from ._utils import tr
```
crowecawcaw, line 30: The API change detector is failing/flagging changes which is expected. But we don't need to make all the imports exports as well. Can you add an __all__ section that flags which interfaces are expected to be public?

API change detector lists these:
```
🆕 ADDED APIs (19):
  + deadline.client.ui.job_bundle_submitter.PreGuiHookContext (alias)
  + deadline.client.ui.job_bundle_submitter.apply_pre_gui_output (alias)
  + deadline.client.ui.job_bundle_submitter.qt_hook_confirmation (alias)
  + deadline.client.ui.job_bundle_submitter.run_pre_gui_hooks (alias)
  + deadline.client.ui.pre_gui_hooks (module)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext (class)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.bundle_dir (attribute)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.farm_id (attribute)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.job_name (attribute)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.parameters (attribute)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.priority (attribute)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.queue_id (attribute)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.storage_profile_id (attribute)
  + deadline.client.ui.pre_gui_hooks.PreGuiHookContext.submitter_name (attribute)
  + deadline.client.ui.pre_gui_hooks.apply_pre_gui_output (function)
  + deadline.client.ui.pre_gui_hooks.logger (attribute)
  + deadline.client.ui.pre_gui_hooks.qt_hook_confirmation (function)
  + deadline.client.ui.pre_gui_hooks.run_pre_gui_hooks (function)
  + deadline.client.ui.pre_gui_hooks.tr (alias)

```

src/deadline/client/ui/pre_gui_hooks.py
```diff
@@ -0,0 +1,249 @@
...
+        submitter_name: The submitter identity passed to hooks (e.g. "maya", "nuke").
+        priority: Initial priority passed to hooks.
+        farm_id / queue_id / storage_profile_id: Passed to hooks; when ``None`` the
+            farm/queue default from config (and the storage-profile setting) is used.
+    """
+
+    bundle_dir: Optional[str] = None
+    job_name: str = ""
+    parameters: Dict[str, Any] = field(default_factory=dict)
+    submitter_name: str = "JobBundle"
```
baxeaz, line 66: Should this be required or is it intended for this to exist as some sort of default (or test) case?

src/deadline/client/ui/pre_gui_hooks.py
```diff
@@ -0,0 +1,249 @@
...
+    priority: int = 50
+    farm_id: Optional[str] = None
+    queue_id: Optional[str] = None
+    storage_profile_id: Optional[str] = None
+
+
+def run_pre_gui_hooks(
+    context: PreGuiHookContext,
+    *,
+    confirm_callback: Optional[Callable[[List["_HookManager"]], bool]] = None,
```
baxeaz: A bit odd that we've got the internal _HookManager in the public API

crowecawcaw (changes requested):

src/deadline/client/ui/pre_gui_hooks.py
leon-li-inspire: Done in 5876bc5. Added `__all__` to `pre_gui_hooks` (`PreGuiHookContext`, `run_pre_gui_hooks`, `apply_pre_gui_output`, `qt_hook_confirmation`) and to `job_bundle_submitter` (`show_job_bundle_submitter`). The snapshot generator treats `__all__` as authoritative for named members, so `logger`, `tr`, and the re-export aliases are no longer reported as public API — only the intended interfaces remain.

src/deadline/client/ui/pre_gui_hooks.py
leon-li-inspire: Intended as defaults, not required. A bare `PreGuiHookContext()` is the common DCC case: env-only sourcing with the default "JobBundle" submitter identity. Keeping every field optional is also what lets us add new pre-GUI inputs later as new fields (with defaults) without breaking existing call sites — the reason the inputs are bundled into this dataclass rather than passed as positional args to `run_pre_gui_hooks`. Callers override the fields they care about (e.g. Maya passes `submitter_name="maya"`, `job_name=...`).

src/deadline/client/ui/pre_gui_hooks.py
leon-li-inspire: Agreed — fixed in 5876bc5. Changed the annotation to `Callable[[list], bool]` for both `run_pre_gui_hooks`'s `confirm_callback` param and `qt_hook_confirmation`'s return, so the internal `_HookManager` no longer appears in the public signature. The callback just receives the list of hook sources to display/confirm and doesn't need the element type exposed.

src/deadline/client/ui/pre_gui_hooks.py
leon-li-inspire: One thing to flag from adding `__all__` to `job_bundle_submitter`: the API detector now also reports ~16 **removed** aliases from that module (e.g. `SubmitJobToDeadlineDialog`, `JobBundleSettings`, `read_yaml_or_json`, `merge_queue_job_parameters`, `logger`, …). These were never intended as public API — they were incidental re-exports of this module's own imports, and the real ones have proper homes (`SubmitJobToDeadlineDialog` in `.dialogs.submit_job_to_deadline_dialog`, `JobBundleSettings` in `.ui.dataclasses`). griffe did **not** flag any of these as *breaking* (only the additive `Fail if API changes` gate tripped, same as before). So this narrows the module to its one intended public symbol (`show_job_bundle_submitter`).

Flagging in case you'd consider any of those accidental paths as de-facto public — if so I can keep them in `__all__`. Otherwise this is the intended cleanup.

src/deadline/client/ui/pre_gui_hooks.py
```diff
@@ -66,7 +73,7 @@ class PreGuiHookContext:
 def run_pre_gui_hooks(
     context: PreGuiHookContext,
     *,
-    confirm_callback: Optional[Callable[[List["_HookManager"]], bool]] = None,
+    confirm_callback: Optional[Callable[[list], bool]] = None,
```
crowecawcaw: This solves the usage of an internal type in code, but seems like the caller will still need that type if they're using types. Should HooksManager be public? Or can we import it from a public place? Is there another constraint that's making this difficult?

baxeaz (dismissed):

src/deadline/client/ui/pre_gui_hooks.py
leon-li-inspire: Fixed in 6923583 by making `HookManager` public. Added a public re-export module `deadline.client.job_bundle.hooks` (`from ._hooks import HookManager; __all__ = ["HookManager"]`) and typed `confirm_callback` / `qt_hook_confirmation` as `Callable[[List[HookManager]], bool]` against it. So the public signature now references a public type, and a DCC author writing a typed callback can `from deadline.client.job_bundle.hooks import HookManager` instead of reaching into the private `_hooks` package. (This reverts the earlier bare-`list` annotation, which had hidden the real runtime type rather than fixing the leak.)

crowecawcaw (approved):

PR #1261, opened 2026-07-09
fix: run environment and bundle submission hooks together

leon-li-inspire:

On the CLI / non-GUI submit path (`create_job_from_job_bundle`), pre/post-submission hooks defined in the **environment** (`DEADLINE_HOOKS_DIR`) and in the **job bundle** (`hooks.yaml`) could not run together.

The code collected both sources into `all_hooks_sources` (environment first, bundle second) and then ran a merge loop that folded them into a single `HookManager`. When both sources had hooks:

- The environment iteration set `hooks = env_hooks` and `hook_manager = env_hook_manager`.
- The bundle iteration took the `if hooks_dir == job_bundle_dir` branch and **overwrote** `hooks = bundle_has_hooks`, but left `hook_manager` still pointing at the environment manager. The `else:` merge branch was therefore unreachable dead code.

Net effect: only the **environment** manager's hooks actually executed, while the confirmation prompt listed the **bundle** hooks — an inconsistent prompt, and one source silently dropped. Environment and bundle submission hooks never ran together.

The preGUI path already had the correct design (`collect_pre_gui_hook_sources`, which returns a *list* of managers); the pre/post-submission path never got the equivalent.

Mirror the preGUI design so all hook phases share one mechanism:

- Add `collect_submission_hook_sources()`, which returns a **list** of `HookManager`s in execution order (environment first, then bundle), deduped when `DEADLINE_HOOKS_DIR` resolves to the bundle directory, with correct "present-but-disabled" / invalid-directory warnings phrased for the submission phase.
- Factor the shared source-selection logic into a private `_collect_hook_sources()` used by both the preGUI and submission collectors.
- The submit flow now shows a single confirmation across all sources and executes pre-/post-submission hooks by iterating every source in order. The pre-submission payload is threaded through the sources so a later source can override an earlier one.

Environment and bundle submission hooks now both run (environment first, then bundle), consistent with the preGUI behavior. The confirmation prompt now lists exactly the hooks that will run. No public API signature changes; behavior only changes for the previously-broken both-sources case.

- Added regression tests that a `DEADLINE_HOOKS_DIR` hook and a bundle hook **both** run, in order, for pre- and post-submission (`TestEnvAndBundleSubmissionHooks`), plus single-source and disabled-source cases.
- Added source-selection unit tests for `collect_submission_hook_sources` (`TestCollectSubmissionHookSources`), mirroring the existing preGUI coverage.
- Verified the new tests fail against the pre-fix behavior and pass with the fix.
- Drove `create_job_from_job_bundle` end-to-end with both sources enabled and confirmed all four hooks executed in env→bundle order for both phases.
- `ruff check`, `ruff format`, and `mypy` clean on the changed files.

- [x] Have you run the unit tests? Yes (hook + submission suites pass).
- [ ] Have you run the integration tests? Not run (no AWS resources in this environment).

- Docstrings for the new/refactored helpers are added and the existing preGUI docstrings reference them. No README/CLI-argument changes.

No. No public contract changes; this only fixes the both-sources execution path.

No new files/directories or permission changes; hook execution semantics are unchanged (same gating settings `allow_bundle_hooks` / `allow_environment_hooks`).

----

### DIFF
--- a/src/deadline/client/api/_submit_job_bundle.py
+++ b/src/deadline/client/api/_submit_job_bundle.py
@@ -36,7 +36,11 @@ from ..job_bundle.loader import (
     parse_yaml_or_json_content,
     validate_directory_symlink_containment,
 )
-from ..job_bundle._hooks import HookManager, HookMetadata, _generate_hooks_confirmation_message
+from ..job_bundle._hooks import (
+    HookMetadata,
+    _generate_hooks_confirmation_message,
+    collect_submission_hook_sources,
+)
 from ..job_bundle.parameters import (
     apply_job_parameters,
     merge_queue_job_parameters,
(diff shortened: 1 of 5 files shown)


test/unit/deadline_client/job_bundle/test_hooks.py
```diff
@@ -1535,6 +1536,77 @@ def test_stdout_absolute_path_parameter_is_respected(self, fresh_deadline_config
...
+            )
+        with open(os.path.join(directory, "hooks.yaml"), "w", encoding="utf8") as f:
+            yaml.dump(
+                {
+                    "version": "1.0",
+                    "preSubmission": [{"command": sys.executable, "args": [script_name, "pre"]}],
+                    "postSubmission": [{"command": sys.executable, "args": [script_name, "post"]}],
+                },
+                f,
+            )
```
crowecawcaw, line 1900: Nit: might be easier to read and audit the tests if these were stored as real job bundles instead of inline strings. Maybe a future improvement if you touch hooks again

crowecawcaw (dismissed):

src/deadline/client/api/_submit_job_bundle.py
```diff
@@ -501,81 +505,47 @@ def create_job_from_job_bundle(
...
-    if bundle_has_hooks and (bundle_has_hooks.pre_submission or bundle_has_hooks.post_submission):
-        if allow_bundle_hooks:
-            all_hooks_sources.append(("bundle", job_bundle_dir))
-        else:
+    # Show confirmation (once) if any hooks will run, listing every source's hooks.
+    if hook_sources and not config_file.str2bool(
+        get_setting("settings.auto_accept", config=config)
+    ):
+        hooks_message = "".join(
+            _generate_hooks_confirmation_message(manager.hooks, manager._original_bundle_dir)
```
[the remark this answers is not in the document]
leon-li-inspire: Good catch — fixed in 5c4e8f3. Each `HookManager` now carries a `source_label` ("job bundle" vs "environment (DEADLINE_HOOKS_DIR)"), set by the collector, and `_generate_hooks_confirmation_message` takes that label so both the header and the directory line identify the true origin. The prompt now renders, e.g.:

```
This environment (DEADLINE_HOOKS_DIR) contains submission hooks that will execute on your machine:
  Pre-submission hooks:
    [1] python3 x.py
  Location: /studio/hooks
This job bundle contains submission hooks that will execute on your machine:
  ...
  Location: /path/to/bundle
```

so a DEADLINE_HOOKS_DIR source is no longer shown as if it came from the job bundle. Both the CLI and GUI confirmation paths pass the label; added unit coverage in `test_confirmation_message_labels_environment_source` and the collector ordering test.

test/unit/deadline_client/job_bundle/test_hooks.py
leon-li-inspire: Agreed — that's a nice readability improvement. I'd prefer to keep this PR focused on the fix, so I've left the inline strings as-is for now and noted it as a follow-up for the next time hooks are touched. Happy to file a tracking issue if you'd like.

crowecawcaw (approved):

PR #1313, opened 2026-07-27
feat: record farm-failed tasks as farm_failed in status file

gianelli-lagos-aws:
> **Note:** This PR is stacked on top of [#PR2](https://github.com/aws-deadline/deadline-cloud/pull/1223) and [#PR3](https://github.com/aws-deadline/deadline-cloud/pull/1258). Please review/merge those first — the base branch here is `download-status-per-task`, so only the last commit belongs to this PR.

The download status file recorded a task as `failed` only when a *download* errored, and left tasks that **failed on the farm** (the render itself failed, producing no output) entirely absent from the file. With no stable file entry, the Deadline Cloud Monitor had to infer those failures from live run-status data — which flips back to a non-failed state when a task is requeued, so the failure signal was lost. A farm failure and a download failure also have different causes and fixes, but the artist couldn't tell them apart.

Detect `FAILED` `taskRun` session actions while retrieving sessions and record their task IDs as a distinct `farm_failed` status in the status file:

- `_retrieve_session_actions_for_session` now also collects the task IDs of `FAILED` taskRun actions into an output set. These are never added to the manifest-download list (they have no output) — only recorded for reporting.
- `_get_job_sessions` returns `(job_sessions, farm_failed_task_ids)`; the per-thread collection merges under a lock since it mutates a shared dict/set across worker threads.
- After per-task download results are built, farm-failed tasks are written as `farm_failed` entries (`0` files, no error code), skipping any task that already has a real download result this run.
- `_build_status_file_content` honors an explicit `download_status` (e.g. `farm_failed`) and otherwise derives it from the download error, as before.

`farm_failed` is deliberately distinct from `failed` (a download error) and from a succeeded-but-no-output task (which stays absent from the dict).

Additive, backwards-compatible change to the status file within schema v1: adds a new `download_status` value `farm_failed` on per-task entries. The sole consumer (Deadline Cloud Monitor) already reads this value under `schema_version === 1`. No CLI or Python API signature changes to public contracts. Stacked on top of the per-task PR (`download-status-per-task`).

Yes. Added unit tests in `test_cli_download_status_file.py` covering farm-failed recording, the distinction from download failures, and the skip-if-real-result-exists path. `hatch run test`, `hatch run lint`, and `hatch run fmt` all pass.

Docstrings for `_retrieve_session_actions_for_session` and `_get_job_sessions` describe the new farm-failed output.

Not needed; no CLI argument changes.

No.

No. It adds a new additive `download_status` value within the existing schema v1; no public contract is modified in a backwards-incompatible way.

No — it reads existing session-action metadata and writes to the same status file created by the existing feature; no new files, directories, or permission surfaces.

----

### DIFF
--- a/src/deadline/client/cli/_download_status_file.py
+++ b/src/deadline/client/cli/_download_status_file.py
@@ -240,6 +240,7 @@ def _build_status_file_content(
     existing_jobs: Optional[dict[str, Any]] = None,
     job_download_results: Optional[dict[str, dict[str, Any]]] = None,
     task_download_results: Optional[dict[str, dict[str, dict[str, Any]]]] = None,
+    succeeded_task_ids: Optional[dict[str, set[str]]] = None,
 ) -> dict[str, Any]:
     """
     Builds the full status file JSON structure by merging existing job entries
@@ -274,6 +275,18 @@ def _build_status_file_content(
         existing_tasks: dict[str, Any] = (existing_entry or {}).get("tasks", {})
         new_tasks: dict[str, Any] = (task_download_results or {}).get(job_id, {})
         merged_tasks = dict(existing_tasks)
+        # A task that SUCCEEDED this run with no output must clear any stale farm_failed
+        # entry carried forward from a prior run. The farm FAILED action is reported by the
+        # API indefinitely, so _get_job_sessions subtracts succeeded IDs from
+        # farm_failed_task_ids — preventing a new entry — but the prior disk entry is
+        # preserved by the merge above unless we explicitly remove it here.
+        job_succeeded_ids: set[str] = (succeeded_task_ids or {}).get(job_id, set())
+        for task_id in job_succeeded_ids:
+            if (
+                task_id not in new_tasks
+                and merged_tasks.get(task_id, {}).get("download_status") == "farm_failed"
+            ):
+                del merged_tasks[task_id]
         for task_id, r in new_tasks.items():
             # Honor an explicit download_status (e.g. "farm_failed" for a task that
             # failed on the farm); otherwise derive it from whether the download errored.
@@ -428,6 +441,7 @@ def write_download_status_file(
     checkpoint_dir: str,
     job_download_results: Optional[dict[str, dict[str, Any]]] = None,
     task_download_results: Optional[dict[str, dict[str, dict[str, Any]]]] = None,
+    succeeded_task_ids: Optional[dict[str, set[str]]] = None,
     print_function_callback: Callable[[Any], None] = lambda msg: None,
 ) -> None:
     """
@@ -466,6 +480,7 @@ def write_download_status_file(
                     existing_jobs=existing_jobs,
                     job_download_results=job_download_results,
                     task_download_results=task_download_results,
+                    succeeded_task_ids=succeeded_task_ids,
                 )
 
                 _atomic_write_json(status_file_path, status_content)
--- a/src/deadline/client/cli/_download_status_file.py
+++ b/src/deadline/client/cli/_download_status_file.py
@@ -123,6 +123,44 @@ def _make_status_entry(
     }
 
 
+# Job statuses that assert the job is finished and nothing failed to download. An errored task
+# contradicts any of them, so these are the states _reconcile_entry_with_tasks overrides.
+# "in_progress" is excluded on purpose: it claims nothing, and the download may still be retried
+# before the job ends, so the honest report is a failed task row under a still-running job.
+_NO_DOWNLOAD_FAILURE_STATUSES = ("downloaded", "skipped")
+
+
+def _reconcile_entry_with_tasks(entry: dict[str, Any]) -> None:
+    """Forces a job entry back to "failed" while any of its tasks still carries a download error.
+
+    A job's category-derived status only says the job finished on the farm, which is not
+    evidence that its outputs reached disk. Without this, a job whose task failed in an
+    earlier run and which produced no new results this run flips to "downloaded" while its
+    task entry still reads "failed" — a green job badge over a red task, and a real failure
+    silently dropped. The same applies to "skipped": a job stopped mid-download reports
+    "skipped" once it goes inactive, which reads as "nothing to fetch" rather than "a file is
+    missing and here is why". Preferring the task evidence keeps the two levels consistent
+    without claiming a missing file is present.
+
+    Tasks that failed on the farm carry no error_code and are deliberately excluded: the
+    render failed, the download didn't, so they must not drag the job to a download failure.
+    """
+    if entry.get("download_status") not in _NO_DOWNLOAD_FAILURE_STATUSES:
+        return
+    errored = [t for t in entry.get("tasks", {}).values() if t.get("error_code")]
+    if not errored:
+        return
+    entry["download_status"] = "failed"
+    entry["error_code"] = errored[0]["error_code"]
+    entry["error_message"] = errored[0].get("error_message")
+    # A download failure is not a skip. Leaving skip_reason set would produce an entry that
+    # claims both, and the monitor keys its "why is this file missing" copy off skip_reason.
+    entry["skip_reason"] = None
+    missing = sum(t.get("total_files", 0) - t.get("downloaded_files", 0) for t in errored)
+    if missing > 0:
+        entry["failed_files"] = missing
+
+
 def _is_job_fully_complete(job: dict[str, Any]) -> bool:
     """Returns True if the job has ended with no active tasks remaining.
 
@@ -276,7 +314,9 @@ def _build_status_file_content(
                     existing_entry["error_message"] = None
                     existing_entry["failed_files"] = 0
             existing_entry["tasks"] = merged_tasks
+            _reconcile_entry_with_tasks(existing_entry)
         else:
+            _reconcile_entry_with_tasks(new_entry)
             jobs_status[job_id] = new_entry
 
     # Update inactive jobs with non-terminal status to a terminal state
@@ -290,6 +330,7 @@ def _build_status_file_content(
                 # No downloads at all — mark as skipped (job stopped before any output)
                 existing_entry["download_status"] = "skipped"
             existing_entry["last_updated"] = now
+            _reconcile_entry_with_tasks(existing_entry)
 
     # Determine run status — "failed" if any job in this run had errors
     has_failures = any(
--- a/src/deadline/client/cli/_download_status_file.py
+++ b/src/deadline/client/cli/_download_status_file.py
@@ -281,13 +281,15 @@ def _build_status_file_content(
                 "download_status",
                 "downloaded" if r.get("error_code") is None else "failed",
             )
-            # A prior-run success is terminal: its output is on disk. A FAILED taskRun from an
-            # earlier attempt keeps being reported by the API on every no-op run, so without this
-            # guard a task that was requeued, succeeded, and downloaded would flip back to
-            # farm_failed forever. Never let farm_failed clobber an already-downloaded task.
-            if status == "farm_failed" and existing_tasks.get(task_id, {}).get(
-                "download_status"
-            ) in ("downloaded", "failed"):
+            # A prior-run download success is terminal: the output is on disk. Never let a stale
+            # farm_failed action (the API reports it indefinitely) overwrite it.
+            # "failed" is intentionally not guarded here: a task may have succeeded on the farm
+            # but its download failed, so a subsequent run re-collects the FAILED action as
+            # farm_failed and should overwrite the stale download error.
+            if (
+                status == "farm_failed"
+                and existing_tasks.get(task_id, {}).get("download_status") == "downloaded"
+            ):
                 continue
             merged_tasks[task_id] = {
                 "download_status": status,
--- a/src/deadline/client/cli/_download_status_file.py
+++ b/src/deadline/client/cli/_download_status_file.py
@@ -235,19 +235,29 @@ def _build_status_file_content(
         # Merge task entries: preserve existing tasks, overwrite only tasks seen this run
         existing_tasks: dict[str, Any] = (existing_entry or {}).get("tasks", {})
         new_tasks: dict[str, Any] = (task_download_results or {}).get(job_id, {})
-        merged_tasks = {
-            **existing_tasks,
-            **{
-                task_id: {
-                    "download_status": "downloaded" if r.get("error_code") is None else "failed",
-                    "total_files": r["total_files"],
-                    "downloaded_files": r["downloaded_files"],
-                    "error_code": r.get("error_code"),
-                    "error_message": r.get("error_message"),
-                }
-                for task_id, r in new_tasks.items()
-            },
-        }
+        merged_tasks = dict(existing_tasks)
+        for task_id, r in new_tasks.items():
+            # Honor an explicit download_status (e.g. "farm_failed" for a task that
+            # failed on the farm); otherwise derive it from whether the download errored.
+            status = r.get(
+                "download_status",
+                "downloaded" if r.get("error_code") is None else "failed",
+            )
+            # A prior-run success is terminal: its output is on disk. A FAILED taskRun from an
+            # earlier attempt keeps being reported by the API on every no-op run, so without this
+            # guard a task that was requeued, succeeded, and downloaded would flip back to
+            # farm_failed forever. Never let farm_failed clobber an already-downloaded task.
+            if status == "farm_failed" and existing_tasks.get(task_id, {}).get(
+                "download_status"
+            ) in ("downloaded", "failed"):
+                continue
+            merged_tasks[task_id] = {
+                "download_status": status,
+                "total_files": r["total_files"],
+                "downloaded_files": r["downloaded_files"],
+                "error_code": r.get("error_code"),
+                "error_message": r.get("error_message"),
+            }
         new_entry["tasks"] = merged_tasks
 
         if (
--- a/src/deadline/client/cli/_incremental_download.py
+++ b/src/deadline/client/cli/_incremental_download.py
@@ -630,6 +630,7 @@ def _retrieve_session_actions_for_session(
     queue_id: str,
     job_id: str,
     output_session: dict[str, Any],
+    output_farm_failed_task_ids: Optional[set[str]] = None,
 ):
     """
     Args:
@@ -641,6 +642,8 @@ def _retrieve_session_actions_for_session(
         queue_id: The queue id for the operation.
         job_id: The job id to process.
         output_session: The session to populate with a sessionActions field.
+        output_farm_failed_task_ids: Optional set populated with task IDs of taskRun actions that
+            FAILED on the farm. These produce no output to download and are recorded separately.
     """
     session_actions_paginator = deadline_client.get_paginator("list_session_actions")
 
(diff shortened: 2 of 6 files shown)


src/deadline/client/cli/_incremental_download.py
```diff
@@ -1174,15 +1364,136 @@ def _incremental_output_download(
...
+        path_mapping_rule_appliers,
     )
+    # Correlate manifests_to_download with downloaded_manifests by position to attribute each
+    # downloaded manifest to its job (and task). Both lists come from _get_manifests_to_download
+    # with identical inputs, so their lengths match in practice. If they ever diverge we skip only
+    # the per-job/per-task attribution — the downloads still proceed via the fallback bucket below,
+    # so a mismatch degrades reporting but never causes a silent zero-download run. (Downloads are
+    # decoupled from attribution: what gets downloaded comes from downloaded_manifests either way;
+    # attribution is best-effort layered on top.)
+    skip_attribution = len(manifests_to_download) != len(downloaded_manifests)
```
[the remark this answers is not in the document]
gianelli-lagos-aws: The ordering is guaranteed, so there's no mis-attribution. _download_all_manifests_with_absolute_paths builds downloaded_manifests by calling _get_manifests_to_download internally and writing each result back at its source index. The CLI then calls _get_manifests_to_download again with the same inputs — it's deterministic (same iteration, no sort) — so [i] lines up exactly in both lists.

One correction: it's not sorted oldest-to-newest here; that sort is in _merge_absolute_path_manifest_list, which isn't used on this path. I'll fix that misleading comment.

Good idea on correlating by S3 key as a safety net — I'll note it as a follow-up.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -1191,61 +1502,352 @@ def _incremental_output_download(
...
             for session in session_list
         ),
-        "downloaded_files": len(manifest_paths_to_download),
-        "downloaded_bytes": sum(path.size for path in manifest_paths_to_download),
+        # On dry runs job_download_results is empty — fall back to job_manifest_paths so the
+        # summary reports the count/size of files that would be downloaded (the preview value).
+        "downloaded_files": sum(
+            r.get("downloaded_files", 0)
+            for job_id, r in job_download_results.items()
+            if r.get("error_code") is None and job_id in job_manifest_paths
```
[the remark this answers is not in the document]
gianelli-lagos-aws: Nice catch, this was genuinely under-counting. Fixed it so downloaded_files sums each job's own count regardless of error_code — a partially-failed job's successful tasks now get counted (still keeping the job_id in job_manifest_paths guard so deduped jobs don't get double-counted). Did the same for downloaded_bytes: full sizes for successful jobs, on-disk-only for partial ones.

Added a test with a 3-task job where one task fails — it now reports 2 files, and I confirmed it reads 0 without the fix. Only the summary/telemetry numbers were affected; the status file counts were already right.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -1191,61 +1504,359 @@ def _incremental_output_download(
...
+                    last_call_time = time.time()
+                return sigint_handler.continue_operation
+
+            job_task_results: dict[str, dict[str, Any]] = {}
+            job_error_code: Optional[str] = None
+            job_error_message: Optional[str] = None
+
+            task_map = job_task_manifest_paths.get(job_id, {})
+            fallback_succeeded = False
+            if task_map:
```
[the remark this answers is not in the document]
gianelli-lagos-aws: Good catch, you're right. I thought this couldn't happen at first but it can.

Chunked-step manifests have S3 keys with no task-<id>, so their files land in job_files but never in task_map. A job mixing task-based and chunked manifests would download only the per-task files and silently drop the chunked ones — still reporting success and advancing the checkpoint, so no retry.

Fixed it: the per-task branch now also grabs any leftover files not covered by a task, counts them, and sets the error code on failure so the checkpoint won't skip past a lost file. Added a regression test that mixes the two manifest types and checks the chunked file actually downloads.

src/deadline/client/cli/_download_status_file.py
```diff
@@ -201,15 +227,77 @@ def _build_status_file_content(
...
+            if new_entry["download_status"] != existing_entry["download_status"]:
+                existing_entry["download_status"] = new_entry["download_status"]
+                existing_entry["last_updated"] = new_entry["last_updated"]
+                if new_entry["download_status"] != "failed":
+                    existing_entry["error_code"] = None
+                    existing_entry["error_message"] = None
+                    existing_entry["failed_files"] = 0
+            existing_entry["tasks"] = merged_tasks
+        else:
+            jobs_status[job_id] = new_entry
```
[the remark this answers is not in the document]
gianelli-lagos-aws: Intentional. The tasks map is cumulative on purpose — each task keeps its status across syncs even when a later incremental run doesn't re-download it (DCM's per-task view relies on that). The job-level total_files/downloaded_files are per last sync, so the two aren't meant to reconcile and consumers shouldn't sum the tasks map against the job total. DCM already treats them separately — its progress bar reads the tasks map, not the job-level counts.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -593,12 +654,21 @@ def _retrieve_session_actions_for_session(
...
-            is_task_run = "taskRun" in session_action.get("definition", {})
-            if succeeded and is_task_run:
+            definition = session_action.get("definition", {})
+            if "taskRun" not in definition:
+                continue
+            status = session_action.get("status")
+            if status == "SUCCEEDED":
+                # Succeeded taskRun — has output to download.
                 session_action_list.append(session_action)
+            elif status == "FAILED" and output_farm_failed_task_ids is not None:
```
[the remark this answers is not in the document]
gianelli-lagos-aws: Good catch — this is a real bug. Confirmed all three legs: the FAILED path records unconditionally before the session_completed_index filter, the synthesis loop's if task_id not in job_task_results guard only covers the current run, and the new_tasks spread in _build_status_file_content overwrites the existing entry.

I fixed it at the merge point in _build_status_file_content rather than by index-filtering the FAILED path. The index-filter approach only fixes the same-session case in the example — but a requeue usually lands in a new session, so the stale FAILED action would still slip through cross-session. The merge-layer rule holds regardless of session or run timing: a farm_failed result never overwrites a task already recorded as downloaded or failed, because both mean the task produced output that's on disk. A prior success is terminal and outranks a later farm-failure re-report.

Added a regression test (test_farm_failed_does_not_clobber_previously_downloaded_task) covering the requeue→succeed→no-op-run sequence — it flips to farm_failed without the guard and stays downloaded with it.

crowecawcaw (dismissed):

test/unit/deadline_client/cli/test_cli_queue_incremental_download.py
phil-IO-p, line 1643: Would it be worth adding one end-to-end test as well? As far as I can tell the wiring between them (the `_get_job_sessions` tuple, the cross-thread merge, and the recording loop at `_incremental_download.py:1913-1930`) isn't currently exercised by anything, so a regression in the plumbing would still leave the suite green even though `farm_failed` never reaches the file.

`test_incremental_output_download_per_task_error_isolation` looks like it already has most of the scaffolding: a session with one `SUCCEEDED` and one `FAILED` `taskRun`, then asserting `tasks[failed_id]["download_status"] == "farm_failed"` in the written file would cover the tuple unpack, the merge, the loop, and the write in one pass.

src/deadline/client/cli/_download_status_file.py
```diff
@@ -235,19 +235,29 @@ def _build_status_file_content(
...
+                "downloaded" if r.get("error_code") is None else "failed",
+            )
+            # A prior-run success is terminal: its output is on disk. A FAILED taskRun from an
+            # earlier attempt keeps being reported by the API on every no-op run, so without this
+            # guard a task that was requeued, succeeded, and downloaded would flip back to
+            # farm_failed forever. Never let farm_failed clobber an already-downloaded task.
+            if status == "farm_failed" and existing_tasks.get(task_id, {}).get(
+                "download_status"
+            ) in ("downloaded", "failed"):
+                continue
```
phil-IO-p: 1. I think there may be a gap in the way out of `farm_failed`. The guard here clears it when the task is later `downloaded` or `failed`, but a task that succeeds on a retry and produces no output never gets an entry in `task_download_results` at all, so nothing overwrites the `farm_failed`. `_filter_session_actions_without_manifests_from_job_sessions` suggests output-free steps are expected (validation steps, steps that save to S3 or a shared drive). If that's right, the file would keep showing a farm failure for a task that has since succeeded, which is close to the stale-signal problem this change is fixing. One option: since the `SUCCEEDED` task IDs are already in hand at `_incremental_download.py:662-664`, you could clear an existing `farm_failed` for those IDs even when there's nothing to download. Or stamping the entry with the failing `sessionActionId` would let a newer succeeded action supersede it. Curious whether you'd already considered this and ruled it out.

2.  Should the guard be limited to `("downloaded",)`? For `downloaded` the "output is already on disk" reasoning in the comment above is clearly right, but `failed` means the *download* errored, so nothing succeeded and there's nothing terminal to protect. The sequence I have in mind: task succeeds on the farm, download fails (entry is `failed` + `error_code`), task is requeued, farm-fails. Since the session index at `_incremental_download.py:1092-1101` advances regardless of whether the download succeeded, the old `SUCCEEDED` action gets filtered out next run, so there's no fresh download result and the guard keeps the stale error code. Meanwhile the job-level entry clears its own error fields in that same run, so the job reads "downloaded, no error" while its task reads "failed" with an old message. If `failed` is meant to be sticky, a test naming that intent would help. Either way, parametrizing `test_farm_failed_does_not_clobber_previously_downloaded_task` over both values (and asserting `error_code` in the `failed` case) would be cheap and would pin whichever behavior you want.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -1883,6 +1910,25 @@ def _update_download_progress(
...
+    # to download, so they'd otherwise be absent from the status file — forcing the FE to rely
+    # on live run-status data that flips on requeue. A stable file entry fixes that.
+    # "farm_failed" (distinct from "failed", which is a download error) tells the artist the
+    # render itself failed — a different cause and fix than a download failure. It's also
+    # distinct from a task that succeeded with no output (those stay absent from the dict).
+    # Skip a task if a real download result already exists for it this run.
+    for job_id, task_ids in farm_failed_task_ids.items():
+        job_task_results = task_download_results.setdefault(job_id, {})
+        for task_id in task_ids:
+            if task_id not in job_task_results:
```
phil-IO-p: Small correction to the PR description rather than the code: the "skip-if-real-result-exists" path is the `if task_id not in job_task_results` check here, but `test_farm_failed_does_not_clobber_previously_downloaded_task` goes through `_build_status_file_content` with `existing_jobs=`, which is the cross-run guard in the other module. I don't think anything currently reaches `:1923`.

That's the within-run case (a task with a stale `FAILED` attempt *and* a successful download in the same sync), which seems worth pinning since it's what happens whenever a task fails and succeeds inside one window. The end-to-end test from my other comment would cover it if the task appears in both the farm-failed set and `task_download_results`.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -1883,6 +1910,25 @@ def _update_download_progress(
...
+                job_task_results[task_id] = {
+                    "total_files": 0,
+                    "downloaded_files": 0,
+                    "error_code": None,
+                    "error_message": None,
+                    "download_status": "farm_failed",
+                }
```
phil-IO-p: This loop sits between two blocks that are both inside `if not dry_run:` but isn't gated itself, so on a dry run it populates `task_download_results`, which makes the synthesis at `:1935-1950` fabricate job results and push those job IDs into `record_successes`.

I traced it and it's harmless today (the tracker `save()` is gated, the file write is gated, and the dry-run stats take the other branch), so this is about intent rather than a live bug. If the omission is deliberate, a short clause saying so would help; otherwise gating it like its neighbours would make the dry-run contract structural instead of depending on a gate 50 lines further down.

src/deadline/client/cli/_incremental_download.py
phil-IO-p: The shape comment here still lists four keys, but the new producer at `:1924-1930` adds `download_status` as a fifth. Since it's optional, set by only one of three producers, and read through a `.get()` with a fallback, that comment is the only place a reader could learn it exists, and `Any` typing means a typo in the key name wouldn't be caught. Worth updating this line (and possibly `:1932`, which now also covers farm-failed-only jobs), or promoting the inner dict to a `TypedDict` with `download_status` as a documented optional field.

src/deadline/client/cli/_incremental_download.py
```diff
@@ -651,12 +654,21 @@ def _retrieve_session_actions_for_session(
...
-            is_task_run = "taskRun" in session_action.get("definition", {})
-            if succeeded and is_task_run:
+            definition = session_action.get("definition", {})
+            if "taskRun" not in definition:
+                continue
+            status = session_action.get("status")
+            if status == "SUCCEEDED":
+                # Succeeded taskRun — has output to download.
                 session_action_list.append(session_action)
+            elif status == "FAILED" and output_farm_failed_task_ids is not None:
```
phil-IO-p, line 684: We're handling `FAILED` and `SUCCEEDED` statuses but what about other terminal session action statuses like `CANCELED`, `INTERRUPTED`, and `NEVER_ATTEMPTED`? They'd currently be dropped silently, and given the motivation is that live run-status data isn't reliable, it seems worth a deliberate decision on whether any of them also deserve a stable entry, even if the answer is that `FAILED` is the only one that matters.

phil-IO-p (changes requested):

test/unit/deadline_client/cli/test_cli_queue_incremental_download.py
gianelli-lagos-aws: Done — added test_incremental_output_download_farm_failed_task_reaches_status_file. It uses a session with one SUCCEEDED and one FAILED taskRun, then reads the written JSON and asserts tasks[task_ids[1]]["download_status"] == "farm_failed". Covers the tuple unpack at :1349, the cross-thread merge, the recording loop, and the file write in one pass.

src/deadline/client/cli/_download_status_file.py
gianelli-lagos-aws: Both addressed. For the stale-farm_failed gap: _retrieve_session_actions_for_session now also collects SUCCEEDED task IDs, and _get_job_sessions removes those from farm_failed_task_ids before returning — so a task that retried and succeeded on the farm can't show as farm_failed even when it produced no output.

For the guard narrowing: changed to == "downloaded". You're right that "failed" isn't terminal — a task that SUCCEEDED on the farm but whose download failed should have its stale error overwritten by a later farm_failed if the task subsequently farm-fails on requeue. The comment in the code explains the reasoning.

src/deadline/client/cli/_incremental_download.py
gianelli-lagos-aws: Correct — test_farm_failed_does_not_clobber_previously_downloaded_task exercises the cross-run guard in _build_status_file_content, not the within-run if task_id not in job_task_results check. The e2e test added for comment 1 covers the within-run case: if a task appears in both the farm-failed set and task_download_results, the real download result takes precedence.

src/deadline/client/cli/_incremental_download.py
gianelli-lagos-aws: Gated under if not dry_run: now, consistent with its neighbours. Added a short comment explaining why — keeping task_download_results clean on dry runs avoids the synthesis block fabricating job results from farm-failed entries.

src/deadline/client/cli/_incremental_download.py
gianelli-lagos-aws: Updated the comment to show download_status? as an optional fifth key with a one-liner explaining when it's set and what the absence means. Held off on a TypedDict for now since download_status is the only optional key and the dict is internal — happy to promote it if you think the typing safety is worth it.

src/deadline/client/cli/_incremental_download.py
gianelli-lagos-aws: Deliberate decision to leave these dropped. INTERRUPTED is a preemption — the task reruns automatically, so recording it would create noise that the next run corrects anyway. CANCELED and NEVER_ATTEMPTED are user intent or a consequence of cancellation — if a job was canceled partway through, sync-output downloads what succeeded and shows those as downloaded; the canceled tasks produced nothing and the artist isn't expecting them. No stable entry is more honest than a misleading one.

src/deadline/client/cli/_download_status_file.py
```diff
@@ -235,19 +273,31 @@ def _build_status_file_content(
...
-                    "download_status": "downloaded" if r.get("error_code") is None else "failed",
-                    "total_files": r["total_files"],
-                    "downloaded_files": r["downloaded_files"],
-                    "error_code": r.get("error_code"),
-                    "error_message": r.get("error_message"),
-                }
-                for task_id, r in new_tasks.items()
-            },
-        }
+        merged_tasks = dict(existing_tasks)
```
phil-IO-p, line 277: Thank you. The `SUCCEEDED` collection is a better fix than what I suggested, and I like that it's collected before the index filter at `:677-681`, so it works on any enumerated session regardless of checkpoint state.

I think there's still a residual gap though: the subtraction in `_get_job_sessions` prevents `farm_failed` from being *re-written*, but it doesn't seem like it can clear an entry that's already on disk from an earlier run.

The path I'm looking at in `_build_status_file_content`: `merged_tasks = dict(existing_tasks)` at this line copies every prior task entry forward, and the loop below only overwrites task IDs that appear in `new_tasks`. Once the task ID is subtracted out of `farm_failed_task_ids`, it's absent from `new_tasks`, so the existing `farm_failed` entry is carried through untouched. As far as I can tell nothing in the merge deletes or rewrites a task entry, so there's no path back out.

In practice: run 1 the task farm-fails and gets written as `farm_failed`. It's requeued and succeeds as a validation-only step, producing no output. Run 2 collects the `SUCCEEDED` action and correctly subtracts the ID, so nothing new is written, but the run-1 `farm_failed` entry is still what the monitor reads.

`test_incremental_output_download_succeeded_task_with_no_output_is_not_a_failure` looks like it covers the "never recorded" case (fresh `checkpoint_dir`, asserts `tasks == {}`), which is the right assertion for that scenario. The case I'm describing needs a pre-existing status file with a `farm_failed` entry going in, then asserting it's gone or `downloaded` coming out.

If clearing it is the intent, the `SUCCEEDED` IDs you're already collecting would be enough, they'd just need to reach the merge as something that removes the entry rather than being subtracted before it. Alternatively, if a stale `farm_failed` on an output-free retry is acceptable in practice, that's a reasonable call and I'd just suggest saying so in the docstring so the next reader doesn't treat it as a bug.

phil-IO-p (changes requested):

src/deadline/client/cli/_download_status_file.py
gianelli-lagos-aws: Good catch — you're right that the subtraction only prevented new writes but left the prior disk entry untouched. Fixed by threading succeeded_task_ids all the way through to _build_status_file_content, where the merge now explicitly removes any farm_failed entry for a task that SUCCEEDED this run with no output. Added a test that pre-populates the status file with a farm_failed entry then runs a sync where that task SUCCEEDS with no manifest — asserts the entry is gone afterward.

phil-IO-p (approved):
Thanks for addressing feedback, looks good!

PR #1322, opened 2026-08-05
fix: route deadline: pre-GUI hook properties to shared values

leon-li-inspire:
Fixes: *N/A — follow-up to #1255 (the pre-GUI hook feature), caught after it merged.*

A pre-GUI hook that returns a `deadline:`-prefixed job property — e.g.
`{"parameters": {"deadline:priority": 88}}` — was **silently dropped** on the
job-bundle-submitter path (`deadline bundle gui-submit`, used by the DCC submitters).
`name` / `description` and ordinary template parameters applied correctly; only
`deadline:*` job properties (priority, initialStatus, maxRetriesPerTask,
maxFailedTasksCount) were lost.

Why it happened: `read_job_bundle_parameters` keeps `deadline:*` values as entries in the
parameters list (`job_bundle/parameters.py`, "Keep the other parameter values around …
`deadline:*`"). So `deadline:priority` **is** in `template_param_names`, and
`apply_pre_gui_output` routed the hook's value into `initial_settings.parameters` — which
the GUI never reads for priority (`openjd_parameters_widget.py` skips any name containing
`:`) — while the stale bundle value stayed in `initial_shared_parameter_values`, which is
what the dialog spinner and the final CreateJob actually use.

In `apply_pre_gui_output`, route `deadline:`-prefixed names to
`initial_shared_parameter_values` **before** the template-name check — they are shared job
properties, not template parameters. The `cli_provided_param_names` check stays ahead of
it, so CLI `--parameter` still wins.

Fixes `priority` / `initialStatus` / `maxRetriesPerTask` / `maxFailedTasksCount` overrides
via pre-GUI for **all** DCC submitters at once. DCC settings without a `.parameters` list
were already unaffected. No public API or CLI contract changes.

- Have you run the unit tests? **Yes.** Added unit regression tests in
  `test/unit/deadline_client/ui/test_pregui_hooks.py`: a `deadline:priority` present in the
  parameters list (as `read_job_bundle_parameters` leaves it) with a stale seeded shared
  value is overridden by the hook and lands in the shared values; CLI `--parameter` still
  wins over the hook for a `deadline:` property; the non-`deadline:` template-parameter
  branch is unchanged.
- Have you run the integration tests? **Yes (UI/xa11y).** Added an end-to-end xa11y test in
  `test/ui/test_bundle_gui_submit_hooks.py` that drives the real GUI: a pre-GUI hook returns
  `deadline:priority` 88 over a bundle whose `parameter_values` seed a stale 50, and asserts
  the value reaches both the visible Priority spinner and the final CreateJob. Also verified
  manually end-to-end on an After Effects test box before the fix (hook returned 88, spinner
  stayed 50, created job came out `priority: 50`) and after (created job `priority: 88`).

- Are relevant docstrings in the code base updated? **Yes** — the `apply_pre_gui_output`
  docstring/comments now explain the `deadline:*` routing.
- Has the README.md been updated? **N/A** — no CLI arguments or user-facing options changed.

No. This is a bug fix that restores intended behavior; it does not modify any public
contract in a backwards-incompatible way.

- Does the change need to be threat modeled? **No** — it does not create or modify
  files/directories or change any trust boundary; it only routes an already-executed hook's
  output to the correct in-memory dict.

----

### DIFF
--- a/src/deadline/client/ui/pre_gui_hooks.py
+++ b/src/deadline/client/ui/pre_gui_hooks.py
@@ -243,14 +243,21 @@ def apply_pre_gui_output(
         # CLI --parameter values take precedence over hook values.
         if param_name in cli_provided_param_names:
             continue
-        if param_name in template_param_names:
+        if param_name.startswith("deadline:"):
+            # Shared job property (priority, initialStatus, maxRetriesPerTask, …). These are
+            # NOT template parameters even though read_job_bundle_parameters keeps them in the
+            # parameters list — the dialog spinner and CreateJob read them from the shared
+            # values dict, so route them there. Checking this before the template-name test
+            # is what lets a pre-GUI hook actually override e.g. deadline:priority.
+            initial_shared_parameter_values[param_name] = param_value
+        elif param_name in template_param_names:
             # Job template parameter — update initial_settings.parameters in-place
             for p in template_parameters:
                 if p["name"] == param_name:
                     p["value"] = param_value
                     break
         else:
-            # Shared job property (deadline: keys, queue parameters, etc.)
+            # Other shared job property (queue parameters, etc.)
             initial_shared_parameter_values[param_name] = param_value
     if "name" in pre_gui_output:
         initial_settings.name = pre_gui_output["name"]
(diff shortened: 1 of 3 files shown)


src/deadline/client/ui/pre_gui_hooks.py
```diff
@@ -243,14 +243,21 @@ def apply_pre_gui_output(
         # CLI --parameter values take precedence over hook values.
         if param_name in cli_provided_param_names:
             continue
-        if param_name in template_param_names:
+        if param_name.startswith("deadline:"):
```
crowecawcaw, line 246: Would be nice to define this as a constant some place with some more context around it.

crowecawcaw (approved):

PR #1325, opened 2026-08-06
fix: warn about long Windows paths regardless of registry setting

phil-IO-p:
## What changed

Two CLI long-path warnings were gated on the machine-wide `LongPathsEnabled` registry setting:

- `manifest_group.py` — the `deadline manifest snapshot` warning about a long manifest path
- `job_group.py` — `_check_and_warn_long_output_paths` in `deadline job download-output`

Both included `and not _is_windows_long_path_registry_enabled()`, which suppressed the warning whenever that setting was on.

That setting only takes effect for processes declaring `longPathAware` in their application manifest. It says nothing about whether the path is safe for the tools and applications that ultimately open these files — DCC executables such as `AfterEffects.exe` and `Cinema4D.exe` do not declare it and remain capped at `MAX_PATH` regardless.

The result was backwards: the warnings went quiet on exactly the hosts where paths are longest, which are the hosts that have the registry setting enabled. Warn whenever the path exceeds the limit on Windows.

The now-unused `_is_windows_long_path_registry_enabled` import is removed from both modules.

## Relationship to the job attachments fix

Companion to [deadline-cloud-job-attachments#67](https://github.com/aws-deadline/deadline-cloud-job-attachments/pull/67), which removes the same incorrect assumption from `_get_long_path_compatible_path`, where it had a functional rather than cosmetic effect (the `\\?\` prefix was skipped on hosts with the registry setting enabled).

This PR is user-facing messaging only. Nothing fails as a result of the current behaviour; users simply are not told about a path that may cause problems.

Note one interaction: with #67 applied, long paths gain the 4-character `\\?\` prefix, and these warnings measure the resulting length. That makes them marginally more likely to fire. Since they are advisory and the underlying path is genuinely near the limit, that seems acceptable, but flagging it in case reviewers would prefer the checks measure the unprefixed length.

## Tests

New `test/unit/deadline_client/cli/test_cli_long_path_warnings.py`:

- `test_registry_helper_no_longer_used` — parametrized over both modules, guards against the registry check being reintroduced
- `TestDownloadOutputLongPathWarning` — invokes the real `deadline job download-output` command with a stubbed downloader, parametrized over both registry states, asserting the warning fires for long paths and not for short ones
- `TestLongPathMessages` — checks the user-facing text names the problem and links the Microsoft guidance

Verified the tests actually catch the defect: with the source changes stashed, 4 of 7 fail; with them applied, all 7 pass.

## Verification

- Full unit suite: 3114 passed, 21 skipped, 0 failed
- `hatch run fmt`: clean
- `hatch run lint`: ruff clean, mypy clean across 315 source files

Not verified on a real Windows host — the warning paths are `sys.platform`-gated and exercised via mocks.

### DIFF
--- a/src/deadline/client/cli/_groups/job_group.py
+++ b/src/deadline/client/cli/_groups/job_group.py
@@ -26,10 +26,7 @@ from ....job_attachments.models import (
     JobAttachmentS3Settings,
     PathFormat,
 )
-from ....job_attachments._utils import (
-    WINDOWS_MAX_PATH_LENGTH,
-    _is_windows_long_path_registry_enabled,
-)
+from ....job_attachments._utils import WINDOWS_MAX_PATH_LENGTH
 from ....job_attachments.progress_tracker import (
     DownloadSummaryStatistics,
     ProgressReportMetadata,
@@ -717,7 +714,10 @@ def _download_job_output(
     def _check_and_warn_long_output_paths(
         output_paths_by_root: dict[str, list[str]],
     ) -> None:
-        if sys.platform == "win32" and not _is_windows_long_path_registry_enabled():
+        # Not gated on the LongPathsEnabled registry setting: that only takes effect for
+        # processes declaring longPathAware in their manifest, so it does not indicate
+        # whether a long path is safe for the applications that will open these files.
+        if sys.platform == "win32":
             for root, paths in output_paths_by_root.items():
                 for output_path in paths:
                     if len(root + output_path) >= WINDOWS_MAX_PATH_LENGTH:
--- a/src/deadline/client/cli/_groups/manifest_group.py
+++ b/src/deadline/client/cli/_groups/manifest_group.py
@@ -27,10 +27,7 @@ from deadline.client.api._session import (
 )
 from deadline.client.config import config_file
 from deadline.job_attachments._diff import pretty_print_cli
-from deadline.job_attachments._utils import (
-    WINDOWS_MAX_PATH_LENGTH,
-    _is_windows_long_path_registry_enabled,
-)
+from deadline.job_attachments._utils import WINDOWS_MAX_PATH_LENGTH
 from deadline.job_attachments.api.manifest import (
     _glob_files,
     _manifest_diff,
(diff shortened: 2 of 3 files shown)


crowecawcaw:
I think the code was correct before but I could be wrong. It looks like these warnings are emitted when someone is downloading job outputs. If they're downloading a file and it has a long path and they don't have long paths enabled, it might fail. The PR reframes the errors as if they're a check before uploading job attachment, but that's not where the warning are emitted. Does that seem right?

andychoquette:
Closing this — on a second look it is not necessary, and as written it is net-negative.

**The argument does not transfer from the job-attachments fix.** [deadline-cloud-job-attachments#67](https://github.com/aws-deadline/deadline-cloud-job-attachments/pull/67) removes the same registry check, and there the reasoning holds: that code runs inside `Cinema4D.exe`, `AfterEffects.exe`, and `pythonservice.exe`, none of which declare `longPathAware`, so `LongPathsEnabled` says nothing about whether a long path will work.

Both call sites this PR touches are CLI-only. `_check_and_warn_long_output_paths` is a closure inside `_download_job_output`; the manifest warning is inside `manifest_group.manifest_snapshot`. They are reachable via the `deadline` / `deadlinew` entry points, `handle_web_url_command`, and `_mcp/tools/job.py` — nothing in `src/deadline/client/ui/` touches them, and the DCC submitters route through `_hash_attachments` / `S3AssetManager` rather than these CLI group functions. So the host process is `python.exe`, which has declared `longPathAware` since CPython 3.6 ([bpo-27731](https://bugs.python.org/issue27731)).

For these two sites the registry check is therefore a *valid* proxy: registry enabled plus a manifest-aware host means long paths genuinely work. This PR turns a correct condition into a false positive.

**The asymmetry I collapsed.** Applying the `\\?\` prefix when it is not needed is free — a no-op for an aware process. Printing a warning when it is not needed is user-visible noise, emitted on precisely the hosts that are correctly configured. Those are not the same change and I treated them as one.

**It also leaves the advice self-contradicting.** The message says "For details and a fix *using the registry*." Firing it unconditionally tells a user who already has `LongPathsEnabled` set to go and set it. `test_download_warning_message_names_the_problem` asserts that URL stays in the text, so this PR would have locked in the misleading advice while removing the condition that suppressed it.

**No sequencing dependency, contrary to what I said on #67.** That PR keeps `_is_windows_long_path_registry_enabled` and is green without this one. This PR only unblocks *deleting* the now-unused helper, which is optional cleanup — a private util that another package imports costs nothing to leave in place. Correcting that comment there.

**The smaller real issue underneath, for anyone picking it up:** the warning's advice is incomplete rather than wrongly gated. A path under a registry-enabled host can still break the DCC that later opens the file, because `LongPathsEnabled` only applies to processes declaring `longPathAware`. The fix is a message-text change — note that downstream applications may still fail — while **keeping** the registry gate, since on a non-enabled host setting the registry really is the right first move. Roughly the inverse of this PR.

PR #1357, opened 2026-09-11
feat: restructure queue sync-output progress report

phil-IO-p:
## What changed

`deadline queue sync-output` prints a multi-section progress report. This restructures it as indented plain-ASCII sections with a closing summary and a single greppable result line, and adds `--verbose`.

Before:

```
Ignoring all storage profiles.
Started incremental download for queue: Test queue
Checkpoint: /Users/me/.deadline/incremental_download/queue-d8bb..._download_checkpoint.json

Checkpoint found
Continuing from: 2026-08-21T10:16:59.319524-07:00

Updating download state across time interval:
    From: 2026-08-21T10:16:59.319524-07:00
      To: 2026-08-21T10:29:54.361383-07:00
  Length: 0:10:55.041859 + 0:02:00 (eventual consistency allowance)

Retrieving updated data from Deadline Cloud...
DEBUG: Got 0 active jobs
DEBUG: Filtered down to 0 active jobs based on SUCCEEDED task filter
...retrieval completed in 0:00:00.606954
Retrieving sessions for 1 jobs...
Using 3 threads
...retrieval completed in 0:00:00.264083
```

After:

```
Queue sync: Test queue
  checkpoint  ~/.deadline/incremental_download/queue-d8bb..._download_checkpoint.json
  storage     ignoring all storage profiles
  state       resumed from checkpoint at 2026-08-21 10:16:59
  window      2026-08-21 10:16:59 to 10:29:54 (10m55s + 2m eventual consistency allowance)

Deadline Cloud
  1 candidate job (0 active, 1 recently ended)
  categorized 1 job (0 in checkpoint)
    NEW       my_render_job (4/4 tasks succeeded)
              job-ec1bc8ac3b14409d8ed010262c7b8b32
              output root: / (posix)
  3 sessions across 1 job
  4 succeeded task runs

Download
  4 asset manifests                                     5.3s
  4 files, 4.58 MB to download
    /tmp/ViewLayer_Camera_output_%04d.png (4 files, 4.58 MB, sequence 1-4)
  downloaded 4 files                                    1.1s

Summary
  files       4 downloaded, 4.58 MB
  jobs        1 downloaded
  task runs   4
  skipped     none
  status file ~/.deadline/.../queue-d8bb..._download_status.json
  checkpoint  saved
  elapsed     7.6s
OK  4 files, 4.58 MB, 1 job in 7.6s
```

## Details

**Plain ASCII only.** Color now marks a problem and nothing else: yellow on `WARNING`, red on `ERROR` and `FAILED`. This matches the three places the CLI already uses color (`job_group.py`, `manifest_group.py`, `bundle_group.py`), all of which are warnings. No other CLI source emitted non-ASCII, so structure is carried by indentation rather than box-drawing characters, which would raise `UnicodeEncodeError` when output is redirected to a file under a legacy Windows codepage.

**A run that finds nothing new prints one line.** This command is built to be polled on a schedule, and most polls find nothing:

```
OK  Test queue: nothing new (window 10m55s, checked 2026-08-21 10:29:54)
```

The new `--verbose` flag forces the full report. A run that finds work, retries a previously failed job, or hits a warning always prints the full report regardless. This changes the default output of a no-op run.

**Failures are repeated in the summary**, so an unattended run does not bury them part-way up a download log:

```
  problems    2 jobs failed
              PERMISSION_DENIED  shot_041_comp (job-abc123def456)
              DISK_FULL          shot_042_comp (job-789abc012def)
  elapsed     41s
FAILED  2 of 7 jobs failed after 41s (12 files, 31.2 MB downloaded)
```

**Durations** render as `0.6s` or `10m55s` instead of `0:00:00.606954`, and are omitted below one second so a genuinely slow step stands out.

**Job wording follows the download status file's vocabulary** (`downloaded`, `in_progress`, `skipped`, `failed`) so the terminal and the JSON report describe a job the same way. The `DONE` tag, which meant "no longer tracked" rather than "downloaded", is now `RETIRED`.

## Bug fixed along the way

Four `DEBUG:` lines used bare `print()` rather than the logger, so they reached stdout even under `--json` and corrupted the JSON output. They are gone, and a regression test asserts `--json` writes nothing to stdout.

## Note for reviewers

Two `job_attachments` helpers are now passed a no-op print callback. Their only output was a thread count, and this command reports those phases in its own format.

Tagged `feat:` rather than `feat!:`. The default output of a no-op run changed, which `AGENTS.md` arguably covers under "default values and behaviors", but `--json` (the documented machine-readable path) is unchanged. Happy to relabel if the team reads that differently.

## Testing

`hatch run lint` (ruff + mypy) clean. `hatch run test`: 3859 passed, 22 skipped, coverage 77.78%. New tests cover the formatter (durations, alignment, ASCII-only output that encodes to cp1252), quiet-by-default, `--verbose`, the problems block, the result line, and the `--json` regression.

`hatch run integ:test` has **not** been run; it needs AWS resources I do not have set up.

### DIFF
--- a/src/deadline/client/cli/_sync_output_format.py
+++ b/src/deadline/client/cli/_sync_output_format.py
@@ -1,0 +1,253 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Terminal formatting for the `deadline queue sync-output` command.
+
+Output is plain ASCII. Color is used only to flag a problem (yellow for warnings, red
+for errors), matching the rest of the CLI, and structure is carried by indentation so
+the log reads the same when redirected, piped, or collected by a log agent.
+"""
+
+from __future__ import annotations
+
+__all__ = [
+    "_ENTRY_DETAIL_INDENT",
+    "_SyncOutputFormatter",
+    "_SyncOutputWriter",
+    "_format_duration",
+    "_format_interval",
+    "_format_path",
+    "_format_timestamp",
+    "_plural",
+]
+
+import os
+from datetime import datetime, timedelta
+from typing import Any, Callable, Optional, Union
+
+import click
+
+# Column that right-aligned step timings are flushed to, in characters.
+_TIMING_COLUMN = 56
+
+# Timings below this are noise: a run reports six sub-millisecond bookkeeping steps whose
+# durations say nothing, and hiding them lets the one genuinely slow step stand out.
+_MIN_REPORTED_SECONDS = 1.0
+
+# Width of the label column used by `field()` and `summary_row()`.
+_LABEL_WIDTH = 12
+
+# Layout of the tagged job entries printed beneath a step, e.g. `NEW       my-job`.
+# Continuation lines indent past the tag so they sit under the entry title.
+_ENTRY_INDENT = 4
+_ENTRY_TAG_WIDTH = 9
+_ENTRY_DETAIL_INDENT = _ENTRY_INDENT + _ENTRY_TAG_WIDTH + 1
+
+
+def _format_duration(duration: Union[timedelta, float, int]) -> str:
+    """Renders a duration compactly: `<0.1s`, `0.6s`, `42s`, `10m55s`, `2h05m`."""
+    seconds = duration.total_seconds() if isinstance(duration, timedelta) else float(duration)
+    seconds = max(seconds, 0.0)
+    if seconds < 0.1:
+        return "<0.1s"
+    if seconds < 10:
+        return f"{seconds:.1f}s"
+    if seconds < 60:
+        return f"{seconds:.0f}s"
+    minutes, whole_seconds = divmod(int(round(seconds)), 60)
+    if minutes < 60:
+        return f"{minutes}m" if whole_seconds == 0 else f"{minutes}m{whole_seconds:02d}s"
+    hours, minutes = divmod(minutes, 60)
+    return f"{hours}h" if minutes == 0 else f"{hours}h{minutes:02d}m"
+
+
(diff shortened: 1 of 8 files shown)


src/deadline/client/cli/_sync_output_format.py
```diff
@@ -0,0 +1,257 @@
...
+def _format_interval(start: datetime, end: datetime) -> str:
+    """Renders a time interval, collapsing the date on the end when both fall on one day."""
+    start_local = start.astimezone()
+    end_local = end.astimezone()
+    if start_local.date() == end_local.date():
+        return f"{_format_timestamp(start)} to {end_local.strftime('%H:%M:%S')}"
+    return f"{_format_timestamp(start)} to {_format_timestamp(end)}"
+
+
+def _format_path(path: str) -> str:
```
kavmur, line 88: will shared drive paths containing unicode be unchanged? (output may not be plain ascii anmore?)

kavmur (dismissed):

src/deadline/client/cli/_sync_output_format.py
phil-IO-p: Good question! Confirming paths pass through untouched, so a shared drive path keeps its own characters and stays one you can copy. Verified + added a new test for it.

You're right that the output wouldn't be plain ASCII overall, and the docstring Claude added was over-claiming. `_format_path` only collapses a leading `$HOME` to `~`; it never touches the rest. What is guaranteed ASCII is the decoration this module adds: headings, labels, tags, separators. I have scoped the docstring to say that, renamed `test_output_is_pure_ascii` to `test_decoration_is_pure_ascii` since that is all it checked, and added `test_values_pass_through_unchanged`.

kavmur (approved):

